PR #1, opened 2016-11-21
Buttons design

koriguy:
@3LOK Please take a look

### DIFF
--- a/src/Button/Button.scss
+++ b/src/Button/Button.scss
@@ -6,7 +6,7 @@
 
     box-sizing:border-box;
     -webkit-font-smoothing: antialiased;
-    border-radius: 15px;
+    border-radius: 26px;
     transition: 
       color 0.3s ease,
       border-color 0.3s ease,
--- a/src/Button/Button.scss
+++ b/src/Button/Button.scss
@@ -7,6 +7,10 @@
     box-sizing:border-box;
     -webkit-font-smoothing: antialiased;
     border-radius: 15px;
+    transition: 
+      color 0.3s ease,
+      border-color 0.3s ease,
+      background-color 0.3s ease;
     height: 36px;
     padding: 0px 24px;
     text-align: center;


src/Button/Button.scss
```diff
@@ -6,7 +6,11 @@
 
     box-sizing:border-box;
     -webkit-font-smoothing: antialiased;
-    border-radius: 15px;
+    border-radius: 26px;
```
mastertheblaster: @koriguy I guess it's not good. According to documentation the buttons can have different heights. So border radius should depend on the height. Now you are changing the radius for all buttons.

Perhaps the button radius could be calculate?

koriguy: @mastertheblaster actually no issue here since this is the highest border-radius of buttons and the browser will reduce this value to the maximum allowed (on smaller buttons).
see here (testes on chrome/FF/safari and edge)
http://codepen.io/anon/pen/dOvejv
all of this since current implementation is by setting specific height manually.

In general I think that the implementation should be with modifier class - small/large so the dev' will not need to set specific height, only set the modifier and we'll apply another class that contains the correct border-radius (or other things if we need)
@3LOK - wdyt?

mastertheblaster: Indeed, browser is amazing :)

3LOK: @koriguy @mastertheblaster Ok, so I've integrated your change, and continued with also changing height to be a string value, linked to a modified class.

__This is, of course, a breaking change to anyone who's been using the height value__

mastertheblaster: @3LOK What if leaving the `height` property untouched. And adding a new one?

```js
<Button
  size="x-small | medium | big | x-big"
  height={100}/>
```

They are a little bit different. Size increases all button properties (font size, paddings).
The `height` increases only the height.

koriguy:
I don't think we need to support other heights, this is a customization
that shouldn't be supported by ui lib

On Nov 22, 2016 10:37 AM, "Mantas Miliukas" <notifications@github.com>
wrote:

*@mastertheblaster* commented on this pull request.
------------------------------

In src/Button/Button.scss <https://github.com/wix/wix-style-react/pull/1>:

> @@ -6,7 +6,11 @@

     box-sizing:border-box;
     -webkit-font-smoothing: antialiased;
-    border-radius: 15px;
+    border-radius: 26px;

@3LOK <https://github.com/3LOK> What if leaving the height property
untouched. And adding a new one?

<Button
  size="x-small | medium | big | x-big"
  height={100}/>

They are a little bit different. Size increases all button properties (font
size, paddings).
The height increases only the height.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/wix/wix-style-react/pull/1>, or mute the thread

src/Button/Button.scss
3LOK: @mastertheblaster - 

The question is - for what purpose? 

If for backwards compatibility, I don't think a week-old package should really hold legacy code for backwards compatibility. 

If as an actual feature: As @koriguy answered in the email: "I don't think we need to support other heights, this is a customization that shouldn't be supported by ui lib". UI Lib limits to specific heights. If your UX has decided to design a button with a different height, he/she should first update the UX UI Lib with it, which will result in an update in this lib as well.

PR #3, opened 2016-11-23
Buttons styles according to ui-lib

koriguy:
@3LOK @mastertheblaster this is clearer I believe and full.
internal div inside the button had different defaults so had to inherit them from the button (and not the default for page, which I'm not sure we need)
buttons are now with correct padding/font size and  radius
image

### DIFF
--- a/src/Button/Button.scss
+++ b/src/Button/Button.scss
@@ -2,7 +2,11 @@
 
 .button {
     @include FontRoman();
-    font-size: 14px;
+    $height: 36px;
+    height: $height;
+    border-radius: $height / 2;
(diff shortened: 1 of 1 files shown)


3LOK:
@koriguy merged, however there was a small issue that when using the 'medium' height, there would be an 'undefined' css class added. I don't think in normal operation this should happen, so fixed it in [this commit](https://github.com/wix/wix-style-react/commit/9635fe304e8f73182a16cf36004ead4b93e5c9af).

koriguy:
@3LOK  Great. Thanks

PR #13, opened 2016-11-24
fix Button testkit to work with enzyme/react-test-utils

yanivefraim:


yanivefraim:
cc @3LOK

3LOK:
@yanivefraim I'm thinking of maybe changing this directory to be called enzyme-test-kit. If someone ever needs a testkit for a different system, then they can create 'blahblah-test-kit' directory. What do you think?

@shahata What do you think?

yanivefraim:
@3LOK - not sure if Enzyme is the correct term here.. This is something coming from [react-test-utils](https://facebook.github.io/react/docs/test-utils.html)

PR #15, opened 2016-11-25
added first test

yanivefraim:


yanivefraim:
cc @shahata @3LOK

yanivefraim:
This is bery initial setup for tests. I will add e2e and more tests, as soon as poss

PR #37, opened 2016-12-05
Remove lodash

nirhart:


src/AutoCompleteInput/AutoCompleteInput.js
```diff
@@ -42,7 +43,7 @@ class AutoCompleteInput extends React.Component {
 
           {this.props.header}
 
-          {_.map(this.props.suggestions, (suggestion, index) => {
+          {map(this.props.suggestions, (suggestion, index) => {
```
shahata: why not `this.props.sugestions.map((suggestion, index) => {`?

src/Slider/Slider.js
```diff
@@ -33,7 +33,7 @@ export default class Slider extends Component {
   }
 
   getMarks() {
-    return _.reduce(this.getRange(), (acc, cur) => {
+    return reduce(this.getRange(), (acc, cur) => {
```
shahata: why not `this.getRange().reduce((acc, cur) => {`?

src/Select/Select.js
```diff
@@ -132,7 +134,7 @@ class Select extends React.Component {
           </svg>
         </div>
         <div data-testkey={(this.props.nameAttr) ? this.props.nameAttr + 'Select' : 'wixMenuSelect'} className={`${styles.options} ${this.state.open ? styles.shown : ''} ${this.props.dropDirectionUp ? styles.up : styles.down}`} ref="options">
-          {_.map(options, (option, idx) => (
+          {map(options, (option, idx) => (
```
shahata: options.map...

3LOK:
@nirhart, @shahata - I have only one major comments from all my comments (under the autocomplete part). Shahar - your comments are major or 'nice to have'?

nirhart:
@3LOK I'm adding @shahata suggestions,
by major are you referring this - ' // TODO this method never gets called, this.props.suggestions[selectedSuggestion] is an object {text: 'fsdfds, ...} ' ?
IMO this is the only major issue

3LOK:
@nirhart - no, I meant:

> this isn't good. it means onMouseDown will re-render on each render.'

having something like:
```javascript
<div onClick={() => this.onClick()}></div>
```
means the div will __always__ get re-renderer as its onClick prop always changes. I'm not sure the previous solution did not have the same issue, but it's worth fixing if touching.

PR #53, opened 2016-12-12
feat(tooltip): proof of concept

mastertheblaster:
Please don't merge yet, double check the approach if it suits your needs.


koriguy:
@3LOK please take a look. tnx

3LOK:
@koriguy thanks - sent comments by mail:

1. storybook needs to be reverted to previous version (unrelated to this pull request, but just FYI).

2. As for the interface, looks good.

3. We have different animations - @mastertheblaster said he'll take a look into which is the correct one.

4. I request a prop that will "lock" the tooltip on "show" (meaning, always show instead of show on hover).

5. I requested a function that will bounce the tooltip - it's an effect in the ui lib. (for example, in case of an error on a button, and the user presses the button again).

Thanks!

koriguy:
We also need the ability to prevent close when the user hovers on the tooltip itself (example: in case there is a link inside the tooltip)

3LOK:
I think that should be the default behaviour.

3LOK:
@koriguy @mastertheblaster you guys will tell me when it's ready to merge, right?

mastertheblaster:
@3LOK Will let you know.

PR #57, opened 2016-12-14
feat: allow hiding magnifying glass

mastertheblaster:
for google address input


mastertheblaster:
@3LOK Please take a look. We need to hide the magnifying glass in our project, so i made it configurable.

PR #58, opened 2016-12-15
remove required from props that should not be required (those that ha…

ronami:
…ve defaults or should have defaults)


ronami:
@3LOK can you please take a look? thanks!

3LOK:
Wait, I'm not sure isRequired and default are XOR. One way to see isRequired is 'this component requires this to work'.

Do you have any reference to this?

3LOK:
https://twitter.com/AmitYoav/status/809439820837687296

The master has spoken.

PR #62, opened 2016-12-19
Implement a basic tooltip

mastertheblaster:


mastertheblaster:
Missing a bunch of tests for the tooltip, will add later.

src/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,237 @@
...
+      onClick: this._chainCallbacks(child.onClick, this._onClick),
+      onMouseEnter: this._chainCallbacks(child.onMouseEnter, this._onMouseEnter),
+      onMouseLeave: this._chainCallbacks(child.onMouseLeave, this._onMouseLeave),
+      onFocus: this._chainCallbacks(child.onFocus, this._onFocus),
+      onBlur: this._chainCallbacks(child.onBlur, this._onBlur)
+    });
+  }
+
+  _chainCallbacks = (first, second) => {
+    return args => {
```
argshook: @mastertheblaster i guess here could be multiple, so maybe `(...args)`?

3LOK:
Hi, I've merged - but note that the current 'bounce' method is unpractical for single bounces people might want.

mastertheblaster:
@3LOK I will introduce a method `bounceOnce` which will do all the logic.
And tooltip will work in controlled and uncontrolled manner.

PR #65, opened 2016-12-20
feature: allow to select/focus google address element

mastertheblaster:
@3LOK


src/GoogleAddressInput/GoogleAddressInput.spec.js
```diff
@@ -47,6 +47,12 @@ describe('GoogleAddressInput', () => {
 
   describe('User Interactions', () => {
 
+    it('should provide select/focus methods', () => {
+      const component = createShallow({Client: GmapsTestClient, countryCode: 'XX'});
+      expect(typeof component.instance().select).toEqual('function');
+      expect(typeof component.instance().focus).toEqual('function');
+    });
```
3LOK: Is it possible to add a spy to the shallow rendered AutocompleteInput and making sure calling select calls the spy (same for focus?)

Issue #68, opened 2016-12-21
Add disabled mode for relevant components


lbelinsk:
For instance, I have a use case where I would like to make the dropDown disabled immediately after a value is chosen.

Today I think the only component with 'disable' props is the checkbox

lbelinsk:
I think we can close this one @nirhart

PR #100, opened 2016-12-30
Added a disable mode for radio buttons. issue #68

lbelinsk:


nirhart:
This all component is not that good, I would split it into 2 js files, one component per file, avoid using context and avoid using lodash.

PR #103, opened 2016-12-30
feat(input theme): allow setting material theme

mastertheblaster:
We need material like design for the inputs.
The complete design (errors, icons, ... are not ready yet). It's still work in progress.
Thus I added only basic support.

Currently `theme=material` will work with a following components:
- `<Input theme="material"/>`
- `<AutoCompleteInput theme="material"/>`
- `<GoogleAddressInput theme="material"/>`


koriguy:
@nirhart this design is required for our onboarding process and is in sync with @nir-k 
@mastertheblaster it should not be in story book so others will not use it till Koso is done with the whole UX implementation of things

mastertheblaster:
@koriguy Removed the examples, and restored README.md

PR #104, opened 2016-12-31
Add optional readOnly property to Input. close #60

### LINKED ISSUE #60: Add optional readOnly property to Input
I am using an Input and want it to be readOnly. since there is no optional readOnly property I am getting Lint warnings: 

 Warning: Failed form propType: You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly`. Check the render method of `Input`.

lbelinsk:


src/Input/Input.js
```diff
@@ -132,7 +134,8 @@ class Input extends React.Component {
 Input.displayName = 'Input';
 
 Input.defaultProps = {
-  theme: 'normal'
+  theme: 'normal',
+  readOnly: false
```
mastertheblaster: No need defaults for the `false` value imho.

src/Input/README.md
```diff
@@ -11,6 +11,7 @@
 | defaultValue | string | - | - | Default value for those who wants to use this component un-controlled |
 | tabIndex  | number | - | - | Standard component tabIndex |
 | placeholder  | string | - | - | Placeholder to display |
+| readOnly  | bool | false | - | sets the input to readOnly |
```
mastertheblaster: The same `Input` is being used by `AutoComplete` and `GoogleAutoComplete` components.
Would be very nice if this feature would provided for them as well.

src/Input/Input.spec.js
```diff
@@ -40,6 +40,17 @@ describe('Input', () => {
     });
   });
 
+  describe('readOnly attribute', () => {
+    it('should pass down to the wrapped input', () => {
```
mastertheblaster: Guess you should also check the default case? That by default it's not readonly.

lbelinsk: I agree.. Will add this test

src/Input/Input.js
lbelinsk: Thanks... I will update this...

src/Input/README.md
lbelinsk: I took a look in their code.
**AutoComplete** pass it automatically with the spreader operator {...props} to the Input.
**GoogleAddressInput** doesn't pass it to AutoComplete, so it requires a slight change in it.

Of course this prop needs to be documented in both of them. I will take care of that.

src/Input/Input.spec.js
lbelinsk: Done.

PR #106, opened 2016-12-31
Enhanced DatePicker component. close #105

### LINKED ISSUE #105: DatePicker changes
- [x]  Enable keyboard navigation on the calendar. DatePicker uses ReactDatePicker which allows this kind of behaviour. We have some bug with our custom Input.

- [x] Automatically blur when a date was selected with the keyboard or a mouse click.

- [x] Add an optional props which enables a readOnly mode where the date can't be selected directly through the Input by typing the desired date.

- [x] Add year dropdown selection. it is annoying to navigate some years forward or backwards because it requires passing through each month on the way.

lbelinsk:
I have added some features to the component which I needed in my project.

This component uses a third party library - react-datepicker which has some problems in it.
I have opened some issues in this project repo regarding some bad behaviours which should be modified.

I have opened an issue which describes these problems. issue #107


src/DatePicker/DatePickerInput.js
```diff
@@ -1,6 +1,7 @@
 import React, {Component} from 'react';
 import Input from '../Input';
 import styles from './DatePickerInput.scss';
+import omit from 'lodash/omit';
```
mastertheblaster: Guess you didn't cleaned up the node_modules. `lodash/omit` is not a dependency of this project.

PR #110, opened 2017-01-03
feat(tooltip): allow setting custom tooltip and arrow styles

mastertheblaster:
- Formatted `TooltipContent` component code to be more readable.
- Added ability to specify custom tooltip and tooltip arrow styles. This is required to allow slightly adjusting the tooltip and it's arrow placements per project needs.


mastertheblaster:
screen shot 2017-01-03 at 09 46 41

PR #117, opened 2017-01-03
Feat/tooltip style to props

mastertheblaster:


mastertheblaster:
Exposed 2 props now:
- `moveBy` allow to move the tooltip relative to current position by given x and y.
- `moveArrowTo` allow to move the arrow to desired placement.

PR #125, opened 2017-01-05
Refactored RadioGroup

lbelinsk:
1. Created a new file for RadioButton so now each component has it's own file.

2. Changed the implementation not to use Context.

3. Not importing the whole lodash package anymore.

4. Important changes to be aware of:
    4.1  It is not allowed anymore to wrapp `<RadioGroup.Radio>` with `<div>`.
    4.2 The vAlign prop should be placed on the `<RadioGroup>` and not on each `<RadioGroup.Radio>`

Usage example:

screen shot 2017-01-06 at 12 02 36 am

**If for any reason on of the changes doesn't sounds good, please let me know**


lbelinsk:
@3LOK Please take a look in theses upcoming changes. If by any chance you see any problematic change, please let me know.

3LOK:
Hi,

Why would you need to use the 'id' prop? Do we have it in other elements?

 - Yoav.

lbelinsk:
This is a composite element (contains other smaller elements) and I saw it on some other elements which place the Id on the outer layer.

PR #139, opened 2017-01-09
added Tag component. close #138

### LINKED ISSUE #138: Tag
Add basic Tag component

lbelinsk:
This commit contains tests + testkit.

### DIFF
--- a/src/Tag/README.md
+++ b/src/Tag/README.md
@@ -8,7 +8,7 @@
 |----------|----------|--------------|------------|-------------|
 | label | string | - | + | The label which will be displayed on the Tag |
 | id | string | - | + | The id of the Tag |
-| erasable | bool | true | - | If the Tag is erasable then it will contain a small clickable X|
-| onRemoveTag | func | - | - | Callback function when removing the Tag |
+| removable | bool | true | - | If the Tag is removable then it will contain a small clickable X|
+| onRemove | func | noop | - | Callback function when removing the Tag |
 | size | small or large | small | - | The height of the Tag |
-| icon | element | - | - | An optional icon to display as part of the Tag |
+| thumb | element | - | - | An optional thumb to display as part of the Tag |
--- a/src/Tag/README.md
+++ b/src/Tag/README.md
@@ -1,0 +1,14 @@
+# Tag component
+
+> A Tag component
+
+## Properties
+
+| propName | propType | defaultValue | isRequired | description |
+|----------|----------|--------------|------------|-------------|
+| label | string | - | + | The label which will be displayed on the Tag |
+| id | string | - | + | The id of the Tag |
+| erasable | bool | true | - | If the Tag is erasable then it will contain a small clickable X|
+| onRemoveTag | func | - | - | Callback function when removing the Tag |
+| size | small or large | small | - | The height of the Tag |
+| icon | element | - | - | An optional icon to display as part of the Tag |
--- a/src/Tag/Tag.js
+++ b/src/Tag/Tag.js
@@ -1,0 +1,35 @@
+import React, {PropTypes} from 'react';
+import styles from './Tag.scss';
+import classNames from 'classnames';
+
+const Tag = ({id, label, icon, erasable, onRemoveTag, size}) => {
+  const className = classNames({
+    [styles.tag]: true,
+    [styles.large]: size === 'large'
+  });
+
+  return (
+    <span className={className} id={id}>
+      {icon && <span className={styles.icon}>{icon}</span>}
+      <span>{label}</span>
+      {erasable && <a className={styles.tagRemoveButton} onClick={() => onRemoveTag(id)}/>}
+    </span>
+  );
+};
+
+Tag.propTypes = {
+  label: PropTypes.string.isRequired,
+  id: PropTypes.string.isRequired,
+  icon: PropTypes.element,
+  onRemoveTag: PropTypes.func,
+  erasable: PropTypes.bool,
+  size: PropTypes.oneOf(['small', 'large']),
+};
+
+Tag.defaultProps = {
+  onRemoveTag: () => {},
+  size: 'small',
+  erasable: true,
+};
+
+export default Tag;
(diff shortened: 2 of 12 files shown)


src/Tag/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+> A Tag component
+
+## Properties
+
+| propName | propType | defaultValue | isRequired | description |
+|----------|----------|--------------|------------|-------------|
+| label | string | - | + | The label which will be displayed on the Tag |
+| id | string | - | + | The id of the Tag |
+| erasable | bool | true | - | If the Tag is erasable then it will contain a small clickable X|
```
mastertheblaster: `erasable` doesn't match the `onRemoveTag`, perhaps you should rename to `removable` ?

src/Tag/README.md
```diff
@@ -0,0 +1,14 @@
...
+| onRemoveTag | func | - | - | Callback function when removing the Tag |
```
mastertheblaster: Perhaps it should be like:

- `removable: true | false`
- `onRemove: () => {}`

src/Tag/README.md
```diff
@@ -0,0 +1,14 @@
...
+| size | small or large | small | - | The height of the Tag |
+| icon | element | - | - | An optional icon to display as part of the Tag |
```
mastertheblaster: What about `thumb` ? Perhaps it describes more meaningfull?

src/Tag/README.md
```diff
@@ -0,0 +1,14 @@
+# Tag component
+
+> A Tag component
+
+## Properties
+
+| propName | propType | defaultValue | isRequired | description |
+|----------|----------|--------------|------------|-------------|
+| label | string | - | + | The label which will be displayed on the Tag |
```
mastertheblaster: Maybe `node` ? However not sure if such a flexibility is required.

src/Tag/Tag.js
```diff
@@ -0,0 +1,35 @@
...
+  label: PropTypes.string.isRequired,
+  id: PropTypes.string.isRequired,
+  icon: PropTypes.element,
+  onRemoveTag: PropTypes.func,
+  erasable: PropTypes.bool,
+  size: PropTypes.oneOf(['small', 'large']),
+};
+
+Tag.defaultProps = {
+  onRemoveTag: () => {},
```
mastertheblaster: As far as I know there is no need to specify default functions for `onClick` handler. React just handles that.

src/Tag/README.md
lbelinsk: sounds reasonable.. will change that

src/Tag/README.md
lbelinsk: sorry, could you explain?

lbelinsk: thumb instead of icon?

mastertheblaster: As far as I know other libraries, frameworks refer this component as Chip 
http://www.material-ui.com/#/components/chip

Personally for me `Icon` refers to some graphics like svg. But what if this `Tag` component would be used for displaying selected user list. So it would make sense to display a user thumb. So I am questioning if `icon` would be a good property name in such case.

src/Tag/README.md
mastertheblaster: In case of `node` you would be able to markup part of the label. For example
```jsx
<Tag label={<span><bold>Hello</bold>&nbsp;world!</span>}/>
```

mastertheblaster: Or perhaps the label should be render using `this.props.children` ?

```js
<Tag removable onRemove={() => {})>
  {tag}
</Tag>
``` 

?

src/Tag/Tag.js
lbelinsk: In some scenarios you might be right, but when you pass some parameters to this function then you will get error for passing argument to undefined function.
In line 15 you can see:

onClick={() => onRemoveTag(id)}/>}

so If this function is undefined then it is not going to work.
This is what happened when you try to remove a Tag with an undefined onRemoveTag (I removed the defaultProps just for this demonstration) 

screen shot 2017-01-09 at 9 24 18 pm

mastertheblaster: Good catch 👍 

```js
onClick={() => onRemoveTag && onRemoveTag(id)}
```

src/Tag/README.md
lbelinsk: I think for now we will stick with string only for the label value.. If such a require will come up in the future it will be pretty easy to refactor this

src/Tag/README.md
lbelinsk: @nirhart what do you think?

src/Tag/Tag.js
lbelinsk: Nice save :)
This could work.. both solutions are valid. I don't see any advantage in either one...

src/Tag/README.md
lbelinsk: Thumb might be a nicer fit. I will add this change

src/Tag/Tag.js
```diff
@@ -2,34 +2,34 @@ import React, {PropTypes} from 'react';
...
+      {thumb && <span className={styles.thumb}>{thumb}</span>}
       <span>{label}</span>
-      {erasable && <a className={styles.tagRemoveButton} onClick={() => onRemoveTag(id)}/>}
+      {removable && <a className={styles.tagRemoveButton} onClick={() => onRemove(id)}/>}
     </span>
   );
 };
 
 Tag.propTypes = {
   label: PropTypes.string.isRequired,
```
mastertheblaster: Just a warning. What if label is a user entered text. Can it wrap? Or perhaps it better to apply the `max-width`. And sorry for being paranoid.

lbelinsk: Please don't be sorry, I am really happy to hear everything that can improve the code :)
today it will wrap the text and will be a wide tag, (no newline).
I am not sure what will be better, because a long text will lead to an ugly tag either way. I can consult with Nir Kosover tomorrow abut this UX issue.

A max-width solution will be bad for some scenarios. for example, I am now refactoring the MultiSelect which uses the Tags. The tags are inserted to a container one after the other, and it will be ugly if their height will be different..

PR #143, opened 2017-01-10
Refactor dropdown and autocopmlete, added InputWithOptions and DropdownLayout

nirhart:
@3LOK please have a look, once it will be merged it will break you.
in here:
1) DropdownLayout component - is the base dialog that pops in dropdown/autocomplete/googleaddressautocomplete and etc..
2) InputWithOptions - this is the similar component the the current Autocomplete, it contains Input and DropdownLayout - its fully controlled. the main difference is in it api - the options must have an ID and you get onSelect when an option is clicked or onManuallyInput when the user input a text regarding his suggestions (entering text in the input and press enter or tab)..
3) Dropdown - extends InputWithOptions - "smart" Select component
4) Autocomplete - extends InputWithOptions - "smart" Autocomplete component


3LOK:
@nirhart I've added another example to the storybook. 

(1) See that blur/focus events are called twice for each event, that's not good.
(2) Unselectable does not work, that's not good.

PR #148, opened 2017-01-10
Radio group enhancement

lbelinsk:
Today the testkit concept scares people off so we end up with a lot of untested components.
After writing three testkits I noticed that there is a lot of code duplication among the different testkits.

Instead of requiring each component to have a testkit, we will have a test-common file which will capture all the similar code.

This will save a lot of time and effort from the components’ developers, but there must be some basic ground rules in order for this concept to work:

1. Each component must have component.driverFactory file which will contain the implementation of it’s driverFactory.

2. Each component must have an id attribute place on it’s most external layer.

3. In component.spec.js file we will need to import 
	-  componentDriverFactory from ‘./component.driverFactory' 
	-  {createDriverWrapper} from './../test-common';

	Then use them use them as follow:
	  const createDriver = createDriverWrapper(componentDriverFactory);

	That’s it. now every ‘it’ block will call createDriver:
		
```
import React from 'react';
import InputDriverFactory from ‘./Input.driverFactory';
import {createDriverWrapper} from './../utils';
import Input from ‘./Input’;

describe('Input', () => {

  const createDriver = createDriverWrapper(inputDriverFactory);

  it('should check that the unit text if passed', () => {
      const unit = '$';

      const driver = createDriver(<Input unit={unit}/>);
      expect(driver.getUnit()).toEqual(unit);
    });
```

**This PR affects only Input and ButtonSelection, but only in the tests that uses their driverFactory.**

### DIFF
--- a/src/RadioGroup/README.md
+++ b/src/RadioGroup/README.md
@@ -10,7 +10,7 @@
 | onChange | func | - | - | Callback function when user selects a different value |
 | disabledRadios | arrayOf(number, string) | [] | - | the values of the disabled radio buttons |
 | vAlign | top, center | top | - | Positioning of the radio bottom compared to the label |
-| direction | vertical or horizontal | horizontal | - | Display direction of the radios |
+| direction | vertical or horizontal | vertical | - | Display direction of the radios |
 | id | string | - | - | An identifier of the component |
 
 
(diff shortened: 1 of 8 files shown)


lbelinsk:
@itaipelles please notice that this change will affect Input and ButtonSelection in Guineapig tests.

PR #150, opened 2017-01-11
feature/input flex (don't merge yet)

mastertheblaster:
Completed the refactoring of the <Input/> using a flex model.
Double checked the Autocompletes, Dropdowns, DatePicker and Timepicker components.


3LOK:
@mastertheblaster Looks good - what's the difference between 'unit' and 'iconLeft'?

mastertheblaster:
@3LOK @nirhart `iconLeft` property was introduced by InterAl and was used inside a datepicker to display the calendar icon on the left. So currently I left it backwards compatible.

3LOK:
@mastertheblaster @nirhart ah, cool - so played around with it - looks good.

PR #153, opened 2017-01-11
Test kit template with data hook

lbelinsk:
After writing three testkits I noticed that there is a lot of code duplication among the different testkits.
Instead of requiring each component to have a testkit, we will have a test-common file which will capture all the similar code.

In the first step, the only component which is going to use this new testkit template is Input. That means that every project that uses the Input testkit in it's tests needs to perform the following simple change:

Instead of doing:

```import {inputTestkitFactory} from 'wix-style-react/dist/testkit'```

change the import to:

```import {inputTestkit} from 'wix-style-react/dist/testkit'```

Alternatively, If you are using enzyme, do:

```import {inputTestkit} from 'wix-style-react/dist/testkit/enzyme'```

### CODE REVIEW

File: src/WixComponent.js [hunk 1]
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import ReactDom from 'react-dom';
+
+class WixComponent extends React.Component {
+
+  constructor(params) {
+    super(params);
+    this.constructor.propTypes['data-hook'] = React.PropTypes.string;
```
lbelinsk: @nirhart I didn't understand why are you doing this here.. In any case there was a linting error because it required a proptypes validation which I added

PR #157, opened 2017-01-12
pass the event to input's on blur callback

itaipelles:
@nirhart please review 

@mastertheblaster 
time input expects to receive the event object in the on blur event.
see https://github.com/wix/wix-style-react/blob/master/src/TimeInput/TimeInput.js


mastertheblaster:
@itaipelles oops, sorry, thanks for the fix

PR #165, opened 2017-01-13
feature/text field chore

mastertheblaster:


src/Label/Label.js
```diff
@@ -21,3 +21,7 @@ Label.propTypes = {
   appearance: React.PropTypes.oneOf(['T1', 'T1.1', 'T1.2', 'T1.3', 'T1.4', 'T2', 'T2.1', 'T2.2', 'T2.3', 'T3', 'T3.1', 'T3.2', 'T3.3', 'T3.4', 'T4', 'T4.1', 'T4.2', 'T4.3', 'T5', 'T5.1']).isRequired,
   children: React.PropTypes.any
 };
+
+Label.defaultProps = {
+  appearance: 'T1.1'
+};
```
nirhart: I should be either default or required, not both, pick one

PR #166, opened 2017-01-13
Feature/error message component

argshook:
hi, this is a generic layout component with title, description and two CTAs. All are optional.

We (promote) need this to display error message in fullscreen and modals. However, since this just renders given components, it could be used for a lot more cases, hence the generic `EndorseContentLayout` name.

did i miss something? if so, let me know.

@koriguy


nirhart:
There are build fails

argshook:
i believe the build failed because of updates in master which you merged? or was it my code? or snyk.io? not familiar with it at all.

PR #178, opened 2017-01-16
Date picker rtl

InterAl:


src/TimeInput/README.md
```diff
@@ -9,3 +9,4 @@
 | defaultValue | moment object | now | false | The control's starting time |
 | onChange | func | - | - | Called upon blur |
 | rtl | bool | - | - | Display in RTL |
+| disableAmPm | bool | - | - | 24h mode |
```
mastertheblaster: Perhaps the property should be smth like:

```js
mode: PropTypes.oneOf(['12h', '24h'])
```

In that case it would be more meaningful. Cause now, `disableAmPm` doesn't mean it will use 24h mode.

stories/DatePicker/ExampleControlledRtl.js
```diff
@@ -0,0 +1,26 @@
...
+    };
+  }
+
+  render() {
+    return (
+      <DatePicker
+        value={this.state.value}
+        onChange={value => this.setState({value})}
+        dateFormat="DD/MM/YYYY"
+        rtl={true}
```
mastertheblaster: FYI, boolean props don't require explicit value.

InterAl: Yeah, I know.

src/TimeInput/README.md
InterAl: Yes, your suggestion is probably better.

PR #184, opened 2017-01-17
Adding ID to Dropdown element if provided, for enzyme testing

giladartzi:
When creating a `<Dropdown/>` component, the wrapping DOM element does not get any ID even if provided as a prop, and thus the enzyme test kit does not work.

The test kit (`Dropdown.enzyme.js`) requires the id: ```const component = wrapper.find(`#${id}`);```


nirhart:
Merged, however, tests are undergo refactoring where you will use data-hook for enzyme and not id, so please be aware of it

PR #190, opened 2017-01-18
Dropdown layout footer header

lbelinsk:
1. Fixes from the previous branch (without MultiSelect).
2. DropdownLayout fixedFooter + fixedHeader.
3. Migration of InputWithOptions to use the testkit template.
4. Migration of DropdownLayout to use the testkit template.
5. Migration of AutoComplete to use the testkit template.
6. Migration of Dropdown to use the testkit template.
7. Migration of Tag to use the testkit template.
8. Added MultiSelect new version with testkit.
9. Changed the usage of Tag. instead of ```<Tag label="something"/>```  use  ```<Tag>something</Tag>```
10. Docs update.


src/DropdownLayout/DropdownLayout.driver.js
```diff
@@ -2,24 +2,26 @@ import React from 'react';
 import DropdownLayout from '../DropdownLayout';
 import ReactTestUtils from 'react-addons-test-utils';
 import ReactDOM from 'react-dom';
-import $ from 'jquery';
+import find from 'lodash.find';
```
mastertheblaster: Perhaps not needed?

```js
[1, 2, 3].find(i => i === 2)
```

Does the same job

lbelinsk: I did a mistake while merging.. I will take a look in a second

src/DropdownLayout/DropdownLayout.scss
```diff
@@ -1,124 +1,137 @@
 @import "../common.scss";
 
 $option_height: 35px;
+$content_container_height: 260px;
 
 :export {
-    option_height: $option_height;
+  option_height: $option_height;
+  content_container_height: $content_container_height;
```
mastertheblaster: @argshook I guess this is how you can export some variables? and use in a 3rd party project.

src/DropdownLayout/DropdownLayout.scss
```diff
@@ -1,124 +1,137 @@
...
+  opacity: 0;
+  overflow: auto;
+  padding: 0;
+  transition: opacity 0.2s ease;
+  visibility: hidden;
+  width: 100%;
+  z-index: 6;
+  position:absolute;
+
+  -webkit-touch-callout: none;
```
mastertheblaster: Guess you don't need them. This should be handled by css prefixer.

src/Input/Input.js
```diff
@@ -28,6 +25,10 @@ class Input extends WixComponent {
 
   state = {
     focus: false
+  };
+
+  componentDidMount() {
+    this.props.autoFocus && this._onFocus();
```
mastertheblaster: Shouldn't this be like:
```js
this.props.autoFocus && this.focus();
```

nirhart: this.focus() don't call onFocus prop, we need that call so the user will know to show the dropdownLayout if exists

src/DropdownLayout/DropdownLayout.driver.js
lbelinsk: this wouldn't work since surprisingly options.childNodes is not an array. it is an object

lbelinsk: I did console.log(typeof options.childNodes) and it showed object

PR #191, opened 2017-01-18
fix(dropdown): to correctly position its place

mastertheblaster:


mastertheblaster:
screen shot 2017-01-18 at 12 30 48

PR #211, opened 2017-01-23
Text area

shaulgo:


src/InputArea/InputAreaSuffix.js
```diff
@@ -0,0 +1,30 @@
+import React, {PropTypes} from 'react';
+
+import SvgExclamation from '../svg/Exclamation.js';
```
shaulgo: the exclamation svg is built dynamically, allowing for different dimensions and color. Although it seems like it can be added to the icons assets and then used from there.

nirhart: The icons are backed by svgs as well, they are dynamically too

PR #229, opened 2017-01-26
fix(input): cropping the bottom part of y letter

mastertheblaster:


mastertheblaster:
Tested on Chrome, Safari and FF. Looks good.

PR #241, opened 2017-01-30
added validation for Input of type number

lbelinsk:


src/Input/Input.js
```diff
@@ -171,6 +171,14 @@ class Input extends Component {
       this.props.onEscapePressed && this.props.onEscapePressed();
     }
   }
+
+  _onChange(e) {
+    if (this.props.type === 'number' && !(/^\d+$/.test(e.target.value))) {
+      return;
```
mastertheblaster: @lbelinsk What about empty values? Guess they should be allowed for number input too?

lbelinsk: right.. i will change the + to *

lbelinsk: Actually it doesn't matter... it only controls if onchange will get triggered.. if nothing is typed it won't even try to trigger it..

mastertheblaster: Let's say i have a number inside the input `1` and i hit backspace button. Will it be ok?

lbelinsk: nice catch... for some reason it is working, but I will change + to * since it is more correct

PR #243, opened 2017-01-30
fixed bad behaviour when pressing select while hovered on disabled o…

lbelinsk:


src/DropdownLayout/DropdownLayout.js
```diff
@@ -108,8 +108,7 @@ class DropdownLayout extends WixComponent {
       }
 
       case 'Tab': {
-        this._onSelect(this.state.hovered);
-        return true;
+        return this._onSelect(this.state.hovered);
```
mastertheblaster: @lbelinsk Shouldn't other handlers behave the same?
For example I see that `Escape` is not returning anything.
Also `Enter` is interesting, cause in some case it returns in another case not.

lbelinsk: The reason why we do this for tab is not to do preventDefault behaviour of tab..
All the other selectors return true / false after the break..

PR #250, opened 2017-01-31
feat(dataTable): add dynamicRowClass

itaibenda:
Allow dynamic row styling according to row content


src/DataTable/DataTable.spec.js
```diff
@@ -68,6 +68,16 @@ describe('Table', () => {
     expect(driver.getRowsWithClassCount(defaultProps.rowClass)).toBe(defaultProps.data.length);
   });
 
+  const getClass = rowData => {
+    return rowData.a.replace(/[\s]+/g, '-');
```
mastertheblaster: @itaibenda What is `a` prop inside `rowData`? Is this ok?

src/DataTable/DataTable.spec.js
```diff
@@ -68,6 +68,16 @@ describe('Table', () => {
...
+  };
+
+  it('should assign a dynamic class to rows', () => {
+    const driver = createDriver(<DataTable {...defaultProps} dynamicRowClass={getClass}/>);
+    expect(driver.getRowsWithClassCount('value-1')).toBe(1);
```
mastertheblaster: @itaibenda Also not sure how this will help you. Cause we use CSS modules.
And passing the custom class from another component wont' be easy.
IMHO you won't be able to write selectors like:

```scss
.value-1 {
  .data-table-row-cell {
     padding: 10px;
  }
}
```

The only thing you will be able to change is row settings.

src/DataTable/DataTable.spec.js
itaibenda: it's a data property from the default data of the UTs

src/DataTable/DataTable.spec.js
itaibenda: I want to apply dynamic style, not pass selectors. We have the rowDataHook property for selectors (which isn't dynamic)

src/DataTable/DataTable.js
```diff
@@ -90,22 +90,27 @@ class DataTable extends WixCopmonent {
...
-        this.props.onRowClick(rowData, rowNum);
+        onRowClick(rowData, rowNum);
       };
       rowClasses.push(s.clickableDataRow);
     }
 
-    if (this.props.rowDataHook) {
-      optionalRowProps['data-hook'] = this.props.rowDataHook;
+    if (rowDataHook) {
+      optionalRowProps['data-hook'] = rowDataHook;
```
mastertheblaster: Please correct me if I am wrong, but all the rows will get the same `data-hook`. 
So how will you be able to change the background of a first cell, using the dynamic row class?
Also keep in mind that I can have a nested table inside a cell.

itaibenda: I don't think it would be a good idea to style the row according to a data-hook. The change here was merely syntactic (extract the prop from this.props).

mastertheblaster: Can you enhance the story book, by showing the example. And i will merge it.

itaibenda: sure

PR #254, opened 2017-02-02
add new location to toast - inplace

ariel-wine:


src/Toast/Toast.js
```diff
@@ -17,7 +17,7 @@ class Toast extends React.Component {
     const {
       show,
       type,
-      position,
+      location,
```
mastertheblaster: @ariel-wine is this good? You may break someones application. Usually we add code for backwards compatibility and print out the warning saying that it will be removed in X days.

https://github.com/wix/wix-style-react/blob/master/src/Input/Input.js#L69

src/Toast/testkit/Toast.js
```diff
@@ -26,7 +26,8 @@ const toastDriverFactory = ({component}) => {
     clickOnClose: () => ReactTestUtils.Simulate.click($component.find(byDataHook('toast-close'))[0]),
     hasTheme: theme => $component.hasClass(theme),
     toastExists: () => $component.parent().find(byDataHook('toast')).length > 0,
-    getTopProperty: () => styleStringToObj($component.attr('style')).top
+    getTopProperty: () => styleStringToObj($component.attr('style')).top,
+    hasLocation: location => $component.attr('class').indexOf(location) >= 0
```
mastertheblaster: @ariel-wine It's better to do like:
```js
$component.get(0).classList.contains(location)
```

src/Toast/Toast.js
nirhart: @ariel-wine ?

ariel-wine: on the contrary,
If you check the documentation, you will see that it says location - backwards computability is kept due to the fact that the example in the documentation works in a field that only has one value.
If we actually start using the 'position' field, it will break all the code that used the instructions.

PR #259, opened 2017-02-02
fix - do not call onSelect when selecting a selected option

### LINKED ISSUE #258: Dropdown doesn't close when clicking on a selected option

lbelinsk:
close #258

### DIFF
--- a/src/DropdownLayout/DropdownLayout.js
+++ b/src/DropdownLayout/DropdownLayout.js
@@ -48,8 +48,9 @@ class DropdownLayout extends WixComponent {
     const {options, onSelect, selectedId} = this.props;
     if (index >= 0 && index < options.length) {
       const newSelectedId = options[index].id;
-      if (newSelectedId !== selectedId && onSelect) {
-        onSelect(options[index]);
+      const isSelectedOptionSelected = newSelectedId === selectedId;
+      if (onSelect) {
+        onSelect(options[index], isSelectedOptionSelected);
         return true;
       }
     }
--- a/src/DropdownLayout/DropdownLayout.js
+++ b/src/DropdownLayout/DropdownLayout.js
@@ -46,15 +46,8 @@ class DropdownLayout extends WixComponent {
 
   _onSelect(index) {
     const {options, onSelect, selectedId} = this.props;
-    if (index >= 0 && index < options.length) {
-      const newSelectedId = options[index].id;
-      const isSelectedOptionSelected = newSelectedId === selectedId;
-      if (onSelect) {
-        onSelect(options[index], isSelectedOptionSelected);
-        return true;
-      }
-    }
-    return false;
+    options[index] && onSelect && onSelect(options[index], options[index].id === selectedId);
+    return !!onSelect && options[index];
   }
 
   _onMouseEnter(index) {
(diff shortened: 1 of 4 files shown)


src/DropdownLayout/DropdownLayout.js
```diff
@@ -48,8 +48,9 @@ class DropdownLayout extends WixComponent {
     const {options, onSelect, selectedId} = this.props;
     if (index >= 0 && index < options.length) {
       const newSelectedId = options[index].id;
-      if (newSelectedId !== selectedId && onSelect) {
-        onSelect(options[index]);
+      const isSelectedOptionSelected = newSelectedId === selectedId;
```
mastertheblaster: If's are good, but not guess to complicated, what if I would suggest you:

1) Exactly the same behavior but with less code:
```jsx
_onSelect(index) {
  const {options, onSelect, selectedId} = this.props;
  if (onSelect && options[index]) {
    onSelect(options[index], options[index].id === selectedId);
    return true;
  }
  return false;
}
```

2) Not sure you need to check `options` and `index`, cause otherwise the event wont be fired, so...
```jsx
_onSelect(index) {
  const {options, onSelect, selectedId} = this.props;
  onSelect && onSelect(options[index], options[index].id === selectedId);
  return !!onSelect;
}
```

src/InputWithOptions/InputWithOptions.js
```diff
@@ -106,15 +106,17 @@ class InputWithOptions extends WixComponent {
...
+  _onSelect(option, isSelectedOption) {
     this.showOptions();
     const {onSelect} = this.props;
 
     if (this.closeOnSelect()) {
       this.hideOptions();
     }
 
-    if (onSelect) {
+    if (isSelectedOption) {
```
mastertheblaster: Not sure if this is good. Cause, `DropdownLayout` is firing `onSelect` even if the same option is selected and `InputWithOptions` not. Guess we should be consistent.

```jsx
isSelectedOption && this.setState({ showOptions: false });
onSelect && onSelect(option);
```

What do you say ?

src/DropdownLayout/DropdownLayout.js
lbelinsk: I will try your second suggestion.. It is shorter but might be a bit less readable.. but it is nice..

src/InputWithOptions/InputWithOptions.js
lbelinsk: There is a slight issue that I rather not call onSelect at all when we select the selected option.. In your suggestion we will call onSelect even if we have selected the selected option. It is a minor issue but I think it is better not to trigger the onSelect  on the selected value at all.. I even added a test that checks that the onSelect callback isn't triggered..

src/DropdownLayout/DropdownLayout.js
lbelinsk: Actually I tried this now and some tests fail..
we must add to the last line:

```return !!onSelect && options[index];```

lbelinsk:
@mastertheblaster thanks for the review!

I have committed a refactor after your review. please merge these changes if you approve. thanks 👍

PR #261, opened 2017-02-03
update buttons to latest design and create storybook generators

vidmantas0010:


src/Button/Button.js
```diff
@@ -37,7 +37,30 @@ Button.defaultProps = {
...
+    'emptygreen',
+    'emptybluesecondary',
+    'emptyblue',
+    'emptypurple',
+    'fullblue',
+    'transparentblue',
+    'whiteblue',
+    'whiteblueprimary',
+    'whitebluesecondary',
+    'close-standard',
```
mastertheblaster: Naming convention is not consistent. Is this by design?

stories/Button/ButtonClose.js
```diff
@@ -0,0 +1,80 @@
...
+  state = {
+    disabled: false,
+    theme: 'close-standard',
+    iconOnly: true,
+    height: 'medium'
+  };
+
+  render() {
+    return (
+      <from className={styles.form}>
```
mastertheblaster: Is this a new tag ?

src/Button/Button.js
vidmantas0010: I believe it should be with dashes, but we can't refactor old ones, because that would break existing projects.

PR #267, opened 2017-02-06
fix(tooltip): don't use state for storing tooltip placement coordinates

mastertheblaster:


mastertheblaster:
@nirhart Changed the code, how the tooltip placement is done. I don't store it inside a state anymore. So there should be no more magic issues.

PR #268, opened 2017-02-07
quick temporary fix for icon positioning in buttons

jakutis:


jakutis:
me and Domas experimented and decided on this

nirhart:
Did you verified that InputSuffix still looks good? it uses the Icon as well

PR #288, opened 2017-02-19
DatePicker testkit and tests

InterAl:


lbelinsk:
@nirhart This isn't the final testkit, right?

PR #298, opened 2017-02-22
feat(tooltip): allow appending tooltip to a parent

mastertheblaster:
to avoid scroll and animation issues of compounding elements


argshook:
👍

PR #303, opened 2017-02-22
Properly handle setProps in drivers

shlomitc:
Allows to extend the current rendered component with new props and render all of the existing children.
rename the current `component` to be `element` to notate that it's a DOM element and not a react component.


nirhart:
build failed

PR #313, opened 2017-02-27
Range input3

boriskaWix:
Added Range component (3.7) - Added Component + Tests + half way in storybook
Field and Selection Component (3.3) - Added Component + Tests


nirhart:
The build is failed

src/Input/Input.js
```diff
@@ -66,7 +67,13 @@ class Input extends Component {
...
       [styles.hasFocus]: forceFocus || this.state.focus,
       [styles.roundInput]: roundInput
-    });
+    };
+
+    if (customClass) {
+      classes[customClass] = true;
+    }
+
+    classes = classNames(classes);
```
nirhart: please do not add a custom class, if a component can get any class from outside its not safe anymore from inner changes

dorraba: So instead would you like me to add a props like "noLeftBorderRadius" and "noRightBorderRadius"? I understand the point where you don't want wrapping components to be able to change css, but this causes that any change we want to make to the input will require us to change the input itself.

PR #340, opened 2017-03-13
feat(SideMenu): add <SideMenu/> and related components

argshook:
migrated from promote-seo project


nirhart:
Thank you for that

PR #341, opened 2017-03-14
Mar12 testkits

assafswix:


src/MessageBox/index.js
```diff
@@ -1,7 +1,7 @@
...
-import MessageBoxLayout2 from './MessageBoxLayout2';
-import HeaderLayout1 from './HeaderLayout1';
-import FooterLayout1 from './FooterLayout1';
+import MessageBoxMarketerialLayout from './MessageBoxMarketerialLayout';
+import MessageBoxFunctionalLayout from './MessageBoxFunctionalLayout';
+import HeaderLayout from './HeaderLayout';
+import FooterLayout from './FooterLayout';
 
-export {MessageBoxLayout1, HeaderLayout1, MessageBoxLayout2, FooterLayout1};
+export {MessageBoxMarketerialLayout, HeaderLayout, MessageBoxFunctionalLayout, MessageBoxFunctionalLayout as MessageBoxLayout2, FooterLayout, FooterLayout as FooterLayout1};
```
nirhart: What about HeaderLayout1,, MessageBoxLayout1?

src/index.js
```diff
@@ -8,7 +8,7 @@ export {default as InputWithOptions} from './InputWithOptions';
 export {default as Checkbox} from './Checkbox';
 export {default as ToggleSwitch} from './ToggleSwitch';
 export {default as Modal} from './Modal';
-export {MessageBoxLayout1, MessageBoxLayout2, HeaderLayout1, FooterLayout1};
+export {MessageBoxMarketerialLayout, MessageBoxFunctionalLayout, MessageBoxFunctionalLayout as MessageBoxLayout2, HeaderLayout, FooterLayout};
 export {default as RadioGroup} from './RadioGroup';
```
nirhart: What about HeaderLayout1, FooterLayout1, MessageBoxLayout1?

src/Button/Button.js
```diff
@@ -33,12 +33,12 @@ class Button extends WixComponent {
   }
 
   render() {
-    const {disabled, onClick, children, type} = this.props;
+    const {disabled, onClick, children, type, dataHook} = this.props;
```
nirhart: Since Button extends WixComponent, it already have a dataHook property on its root, with a prop name dataHook, I'm not sure what will happen but it looks like its going to have 2 dataHook with the same name on two different elements.

 I thought you wanted to add a data-hook on top of the regular one, something like
+        <button onClick={onClick} disabled={disabled} type={type} data-hook="inner-button-hook">

and then get the button by it, cause even if you don't see it in the code, the render will actually going to be
       <ButtonLayout data-hook={dataHook} {...buttonLayoutProps}>
so in order to find the component itself you don't need to do anything, you have it, inside your driver you can look for the <button> with querySelector or with your hard-coded data-hook as I suggested above.

If you want we can talk about it over hangouts

PR #344, opened 2017-03-15
Mar12 testkits

assafswix:


mastertheblaster:
@nirhart this will break our application.
We already use old namings `Layout1` and `Layout2`.

Maybe we can currently just leave 2 namings, to support old applications and announce the grace period during which we need to migrate.

nirhart:
@mastertheblaster you are right, messageboxlayout1/2 are backwards compatible but headerlayout and footerlayout are not, @assafswix please take a look at it, also, this pr have some test failing

PR #358, opened 2017-03-21
fix(InfiniteScroll): remove errors caused by the unused element prop

Jefoss:
Removed the unused element prop (which used to default to 'div') passing just the ref prop to the created div element. Theres not point in keeping that prop just in case someone is gonna use it someday.


nirhart:
@Jefoss  We can't remove those props since someone might uses them. what we should do is deprecate them first with a console warning, and after a month or so, remove them completely

PR #375, opened 2017-03-27
e2e driver and tests for toggleSwitch

mariusju:


src/ToggleSwitch/ToggleSwitch.js
```diff
@@ -19,7 +19,7 @@ class ToggleSwitch extends React.Component {
     });
 
     return (
-      <div className={rootClassName}>
+      <div className={rootClassName} data-hook={this.props.dataHook}>
```
nirhart: Instead of adding a dataHook property, just extend WixComponent, and not React.Component

PR #376, opened 2017-03-29
feat(modal): expose portalClassName prop

mastertheblaster:


nirhart:
@mastertheblaster Can you please explain why would you want to do that? we try to avoid allowing setting class name from outside the comonent

mastertheblaster:
@nirhart https://github.com/wix/wix-style-react/blob/master/src/Modal/Modal.js#L29

We need to disable the scroll for some modals.
So I was thinking about allowing to specify the root class so that I could override the scroll styles.

So regarding your comment: do you suggest creating a property `scrollable={true || false}` instead ?

nirhart:
@mastertheblaster Thats right, by supporting a specific prop we can know exactly what we support and what we are not supporting, and allow us to change the implementation without breaking users who counted on a specific class hierarchy

mastertheblaster:
@nirhart ok, will create another PR.

PR #377, opened 2017-03-29
feat(<Modal/>): add new prop 'scrollable'

mastertheblaster:


mastertheblaster:
@nirhart added the `scrollable` prop

mastertheblaster:
@ArturasCH please follow this one, previous PR was closed.

src/Modal/Modal.js
```diff
@@ -73,7 +77,8 @@ Modal.propTypes = {
   onAfterOpen: React.PropTypes.func,
   horizontalPosition: React.PropTypes.oneOf(Object.keys(positions)),
   verticalPosition: React.PropTypes.oneOf(Object.keys(positions)),
-  closeTimeoutMS: React.PropTypes.number
+  closeTimeoutMS: React.PropTypes.number,
+  scrollable: React.PropTypes.bool
 };
 
 Modal.defaultProps = {
```
nirhart: I think its better to add scrollable true as a default prop, you checked (props.scrollable === false) and it worked since scrollable is undefined, if we would write there (props.scrollable === true) it will be false as well

mastertheblaster: Fixed and pushed

mastertheblaster:
@nirhart please double check updated version. Thanks.

PR #378, opened 2017-03-29
Stretch Card inside Grid (ignore branch name)

amimagid:
Added ability to stretch cards to the 100 percent of the container, and, added support for auto adjustable stretched cards

### DIFF
--- a/src/Grid/Grid.scss
+++ b/src/Grid/Grid.scss
@@ -20,7 +20,9 @@ $grid-row-margin-bottom: 30px;
 		margin-bottom: $grid-row-margin-bottom;
 	}
 }
-
+.flex-container {
+	display: flex;
+}
 .wixContainer {
 	@extend .container;
 
--- a/stories/GridWithCardLayout/ExampleGrid.js
+++ b/stories/GridWithCardLayout/ExampleGrid.js
@@ -28,19 +28,21 @@ export default () =>
   <div data-hook="card-example" className={styles.exampleContainer}>
     <Container>
       <AutoAdjustedRow>
-        <Card>
+        <Card stretchVertically={true}>
           <Card.ButtonHeader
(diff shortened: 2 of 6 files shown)


nirhart:
I think it should be on the row, like .row-eq-height in [here](http://getbootstrap.com.vn/examples/equal-height-columns/)

nirhart:
@amimagid ?

amimagid:
@nirhart - Sorry, had no time to look into it, but I think that this solution is only for Bootstrap's rows, while what I suggested is for any kind of container. If the intention was for Card to be only inside a row then I agree we should stick to the Bootstrap guidelines and classes. What say you?

src/Grid/Grid.scss
```diff
@@ -20,7 +20,9 @@ $grid-row-margin-bottom: 30px;
 		margin-bottom: $grid-row-margin-bottom;
 	}
 }
-
+.flex-container {
+	display: flex;
```
Faradey27: wrong number of spaces?

stories/GridWithCardLayout/ExampleGrid.js
```diff
@@ -28,19 +28,21 @@ export default () =>
   <div data-hook="card-example" className={styles.exampleContainer}>
     <Container>
       <AutoAdjustedRow>
-        <Card>
+        <Card stretchVertically={true}>
```
Faradey27: why not just <Card stretchVertically> ?

nirhart: just what?

Faradey27: sorry) forgot quotes for code block

`<Card stretchVertically>`
...

nirhart: ohh right, sure. @amimagid

PR #379, opened 2017-03-30
Input theming

mariusju:
Theming architecture POC:

The goal of this poc was to separate business logic of input component from the theme and everything related/depending on theme.  For that we created higher order component (HOC) utilising inheritance inversion pattern (https://medium.com/@franleplant/react-higher-order-components-in-depth-cf9032ee6c3e). 

Theme HOC takes a bare input component and wraps it with additional html, which is conditionally rendered based on theme.

During the refactor we encountered another problem. Sometimes the theme-depending logic is deep in the component and can not be just wrapped around it. If that is the case (which was with suffixes of input component), these inner components should be extracted to separate components and wrapped with their own themeHOCs. 

To sum up, these are main constraints we applied:

- The bare input component should not know anything about theme. 
- The theme HOC should not have any logic which is not directly depending on theme
- If theme-depending logic is deep inside the component, that part should be extracted to a separate component with its on themeHOC


src/Input/Input.js
```diff
@@ -327,4 +222,52 @@ Input.propTypes = {
...
+        [styles.rtl]: !!rtl,
+        [styles.disabled]: disabled,
+        [styles.hasError]: !!error,
+        [styles.hasHover]: forceHover,
+        [styles.hasFocus]: forceFocus || this.state.focus,
+        [styles.roundInput]: roundInput,
+        [styles.hasValue]: this.input && !!this.input.value,
+        [noRightBorderRadius]: noRightBorderRadius,
+        [noLeftBorderRadius]: noLeftBorderRadius,
+      };
```
mariusju: we found no use for these two lines:
[noRightBorderRadius]: noRightBorderRadius,
[noLeftBorderRadius]: noLeftBorderRadius,

even more, they do not apply scoped css classes. They add any global css class (not necessarily related to border radius) which is passed to the component under the name noRightBorderRadius or noLeftBorderRadius. Is it really what we need here?

nirhart: You right, its not what we want, RangeInputWithLabelComposite is using them and it keeps the css classes in itself, they probably should be on the input

PR #384, opened 2017-04-03
google address input now supports state

MosheMats:
city has fallback to sublocality in case locality is missing


3LOK:
@MosheMats Thanks. One comment - can you please do something like:
_.omitBy(result, _.isUndefined)

So the result will be exactly like what it was before?

Thanks!

PR #394, opened 2017-04-09
Removing unnecessary direct lodash dep

aryelu:


sdoomz:
I found that plain `lodash` package doesn't exist in `dependencies` at all. 
What do you think about adding `lodash.omitBy` to dependencies and importing methods as in other places?

```
import includes from 'lodash.find';
import omitBy from 'lodash.omitBy';
import isUndefined from 'lodash.isundefined';
```

aryelu:
@sdoomz updated

nirhart:
This is the worst way to import it https://medium.com/@ericchan/best-smallest-way-to-import-install-lodash-6622802dbeb, we should actually just use lodash-es and then we can import like this:
import {includes, omitBy, isUndefined} from 'lodash-es'

aryelu:
jest needed some config http://stackoverflow.com/a/42353679

aryelu:
We need to also transpiling of `node_modules/lodash-es` directory, yoshi ignores it all

roymiloh:
@aryelu [externalUnprocessedModules](https://github.com/wix/yoshi#yoshiexternalunprocessedmodules) should work for you (both in consumer project and wix-style-react itself, where we run redundant webpack step, but still), but generally we might need to find a better solution.

nirhart:
We will do something more general regarding moving to lodash-es

PR #406, opened 2017-04-19
[checkbox] align label appearance according to design system

sdoomz:
Label for checkbox component always uses color scheme: 

```
font-family: HelveticaNeueW01-45;
font-size: 16px;
```
So I think it is reasonable to set this value by default.

TBD: 
1. do we want to have this property as part of public API of `Checkbox` component? 
2. How to deal with `disabled` state? In zeplin text-color of `<Checkbox disabled/>` is `#cbd3dc;` in the other hand we don't have typography with this value, can we use `typography.t1_4` or not?


nirhart:
Tests are failing

sdoomz:
@nirhart 
hmm, so that means that we need to regenerate snapshots or make this prop optional and set value "T1.1" as default?

PR #407, opened 2017-04-19
Feature/text component

argshook:
a `<Text/>` component to be used for typography.

currently there is no one proper way to use wix typography. This component aims to solve it.

it's mainly to avoid `import {h1} from 'wix-style-react/src/dist/Typography.scss'` and instead do `<Text appearance="H0">Hello</Text>`.

depending on whether `appearance` is of type `H0` or `T1` or similar, `<Text/>` will either render `<h1>` - `<h5>`  or default to `<span>`.


argshook:
if this is merged, i'd like to go over [typography examples](https://wix.github.io/wix-style-react/?selectedKind=Common&selectedStory=Typography&full=0&down=0&left=1&panelRight=0) and use `<Text/>` in there instead of importing css classnames. @nirhart wdyt?

nirhart:
Its great, including the typography example.
The driver is not like we do drivers, the test is missing enzyme sanity test (see Label for example) and a protractor driver + test is missing

stories/Text/TextAreaTemplate.js
```diff
@@ -0,0 +1,37 @@
+import React, {PropTypes, Component} from 'react';
```
Faradey27: let use in future propTypes from separate lib, to be able to move to react 16 in future

argshook: this is not related to my changes but yes, we need to use `from 'prop-types'`, we get many warnings now. same for `React.createClass`

argshook: wait wat it is in my changes, though shouldn't be at all! maybe i somehow moved file accidentally. i'll clean it up, thanks for noticing.

PR #423, opened 2017-04-25
added sticky type for Notification

lbelinsk:


lbelinsk:
1. Added sticky type which makes the position fixed.
2. fixed linting errors.
3. Added test for the new type.
4. Updated story book example + Readme.

PR #439, opened 2017-05-07
Add grid classname

liorur:


nirhart:
We are not adding claasName as a prop
Since we can't know how it will be used and we can't support for it
Also, users should use only the accepted UX and not modify their own classes

PR #454, opened 2017-05-13
Update Header.scss

andrey-ushakov:
withDivider instead of with-divider

### DIFF
--- a/src/Card/Header/Header.scss
+++ b/src/Card/Header/Header.scss
@@ -6,7 +6,7 @@ $header-padding: 27px 30px 26px;
 
   padding: $header-padding;
 
-  &.with-divider {
+  &.withDivider {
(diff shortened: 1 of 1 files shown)


nirhart:
@andrey-ushakov why?

andrey-ushakov:
@nirhart In the file `src/Card/Header/Header.js` on the line line 25 `withDivider` class used (and not `with-divider)`:
`[styles.withDivider]: !withoutDivider,`
Otherwise `withDivider` class is not applied to the Card's Header. It means there is no divider between Card's Header and Body independently of the component's input parameter `withoutDivider: bool`

nirhart:
since we have css-loader?camelCase we can write it like this, as you can see it is working in the storybook

PR #455, opened 2017-05-13
inputWrapper instead of input-wrapper

andrey-ushakov:


nirhart:
@andrey-ushakov why?

andrey-ushakov:
@nirhart In the file `src/Input/Input.js` on the line line 96 `inputWrapper` class used (not `input-wrapper)`:
`return (<div className={styles.inputWrapper}>`
Otherwise some important styles are not applied to the div element, especially `width: 100%`.

nirhart:
see #454

PR #458, opened 2017-05-14
add aria-label to input

liorur:


src/Input/Input.driver.js
```diff
@@ -20,6 +20,7 @@ const inputDriverFactory = ({element, wrapper, component}) => {
     getReadOnly: () => input.readOnly,
     getTextOverflow: () => input.style['text-overflow'],
     getType: () => input.type,
+    getProp: propName => input.getAttribute(propName),
```
shlomitc: change to specific attribute (getAriaLabel, getAraiaControls)

src/Input/Input.js
```diff
@@ -88,7 +90,8 @@ class Input extends Component {
         onKeyUp={onKeyUp}
         readOnly={readOnly}
         type={type}
-
+        {...(ariaLabel && {'aria-label': ariaLabel})}
```
shlomitc: Extract to an aria object and spread here

src/Input/InputSuffix.js
```diff
@@ -48,7 +48,7 @@ const InputSuffix = ({theme, errorMessage, error, disabled, help, helpMessage, o
     },
     {
       component: () =>
-        <div onClick={onClear} className={styles.clearButton}>
+        <div onClick={onClear} className={styles.clearButton} tabIndex={0}>
```
shlomitc: remove from the PR

nirhart:
@liorur please see @shlomitc reviews

src/Input/InputSuffix.js
Faradey27: @shlomitc  why tabIndex should be removed? can it be optional and passed via props?

shlomitc: It's an OK addition, just not in this PR. should be a separate one

Faradey27: thank you, understood

PR #460, opened 2017-05-15
add dynamic width to input

nirhart:


src/MultiSelect/InputWithTags.js
```diff
@@ -31,10 +33,18 @@ class InputWithTags extends React.Component {
 
         {tags.map(({label, ...rest}) => <Tag key={rest.id} onRemove={onRemoveTag} {...rest}>{label}</Tag>)}
         <span className={styles.input} data-hook="inner-input-with-tags">
+          <div className={styles.hiddenDiv}>
```
mastertheblaster: @nirhart `div` inside `span` is not good. It should be otherwise.

mastertheblaster: @nirhart The div `classname` IMHO should be dynamic, cause the input itself can be small/medium/large.

src/MultiSelect/InputWithTags.js
```diff
@@ -31,10 +33,18 @@ class InputWithTags extends React.Component {
...
+            {this.state.inputValue}
+          </div>
           <Input
             ref={input => this.input = input}
             placeholder={tags.length === 0 ? placeholder : ''}
             {...desiredProps}
+            width="100px"
```
mastertheblaster: @nirhart This should be 100% ?

src/MultiSelect/InputWithTags.scss
```diff
@@ -1,7 +1,15 @@
 @import '../common';
 
 .input {
-  width: 100px;
+  min-width: 100px;
+}
+
+.hiddenDiv {
+  @include FontRoman();
+  font-size: 16px;
```
mastertheblaster: @nirhart This should depend on the input size? Or input size is frozen (it's always medium?)

mastertheblaster:
@nirhart The UX issue appears when typing multiple spaces.

Faradey27:
@nirhart maybe better just to use some lib for dynamic input width?
as https://github.com/JedWatson/react-input-autosize or something like this, in such way we will have more general and stable solution

src/MultiSelect/InputWithTags.js
nirhart: I agree, however the div is invisible so I don't see any problem with that in this case, what do you think?

nirhart: medium and large act the same, I added a fix for small

src/MultiSelect/InputWithTags.js
nirhart: no, it should be 100px, its the initial input size, I don't want it to fit the whole container

src/MultiSelect/InputWithTags.scss
nirhart: you are right, I added a fix for it

nirhart:
@Faradey27 I really don't want to do it, we already have an Input of our own, that contains our UX demands properties and our accessibility, I think we should reuse our code in that case

nirhart:
@mastertheblaster fixed the spaces issue

mastertheblaster:
@nirhart seems good

PR #471, opened 2017-05-18
add id to tooltip content

liorur:


nirhart:
why do you need it?

liorur:
i dont actually, my bad, pls ignoe

PR #476, opened 2017-05-21
enhance DatePicker

lbelinsk:


lbelinsk:
1. Update react-datePicker to the latest version (some relevant issues are resolved).
2.  Support Input configurations, such as width, theme, placeholder, prefix. (The prefix is required because when i want to use the amaterial theme I don't want to have the Icon since it collides with the title prop..)

PR #480, opened 2017-05-22
chore(build): add module entry to package.json

ipanasenko:
This will allow bundlers like Rollup and Webpack2 to decrease bundle build by applying tree-shaking.


nirhart:
Since more work is needed, I'm not pulling this one yet

PR #481, opened 2017-05-22
chore(TimeInput): refactor deprecated React.createClass to es6 class

argshook:
the only component i found that uses now deprecated `React.createClass` api.

also some cleanup: removed unused `restoreFocus`, changed handlers to use arrow functions (to avoid binding context for each), sorted class methods and such.

no functional nor visual changes.


src/TimeInput/TimeInput.js
```diff
@@ -206,66 +135,96 @@ export default React.createClass({
     const text = this.formatTime(time);
     this.setState({time, am, text});
     this.bubbleOnChange({time, am});
-  },
+  }
+
+  handleAmPmClick = () =>
+    this.updateDate({am: !this.state.am});
```
mastertheblaster: Check this

PR #483, opened 2017-05-23
add viewer component

nirhart:


src/BaseComponents/common.scss
```diff
@@ -0,0 +1,135 @@
+/* Fonts */
+$fonts-fallback: "Helvetica Neue", "Helvetica", "Arial", "メイリオ, meiryo", "ヒラギノ角ゴ pro w3", "hiragino kaku gothic pro", "sans-serif";
```
argshook: @nirhart `common.scss` was moved, it can break some projects that import this file for mixins or color variables. need to check

nirhart: we actually have both now, its duplicated, but I think that no one should use it for external mixins or colors, we never said we support it

src/Viewer/Button/Button.js
```diff
@@ -0,0 +1,118 @@
...
+    return (
+      <ButtonLayout {...buttonLayoutProps} >
+        <button
+          onClick={onClick}
+          disabled={disabled}
+          type={type}
+          onMouseEnter={event => {
+            const target = event.target;
+            if (this.props.theme === ButtonLayout.settings.DESIGN_THEME) {
+              onHoverBackgroundColorOpacity = ButtonLayout.returnSelectedValue(onHoverBackgroundColorOpacity, ButtonLayout.settings.DEFAULT_DESIGN_HOVER_BG_OPACITY);
```
argshook: @nirhart wow this is rough, need to refactor imo

PR #486, opened 2017-05-23
improve(readme): use default postcss.config.js filename + typo

bchmn:


README.md
```diff
@@ -27,7 +27,7 @@ The files are brought 'uncompiled' and 'unpacked'. You will need to make sure we
 ```javascript
     {
         test: /\.scss$/,
-        include:['node_modules/wix-style-react/src')],
+        include:['node_modules/wix-style-react/src'],
```
bchmn: typo

README.md
```diff
@@ -112,7 +112,7 @@ package json - dependency
 "images-require-hook": "^1.0.3",
 
 #### to make css work
-add postcss.config
+add postcss.config.js
```
bchmn: The default filename includes the .js suffix, without this suffix it is required to explicitly define the config file path.

PR #516, opened 2017-06-07
chore: add support for XL button

koriguy:
not sure this is the proper for font-size, please take a look


src/ButtonLayout/ButtonLayout.scss
```diff
@@ -89,9 +88,32 @@
     }
 }
 
+.heightXL {
+    $height: 54px;
+    height: $height;
+    border-radius: $height / 2;
```
mastertheblaster: @koriguy it's `50%` ?

src/ButtonLayout/ButtonLayout.scss
```diff
@@ -89,9 +88,32 @@
...
+    padding: 0px 29px;
```
mastertheblaster: @koriguy base padding and transparent padding differs only by `1px`. We could define it one place, there is no place to maintain variation of `1px`?

src/ButtonLayout/ButtonLayout.scss
```diff
@@ -89,9 +88,32 @@
...
+
+    &.transparent {
+        padding: 0px 30px;
+    }
+
+    &.close-standard, &.close-dark, &.close-transparent {
+        $closeHeight: 26px;
+        height: $closeHeight;
+        width: $closeHeight;
+        border-radius: $closeHeight / 2;
```
mastertheblaster: @koriguy If you would declare `border-radius` as `50%`, you won't need to override the border radius each time.

src/ButtonLayout/ButtonLayout.scss
```diff
@@ -89,9 +88,32 @@
...
+        padding: 0;
+    }
+}
+
 .fullblue,
 .icon-standard {
     @extend .t1_2;
+    &.heightXL {
+        font-size: 20px;
```
mastertheblaster: @koriguy I see a lot of copy paste. Perhaps we could defined `20px` once ?
IMHO you should also use typography mixins for it.

stories/Button/ButtonStandard.js
```diff
@@ -95,6 +95,7 @@ class ButtonStandard extends Component {
                 <RadioGroup.Radio value="small">Small</RadioGroup.Radio>
                 <RadioGroup.Radio value="medium">Regular</RadioGroup.Radio>
                 <RadioGroup.Radio value="large">Large</RadioGroup.Radio>
+                <RadioGroup.Radio value="XL">XL</RadioGroup.Radio>
```
mastertheblaster: @koriguy Perhaps it should follow the naming convention and be lower case?

mastertheblaster: Smth wrong here. You should add new value to props. Otherwise validation won't pass.
and perhaps we could name it like `x-large` ?

src/ButtonLayout/ButtonLayout.scss
mastertheblaster: https://github.com/wix/wix-style-react/blob/master/src/Button/Button.js#L31

Perhaps this code overrides your sizes? Cause inline style has bigger priority.

stories/Button/ButtonStandard.js
koriguy: got the name from Nir but agree it's better x-large

nirhart: I know I agreed to XL but mantas have a point, xlarge seems better

koriguy:
@mastertheblaster please check it out

PR #523, opened 2017-06-09
fix(AutoDocs): do not fail on component classes that have no render method

argshook:
`react-docgen` used in `AutoDocs` throws an error if you try to parse a
react class component lacking render method. such component is possible
if it `extends` another component and uses inherited `render` implicitly.
`react-docgen` doesn't check inheritance by default, so it doesn't see
`render()` in component source, decides it's not a react component and
thus, throws an error.

here i introduce a custom resolver which is same as default one
except that it doesn't throw error and instead just console.warn about
it.

this is something that could be improved a lot. ideally resolver should
properly check and parse inherited classes to see if they have required
methods.

but perhaps implicit `render()` is worse and `console.warn` is a
friendly reminder to not use methods implicilty.

a PR for those up for a challenge is very welcome!


stories/utils/Components/AutoDocs/component-resolver.js
```diff
@@ -0,0 +1,148 @@
...
+  isExportsOrModuleAssignment,
+  isReactComponentClass,
+  isReactCreateClassCall,
+  isStatelessComponent,
+  normalizeClassDefinition,
+  resolveExportDeclaration,
+  resolveToValue
+} = utils;
+
+var ERROR_MULTIPLE_DEFINITIONS =
```
mastertheblaster: @argshook `const` ?

stories/utils/Components/AutoDocs/component-resolver.js
```diff
@@ -0,0 +1,148 @@
...
+
+const {
+  isExportsOrModuleAssignment,
+  isReactComponentClass,
+  isReactCreateClassCall,
+  isStatelessComponent,
+  normalizeClassDefinition,
+  resolveExportDeclaration,
+  resolveToValue
+} = utils;
```
mastertheblaster:
```js
const {
  utils: {
    isExportsOrModuleAssignment,
    isReactComponentClass,
    isReactCreateClassCall,
    isStatelessComponent,
    normalizeClassDefinition,
    resolveExportDeclaration,
    resolveToValue
  }
} = require('react-docgen');
```

PR #527, opened 2017-06-11
Fix disabled status in MultiSelect component

adavidof:


src/MultiSelect/InputWithTags.scss
```diff
@@ -36,3 +36,7 @@
 .error {
   border-color: $paletteR1;
 }
+
+.disabled {
+  border-color: $D55;
```
lbelinsk: @adavidof This color is defined in common.scss so it is better to use the common constants.

src/MultiSelect/InputWithTags.js
```diff
@@ -21,14 +21,15 @@ class InputWithTags extends React.Component {
...
 
     const className = classNames({
       [styles.tagsContainer]: true,
+      [styles.disabled]: disabled,
       [styles.error]: error,
     });
 
-    const desiredProps = omit(inputProps, ['onManuallyInput', 'inputElement', 'closeOnSelect', 'predicate', 'menuArrow', 'onClickOutside', 'fixedHeader', 'fixedFooter', 'dataHook']);
+    const desiredProps = omit(inputProps, ['onManuallyInput', 'inputElement', 'closeOnSelect', 'predicate', 'menuArrow', 'onClickOutside', 'fixedHeader', 'fixedFooter', 'dataHook', 'disabled']);
     const fontSize = (desiredProps.size && desiredProps.size === 'small') ? '14px' : '16px';
```
lbelinsk: disabled is not part of inputProps, so you don't need to omit it. You can remove this part

adavidof:
@lbelinsk Thanks for review! Fixed :)

PR #534, opened 2017-06-12
Turn off remove Tag when the disable status is true

adavidof:


src/Tag/Tag.spec.js
```diff
@@ -49,6 +49,17 @@ describe('Tag', () => {
     expect(onRemove).toBeCalledWith(id);
   });
 
+  it('should not call onRemove function on remove when Tag is disabled', () => {
+    const onRemove = jest.fn();
+
+    const driver = createDriver(<Tag id={id} disabled onRemove={onRemove}>{label}</Tag>);
```
lbelinsk: no need to do disabled={true}. just do disabled.

PR #557, opened 2017-06-22
Tooltip refactor

InterAl:
The Tooltip component now uses Popper.js for calculating the tooltip position.

* The position readjusts according to the available screen space.
* It moves together with the element.
* It realigns itself when the element dimensions change.
* We kept the original API and tests (and added a few of our own), so no breaking changes should occur.

Since it's a pretty heavy refactoring, we'd be glad if you could QA it a bit before merging.


src/Tooltip/Tooltip.js
```diff
@@ -37,365 +20,319 @@ class Tooltip extends WixComponent {
...
-        if (nextProps.active) {
-          this.show();
-        }
-      }
+    this.handleNextActive(nextProps);
+    this.handleNextMoveBy(nextProps);
+  }
+
+  //Schedule popper updates periodically, only when the tooltip is visible (for
+  //tooltip repositioning - e.g. when the target dimensions change).
```
nirhart: I think its not only resizing, but also repositioning

PR #559, opened 2017-06-23
add `displayTooltip` prop to Slider component, defaults to `true`

ipanasenko:


src/Slider/SliderHandle.js
```diff
@@ -46,7 +46,8 @@ export default class SliderHandle extends Component {
   }
 
   toggleTooltip(showTooltip) {
-    this.setState({showTooltip});
+    const {displayTooltip} = this.props;
```
ipanasenko: what's the problem with it?

adavidof: You use `displayTooltip` only once
As I see in the project not provided the rule: to destruct all values

ipanasenko: I don't see a reason not to use destructuring if I want to.
https://github.com/wix/wix-style-react/blob/53dc77588a25358e7a55d9dd0c6fa8ed8c0f01fd/src/Slider/Slider.js#L48-L50

PR #571, opened 2017-06-29
Wix style react TPA

giladartzi:


.storybook/config.js
```diff
@@ -1,5 +1,6 @@
 import {configure} from '@kadira/storybook';
 import {setOptions} from '@kadira/storybook-addon-options';
+import throttle from 'lodash/throttle';
```
nirhart: Why do we need this?

Issue #580, opened 2017-07-03
Wrong button display property for MessageBox HeaderLayout


Faradey27:
You can see that here we have className https://github.com/wix/wix-style-react/blob/master/src/MessageBox/HeaderLayout.js#L15 'close',

which is display flex https://github.com/wix/wix-style-react/blob/master/src/MessageBox/HeaderLayout.scss#L30

but https://stackoverflow.com/questions/35464067/flexbox-not-working-on-button-element-in-some-browsers

As a result, we have ugly close button in IE11(some versions of FF)

PR #583, opened 2017-07-03
Wrong button display property for MessageBox HeaderLayout #580

Faradey27:
Resolved https://github.com/wix/wix-style-react/issues/580

Manually tested in: Chrome, Safari, FF, not directly in IE11(by devtools in some of the projects that uses wix-style-react)


nirhart:
What was the problem?

Faradey27:
@nirhart i tried to explain this in issue,
but in general, if i m using normalize.css or one of my client, then button content will be not centered, so "cross" icon will be at the right part of button, instead of the center

Faradey27:
and thats the reason https://stackoverflow.com/questions/35464067/flexbox-not-working-on-button-element-in-some-browsers,

because button could not be display: flex, it not according to standart

@nirhart 

tommorow i can attach screnshots

nirhart:
No need for screenshot :), I'll recheck it tomorrow with IE and Edge and will merge if everything works good

Faradey27:
@nirhart sounds good, thank you

PR #586, opened 2017-07-04
updated package.json for react-datepicker

iaroslavshvets:
updated package.json for react-datepicker

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,7 +118,7 @@
     "prop-types": "^15.5.8",
     "rc-slider": "^8.0.2",
     "react-autosuggest": "^7.0.1",
-    "react-datepicker": "^0.46.0",
+    "react-datepicker": "^0.48.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -118,7 +118,7 @@
     "prop-types": "^15.5.8",
     "rc-slider": "^8.0.2",
     "react-autosuggest": "^7.0.1",
-    "react-datepicker": "^0.46.0",
+    "react-datepicker": "^0.48.0",
```
Faradey27: this change no make sense, because of ^, am i wrong?

iaroslavshvets: for instance i have 0.46 version in wix-style-react installed. Try it yourself. So it makes a difference.

nirhart: I don't see why it would make a difference, but I don't care merging it as well.

PR #591, opened 2017-07-08
Language picker

lbelinsk:
**Still need to create an interesting storybook example, but we can merge this for now**

@Noam3kCH Please take a look at the buttonWithOption updates and let me know if there is something there that you don't like.. I changed the validations to be more simple.

Regarding the ButtonWithOptions, a small triangle that "ticks" from the icon will improve the design of the component. We have the css for it in Guineapig. 

@nirhart I will send you a link for a feature toggle example in account recovery with this component in it. I don't know why the team city pr is red and I don't have access to see :)

This component doesn't look good enough. We should get UX assistance so I will ask for Lior Dahan's help (account recovery UX).


src/ButtonWithOptions/ButtonWithOptions.js
```diff
@@ -28,8 +28,7 @@ class ButtonWithOptions extends WixComponent {
     this.closeOnSelect = this.closeOnSelect.bind(this);
 
     if (props.children) {
-      this.buttonElement = props.children.find(element => element.type === ButtonWithOptions.Button);
-      this.optionsElement = props.children.filter(element => element.type !== ButtonWithOptions.Button);
+      [this.buttonElement, ...this.optionsElement] = React.Children.toArray(props.children);
```
lbelinsk: @Noam3kCH I only want the button to be first and the options to appear after

src/ButtonWithOptions/ButtonWithOptions.js
```diff
@@ -148,21 +147,16 @@ ButtonWithOptions.propTypes = {
...
   dropdownOffsetLeft: PropTypes.string,
   children: PropTypes.arrayOf((propValue, key) => {
-    if (propValue[key].type === ButtonWithOptions.Button) {
-      propValue.forEach((child, anotherKey) => {
-        if (anotherKey !== key && propValue.type === ButtonWithOptions.Button) {
-          return new Error(`ButtonWithOptions: Invalid Prop children, more than one ButtonWithOptions.Button was given`);
-        }
-      });
-      return;
+    if (key === 0 && propValue[key].type !== ButtonWithOptions.Button) {
```
lbelinsk: @Noam3kCH See if you like this. I thought it is simpler but I can change it back

src/ButtonWithOptions/ButtonWithOptions.protractor.driver.js
```diff
@@ -0,0 +1,10 @@
+const buttonWithOptionsDriverFactory = component => ({
```
lbelinsk: @Noam3kCH I added a protractor driver, just like InputWithOptions.

src/LanguagePicker/LanguagePicker.spec.js
```diff
@@ -0,0 +1,68 @@
...
+    const onSelect = jest.fn();
+    const {driver, dropdownLayoutDriver} = createDriver(languagePicker({onSelect}));
+
+    driver.click();
+    dropdownLayoutDriver.clickAtOption(0);
+
+    expect(onSelect).toBeCalledWith('en');
+  });
+
+  it('should print console warning for bad children format', () => {
```
lbelinsk: @Noam3kCH it will be nice to add this test also for ButtonWithOptions

src/ButtonWithOptions/ButtonWithOptions.js
```diff
@@ -148,21 +147,16 @@ ButtonWithOptions.propTypes = {
...
+      return new Error(`ButtonWithOptions: Invalid Prop children, first child must be ButtonWithOptions.Button`);
     }
-    if (propValue[key].type !== ButtonWithOptions.Option) {
-      const optionsValidator = React.Children.map(propValue[key], item => item.type === ButtonWithOptions.Option);
-      if (optionsValidator.some(optionValid => !optionValid)) {
+
+    React.Children.forEach(propValue[key], item => {
+      if (item.type !== ButtonWithOptions.Option) {
         return new Error(`ButtonWithOptions: Invalid Prop children was given. Validation failed on child number ${key}`);
```
Noam3kCH: this will always throw because the first child is the button...
@lbelinsk

src/ButtonWithOptions/ButtonWithOptions.js
Noam3kCH: this is fine

PR #596, opened 2017-07-10
Update e2e tests

yardnsm:
After the migration to storybook 3 (#579), the classnames have been hashed (in production). This have caused some issues with the e2e test, since some of the tests were relying on classnames to test behaviour. This is why we've added a [hotfix](https://github.com/wix/yoshi/commit/709336e659f69036a20ad527dfc90adc1287e220) for this in yoshi.

This PR attempts to fix that. Some of the e2e tests have been updated to check the hashed classnames instead of the original ones.


nirhart:
build failed

PR #608, opened 2017-07-13
Fixed container bug

dorraba:
The container had a bug after we added animations that the margin bottom disappeared.


shahafan:
after that change, when a row is last, it still gets the margin-bottom..

shahafan:
you've missed a comma
`
>div>span:last-child, .row:last-child {
  margin-bottom: 0;
}
`

PR #611, opened 2017-07-13
Disabled time input

dutzi:
Added disabled state to TimeInput


nirhart:
@dutzi The build is failing, please tune 'npm run build' and see the lint errors

PR #615, opened 2017-07-13
Allow to pass any Button theme to to Button Card Header

el-fuego:
Button is support a lot of themes, instead of two. I need to use already existed on Button at ButtonHeader


el-fuego:
Saw, that theme of ButtonHeader control not just a Buttons theme, but also height (((
Same API will be better for the future...

PR #616, opened 2017-07-13
Allow to pass suffix node to Card Header

el-fuego:
I need to pass custom node to the right side:
http://dl4.joxi.net/drive/2017/07/13/0024/1666/1582722/22/5be2b37c5c.jpg
http://dl3.joxi.net/drive/2017/07/13/0024/1666/1582722/22/9eeabe8598.jpg

Will be better to use components with composition of existed instead of predefined types of huge components (LinkHeader, ButtonHeader)

### DIFF
--- a/src/Card/Header/Header.js
+++ b/src/Card/Header/Header.js
@@ -9,16 +9,18 @@ class Header extends WixComponent {
   static propTypes = {
     title: node.isRequired,
     subtitle: node,
-    withoutDivider: bool
+    withoutDivider: bool,
+    suffix: node
   };
 
   static defaultProps = {
     subtitle: null,
+    suffix: null,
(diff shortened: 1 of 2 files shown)


nirhart:
@el-fuego Right now only button and link are allowed by the library UI, please refer your UX to talk to Nir Kosover so they will decide how to solve it

nirhart:
@el-fuego anything new?

argshook:
FYI

i think UX guys and wix-style-react devs miscommunicated at some point. Card header can contain various kinds of actions but we implemented only very specific ones and think only those are allowed, but that's not entirely true.

this is from ui-lib:

pasted image at 2017_07_18 04_17 pm

PR #624, opened 2017-07-17
Eslint plugin lodash

eddierl:
* add eslint-plugin-lodash
* cherty pick all lodash modules


shahata:
@eddierl new rules should always go to https://github.com/wix/eslint-config-wix, never to specific projects

eddierl:
Got it, thanks

PR #635, opened 2017-07-19
add disabled documentation for DatePicker

### LINKED ISSUE #634: DatePicker readOnly prop doesn't work (disable prop does)
In the StoryBook it says this prop should disable\enable modifying the date

It seems that this prop is propagated to ReactDatePicker component, but is not supported by it.
'disabled' prop, however, exists on ReactDatePicker and is working as expected.

lbelinsk:
@tomer-sharon 

close #634


tomer-sharon:
Thanks @lbelinsk

PR #642, opened 2017-07-20
Add info icon with tooltip component

el-fuego:
Can be used to describe form fields

http://dl3.joxi.net/drive/2017/07/18/0024/1666/1582722/22/a76bf39604.jpg


nirhart:
there are lint errors in here, please run 'npm run build'

PR #643, opened 2017-07-21
Fixed bug when tooltip is not visible when it inside of Modal

Faradey27:


src/Tooltip/Tooltip.js
```diff
@@ -82,7 +82,7 @@ class Tooltip extends WixComponent {
     hideTrigger: 'mouseleave',
     showDelay: 200,
     hideDelay: 500,
-    zIndex: 2000,
+    zIndex: 10011, // to be able to overlap modals, react-modal has 10010 z-index
```
mastertheblaster: @Faradey27 It's not very good, otherwise tooltip displayed in a main document would be overlapping the modal.
We have the same issue in our product, so a best solution was to use `appendToParent` property.

Faradey27: the problem that i m using RichTextEditor, which is using Tooltip, and most of functionality of richTextEditor become unworkble for user(

so i see to solution, try to modify RIchTextEditor with appendTooltipToParent property or update z-index

@mastertheblaster what do you think?

mastertheblaster: Maybe `RichTextEditor` should always append tooltips to parent?
And what do you mean by saying `or update z-index` ? Where ?

Faradey27: maybe, will try to do this and check

"And what do you mean by saying or update z-index ? Where ?" - this is what i have done in this pull request) i updated z-index of tooltip

Faradey27:
@mastertheblaster can you pls re-check?
i updated

PR #644, opened 2017-07-22
Typescripted wix style react

assafswix:


nirhart:
please re submit when ready

PR #652, opened 2017-07-24
Fix for nesting Container > Row > Container > Row

el-fuego:
Reverting old Container as RawContainer
Usage example http://joxi.ru/bmoR817UM3Yvkr.jpg

### DIFF
--- a/src/Grid/Grid.js
+++ b/src/Grid/Grid.js
@@ -5,6 +5,21 @@ import classNames from 'classnames';
 import PropTypes from 'prop-types';
 import Animator from '../Animator';
 
+class RawContainer extends Component {
+
+  static propTypes = {
+    children: PropTypes.node
+  };
+
+  render() {
+    return (
+      <div className={styles.container}>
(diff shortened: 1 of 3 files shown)


nirhart:
Sorry, I don't understand why would you need container -> row -> container

el-fuego:
1. Because of this http://joxi.ru/Vm6KnBWHx4pWZ2.jpg
real design http://joxi.ru/bmoR817UM3Yvkr.jpg

2. Adding Animator to Container was bracking change for margins

3. Margins can`t be fixed just by styles like this https://github.com/wix/wix-style-react/blob/master/src/Grid/Grid.scss#L26 (however, now it does not work because of Animator structure changes)

4. With Container, nested Rows and Cols are more semantic and section parts can be sepatared with return (<Container> ... </Container>) (single node) instead of (<Row>...</Row><Row>...</Row>)

5. I don`t want any animation on each <Row />. It will be wery slow

Better way is split Container and Animator and use Animator just if needed

PR #682, opened 2017-08-07
dev(inputArea): allow custom dataHook

alonbt:


src/InputArea/InputArea.js
```diff
@@ -198,7 +198,8 @@ InputArea.propTypes = {
   ariaDescribedby: PropTypes.string,
   ariaControls: PropTypes.string,
   tooltipPlacement: PropTypes.string,
-  onTooltipShow: PropTypes.func
+  onTooltipShow: PropTypes.func,
+  dataHook: PropTypes.string
 };
```
lbelinsk: I think maybe in order not to break existing usages, we can add:
```InputArea.defaultProps = {dataHook: 'textarea-div'}```
Then your changes won't be needed in the test file. (but you would need to add a test for this prop anyway).

wdyt?

alonbt: It's possible but I'm pretty sure it's a bug, so better to remove it then to leave it there. I've checked and no one in our repositories is using it, I doubt someone from the big web is using it..

lbelinsk: I agree, this is not a good practice.
So just to be on the safe side, can you just announce it on the slack channel?

alonbt: announced :)

PR #684, opened 2017-08-07
fix input, dropdown disabled styles

el-fuego:
https://app.zeplin.io/project/5864e02695b5754a69f56150/screen/5871f72e355394883366a692


lbelinsk:
@el-fuego is this approved by Nir Kosover or Ben Benhorin?
They are the UX of wix-style-react.

el-fuego:
I attached the link, as I understand, to current version of design, what need to be implemented (is it what you are using?)
Nir also told, that current code implementation is wrong and need to be fixed regarding to specification

lbelinsk:
Yeah this is the zeplin that Kosover prepared.
If this is what Nir instructed then I will merge this.

@nirhart if there is something that isn't what you talked about in the end so please be aware of this merge so you could change it with @el-fuego after. For me it looks good..

PR #686, opened 2017-08-07
Add link theme for ButtonLayout

el-fuego:
LinkLayout can`t be used in my case.
1. For accessibility I need to use <button/> or <a/> for active elements
2. I don`t need for some cases href at all, just a click - need to use button only
3. <LinkLayout><button/></LinkLayout> will have borders and other stuff from button

Here is an example http://joxi.ru/EA4d3B5CDpMe1A.jpg

### DIFF
--- a/src/ButtonLayout/ButtonLayout.js
+++ b/src/ButtonLayout/ButtonLayout.js
@@ -69,7 +69,9 @@ ButtonLayout.propTypes = {
     'icon-standard',
     'icon-standardsecondary',
     'icon-white',
-    'icon-whitesecondary'
+    'icon-whitesecondary',
+    'link',
+    'link-purple'
   ])
 };
 
--- a/src/ButtonLayout/ButtonLayout.scss
+++ b/src/ButtonLayout/ButtonLayout.scss
@@ -310,6 +310,30 @@
     color: $D10;
 }
 
+.link,
+.link-purple {
+    @extend .t1_3;
+    height: 24px;
+    padding: 0;
+
+    &.heightsmall {
+        @extend .t3_3;
+        height: 18px;
+        padding: 0;
+    }
+
+    background-color: transparent;
+    border: none;
+}
+
+.link {
+    color: $B10;
+}
+
+.link-purple {
+    color: $P10;
+}
+
 .fullgreen:hover, .fullgreen.hover,
 .emptygreen:hover, .emptygreen.hover {
     @extend .t1_2;
@@ -416,6 +440,11 @@
     background-color: transparent;
     border-color: transparent;
     color: $B20;
+}
+
+.link:hover, .link.hover,
+.link-purple:hover, .link-purple.hover {
+    text-decoration: underline;
 }
 
 .icon-greybackground:active, .icon-greybackground.active,
@@ -591,5 +620,10 @@
     pointer-events: none;
     background-color: transparent;
     border-color: transparent;
+    color: $D55;
+}
+
+.link.disabled,
+.link-purple.disabled {
     color: $D55;
 }


el-fuego:
Here is TextLinkLayout
http://joxi.ru/Q2Kga9Mc9vWpR2.jpg

```
            <TextLinkLayout>
              <button onClick={_.noop}>
                button
              </button>
            </TextLinkLayout>

            <TextLinkLayout>
              <a onClick={_.noop} href="#">
                a
              </a>
            </TextLinkLayout>
```

Button with link theme  - is what I need.
It is very common style (see Bootstrap as example), what will be used without any additional hack-wrappers

lbelinsk:
I don't understand what is missing here for you.
You have ButtonLayout and TextLinkLayout which simply wraps their children prop with the relevant  style accordingly.

Do you mean that you don't think that this wrapper model is good? Or is there still something missing for you practically?

el-fuego:
I need <button/> or <a/> inside for accessibility and semantics. How it looks you can see at previous message (looks wrong)

Also, I need two font-sizes (14px and 16px)

lbelinsk:
In my opinion we should improve the current approach and not to add something similar.
I think that instead of this new code it can be integrated inside the current layouts in a way that it will work as expected.

For example, why to use a button inside a textlinkLayout? maybe a ```<div>``` element with onClick will be better, or maybe not.

If you don't think it is the right approach, we can wait for @nirhart for that if it is not urgent..

el-fuego:
In any case, at the future we wil need to use semantic clickable elements.. so I vote for adding theme like it done at Bootstrap (and used by a lot of worldwide users)
How long we need to wait?

lbelinsk:
Nir returns fro a vacation next week. Will it be ok to wait until then?

el-fuego:
Ok, will wait

kupriyanenko:
@harelig, @lbelinsk, @el-fuego I think we should close this PR. As for me, all buttons, event with link style, should have the same layout (paddings, margins, etc). Look in bootstrap: https://getbootstrap.com/docs/3.3/css/#buttons-options

If we will have different layout for different styles of buttons, we will have a mess and button composing will be too complex.

kupriyanenko:
@lbelinsk can we add to `TextLinkLayout` `onClick` handler?

kupriyanenko:
Or, we can leave this PR opened, because it can be used, but change a little bit. We shouldn't change layout for it, we can change only colors.

nirhart:
@kupriyanenko @el-fuego wix-style-react is not bootstrap, its very restricted to wix theming, and wix theming of link looks different from a button, thats why we separated it. I agree that for accessibility we will need some kind of styleless button to be inside TextLinkLayout, but design wise, TextLinkLayout can do exactly what you need.

el-fuego:
About using of current TextLinkLayout or Button:
https://stackoverflow.com/questions/37147842/a-vs-button-vs-div-vs-input-type-button
https://davidwalsh.name/html5-buttons

nirhart:
@el-fuego TextLinkLayout do not restrict you to <a> or to <button>, you can do whatever you want with it.

el-fuego:
See examples at previous messages, where button was wrapped with this component.

Another option is make it as styled button not inside Button, but TextLinkLayout and will support all events

lbelinsk:
@kupriyanenko You shouldn't put ```onClick```  on ButtonLayout, instead put it on it's child component.

Stupid example:

```
<ButtonLayout>
    <div onClick../>
</ButtonLayout>
```

el-fuego:
@lbelinsk https://github.com/wix/wix-style-react/pull/686#issuecomment-320755851 and not div beause of semantics

PR ussless? Ok, will wait 2 hours and close it with realizing all this stuff at my project

PR #691, opened 2017-08-08
dev(cssTransitionGroup): revert from v2.x to v1.x since some animatio…

alonbt:
…ns were broken and could not have been fixed


lbelinsk:
@alonbt is this ready for merge?

PR #692, opened 2017-08-08
PopoverMenu Testkit

skyiea:


src/PopoverMenu/PopoverMenu.driver.js
```diff
@@ -0,0 +1,44 @@
...
+  const driver = {
+    init: {
+      menuItemDataHook: dataHook => {
+        menuItemDataHook = dataHook;
+
+        return driver;
+      },
+    },
+    trigger: {
+      exists: () => !!element,
```
lbelinsk: We have some framework helpers which requires the exists method to sit directly on the driver object, not on the driver.trigger.

src/PopoverMenu/PopoverMenu.driver.js
```diff
@@ -0,0 +1,44 @@
...
+      throw new Error('Before accessing menu fields init menu item data hook with "driver.given.menuItemDataHook(\'myDataHook\')" command');
+    }
+
+    return fn(...args);
+  };
+
+  const itemsArray = () => [...document.body.querySelectorAll(`[data-hook="${menuItemDataHook}"]`)];
+
+  const driver = {
+    init: {
```
lbelinsk: Why do we need the init method If we can simply put a dataHook on the real component which inherits from wixComponent?

src/PopoverMenu/PopoverMenu.spec.js
```diff
@@ -0,0 +1,125 @@
...
+
+    await waitFor(() => {
+      expect(driver.menu.isShown()).toBe(true);
+    });
+
+    expect(driver.menu.itemsLength()).toBe(2);
+    expect(driver.menu.itemContentAt(0)).toBe(menuItem1Text);
+
+    driver.menu.clickItemAt(0);
+    expect(menuItem1Listener).toBeCalled();
```
skyiea: WDYM?
There is no arguments passed into callback. Each menu item has it's own callback, thus no way to be more specific.

src/PopoverMenu/PopoverMenu.driver.js
skyiea: I see. Would be great if it will be documented somewhere for the future testkit developers.

src/PopoverMenu/PopoverMenu.driver.js
skyiea: Unsure I understand you.
There *is* `dataHook` on the `<PopoverMenu/>` too.
Why we need another `dataHook` is because `PopoverMenu` (under the hood) uses `Tooltip`, which works like this:
- renders *trigger* button (in case of `PopoverMenu` it is three-dots-button)
- on click on *trigger* button it renders actual tooltip into `document.body`

Thus you can't access tooltip content from `Tooltip` component via descendants query (take a look at `Tooltip.driver.js`). So it means root `dataHook` can be used only to access *trigger* button, and not tooltip itself.
You can't do document query for `.tooltip` (like `Tooltip.driver.js` is doing, which is wrong) because:
a) driver won't know which one of the tooltips (in case of multiple tooltips rendered) is the one that was rendered by driver's trigger-element. (Probably I should avoid using tooltip-driver's `isShown` for the same reason)
b) in case of e2e test `.tooltip` class name most probably will be transformed into something else.

So I was forced to introduce another `dataHook` for menu item (which is known for the users of the driver), which solves this issue.

src/test-common.js
```diff
@@ -21,14 +21,14 @@ export const testkitFactoryCreator = driverFactory => ({wrapper, dataHook}) => {
 // enzyme
 export const enzymeTestkitFactoryCreator = driverFactory => ({wrapper, dataHook}) => {
   const regexp = new RegExp(`^<[^>]+data-hook="${dataHook}"`);
-  const component = wrapper.findWhere(n => !n.props().dataHook && (regexp).test(n.html()));
+  const component = wrapper.findWhere(n => typeof n.type() === 'string' && (regexp).test(n.html()));
```
skyiea: I assumed that `!n.props().dataHook` condition was there, as you've said, to check if component is DOM element. Better way to check this is to use `n.type()`, which returns *string* in case of DOM component, and *function* (constructor) in case it is React component (http://airbnb.io/enzyme/docs/api/ReactWrapper/type.html).

Yes, I had an issue with previous condition. `PopoverMenu` uses `Tooltip`, passing `<Button><Dots/></Button>` as trigger element. In such scenario `wrapper.findWhere(n => !n.props().dataHook && (regexp).test(n.html()));` returned array with *multiple* DOM components (5, all of them were the same component - `<button>`). Seems that there is some deep problem with enzyme's `.find` & `.findWhere` methods (both returned incorrect results in this case). Found similar unresolved issue: https://github.com/airbnb/enzyme/issues/701.

skyiea:
@lbelinsk Updated according to your comments.

src/PopoverMenu/PopoverMenu.spec.js
lbelinsk: Sorry, you are right.
I confused it with the dropdownLayout where it i being called with the item's index.

src/test-common.js
lbelinsk: Ok, sounds good then.
I can't think of a scenario where this change won't work, and the tests are green, so let's go with that..

skyiea:
Why not merge? Anything else?

PR #695, opened 2017-08-08
fix Tooltip bottom padding according to design

el-fuego:
screen shot 2017-08-08 at 6 45 48 pm
screen shot 2017-08-08 at 6 45 42 pm

Probably 18px bottom was made by mistake


lbelinsk:
@nir-k It sounds reasonable that the margin bottom will be 12 like the margin top.
I am merging it but if for some reason someone did it on purpose let me know

PR #699, opened 2017-08-10
Data table details

iluxav:


src/DataTable/DataTable.driver.js
```diff
@@ -2,14 +2,15 @@ import React from 'react';
...
-const dataTableDriverFactory = ({element, wrapper, component}) => {
+const dataTableDriverFactory = ({ element, wrapper, component }) => {
 
   const getHeader = () => element.querySelector('thead');
   const hasHeader = () => !!getHeader();
 
   const getRows = () => element.querySelectorAll('tbody tr');
   const getRowsCount = () => getRows().length;
   const getRow = rowIndex => getRows()[rowIndex];
+  const getRowDetails = index => element.querySelector(`tbody tr td[data-hook="${index}_details"]`);
```
lbelinsk: Maybe you should expose this because in case the rowDetails will consist of not only text, but buttons or other stuff then we want to allow to access them.

src/DataTable/DataTable.scss
```diff
@@ -22,6 +22,18 @@
   border-bottom: solid 1px $D70;
 }
 
+.table td.details {
+  padding: 0px;
+  height: 0px;
+  border-bottom: none;
+}
+
+.table td.active {
```
lbelinsk: It won't be a blocker, but please consult with Alon Bar-tur to assist with some animation.

src/DataTable/DataTable.spec.js
```diff
@@ -119,23 +119,57 @@ describe('Table', () => {
...
+      expect(driver.hasRowDetails(0)).toBe(true);
+      expect(driver.getRowDetailsText(0)).toBe('');
+      driver.clickRow(0);
+      expect(driver.getRowDetailsText(0)).toBe(defaultProps.data[0].a);
+      driver.clickRow(0);
+      expect(driver.hasRowDetails(0)).toBe(true);
+      expect(driver.getRowDetailsText(0)).toBe('');
+    });
+
+    it('should expand with correct content and collapse', () => {
```
lbelinsk: Duplicated test... Please remove it

stories/DataTable/ExampleWithRowDetails.js
```diff
@@ -0,0 +1,58 @@
...
+class ExampleWithRowDetails extends React.Component {
+  render() {
+    return (
+      <div style={style}>
+        <DataTable
+          dataHook='story-data-table'
+          data={generateData()}
+          rowDetails={(row, rowNum) => <MyRowDetailsComponent {...row} />}
+          allowMultiDetailsExpansion={true}
+          onRowClick={(row, rowNum) => {
```
lbelinsk: can remove this empty callback and also itemPerPage prop

src/DataTable/DataTable.js
```diff
@@ -93,23 +94,24 @@ class DataTable extends WixComponent {
...
 
-    handlers.forEach(({rowEventHandler, eventHandler}) => {
+    handlers.forEach(({ rowEventHandler, eventHandler }) => {
       if (rowEventHandler) {
         optionalRowProps[eventHandler] = event => {
           if (event.isDefaultPrevented()) {
             return;
           }
           rowEventHandler(rowData, rowNum);
+          this.tryToggleRowDetails(eventHandler, rowNum);
```
lbelinsk: Suggestion: 
Instead of calling tryToggleRowDetails which checks if the ```eventHandler === 'onClick'```, lets just call it from row 120 inside if (onRowClick)

stories/DataTable/ExampleWithRowDetails.js
```diff
@@ -0,0 +1,55 @@
...
+class ExampleWithRowDetails extends React.Component {
+  render() {
+    return (
+      <div style={style}>
+        <DataTable
+          dataHook="story-data-table"
+          data={generateData()}
+          rowDetails={(row, rowNum) => <MyRowDetailsComponent {...row}/>}
+          allowMultiDetailsExpansion={true}
+          itemsPerPage={20}
```
lbelinsk: itemsPerPage is relevant only for infiniteScroll. Please remove it to prevent confusion

stories/DataTable/ExampleWithRowDetails.js
```diff
@@ -0,0 +1,55 @@
...
+
+class ExampleWithRowDetails extends React.Component {
+  render() {
+    return (
+      <div style={style}>
+        <DataTable
+          dataHook="story-data-table"
+          data={generateData()}
+          rowDetails={(row, rowNum) => <MyRowDetailsComponent {...row}/>}
+          allowMultiDetailsExpansion={true}
```
lbelinsk: allowMultiDetailsExpansion  
remove the ```without ={true}```

src/DataTable/DataTable.js
```diff
@@ -110,6 +111,7 @@ class DataTable extends WixComponent {
...
```
lbelinsk: What About my previous suggestion?
Moving it to row 120? and rename the ```tryToggleRowDetails``` to ```toggleRowDetails``` and delete remove the condition inside this func

nirhart:
@iluxav any updates on it?

src/DataTable/DataTable.scss
iluxav: Probably will be done later. I want Tomer to start working with it and we will add the animation later.

src/DataTable/DataTable.js
iluxav: Line 120 is a `renderRow` function that called for each row, but what we need in this case is click handler inside `ptionalRowProps[eventHandler] = event => {}` where eventHandler should be onClick

PR #702, opened 2017-08-10
Adding Sorting

iluxav:


src/DataTable/DataTable.spec.js
```diff
@@ -121,21 +121,75 @@ describe('Table', () => {
     });
   });
 
+
+  describe('Sortable column titles', () => {
+    const tests = [
```
lbelinsk: You are not using tests variable..

src/DataTable/DataTable.spec.js
```diff
@@ -121,21 +121,75 @@ describe('Table', () => {
...
+      { handler: 'onRowClick', driverMethod: 'clickRow' },
+      { handler: 'onMouseEnterRow', driverMethod: 'mouseEnterRow' },
+      { handler: 'onMouseLeaveRow', driverMethod: 'mouseLeaveRow' }
+    ];
+
+    it('should display sortable title', () => {
+      const props = {
```
lbelinsk: you are defining const props 3 times in each it block. Better to define it in the describe block once

src/DataTable/DataTable.spec.js
```diff
@@ -121,21 +121,75 @@ describe('Table', () => {
...
+          { title: 'B', render: row => row.b }
+        ],
+        onSortClick: jest.fn()
+      };
+      const driver = createDriver(<DataTable {...props} />);
+      driver.clickSort(2);
+      expect(props.onSortClick).not.toHaveBeenCalled();
+    });
+  });
+
```
lbelinsk: I am missing a test to check that the icon is getting flipped onClick. wdyt?

stories/DataTable/ExampleSortable.js
```diff
@@ -0,0 +1,68 @@
...
+    return this.state.data.sort((a, b) => desc ? ~~(field ? a[field] < b[field] : a < b) : ~~(field ? a[field] > b[field] : a > b));
+  }
+
+  render() {
+    return (
+      <div style={style}>
+        <DataTable
+          dataHook='story-data-table'
+          data={this.state.data}
+          onRowClick={(row, rowNum) => {
```
lbelinsk: empty callback. can remove it

stories/DataTable/ExampleSortable.js
```diff
@@ -0,0 +1,68 @@
...
+          }}
+          onSortClick={(col, colNum) => this.handleSortClick(colNum)}
+          infiniteScroll
```
lbelinsk: please remove InfinitScrool and ItemsPerPage, not relevant and can confuse the users

stories/DataTable/ExampleSortable.js
```diff
@@ -0,0 +1,68 @@
...
+    const filelds = {
+      '1': 'firstName',
+      '2': 'lastName'
+    };
+    const sortedData = this.sortDataByField(filelds[colNum], desc);
+    this.setState({ sort, data: sortedData });
+  }
+
+  sortDataByField(field, desc) {
+    return this.state.data.sort((a, b) => desc ? ~~(field ? a[field] < b[field] : a < b) : ~~(field ? a[field] > b[field] : a > b));
```
lbelinsk: I am not familiar with this ~~ syntax :)
What is it doing?

src/DataTable/DataTable.js
```diff
@@ -4,6 +4,7 @@ import s from './DataTable.scss';
 import classNames from 'classnames';
 import InfiniteScroll from './InfiniteScroll';
 import WixComponent from '../BaseComponents/WixComponent';
+import ArrowVertical from '../Icons/dist/components/ArrowVertical';
```
lbelinsk: I would think the ArrowDownThin will be more suitable here.
You should consult with your UX, but most importantly with the new UX of wix-style-react Ben Benhorin.

src/DataTable/DataTable.js
```diff
@@ -132,32 +133,46 @@ class DataTable extends WixComponent {
...
       height: this.props.thHeight,
-      fontSize: this.props.thFontSize
+      fontSize: this.props.thFontSize,
+      cursor: column.sortable === undefined ? 'arrow' : 'pointer'
     };
-    return <th style={style} key={colNum}>{column.title}</th>;
+
+    let optionalHeaderCellProps = {};
+    if (column.sortable) {
+      optionalHeaderCellProps.onClick = () => this.props.onSortClick(column, colNum);
```
lbelinsk: the onSortClick prop is not "isRequired", doesn't have default noop prop. You can add one of them or wrap this usage with an if condition

src/DataTable/DataTable.js
```diff
@@ -132,32 +133,46 @@ class DataTable extends WixComponent {
...
-    const classes = classNames({[s.important]: column.important});
+    const classes = classNames({ [s.important]: column.important });
     return <td className={classes} key={colNum}>{column.render && column.render(rowData, rowNum)}</td>;
   };
 
+  renderSortingArrow = (sortDescending, colNum) => {
+    if (sortDescending === undefined) {
+      return null;
+    }
+    const sortDirectionClassName = sortDescending === true ? s.sortArrowAsc : s.sortArrowDesc;
```
lbelinsk: === true is not really needed, up to you..

src/DataTable/DataTable.spec.js
iluxav: Icon position reflects the true/false prop that passed from outside. So I'm not sure if I really need to test it css, not sure...

PR #704, opened 2017-08-11
Support collapsing by header click && fix onCollapsedChange at Collap…

el-fuego:
…sedHeader

First one was approved by Nir Kosover

### DIFF
--- a/src/Card/CollapsedHeader/CollapsedHeader.js
+++ b/src/Card/CollapsedHeader/CollapsedHeader.js
@@ -46,10 +46,15 @@ class CollapsedHeader extends WixComponent {
     }
   }
 
-  toggleCollapsed = event => {
+  stopPropagation(event) {
+    event.stopPropagation();
+  }
+
+  toggleCollapsed = () => {
     const {onCollapsedChange} = this.props;
-    this.setState({isCollapsed: !this.state.isCollapsed});
-    onCollapsedChange && onCollapsedChange(event);
+    const isCollapsed = !this.state.isCollapsed;
+    this.setState({isCollapsed});
+    onCollapsedChange && onCollapsedChange(isCollapsed);
   };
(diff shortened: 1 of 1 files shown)


src/Card/CollapsedHeader/CollapsedHeader.js
```diff
@@ -46,10 +46,15 @@ class CollapsedHeader extends WixComponent {
...
+    event.stopPropagation();
+  }
+
+  toggleCollapsed = () => {
     const {onCollapsedChange} = this.props;
-    this.setState({isCollapsed: !this.state.isCollapsed});
-    onCollapsedChange && onCollapsedChange(event);
+    const isCollapsed = !this.state.isCollapsed;
+    this.setState({isCollapsed});
+    onCollapsedChange && onCollapsedChange(isCollapsed);
```
lbelinsk: Why did you call onCollapsedChange with a boolean value?

el-fuego: Because, I need to know current state. With checkbox it will work, but not with button

lbelinsk: Ok, so you need to update it in the docs.
onCollapsedChange is a prop that the user needs to know what kind of callback to pass, so it might be a breaking change because if someone is already passing:

```onCollapsedChange={event => event.target.blabla}```
He will get an error because event is a boolean now.

I looked in the wix-private github and I don't see that anyone passes this callback prop, but please document it.

el-fuego: For a now, onCollapsedChange was not docummented at all

nirhart: @el-fuego it still needs to get documentation

PR #706, opened 2017-08-11
feat(badge): add rectangle shaped badge

tomerla:


src/Badge/Badge.js
```diff
@@ -53,14 +54,18 @@ Badge.propTypes = {
     'T5', 'T5.1'
   ]).isRequired,
 
+  /** set the shape */
+  shape: oneOf(['ellipse', 'rectangle']).isRequired,
```
lbelinsk: isRequired is redundant because you set a defaultProp of alipse

src/Badge/Badge.spec.js
```diff
@@ -27,4 +27,8 @@ describe('Badge', () => {
     expect(driver.isOfAppearance('H1')).toBeTruthy();
   });
 
+  it('should allow specifying a shape', () => {
```
lbelinsk: I would add a test that checks that the default shape is elipse

src/Badge/Badge.js
tomerla: agreed
but I just follow the existing pattern -  all other props has a `defaultProp` and they are all marked as `isRequired`
do you want me to change them all?

src/Badge/Badge.spec.js
tomerla: we have already a e2e tests for it
but sure.. why not :)

tomerla:
@lbelinsk so I updated a test for a default ellipse shape and updated the docs
please let me know what do you want to do with the `isRequired` (wrote to you in the line comment)
tnx!

src/Badge/Badge.js
lbelinsk: Yeah, it will be great! I guess it is a trailing error :)

tomerla:
tnx @lbelinsk !

PR #731, opened 2017-08-17
Badge driver - add isBadge method

tomerla:


tomerla:
of course, consider the following scenario:

a developer (by mistake) might replace this line:
`import Badge from 'wix-style-react/dist/src/Badge'`

with this line:
`import Badge from 'wix-style-react/dist/src/Text'`

all tests will pass but the app ui will be broken :)

PR #736, opened 2017-08-17
Talh/stylable conversion

talh78:
Waiting for new Yoshi version with updated versions of stylable-integration and stylable


lbelinsk:
Het @talh78 this pr is getting really really big.
Do you have any estimation when would you want to start discussing about merging it?

Maybe it will be better to pause the progress and focus on a specific simple component to merge. There is no way we could review such pr. The only way will be to close this one and to create a separate pr for each component.

If you are not ready and you are just using the team city pr, then it is ok as well, just update please what is the status.

Thanks..

talh78:
Yeah, this is just the big branch where everything could be managed in the same place. When I'm ready to merge, we will do it in smaller chunks. Thanks!

PR #738, opened 2017-08-18
fix manual input to multiselect

el-fuego:
1. Type in new tag
2. Press enter to add

ACTUAL RESULT:
  tag is erased
EXPECTED RESULT:
  tag should be added

https://jira.wixpress.com/browse/VOD-1031


el-fuego:
@lbelinsk Can you or somebody take care of this PR ?
We roll-back our app because of this bug..

PR #751, opened 2017-08-25
Add `data-hook` to `Card` and `Container` components

adavidof:


nirhart:
As Lior said, its redundant

PR #756, opened 2017-08-28
fix: yellow autocomplete background

Varnius:
Fixes Crome adding yellow background to the input in case of autocomplete.
https://www.screencast.com/t/j67lzTYkphO

### DIFF
--- a/src/Input/Input.js
+++ b/src/Input/Input.js
@@ -200,6 +200,7 @@ Input.defaultProps = {
 
 Input.propTypes = {
   id: PropTypes.string,
+  name: PropTypes.string,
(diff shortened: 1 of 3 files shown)


nirhart:
@Varnius lint is failing, please run "npm run build" to see lint errors

src/Input/Input.js
```diff
@@ -198,6 +200,7 @@ Input.defaultProps = {
 
 Input.propTypes = {
   id: PropTypes.string,
+  name: PropTypes.string,
```
nirhart: What is name for?

Varnius: Google autofill uses it to tell apart fields, along with "autocomplete" attrib. Actually, I'm not sure if we really need it, so added just in case.

PR #761, opened 2017-08-28
Feature/date picker

malcolmmadsheep:
@nirhart Can you, please, take a look at this?


testkit/index.js
```diff
@@ -167,3 +167,6 @@ export const popoverMenuTestkitFactory = testkitFactoryCreator(popoverMenuDriver
 
 import badgeDriverFactory from '../src/Badge/Badge.driver';
 export const badgeTestkitFactory = testkitFactoryCreator(badgeDriverFactory);
+
+import datePickerDriverFactory from '../src/DatePicker/DatePicker.driver';
```
lbelinsk: Export the enzyme testkit from 'testkit/enzyme as well'

src/DatePicker/DatePicker.driver.js
```diff
@@ -1,51 +1,22 @@
...
-          this.navigateToNextMonth();
-        } else {
-          this.navigateToPreviousMonth();
-        }
-      }
+  const getDatePickerOption = (index = 0) => element.querySelectorAll('[role="option"]:not([class*="outside-month"])')[index];
+  const getElementByDataHook = dataHook => element.querySelector(`[data-hook="${dataHook}"]`);
+  const getDatePickerInput = () => element.querySelector('input');
+  const getDatePickerInputValue = () => getDatePickerInput().value;
+  const getDatePicker = () => element.querySelector('.react-datepicker');
```
lbelinsk: getCalendar will be more accurate. Datepicker is the component's name including the input..

src/DatePicker/DatePicker.spec.js
```diff
@@ -1,57 +1,108 @@
...
+    it('should set inputDataHook from props', () => {
+      const driver = createDriver(<DatePicker onChange={onChange} inputDataHook={'dataHook'}/>);
 
-    driver = datePickerDriverFactory(datePickerRoot);
-  });
+      expect(driver.getElementByDataHook('dataHook')).toBeTruthy();
+    });
+
+    it('should set datePicker disabled from props', () => {
+      const driver = createDriver(<DatePicker onChange={onChange} disabled={true}/>);
```
lbelinsk: disabled is enough, no need to write disabled={true}

src/DatePicker/DatePicker.spec.js
```diff
@@ -1,57 +1,108 @@
...
+
+      expect(driver.getDatePickerInput().hasAttribute('disabled')).toBeTruthy();
```
lbelinsk: better to check that the calendar is not opened on click and that the input and use the inputTestkit.isDisabled() method for that..
https://github.com/wix/wix-style-react/blob/master/src/Input/Input.driver.js#L46

src/DatePicker/DatePicker.spec.js
```diff
@@ -1,57 +1,108 @@
...
   });
 
-  describe('calendar', () => {
-    it('calendar is not shown', () => {
-      expect(driver.getDatePickerPopup()).toBeNull();
+  describe('date picker popover', () => {
+    it('should not render datePicker on start', () => {
+      const driver = createDriver(<DatePicker onChange={onChange}/>);
+
+      expect(driver.getDatePicker()).toBeFalsy();
```
lbelinsk: Maybe we can add isVisible method.. might be very useful for other projects as well..

src/DatePicker/DatePicker.spec.js
```diff
@@ -1,57 +1,108 @@
...
-      expect(driver.getDatePickerPopup()).not.toBeNull();
+    it('should call onChange on click on available day in datePicker', () => {
+      const driver = createDriver(<DatePicker onChange={onChange}/>);
+      const input = driver.getDatePickerInput();
+      driver.clickOn(input);
+      const option = driver.getDatePickerOption();
+      driver.clickOn(option);
+
+      expect(onChange).toHaveBeenCalled();
     });
```
lbelinsk: What about selecting with keyboard press.. not only with mouse

src/DatePicker/DatePicker.spec.js
```diff
@@ -1,57 +1,108 @@
...
-      driver.showDatePickerModal();
-      expect(driver.getDatePickerPopup()).not.toBeNull();
+    it('should call onChange on click on available day in datePicker', () => {
+      const driver = createDriver(<DatePicker onChange={onChange}/>);
+      const input = driver.getDatePickerInput();
+      driver.clickOn(input);
+      const option = driver.getDatePickerOption();
+      driver.clickOn(option);
+
+      expect(onChange).toHaveBeenCalled();
```
lbelinsk: Can be more specific - toHaveBeenCalledWith(....)

src/DatePicker/DatePicker.spec.js
```diff
@@ -1,57 +1,108 @@
...
+      driver.clickOn(option);
 
-      it('onChange reports the selected date', () => {
-        driver.selectDate(moment('2016-03-16'));
-        const changeDateCallbackValue = onChangeMock.getCall(0).args[0];
-        expect(changeDateCallbackValue.toISOString()).toEqual(moment('2016-03-16').toISOString());
-      });
+      expect(driver.getDatePicker()).toBeFalsy();
+    });
+
```
lbelinsk: This is a good test.. Add a similar one that checks that the calendar doesn't close if `shouldCloseOnSelect={false}` is provided..

malcolmmadsheep:
@nirhart Thanks for review. Yeap, it's a great idea to split functionality by several drivers and we will probably do it. We make this PR to ask about if we move the right way cause it's our first time with wix-style-react and first of all we wanted to adapt to this project and it's testing flow.

lbelinsk:
Yeah, I think it is good that you shared your progress, that is why I shared my thoughts as well..
Feel free to ask for more semi-reviews like this one whenever you feel like you need one..

nirhart:
@lbelinsk please merge this pr when you think its good enough

malcolmmadsheep:
@lbelinsk We've just pushed updated PR with split driver logic and updated tests according to it. Now we continue to work on other props, that we haven't covered yet. We also would like to know about what should be tested with e2e tests. We think that we should test here some key features like opening with focus on element or selecting date with keys. It seems like some tests may have the same logic as unit tests and we are confused about how should we split logic between the tests. WDYT about it?

src/DatePicker/DatePicker.spec.js
```diff
@@ -1,57 +1,234 @@
...
+      const date = new Date(2017, 9, 2);
+      const {inputDriver} = createDriver(<DatePicker
+        onChange={onChange} value={moment(date)}
+        dateFormat={'DD/MM/YYYY'}
+        />);
+
+      expect(inputDriver.getValue()).toBe('02/10/2017');
+    });
+
+    it('should set placeholder from props if value is not set', () => {
```
lbelinsk: It is ok, but lets talk about it tomorrow, I am not sure if we want to support this.
I think maybe the best usage will be instead of doing:
```
<DatePicker placeHoderText="text"/> 
```

To do:
```
<DatePicker customInput={<Input placeholder="text"/>}/>
``` 

In this way we don't mix props fof input with props for datepicker
I am not sure, let's think together tomorrow about it..

src/DatePicker/DatePicker.spec.js
```diff
@@ -1,57 +1,234 @@
...
+    it('has custom prefix', () => {
+      const {inputDriver} = createDriver(<DatePicker
+        onChange={onChange}
+        prefix={<span>#</span>}
+        />);
+
+      expect(inputDriver.hasPrefix()).toBe(true);
+    });
+
+    it('should show calendar on select with ArrowUp key', () => {
```
lbelinsk: This test is great, but it shouldn't be under this describe IMHO..

src/DatePicker/DatePicker.spec.js
```diff
@@ -1,57 +1,234 @@
...
-    onChangeMock = sinon.spy();
+  describe('date picker input', () => {
+    it('should exist', () => {
+      const {calendarDriver} = createDriver(<DatePicker onChange={onChange}/>);
 
-    mount(
-      <DatePicker onChange={onChangeMock}/>,
-      {attachTo: datePickerRoot}
-    );
+      expect(calendarDriver.exists()).toBeTruthy();
```
lbelinsk: You meant inputDriver, right?

src/DatePicker/DatePicker.spec.js
```diff
@@ -1,57 +1,234 @@
...
+          value={value}
+          onChange={onChange}
+          />
+      );
+      inputDriver.trigger('click');
+      calendarDriver.triggerOnNthDay({trigger: 'click', n: 1});
+
+      const newValue = onChange.mock.calls[0][0];
+
+      expect(onChange).toHaveBeenCalled();
```
lbelinsk: I like better the syntax toHaveBeenCalledWith() instead of the additional line newValue.diff(expectedValue...

src/DatePicker/DatePicker.spec.js
```diff
@@ -1,57 +1,234 @@
...
+    it('should close calendar on select with click', () => {
+      const {inputDriver, calendarDriver} = createDriver(<DatePicker onChange={onChange}/>);
+
+      inputDriver.trigger('click');
+      calendarDriver.triggerOnNthDay({trigger: 'click'});
+
+      expect(calendarDriver.isVisible()).toBe(false);
+    });
+
+    it('should close calendar on select with Enter key', () => {
```
lbelinsk: What about Tab, Escape keys...

src/DatePicker/DatePicker.spec.js
```diff
@@ -1,57 +1,234 @@
...
+          />
+      );
+
+      inputDriver.trigger('click');
+      calendarDriver.triggerOnNthDay({trigger: 'click'});
+
+      expect(onChange).not.toHaveBeenCalled();
+      expect(calendarDriver.isVisible()).toBe(true);
     });
+  });
```
lbelinsk: Tests are really great!
I have 2 more in mind:

1. Test that onSelect is not called if trying to select a selected value.. This is an optimization we try to keep in all components..

2. We didn't test the yearDropdown or the MonthDropdown at all..

lbelinsk:
About the e2e tests, they should be in high level..
Test a basic usage, open calendar, select a date, change the date withe keyboard and mouse click.

No need to test if the calendar is opening on click, or other unit tests like this because it is already covered by the component tests you wrote..

Try to expose some useful methods in the protractor driver, even if you are not using them... but make sure they work though :)

src/DatePicker/DatePickerInput.js
```diff
@@ -14,7 +12,8 @@ export default class DatePickerInput extends Component {
     dateFormat: PropTypes.func,
     style: PropTypes.object,
     onEnterPressed: PropTypes.func,
-    customInput: PropTypes.node
+    customInput: PropTypes.node,
+    dataHook: PropTypes.string
```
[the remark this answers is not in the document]
ihork: We are using `inputDataHook` property to pass it in customInput as dataHook property

src/DatePicker/DatePicker.js
```diff
@@ -62,7 +65,7 @@ export default class DatePicker extends Component {
 
   render() {
     return (
-      <div className={css.wrapper}>
+      <div className={css.wrapper} data-hook={this.props.dataHook}>
```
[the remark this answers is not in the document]
ihork: Looks great, thanks

src/DatePicker/DatePicker.spec.js
ihork: But how do you suggest to compare objects? When we click on next day `calendarDriver.triggerOnNthDay({trigger: 'click', n: 1});` new moment object will be created. I can't see a problem with `onChange.mock.calls[0][0]` - it's actually the same. We even don't need `expect(onChange).toHaveBeenCalled();` in this case.

src/DatePicker/DatePicker.spec.js
ihork: Yeap, we wanted to start tests in all directions, and now we 're going to cover it with all use cases.

src/DatePicker/DatePicker.spec.js
ihork: 1. Correct, will be done
2. We thought about should we need to cover it in e2e or not

lbelinsk:
Failing on ci due to lint errors

lbelinsk:
@malcolmmadsheep I am merging this pr after the build will pass..
Don't forget to update the documentation for the updated testkit methods in the next pr..

src/DatePicker/DatePicker.spec.js
```diff
@@ -252,11 +252,32 @@ describe('DatePicker', () => {
...
       expect(calendarDriver.isVisible()).toBe(true);
     });
+
+    describe('with year dropdown', () => {
+      it('should give a possibility to choose date from another year', () => {
+        const date = moment(new Date(2015, 9, 2));
+        const {calendarDriver, inputDriver} = createDriver(
+          <DatePicker
+            onChange={onChange}
+            showYearDropdown={true}
```
lbelinsk: no need for the ={true}

lbelinsk:
You have again lint errors...
Please run yarn build before you push..

One more thing - dataHook={'datepicker-default'}`
Please change to 'story-datepicker-default' as this is the convention across the lib..

The build should be green after fixing the linting errors

ihork:
Yeap, sorry for that. We had started e2e and we weren't prepared to merge, so we pushed just to sync between us.

lbelinsk:
ok.. Let's try to make this branch green.. It is red from the beginning (mainly because of yoshi) so we will be able to merge after the e2e's are ready

src/DatePicker/DatePicker.protractor.driver.js
```diff
@@ -0,0 +1,33 @@
...
+  const getCalendar = () => component.$('.react-datepicker');
+  const getNthAvailableDay = n => component.$$('[role="option"]:not([class*="outside-month"])').get(n);
+  const getYearDropdown = () => component.$('[class$="year-read-view"]');
+  const getNthYear = n => component.$$('[class*="year-option"]').get(n);
+  const getMonthsDropdown = () => component.$('[class$="month-read-view"]');
+  const getNthMonth = n => component.$$('[class*="month-option"]').get(n);
+
+  return {
+    input: {
+      exists: () => getInput().isPresent(),
```
lbelinsk: I think that I would want a user to use it like: `datePickerDriverFactory({dataHook: data-hook}).inputDriver.getValue()`
So rename the input to inputDriver..

src/DatePicker/DatePicker.protractor.driver.js
```diff
@@ -0,0 +1,33 @@
...
+      isDisplayed: () => getInput().isDisplayed(),
```
lbelinsk: It is more common (at least in this lib) to call it isVisible instead of isDisplayed..

src/DatePicker/DatePicker.e2e.js
```diff
@@ -0,0 +1,137 @@
...
+  beforeEach(() => {
+    browser.get(storyUrl);
+  });
+
+  eyes.it('should not open calendar when disabled', () => {
+    const {input, calendar} = datePickerTestkitFactory({dataHook: 'story-datepicker-disabled'});
+
+    input.click();
+
+    expect(calendar.exists()).toBe(false);
```
lbelinsk: isVisible instead of exists

src/DatePicker/DatePicker.e2e.js
```diff
@@ -0,0 +1,137 @@
...
+    });
+
+    eyes.it('should close calendar on Escape key', () => {
+      const {input, calendar} = testkit;
+
+      input.click();
+      expect(calendar.exists()).toBe(true);
+
+      input.pressEscKey();
+      expect(calendar.exists()).toBe(false);
```
lbelinsk: isVisible

src/DatePicker/DatePicker.e2e.js
```diff
@@ -0,0 +1,137 @@
...
+      calendar.clickOnNthAvailableDay(1);
+
+      expect(calendar.exists()).toBe(false);
+    });
+
+    eyes.it('should close calendar on Escape key', () => {
+      const {input, calendar} = testkit;
+
+      input.click();
+      expect(calendar.exists()).toBe(true);
```
lbelinsk: isVisible

lbelinsk:
Build is red only because eyes.it fails on the new datePicker examples, so it is all good.. will merge after the master build will be green.
Thanks a lot  @ihork and @malcolmmadsheep really great work!

nirhart:
@lbelinsk

PR #764, opened 2017-08-28
feat(proper-imports): exports dist components from root

argshook:
this naively reads dist/src, takes all folders that have index.js and
puts componentName.js in root with `module.exports = require('./dist/src/componentName')`

this way it's possible to `import componentName from 'wix-style-react/componentName';`

@ranyitz something what we spoke about today, maybe good enough for starters


ranyitz:
@argshook, That looks wonderful.

What left to do IMHO:
1. Verify that all components follow the convention.
2. Test the script.
3. Think about Whether to put it on `.gitignore` or not.
If it will be on git, consider a pre-push hook, if it's not, consider a pre-release script.

nirhart:
@argshook do you see any case which it will break the current users?

argshook:
@nirhart it shouldn't because i only add new files and do not change any existing ones, all previous paths remain the same

ranyitz:
@argshook, @nirhart, What do you say about turning wix-style-react into a [monorepo](https://github.com/lerna/lerna) - every package would be isolated, development would be on one repository. there is no use case of taking the whole wix-style-react anyway.

nirhart:
@ranyitz I think that the downside of it its a much more difficult development process, and this library is under heavy development these days

PR #765, opened 2017-08-29
allow sending additional classnames

eyaleizenberg:


nirhart:
I'm sorry but I have to reject it. Wix-Style-React should support all the design you need and unify the design across Wix, if you have a different UX then the library then either we did it wrong or you.
It should go to the library UX Ben Benhorin first, and he can either approve it and we will all use your design or reject it and you will need to use the library design.

PR #767, opened 2017-08-29
fix: FieldWithSelection style

Varnius:
Fixed:
> 3.3 Field with selection check box size is 20px. Please update it to the right size 16px. Plus, checkbox label style should be T3.1


Varnius:
@nirhart Well, the actual component is /stories/FieldWithSelectionComposite and internally it uses FieldWithSelection as a template but template does not specify any of its children, including the checkbox I fixed.

PR #769, opened 2017-08-29
Underline Style doesn't work on TextLink in storybook

iaroslavshvets:
Underline Style doesn't work. When Underline style is on, the underline must be always visible


nirhart:
You changed a prop name, it will break people that uses it now, lets keep the props of the component the same

nirhart:
@iaroslavshvets

iaroslavshvets:
But we are missing 3rd state in this case. Boolean doesn't cover it. That's why I've to change it to string and give different name. Or as a compromise I can leave old prop as well,  but Mark it deprecated.

nirhart:
@iaroslavshvets Its not a real error, its just a bad storybook, please see this - https://github.com/wix/wix-style-react/pull/779

PR #776, opened 2017-08-30
tooltip onHide trigger

galmorad:


src/Tooltip/Tooltip.js
```diff
@@ -265,7 +270,7 @@ class Tooltip extends WixComponent {
       clearTimeout(this._showTimeout);
       this._showTimeout = null;
     }
-
+    this.props.onHide && this.props.onHide();
```
nirhart: This is not the actual place of hide, it should go where ReactDom.unmountComponent is (a couple of rows down)

galmorad: somehow in tests, it does not unmount, just sets visible to false , causing test to fail

galmorad: I added the call in two places, after unmount and in case it was called and visible state is false.
I checked in stroybook and it is triggered only once

src/Tooltip/Tooltip.js
```diff
@@ -283,6 +286,8 @@ class Tooltip extends WixComponent {
           this.setState({visible: false});
         }
       }, this._unmounted ? 0 : this.props.hideDelay);
+    } else {
+      this.props.onHide && this.props.onHide();
```
nirhart: Why would you call onHide if you are already hidden?

nirhart: @galmorad

PR #781, opened 2017-08-31
Project structure improvements

ranyitz:
* move .md files into docs direcotry
* fix gitignore script to ignore autogenerated components
* add `/n` in the end of each autogenerated component for the linter


scripts/export-components.js
```diff
@@ -17,7 +17,7 @@ fs
 
     return [
       `./${name}.js`,
-      `module.exports = require('./${componentPath}');`
+      `module.exports = require('./${componentPath}');\n`
```
argshook: @ranyitz does it work for you?

`\n` for me only works if i also add a regular line break, like so:
```
        `module.exports = require('./${componentPath}');
\n`
```

.gitignore
```diff
@@ -13,3 +13,6 @@ storybook-e2e
 *.iml
 .vscode
 test/e2e
+/*.js
```
argshook: @ranyitz i think you don't need forward slash

PR #785, opened 2017-08-31
PopoverMenu: testkit exists fix

skyiea:
`wrapper` always exists, thus it caused an issue when `testkit.exists()` always returned true.


skyiea:
Hold on with it pls.

src/PopoverMenu/PopoverMenu.spec.js
```diff
@@ -81,11 +81,13 @@ describe('Testkits', () => {
     </PopoverMenu>
   );
 
-  describe('Using ReactTestUtils testkit', () => {
-    expect(isTestkitExists(genComponent(), popoverMenuTestkitFactory)).toBe(true);
+  it('Using ReactTestUtils testkit', () => {
+    expect(isTestkitExists(<span/>, popoverMenuTestkitFactory)).toBe(false);
```
lbelinsk: This is a nice idea to check the negation..
Let's not treat this component by it's own, but instead we can add some wrapper function, something like `ValidateTestkit()` that will check the 2 things you are checking (true and false with `<span/>`).
What do you think? It might reduce a lot of future code duplication..

skyiea: Absolutely vote for it.
When I was creating this testkit, it wasn't obvious to include 'should not exist' test-case. Might be the same issue in the future (or even existing) testkits.

PR #796, opened 2017-09-07
Handling row click in case of having rowDetails passed and onRowClick…

iluxav:
… prop not specified


src/DataTable/DataTable.js
```diff
@@ -173,10 +178,7 @@ class DataTable extends WixComponent {
     }
   }
 
-  tryToggleRowDetails = (eventHandler, selectedRow) => {
-    if (eventHandler !== 'onClick') {
-      return;
-    }
+  toggleRowDetails = (eventHandler, selectedRow) => {
```
lbelinsk: you call this function with incompatible params.. should be (rowData, rowNum)
why eventHandler? you know that this is onClick now, not just some handler..

src/DataTable/DataTable.js
```diff
@@ -103,13 +103,19 @@ class DataTable extends WixComponent {
...
+  }
+
   renderRow = (rowData, rowNum) => {
     const {onRowClick, onMouseEnterRow, onMouseLeaveRow, rowDataHook, dynamicRowClass, rowDetails} = this.props;
     const rowClasses = [this.props.rowClass];
     const optionalRowProps = {};
 
     const handlers = [
-      {rowEventHandler: onRowClick, eventHandler: 'onClick'},
+      {rowEventHandler: this.onRowClick.bind(this, rowData, rowNum), eventHandler: 'onClick'},
```
lbelinsk: Please do the bind in the constructor instead..

iluxav: Well we still need to pass the params, so it might be like () => this.onRowClick(rowData, rowNum) but this is also bad, let's have a talk on Monday, I'll explain.

src/DataTable/DataTable.js
iluxav: Correct. Actually, rowNum will be enough.

src/DataTable/DataTable.js
lbelinsk: I don't understand.. we used to write:
```{rowEventHandler: onRowClick, eventHandler: 'onClick'}```

Now why can't we just write: 
```{rowEventHandler: this.onRowClick, eventHandler: 'onClick'}```
And if you need to bind something just do it in the constructor.

Comment (iluxav) [reply src/DataTable/DataTable.js]: On Click need the selectedRow number, and obviously it's dynamic, and we can get it once renderRow function is called. How can you bind rowNumber in contractor? Or I'm missing something?

File: src/DataTable/DataTable.js [hunk 3]
```diff
@@ -173,10 +178,7 @@ class DataTable extends WixComponent {
     }
   }
 
-  tryToggleRowDetails = (eventHandler, selectedRow) => {
-    if (eventHandler !== 'onClick') {
-      return;
-    }
+  toggleRowDetails = selectedRow => {
```
lbelinsk: Looks great.. I wanted to suggest an alternative implementation, but for some reason I don't have permissions to push to this branch :/

```
toggleRowDetails = rowNum => {
    const {selectedRows} = this.state;

    let newSelectedRows = this.props.allowMultiDetailsExpansion ?
      {...selectedRows, [rowNum]: !selectedRows[rowNum]} :
      {[rowNum]: !selectedRows[rowNum]};

    this.setState({selectedRows: newSelectedRows});
  }
```

Please add it or say it is not good, then I will merge this branch..
Thanks

iluxav: :) well it will be almost identical after transpiling it to es5, but sure is a bit shorter. The only issue that I personally have with these approach is that many operations done in the same line, "if" statement and object extension, therefore could be more annoying to debug. But that just my preference. You can merge it if you want.

lbelinsk: Your pr, your call :)
As you said, it is a manner of preference..

Will merge it on Sunday

iluxav: Thats fine. Push the button!

PR #797, opened 2017-09-07
fix conditional PopoverMenuItem rendering

ipanasenko:


src/PopoverMenu/PopoverMenu.spec.js
```diff
@@ -48,6 +48,12 @@ describe('PopoverMenu', () => {
           dataHook={menuItemDataHook}
           text="Menu Item #2"
           />
+        {false && (
```
lbelinsk: It looks like a healthy fix, and it is tested even if you don't add any test because we check that the items length is 2 and not 3.

But maybe adding an explicit test will make things more straight forward to understand. it('should filter falsy items...')

ipanasenko:
okay, I will add it, thanks.

ipanasenko:
@lbelinsk done

PR #798, opened 2017-09-07
update yarn.lock

ipanasenko:


lbelinsk:
@ipanasenko We still don't have a good solution for maintaining the yarn.lock file.
The current decision is to keep the yarn.lock file empty..

ipanasenko:
What's the problem with it? Can you give more details?

lbelinsk:
The idea is that we don't want to have outdated yarn.lock file, because then what runs on ci will not be the same like running locally.

We don't have a good solution today of how to maintain the yarn.lock, because for example yoshi's package is being updated after every successful build.

On ci we run with --no-lockfile, so if the yarn.lock is empty, it will be the same when running locally.
Not the best solution, but still didn't came up with a better solution.

If you wish to continue this discussion, write on fed-infra slack channel..

PR #802, opened 2017-09-07
Fail test after errors inside

kupriyanenko:
I want to prevent incorrect using of props in my project, and for that I need to have valid using of prop types in my dependencies.

Also it will help to avoid potential issues with API changing. 

**Before we will merge this PR, we need to fix all errors**

### DIFF
--- a/package.json
+++ b/package.json
@@ -125,6 +125,7 @@
     "setupFiles": [
       "<rootDir>/node_modules/babel-polyfill"
     ],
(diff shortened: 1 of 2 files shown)


nirhart:
@kupriyanenko thats great, I changed a bit the setup-jest to work well with wallaby as well and fixed all current errors, merged it in a different branch https://github.com/wix/wix-style-react/pull/808

PR #806, opened 2017-09-10
Change dropdownlayout default tabindex to 0

liorur:


lbelinsk:
@liorur can you just explain what is the reason for this change?

shlomitc:
@liorur - Please update the MD file as well for the default value.

liorur:
generally it is a bad practice to use tabindex=1, as a general guideline we refrain  from using any tabindex other than 0 or -1.

specifically i am making use of googleAddressInput that uses dropdownLayout internally, i have no way to pass tabindex to it

PR #831, opened 2017-09-15
fix: InputWithOptions composition mode (fixes #555)

### LINKED ISSUE #555: Bug in Japanese typing for Google Address Input
Problem is that in Japanese, we type the address and need to press Space, and Enter in order to chose Kanji (Chinese Letter). 
Then, the moment you press enter, the typing cursor is out of the entry field and you need to click on it again to continue typing. And we can not complete the entire home address.

Debug:
1) Set keyboard to "Hiragana" (Japanese)
2) Type Below;
Kumamotoken> Space, Enter > Kumamotoshi> Space, Enter> Higashiku> Space, Enter> Enokimachi> Space, Enter> 23-26
You will be rejected to type "23-26"
If you need to show to the Developers, just feel free to contact me Takanori Kawaharada
I will come to Debug in Japanese.

Thanks!
Taka & Kazu

Varnius:
Fixed InputWithOptions to react to external compositor events (used with languages such as Japanese, Chinese etc). Fixes #555.

### DIFF
--- a/src/DropdownLayout/DropdownLayout.js
+++ b/src/DropdownLayout/DropdownLayout.js
@@ -79,7 +79,7 @@ class DropdownLayout extends WixComponent {
   }
 
   _onKeyDown(event) {
-    if (!this.props.visible) {
+    if (!this.props.visible || this.props.isComposing) {
       return false;
     }
 
--- a/src/DropdownLayout/DropdownLayout.spec.js
+++ b/src/DropdownLayout/DropdownLayout.spec.js
@@ -131,6 +131,13 @@ describe('DropdownLayout', () => {
     expect(onSelect).toBeCalledWith(options[5], false);
   });
 
+  it('should not call onSelect when composing text via external means', () => {
+    const onSelect = jest.fn();
+    const driver = createDriver(<DropdownLayout visible options={options} onSelect={onSelect}/>);
+    driver.pressEnterKey();
+    expect(onSelect).not.toBeCalled();
+  });
+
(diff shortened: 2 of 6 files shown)


Takawix:
Hi,

I confirmed from my environment in Israel, but my collegue tied in US and it's not working.

I have tried by VPN-US and it doesn't work properly. Google API doesn't find the address.

Why does it happen?

Taka

Varnius:
@Takawix can you elaborate?
Not sure how VPN is connected to this issue. If you see (and are able to enter) the correct symbols in the input but the address suggestion won't work then it's probably unrelated issue.

Takawix:
Hi @Varnius 

Kazu/Marketing writer in US tried to input address using VPN(location in Japan) and it worked.

You can see here: 
Example of succeeding to find address in ISR,JP: https://www.screencast.com/t/Dx5cv8wnK

Failure example from US environment: (Kazu's comment)
I tested the following address for the famous Tokyo Tower (so we know it exists): 東京都港区芝公園4-2-8
When I input 港区 and hit enter, it auto-fills 南青山 which is not my intended address. When I remove 南青山 and continue typing, I get rejected after reaching 4-2-8.
Here is the screen capture: 
https://drive.google.com/a/wix.com/file/d/0B1h-i96AoGiAdDBLN3JGa2NOdTg/view?usp=sharing
Time stamp: 
0:06 - Auto-fills 南青山
0:15 - Rejects 4-2-8

Are you located in Israel?
If you have any question, just we can meet and talk. 👍

PR #836, opened 2017-09-18
Datepicker/current day

justame:
Update current day color for the date picker

image


lbelinsk:
@justame the build is red only because of the eyes.it, so it is ok..
About the styles, I think this is not aligned properly:
image

I think the grey circle background should be smaller + the number inside is not exactly aligned.

Please get Ben's approval for this, then we can merge..

PR #839, opened 2017-09-19
Fixed onCancel and onConfirm handlers. Add support of drag/click events for picking colors.

vladik7244:


src/ColorPicker/color-picker-actions.js
```diff
@@ -8,11 +8,11 @@ import css from './color-picker-actions.scss';
 
 const ColorPickerActions = ({onCancel, onConfirm}) =>
   <div className={css.root}>
-    <Button height="small" theme="icon-standardsecondary">
-      <Close size="12px" onClick={onCancel}/>
+    <Button height="small" theme="icon-standardsecondary" onClick={onCancel}>
```
mastertheblaster: Good catch, missed that :)

src/ColorPicker/color-picker-hsb.js
```diff
@@ -12,15 +15,48 @@ export default class ColorPickerHsb extends WixComponent {
...
     const {current} = this.props;
     const hue = current.saturationv(100).lightness(50);
     const style = {
       left: `${current.saturationv()}%`,
-      top: `${100 - current.lightness()}%`
+      top: `${100 - current.value()}%`
     };
     return (
-      <div className={css.root}>
+      <div className={css.root} ref="gradient" onMouseDown={this.onMarkerDragStart}>
```
mastertheblaster: @vladik7244 Guess you need to `bind(this)` function in constructor, otherwise `this` reference will be lost.

src/ColorPicker/color-picker-hsb.js
```diff
@@ -12,15 +15,48 @@ export default class ColorPickerHsb extends WixComponent {
...
+    this.gradientRect = getBoundingRect(this.refs.gradient);
+    this.setNewColorByMouseEvent(e);
+  }
+
+  onMarkerDrag = e => {
+    this.setNewColorByMouseEvent(e);
+  }
+
+  onMarkerDragEnd = () => {
+    window.removeEventListener('mousemove', this.onMarkerDrag);
```
mastertheblaster: @vladik7244 Guess the events should be removed when component is being destroyed too?

src/ColorPicker/color-picker-hsb.js
vladik7244: @mastertheblaster These methods would be automatically bound because it declared as lambda expression

src/ColorPicker/color-picker-hue.js
```diff
@@ -12,11 +15,39 @@ export default class ColorPickerHue extends WixComponent {
...
+    const h = this.getHueByMouseEvent(e);
+    onChange(color({h, s: current.saturationv(), v: current.value()}));
+  }
+
   render() {
     // HUE is an intenger value from 0 to 360.
     const percentage = (this.props.current.hue() / 360) * 100;
     return (
-      <div className={css.root}>
+      <div className={css.root} ref="slider" onMouseDown={this.onMarkerDragStart}>
```
mastertheblaster: @vladik7244 If i correctly remember `ref="..."` is obsolete and you should use a function instead?
```js
<div ref={e => this.slider = e}/>
```
The good thing is you don't need the `ReactDOM`

src/ColorPicker/color-picker-hsb.js
mastertheblaster: Yep, you are right. Missed that.

src/ColorPicker/color-picker-hue.js
vladik7244: @mastertheblaster Okay, will fix it

src/ColorPicker/color-picker-hsb.js
vladik7244: @mastertheblaster yes, you are right

PR #840, opened 2017-09-19
date-picker: triggering open and close

varzager:
using ref prop on DatePicker, will allow you to trigger 'open' and 'close' of the datePicker dialog.
this does not allow control of dialog, but triggers it to open or close


lbelinsk:
@varzager I am closing it to think about the alternative solution in react-date-picker.
If it will be relevant I will reopen and merge this..

.nvmrc
```diff
@@ -1 +1 @@
-6.10.0
```
lbelinsk: Maybe this is not relevant anymore...

stories/DatePicker/ExampleTriggerOpen.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import {DatePicker} from 'wix-style-react';
+import moment from 'moment';
+
+class ExampleTriggerOpen extends React.Component {
```
lbelinsk: @argshook I brought this component alive back from the dead. It was waiting for this discussion to end:
https://github.com/Hacker0x01/react-datepicker/pull/1048

We have `datePicker.focus()` and `datePicker.blur()`. How can we document it inside the new autoDoc mechanism?

src/DatePicker/DatePicker.spec.js
```diff
@@ -313,6 +313,25 @@ describe('DatePicker', () => {
         expect(newDate.year()).not.toEqual(date.year());
       });
     });
+
+    describe('When trigger open and close', () => {
+      it('should open calendar using ref', () => {
+        const {calendarDriver} = createDriver(<DatePicker onChange={onChange}/>);
+
+        calendarDriver.componentInstance().open();
```
lbelinsk: I would prefer a simpler API since this is too low level.
How about `calendarDriver.triggerOpenCalendar()`?

same for close

src/DatePicker/DatePicker.driver.js
```diff
@@ -22,7 +22,8 @@ const datePickerDriverFactory = ({element, wrapper}) => {
     clickOnYearDropdown: () => ReactTestUtils.Simulate.click(getYearDropdown()),
     clickOnNthYear: (n = 1) => ReactTestUtils.Simulate.click(getNthYear(n)),
     clickOnPrevMonthButton: () => ReactTestUtils.Simulate.click(getPrevMonthButton()),
-    clickOnNextMonthButton: () => ReactTestUtils.Simulate.click(getNextMonthButton())
+    clickOnNextMonthButton: () => ReactTestUtils.Simulate.click(getNextMonthButton()),
+    componentInstance: () => componentInstance
```
lbelinsk: See my comment below about the low level api.

stories/DatePicker/ExampleTriggerOpen.js
argshook: @lbelinsk once #1013 is merged, `<AutoDocs/>` will also display list of methods.

comment above method implementation same as you would do with prop and it will show up.

src/DatePicker/DatePicker.js
```diff
@@ -64,11 +64,20 @@ export default class DatePicker extends WixComponent {
     );
   }
 
+  open() {
```
lbelinsk: @argshook so here we just need to add a comment with explanation?
that's it?

argshook:
```js
/** this method is the greatest */
open() {
  // ...
}

/** this method is the worst */
crash(everything) {
  // ...
}
```

this will result in something like (printed below props table):

* __open()__: this method is the greatest
* __crash(everything)__: this method is the worst

markdown is also supported in comments

lbelinsk:
@varzager have any estimation about when will you be able to do these small changes?

PR #851, opened 2017-09-24
date-picker: mouse hover > keyboard flow

varzager:
**“Do not merge: For liorbe”**
selected > mouse hover > keyboard flow > current day
image


lbelinsk:
@justame In case we will merge this, can you confirm it is not conflicting with your changes from last week?

justame:
it seems to be o.k

PR #852, opened 2017-09-25
fix(FieldWitSelectionComposite): blur event propagation on input element

alex5il:

Inside `FieldWithSelectionComposite.js` - _onTextInputBlur method calls onBlur if it was received from the props object and changes the state of focus as it previously did.
- [x] Explain why the change is needed.
At the moment focus and blur callbacks on the input prop dont do anything since `_onTextInputBlur` and `_onTextInputFocus` override them.
I havent changed the `_onTextInputFocus` since there is a bug in it if we use the `onFocus` callback and call `focus()` method explicitly on the input element - needs further checking.
- [x] Make sure the build is passing on ci.


src/Composite/FieldWithSelectionComposite/FieldWithSelectionComposite.spec.js
```diff
@@ -33,6 +33,16 @@ describe('FieldWithSelectionComposite', () => {
     expect(driver.hasSelectionInput()).toBe(true);
   });
 
+  describe('input properties', () => {
```
lbelinsk: This test doesn't really test anything, you spy on input.blur and you simply calling it on your own :)


The test should start like:

```
const onBlur =  jest.fn();
const driver = createCompositeDriverFactory(<FieldWithSelectionComposite><Label/><Input onBlur={onBlur}/><Checkbox/></FieldWithSelectionComposite>);
```

Then you need to actually focus and blur, using the driver's methods, and expect **onBlur** toHaveBeenCalled

PR #856, opened 2017-09-25
PopoverMenu testkit: fixing lookup for multiple data hooks

skyiea:
In #842 ability to use multiple data-hooks on PopoverMenu items was introduced.
This PR adds an ability to pass multiple space-separated data-hooks as `menuItemDataHook`. Inclusive `~=` attribute selector operator [won't work if value contains spaces](https://www.w3.org/TR/css3-selectors/#attribute-representation), thus need to workaround.


argshook:
@skyiea perhaps selecting with `*=` would work?

skyiea:
@argshook it won't in this case:
```
<PopoverMenuItem dataHook="item index=0 remove"/>

document.body.querySelectorAll('[data-hook*="item remove"]'); // returns []
```

argshook:
@skyiea why are multiple data hooks required, what's the use case?

i think `data-hook` should be treated as `id`: unique value, no whitespace.

```
  <PopoverMenuItem dataHook="menu-item-1">
  <PopoverMenuItem dataHook="menu-item-2">
  <PopoverMenuItem dataHook="menu-item-3">
```

then select individually or all with `document.querySelectorAll(`[data-hook*="menu-item"]`)`

this way you don't need special handling for multiple data hooks and no special syntax inside them (refering to `index=0`, `other.item remove.this`)

would this work for you? wdyt?

@liorbe

skyiea:
@argshook 
the use case is that in order to make `PopoverMenu` testkit work you need both `PopoverMenu` & `PopoverMenuItem` data-hooks. Each `PopoverMenuItem` within the same `PopoverMenu` should have same data-hook (testkit requirement). Everything seems ok, until you want to click on *certain* menu item. You have next options:
1) query by common menu items data-hook, and pick the one you want from returned elements array by *index*; that's bad (implementation might change ordering, then you'll need to update all tests)
2) add *multiple* data-hooks to menu items, making it work both for testkit initialization, and to query it directly (without guessing it's index)

At the moment `wix-style-react` is designed in such a way that `data-hook` is treated as kind of unique `id`, as you've said. But this approach has its own limitations (not only because of specific requirements of `PopoverMenu` testkit). Simple example:
Assume you have list of items, and in different tests you want to be able to click on the item based on its 1) index, 2) id, 3) title, 4) whatever else attribute. If `data-hook` is "classy", it will be as simple as:
```
<li data-hook="item index=4 id=guid1 action=update"/>

document.querySelector('[data-hook~="item"][data-hook~="id=guid"]').click();
document.querySelector('[data-hook~="item"][data-hook~="index=4"]').click();
document.querySelector('[data-hook~="item"][data-hook~="action=update"]').click();
```

But back to PR - it's just a tiny improvement of already existing code. Tilde operator is already there (for `PopoverMenu` testkit). Don't think we should spend much time on it.
I would gladly continue `data-hook` talk in some other thread, if you want :)

argshook:
hey, thanks for thorough explanation!

i don't like the idea of multiple data hooks. in my view they should be really simple, with no place for special syntax that might be confusing for others.

in your case i think it should be enough to use unique data hooks for each popover item and perhaps in driver have something like `getItemByDataHook` or similar. it could utilize `*=` attribute selector to either get all items or just specific one.

`PopoverMenu` testkit has such requirement because, if im not mistaken, you implemented it? if so, why not simplify.

@liorbe i'd like to know what you think about this

skyiea:
@argshook 
Honestly, I don't think that "classy" attribute should be confusing for FEDs. Yes, unique value is easier to query, but as I've said, you are quite limited with it when it comes to "descriptor" with different "attributes".
I think it's great idea to be able to hook some item from the list bases on its class and some attribute.

You are right, I've implemented it. FYI, `PopoverMenu` is a small layer on top of `Tooltip`, thus all weirdnesses and complications of the latter spread on `PopoverMenu` as well, and the only viable solution I came up with (I didn't want to touch `Tooltip`) was additionally binding `PopoverMenu` to its items via one more "classy" data-hook. You can find detailed explanation here: https://github.com/wix/wix-style-react/pull/692#discussion_r132141120

Not sure I understand your idea with `getItemByDataHook`.

nirhart:
@argshook is it approved by you?

argshook:
@nirhart @skyiea i don't like the idea of multiple data hooks and would like to avoid it. but as we don't have a clear solution of how to properly target nested specific components, let's not delay, i'll merge this

nirhart:
@argshook you can merge it and open an issue to solve it later

PR #898, opened 2017-10-03
feat(AutoExample): component for easy interactive examples

argshook:
introducing `<AutoExample/>` - sister of `<AutoDocs/>`.

example for `<ColorPicker/>`

in storybook, give this:
screenshot 2017-10-03 18 39 25

and you'll get this:
screenshot 2017-10-03 18 38 27

this is a POC, some things are still missing.
please feedback and ideas :)

this PR also includes `<AutoExample/>` usage for `<DatePicker/>`, run storybook to see it.

@lbelinsk


src/DatePicker/README.TESTKIT.md
```diff
@@ -1,6 +1,15 @@
...
+```js
+import datePickerPolyfills from 'wix-style-react/dist/src/DatePicker/Polyfills';
+```
+
+Before the tests and after each jsdom cleanup -
+```js
+datePickerPolyfills(window, global);
+```
 
 ## DatePicker TestKit API
```
lbelinsk: Maybe this line can also be autoGenerated?
<ComponentDisplayName> Testkit API

src/DatePicker/README.md
```diff
@@ -1,48 +1,7 @@
-# DatePicker component
+# DatePicker
 
-> A DatePicker component
+##### [source](https://github.com/wix/wix-style-react/blob/master/src/DatePicker/DatePicker.js)
```
lbelinsk: image

Not sure how configurable is it, but github link on the right to the **DatePicker** Header will be a perfect location. Easy to miss it now.

src/DatePicker/README.md
```diff
@@ -1,48 +1,7 @@
...
 
-## Testkit
-When testing DatePicker you might see some errors with unknown functions, to fix it - 
-```javascript
-import datePickerPolyfills from 'wix-style-react/dist/src/DatePicker/Polyfills';
-```
+`import DatePicker from 'wix-style-react/DatePicker';`
```
lbelinsk: This is really important to show this line. but it is not the correct path, right?
shouldn't it be `wix-style-react/src/DatePicker`

Also, we can probably automate it as well in the next phase..

stories/ColorPicker/index.js
```diff
@@ -1,21 +1,21 @@
...
-  .add('Color Picker', () => (
+  .add('ColorPicker', () => (
     <TabbedView tabs={['Example', 'API']}>
-      <div>
-        <InteractiveCodeExample title="Customize <ColorPicker/>" autoExpand={false}>
-          <ExampleDefault/>
-        </InteractiveCodeExample>
-      </div>
-      <AutoDocs source={ColorPicker}/>
+      <AutoExample
```
lbelinsk: Amazing!

stories/DatePicker/index.js
```diff
@@ -20,11 +27,20 @@ import ExampleControlledYearMonthDropdownsRaw from '!raw-loader!./ExampleControl
 
 storiesOf('Core', module)
   .add('DatePicker', () => (
-    <TabbedView tabs={['API', 'TestKit']}>
+    <TabbedView tabs={['Example', 'API', 'TestKit']}>
```
lbelinsk: Maybe "Usage" instead of example?
Your call..

stories/utils/Components/Form/index.js
```diff
@@ -0,0 +1,120 @@
+import React from 'react';
```
lbelinsk: Form is a bit confusing name imo.
Maybe FormComponents?

src/DatePicker/README.md
argshook: Nope, `wix-style-react/DatePicker` is good, no need for src dist or anything

stories/utils/Components/Form/index.js
argshook: Yes naming is quick and dirty. I think this all is a good candidate for extraction, if that's done it'd be good to split components to dedicated files, 'cause now it's like a bucket

src/DatePicker/README.TESTKIT.md
argshook: I think it's possible but I'll leave this for another PR

lbelinsk:
AutoExample is the hot sister :)

Maybe rename "interactive props" to "props"?
image

src/DatePicker/README.md
argshook: im thinking of improving `stories/index.js` so that it would have `const stories = [ ...{ component: DatePicker, source: RawDatePickerSource, github: 'github.com/wix/wix-style-react/...' } ]`

then iterate through `stories` and do all we need: configure storybook sidebar, display link to github, use AutoExample, use AutoDocs, display tesktit documentation, in future - set theme, and also probably something custom.

but i don't want to hold this PR for long so i'll do something like that in separate PR

PR #901, opened 2017-10-04
Fix `Row` CSS styles

adavidof:
CSS bug in the `before`, `after` with `display: table` in `Row` component

Gist:
https://gist.github.com/adavidof/30ebdb1e2fd5e3c21621ea6fef7ef2a9.js

Video:
https://screencast.com/t/drjRurfv
https://screencast.com/t/Z6VIskG0Ol

### DIFF
--- a/src/Grid/Grid.scss
+++ b/src/Grid/Grid.scss
@@ -83,3 +83,12 @@ $grid-row-margin-bottom: 30px;
 @media (min-width: $screen-lg-min) {
 	@include rtl-float-grid-columns(lg);
 }
+
+.row {
+	display: flex;
+
+	&:before,
+	&:after {
+		content: none;
+	}
+}


src/Grid/Grid.scss
```diff
@@ -83,3 +83,12 @@ $grid-row-margin-bottom: 30px;
 @media (min-width: $screen-lg-min) {
 	@include rtl-float-grid-columns(lg);
 }
+
+.row {
+	display: flex;
+
+	&:before,
+	&:after {
+		content: none;
```
argshook: @adavidof `content: none;` makes pseudo elements to not render entirely, and this breaks grid examples in storybook.

lbelinsk:
@adavidof what is going on?
Can we close it?
We don't want you to end up performing any hack on your code, it is important to solve it here.
Please see Arijus's comment.

lbelinsk:
@argshook do you have any other solution in mind?

argshook:
@lbelinsk i see the problem and, well, would think of it as really minor.

@adavidof if this is breaking you and you need to hack around, please tell.

in future i want to get rid of bootstrap grid entirely and rely on some more lightweight solution (with no clearfixes). hence im not jumping on to fixing this now as i see it as a very minor issue. please tell if it's not.

adavidof:
@argshook ok, thanks

PR #920, opened 2017-10-09
feat(DataTable): adding support for dynamic rowDataHook

selama:

- [ ] Added tests.
- [ ] Update documentation.

### DIFF
--- a/src/DataTable/DataTable.driver.js
+++ b/src/DataTable/DataTable.driver.js
@@ -17,7 +17,7 @@ const dataTableDriverFactory = ({element, wrapper, component}) => {
   return {
     getRowsCount,
     getRowsWithClassCount: className => values(getRows()).filter(elem => elem.classList.contains(className)).length,
-    getRowsWithDataHookCount: dataHookName => element.querySelectorAll(`[data-hook="${dataHookName}"]`).length,
+    getRowsWithDataHook: dataHookName => element.querySelectorAll(`[data-hook="${dataHookName}"]`),
     getRowWithDataHook: dataHookName => element.querySelector(`[data-hook="${dataHookName}"]`),
     getRowText: index => values(getRows()[index].querySelectorAll('td')).map(td => td.textContent),
     getRowClasses: index => values(getRows()[index].classList),
--- a/src/DataTable/DataTable.driver.js
+++ b/src/DataTable/DataTable.driver.js
@@ -17,6 +17,8 @@ const dataTableDriverFactory = ({element, wrapper, component}) => {
   return {
     getRowsCount,
     getRowsWithClassCount: className => values(getRows()).filter(elem => elem.classList.contains(className)).length,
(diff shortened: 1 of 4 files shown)


selama:
added support for dynamic rowDataHook.
it simplifies the ability to check whether the DataTable is rendered with correct data.

src/DataTable/DataTable.driver.js
```diff
@@ -17,6 +17,8 @@ const dataTableDriverFactory = ({element, wrapper, component}) => {
   return {
     getRowsCount,
     getRowsWithClassCount: className => values(getRows()).filter(elem => elem.classList.contains(className)).length,
+    getRowsWithDataHookCount: dataHookName => element.querySelectorAll(`[data-hook="${dataHookName}"]`).length,
```
lbelinsk: Code duplication here with the function
```
dataHookName => element.querySelectorAll(`[data-hook="${dataHookName}"]`)
```

Please extract it to a separated internal func at the top and use it.

src/DataTable/DataTable.spec.js
```diff
@@ -76,6 +76,21 @@ describe('Table', () => {
     expect(driver.getRowsWithClassCount(defaultProps.rowClass)).toBe(defaultProps.data.length);
   });
 
+  it('should assign dataHook to rows', () => {
+    const rowDataHook = 'row-data-hook';
+    const props = Object.assign({}, defaultProps, {rowDataHook});
+    const driver = createDriver(<DataTable {...props}/>);
+    expect(driver.getRowsWithDataHookCount(rowDataHook)).toBe(defaultProps.data.length);
```
lbelinsk: Please add another expect to check that the dataHook is correct. I mean add something similar to the next tets you wrote (line 91 for example)

src/DataTable/DataTable.js
```diff
@@ -80,8 +79,7 @@ class DataTable extends WixComponent {
         loader={this.props.loader}
         useWindow={this.props.useWindow}
         scrollElement={this.props.scrollElement}
-        >
-        {table}
+        >{table}
```
lbelinsk: why?

src/DataTable/DataTable.driver.js
selama: it isn't exactly a duplication: in getRowsWithDataHookCount querySelectorAll is used, and in getRowWithDataHook querySelector is used.

I could used querySelectorAll for getRowWithDataHook as well(and pop the first result), but I'm not sure it is good practice.

what do you say?

src/DataTable/DataTable.spec.js
selama: on it.

src/DataTable/DataTable.js
selama: I was strangling the es linter - changing it back to the original code style.

src/DataTable/DataTable.driver.js
lbelinsk: Right.. missed that. just leave it like that then...

PR #922, opened 2017-10-09
Docs/refactor storybook

argshook:
this add `createStory` function which can be used to easily create story page. With simple config object you get:

1. storybook config (calls `storyOf` so you don't have too)
1. `Usage`, `API`, `TestKits` tabs
1. link to github
1. markdown readme (if exists)
1. `AutoExample`
1. `AutoDocs`
1. testkits markdown
1. anything special you need for special component


.storybook/create-story.js
```diff
@@ -0,0 +1,54 @@
...
+  readme,
+  readmeTestkit,
+  source,
+  component,
+  componentProps = {},
+  github,
+  children = null
+}) =>
+  storiesOf(category, module).add(name, () =>
+    <TabbedView tabs={['Usage', 'API', 'TestKit']}>
```
lbelinsk: Maybe, just maybe, we want to allow tabs to be configurable?
It will be an optional prop with default value of ['Usage', 'API', 'TestKit'].

The thig is that it all looks super nice and generic, and I have no doubt it will reduce a lot of code duplication.

The thing is that if in the future we would want to start adding "accessibility" tab, so we won't be able to add it to all components at once, so some components would have the tab and some won't have.

.storybook/create-story.js
```diff
@@ -0,0 +1,54 @@
...
+      <div className={styles.usage}>
+        {readme ?
+          <Markdown source={readme} /> :
+          <Markdown source={`# \`<${name}/>\``}/>
+        }
+
+        {name &&
+          <div className={styles.githubLink}>
+            <TextLink
+              link={`https://github.com/wix/wix-style-react/blob/master/src/${name}`}
```
lbelinsk: Amazing!

PR #923, opened 2017-10-09
TextLink: added onClick

Roeia1:
added an onClick attribute to enable additional functionality for TextLinks (the most common is BI events when links clicked..)
when links are disabled the onClick disabled as well and won't be triggered.


src/BaseComponents/TextLink/TextLink.driver.js
```diff
@@ -8,6 +8,7 @@ const textLinkDriverFactory = ({element, wrapper, component}) => {
 
   return {
     exists: () => !!element,
+    click: () => ReactTestUtils.Simulate.click(element),
```
lbelinsk: Missing documentation for this method in README.TESTKIT.md

src/Backoffice/TextLink/TextLink.js
```diff
@@ -22,11 +23,11 @@ export default class TextLink extends WixComponent {
     disabled: false,
     download: false,
     rel: null,
-    target: null
+    target: null,
+    onClick: () => {}
```
lbelinsk: Redundant because you added it in BaseTextLink.defaultProps, so you can delete this line.

src/Backoffice/TextLink/TextLink.js
```diff
@@ -13,7 +13,8 @@ export default class TextLink extends WixComponent {
       download: PropTypes.bool,
       rel: PropTypes.string,
       target: PropTypes.oneOf(['_blank', '_parent', '_self', '_top', 'framename']),
-      ariaLabel: PropTypes.string
+      ariaLabel: PropTypes.string,
+      onClick: PropTypes.func
```
lbelinsk: Redundant because you added it in BaseTextLink.propTypes, so you can delete this line.

src/BaseComponents/TextLink/TextLink.js
```diff
@@ -25,16 +26,17 @@ export default class BaseTextLink extends WixComponent {
...
   );
 
   render() {
-    const {ariaLabel, disabled, link, children, download, rel, target, onMouseEnter, onMouseLeave} = this.props;
+    const {ariaLabel, disabled, link, children, download, rel, target, onMouseEnter, onMouseLeave, onClick} = this.props;
     const props = {
       download,
       href: `${link}`,
-      onClick: event => disabled && event.preventDefault(),
+      onClick: event => disabled ? event.preventDefault() : onClick(),
```
lbelinsk: Not sure, but maybe we should pass the event to onClick?
onClick(event)

WDYT?

src/BaseComponents/TextLink/TextLink.spec.js
```diff
@@ -79,6 +80,15 @@ describe('TextLink', () => {
...
     expect(driver.getTarget()).toBe('_blank');
   });
+
+  it('should call the onClick func when clicked', () => {
+    const onClickFunc = spy();
+    const driver = createDriver(<TextLink link="" onClick={onClickFunc}/>);
+
+    driver.click();
+
+    expect(onClickFunc.calledOnce).toEqual(true);
```
lbelinsk: 1. We prefer to use jest.fn() instead of sinon spy, but it is not mandatory.
2. If we decide we do want to pass the event to onClick, this test should check that it was called with an event.

src/BaseComponents/TextLink/TextLink.spec.js
```diff
@@ -79,6 +80,15 @@ describe('TextLink', () => {
...
+  });
```
lbelinsk: Need to add a test that checks that onClick is not being called on a disabled link.

Roeia1:
Thanks, Lior!
Good catch on the TextField, didn't notice that.
I reverted all the files back that was about that pr,
and Ben didn't give permission for that.
I'm waiting for a response from my company UX about it,
will update you as soon as I can.

src/BaseComponents/TextLink/TextLink.js
Roeia1: yeah, it can be useful in some cases.
I will add it.

src/BaseComponents/TextLink/TextLink.spec.js
Roeia1: I saw another place in the project that Sinon spy was used.
I don't know how to check if the event obj itself was sent as an argument,
so I'm checking if the num of args is 1 in the test.
Is it enough?

lbelinsk: I think that you can just remove this line in that case. It is not really important.

lbelinsk:
@Roeia1 is this ready to merge?

lbelinsk:
Going to merge it since it looks good IMO.
Thanks a lot, turns out some people was waiting for this :)

PR #956, opened 2017-10-12
Dropdown updates on each props change.

aarnoldaas:
This check was preventing component to update displayed value. We have use case where value changes even selectedId is not changed.

We have filters with count and count can be changed without selecting new dropdown value: 
screen shot 2017-10-12 at 16 40 35


lbelinsk:
@aarnoldaas I understand the problem.
What you did will fix the problem, but it will also harm the performance because now for any prop that changes, the `update` function will be called.

Maybe you can try to use `shouldComponentUpdate` lifecycle method in order to optimize this solution a little bit?
I thought that you can leave your change (call update every time) and in `shouldComponentUpdate` you can return true only if at least one item was **deeply** changed (deep-equal).

Wdyt?
I think it worth giving it a try.
If it will not be that straightforward you can at least open an issue for that so it will be taken care of in the future.

aarnoldaas:
@lbelinsk 
I didn't managed to write failing test. 
Maybe you can advise? Tried something like this:
  it('should update text when selected option changes', () => {
    const {driver, inputDriver, dropdownLayoutDriver} = createDriver(<Dropdown options={options}/>);
    driver.focus();
    dropdownLayoutDriver.clickAtOption(0);
    expect(inputDriver.getValue()).toBe('Option 1');
    options[0].value = 'Updated';
    driver.setProps({ options, selectedId:0 });
    expect(inputDriver.getValue()).toBe('Updated');
  });

driver.setProps() created new component.
How do you test components when props changes?

lbelinsk:
@aarnoldaas this test looks fine. I will be happy to help to debug it. First thing I would do is to add `console.log` inside the `componentWillReceiveProps` just to make sure it is being called.
Maybe we have a problem with the setProps function...
The setProps doesn't really create a new component, I remember reading about it in the past and ReactDOM.render is smarter than it looks. It will know how to update the component and will not unmount + mount it.

Let me know if this is even the problem (setProps).
Feel free to contact me tomorrow.

lbelinsk:
@aarnoldaas any updates?

aarnoldaas:
@lbelinsk When I was writing that test i tried console log but it didn't get called. I didn't had time to check again. Will do that tomorrow and let you know.

src/Dropdown/Dropdown.js
```diff
@@ -31,9 +31,12 @@ class Dropdown extends InputWithOptions {
...
 
   componentWillReceiveProps(nextProps) {
-    if (this.props.selectedId !== nextProps.selectedId) {
-      this.update(nextProps, {isFirstTime: false});
-    }
+    this.update(nextProps, {isFirstTime: false});
+  }
+
+  shouldComponentUpdate(nextProps) {
+    const {value} = (nextProps.options.find(({id}) => id === nextProps.selectedId) || {});
```
lbelinsk: @aarnoldaas looks like a good direction. Unfortunately, this component is semi-controlled. It means that we support both controlled and uncontrolled mode.

It means that people might be using `value` and `onSlect` prop, without using the `selectedId`. It is not good, we want to change it, but this is a living fact.

I need to think about it.
I am not sure if we want to risk it.
The best solution is to refactor this component to support only controlled mode, then this solution will not be needed probably, but in any case, it will work with it for sure, not like now.

Maybe we will just go with your previous solution, with no optimization. I mean, simply remove the shouldComponentUpdate.

After me sharing my doubts, I will be happy to hear what do you think.
FYI @argshook

src/Dropdown/Dropdown.spec.js
```diff
@@ -44,6 +44,16 @@ describe('Dropdown', () => {
     expect(inputDriver.getValue()).toBe('Option 1');
   });
 
+  it('should update text when selected option changes', () => {
```
lbelinsk: Test looks great

src/Dropdown/Dropdown.js
aarnoldaas: @lbelinsk I never like premature optimizations. It seems that this component can be used in many different ways. Implementing strict shouldComponentUpdate can break some use cases as you pointed it out. 
I hope somebody will register issue if this will be performance bottleneck in some applications.

aarnoldaas:
@lbelinsk I have removed shouldComponentupdate. 
Refactored "const options" to "getOptions()", tests was modifying original "options", there was possibility test sequence can break tests. 
I think this is ready to merge.

lbelinsk:
In this case, as long as we don't refactor then let's just remove the `shouldComponentUpdate`.
Sorry that I didn't think of the "uncontrolled" case before.

PR #994, opened 2017-10-17
Chore/backoffice button

argshook:
this includes two things:

1. i noticed that we have `ButtonLayout` and `src/Backoffice/ButtonLayout`, however, they look and behave the same. Thus, i remove `src/Backoffice/ButtonLayout` and reexport to `src/ButtonLayout`

1. before this, `Backoffice/Button` would use `Backoffice/ButtonLayout`. after, `Backoffice/Button` is standalone, does not use button layout. And the `ButtonLayout` itself reuses styling and proptypes from `Button`. this way i "inverse" dependencies: `Button` becomes standalone and `ButtonLayout` depends on it, not the other way around.

this makes it easier to think about themes, we don't need to pass props to `ButtonLayout`. components become separate, we don't need to nest them anymore


src/Backoffice/Button/Button.js
```diff
@@ -1,67 +1,107 @@
...
+    icon ?
+      <div className={className} data-hook={dataHook}>
+        {React.cloneElement(icon, {size: iconSize})}
+      </div> :
+      null
+  );
+};
 
 class Button extends WixComponent {
+  static displayName = 'Button'
```
nktssh: Since it's named class. maybe, it would be good to remove `displayName` property?

argshook:
if you received "unhandled pull request" email from github, ignore this.

Issue #969, opened 2017-10-15
Improve contribution doc


lbelinsk:
- [x] Explain how to run specific component tests with jest
- [x] Explain how to run specific e2e test

argshook:
contribution docs improved in #1018, so closing

PR #1001, opened 2017-10-18
Jest interactive mode

shlomitc:
Allow `jest` to run in interactive mode during watch
The great benefit is that it allows running specific test files or test patterns.
This will solve #969 first part


argshook:
simply changing order, wonderful! thank you!

PR #996, opened 2017-10-17
Google address input credit footer

Roeia1:
Giving the option to give credit in the GoogleAddressInput component.

PR #1004, opened 2017-10-18
Feat/google address input credit footer

argshook:
@Roeia1 extracted from #996  

---

Giving the option to give credit in the GoogleAddressInput component.


src/GoogleAddressInput/GoogleAddressInput.js
```diff
@@ -57,6 +59,15 @@ class GoogleAddressInput extends React.Component {
       });
     }
 
+    let poweredByGoogleFooter;
+    if (this.props.poweredByGoogle) {
```
argshook: @Roeia1 i would move this footer component to be class method maybe. then you don't need `let` and `if`

src/GoogleAddressInput/GoogleAddressInput.js
```diff
@@ -69,6 +80,7 @@ class GoogleAddressInput extends React.Component {
           onManuallyInput={this.onManuallyInput}
           value={value}
           options={options}
+          fixedFooter={this.props.poweredByGoogle ? poweredByGoogleFooter : null}
```
argshook: @Roeia1 this could be `fixedFooter={this.props.poweredByGoogle ? this.getGoogleFooter() : null}`

src/GoogleAddressInput/GoogleAddressInput.spec.js
```diff
@@ -65,6 +65,14 @@ describe('GoogleAddressInput', () => {
         disabled: true
       }]);
     });
+    it('should show the powered by google footer', () => {
+      const component = createShallow({Client: GmapsTestClient, poweredByGoogle: true});
+      expect(component.instance().props.poweredByGoogle).toEqual(true);
```
argshook: @Roeia1 are you sure you actually test for footer existance here? looks like you give a prop, and check that the prop is true. But does it actually render? you set `data-hook="google-footer"`, you can target it and see if it exists.

Roeia1: yeah, that's what I initially did, Nir Hartman changed it for some reason.

src/GoogleAddressInput/GoogleAddressInput.js
Roeia1: @argshook is this ok?
```const poweredByGoogleFooter = this.props.poweredByGoogle ?
      (<div className={styles['google-footer']} data-hook="google-footer">
        <img src={poweredByGoogleFooterImage} className={styles['google-img']}/>
      </div>) :
    null;
```

src/GoogleAddressInput/GoogleAddressInput.js
Roeia1: ```fixedFooter={poweredByGoogleFooter}```

src/GoogleAddressInput/GoogleAddressInput.js
argshook: that's fine but i would move it out of `render`

Roeia1: @argshook oh class method, ok got confused... yeah ofc.
should I make the getGoogleFooter method static?
it only returns a div with an img...

argshook: you can, of course, but for me it's good either way

PR #1006, opened 2017-10-18
feat: interactive auto example

vidmantas0010:
interactive auto example, added node and function


stories/utils/Components/AutoExample/index.js
```diff
@@ -178,12 +198,46 @@ export default class extends Component {
...
+            return acc;
+          }, {})
+      )
+    };
+
+    const codeProps = {
+      ...this.state.propsState,
+      ...(
+        Object
+          .keys(this.props.exampleProps)
```
nktssh: Please, specify the default value for `exampleProps`, otherwise, it will break the code

Issue #959, opened 2017-10-13
`AutoDocs` should display composed props with full data


argshook:
#931 & #932 are similar but not exactly.

our `AutoDocs` currently takes some component source as string but does not resolve any imports within that source.

If component has `propTypes = { ...propTypesFromOtherFile }` all `AutoDocs` does now is adds one row to generated table of props with something like "also includes props from: propTypesFromOtherFile".

what it should do is gather all those spread props and include them in generated props table (including description, isRequired, default value etc)

argshook:
WIP @ #997

argshook:
closed by #997 & #1009

PR #1009, opened 2017-10-19
Fix story function

nktssh:
Related to #959.
Fix bug introduced into #997.


stories/utils/Components/ComponentMetaInfoGetter/index.js
```diff
@@ -72,7 +72,7 @@ export default class ComponentMetaInfoGetter extends React.PureComponent {
           newSourcePath = match;
         });
 
-        newSourcePath = newSourcePath.replace(/^\.\//, '');
+        newSourcePath = newSourcePath.replace(/^\.\//, '/');
```
argshook: @nktssh how did this happen? :)

nktssh: It's allowing the ComponentMetaInfoGetter to get source recursively deep.
Before it was only 1 level deep, so all was fine.

PR #1018, opened 2017-10-20
Docs/story and friends

argshook:
some markdown updates


nktssh:
👍
Like the section about black magic)

PR #1020, opened 2017-10-20
enhancement(Input): introduced "clearButton" property

skyiea:
Introduced `clearButton` property (in addition to existing `onClear` callback); made *clear button* automatically clear input value and focus input.

This is a part of ongoing `Search` component implementation.

Note: will update `Input`'s Storybook stuff after formal approve.


src/Input/Input.spec.js
```diff
@@ -376,26 +401,125 @@ describe('Input', () => {
...
+        <Input
+          value="some value"
+          onClear={() => null}
+          />
+      );
+      expect(driver.hasClearButton()).toBe(true);
+    });
+
+    it('should invoke callback', () => {
+      const onClear = sinon.spy();
```
lbelinsk: Why did you import sinon if you have jest.fn which you just use in the previous tests?

PR #1021, opened 2017-10-21
enhancement(Input): story update

skyiea:
Input story update with `clearButton` property.


src/Input/Input.js
```diff
@@ -230,7 +230,8 @@ Input.defaultProps = {
   textOverflow: 'clip',
   maxLength: 524288,
   width: 'initial',
-  withSelection: false
+  withSelection: false,
```
lbelinsk: Actually we can remove both lines, no need to say default false

PR #1025, opened 2017-10-22
added a test kit for ButtonLayout, and a method to check for the comp…

naamagr:
The change is: 
1. added a testkit to the ButtonLayout component.
2. added a data-hook property to the ButtonLayout component.
3. Added a new method in the driver to check for the component theme.
4. added tests+documentation for the testkits.


src/ButtonLayout/ButtonLayout.spec.js
```diff
@@ -116,6 +120,7 @@ describe('ButtonLayout', () => {
       const driver = createDriver(someDivWithLayout({theme}));
 
       expect(driver.doesComponentHasClass(theme)).toBeTruthy();
```
lbelinsk: Can remove line 122, redundant

src/ButtonLayout/ButtonLayout.spec.js
```diff
@@ -124,4 +129,31 @@ describe('ButtonLayout', () => {
       expect(driver.doesComponentHasClass('hover')).toBeTruthy();
     });
   });
+
+  describe('testkit', () => {
```
lbelinsk: Better to use `isTestkitExists` helper func.
For example:
https://github.com/wix/wix-style-react/blob/f8e41fd233743738eb0034ce2d36b152d2dae112/src/FilePicker/FilePicker.spec.js#L26-L34

If it not an immediate change due to some unexpected issue please let me know, don't waste time

PR #1027, opened 2017-10-23
enhancement(InputWithOptions): introduced "showOptionsIfEmptyInput" property

skyiea:
Introduced `showOptionsIfEmptyInput` property (which is `true` by default, since it was default behavior before).

This is a part of ongoing `Search` component implementation, and the functionality brought with this property is necessary in order to implement `Search`.

### DIFF
--- a/src/InputWithOptions/InputWithOptions.spec.js
+++ b/src/InputWithOptions/InputWithOptions.spec.js
@@ -1,14 +1,17 @@
 import React from 'react';
 import ReactTestUtils from 'react-dom/test-utils';
+import {mount} from 'enzyme';
+
 import inputWithOptionsDriverFactory from './InputWithOptions.driver';
 import InputWithOptions from './InputWithOptions';
 import {createDriverFactory} from '../test-common';
+import {makeControlled} from '../../test/utils';
 import {inputWithOptionsTestkitFactory} from '../../testkit';
 import {inputWithOptionsTestkitFactory as enzymeInputWithOptionsTestkitFactory} from '../../testkit/enzyme';
-import {mount} from 'enzyme';
 
 const runInputWithOptionsTest = driverFactory => {
   describe('InputWithOptions', () => {
+    const ControlledInputWithOptions = makeControlled(InputWithOptions);
 
     const createDriver = createDriverFactory(driverFactory);
     const options = [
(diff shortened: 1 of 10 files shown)


src/InputWithOptions/InputWithOptions.spec.js
```diff
@@ -39,6 +42,48 @@ const runInputWithOptionsTest = driverFactory => {
       expect(dropdownLayoutDriver.isShown()).toBeTruthy();
     });
 
+    describe('showOptionsIfEmptyInput=false property', () => {
```
lbelinsk: I would separate it like this:
```
describe('showOptionsIfEmptyInput prop', () => {
   it('should show DropdownLayout on focus by default', () => {...})

//your new tests going here
```

src/InputWithOptions/InputWithOptions.spec.js
```diff
@@ -39,6 +42,48 @@ const runInputWithOptionsTest = driverFactory => {
...
+      it('should not show DropdownLayout if input is empty', () => {
+        const driver = createDriver(
+          <ControlledInputWithOptions
+            showOptionsIfEmptyInput={false}
+            options={options}
+            />
+        );
+
+        driver.inputDriver.focus();
+        expect(driver.inputDriver.isFocus()).toBe(true);
```
lbelinsk: No need (line 70) (trust the driver :))

src/InputWithOptions/InputWithOptions.spec.js
```diff
@@ -39,6 +42,48 @@ const runInputWithOptionsTest = driverFactory => {
...
+
+        driver.inputDriver.trigger('keyDown', {
+          key: 37  // <Left Arrow> key code
+        });
+        expect(driver.dropdownLayoutDriver.isShown()).toBe(false);
+
+        driver.inputDriver.enterText('some value');
+        expect(driver.inputDriver.getValue()).toBe('some value');
+        expect(driver.dropdownLayoutDriver.isShown()).toBe(true);
+        driver.inputDriver.clearText();
```
lbelinsk: This test is too long. Can create a new test for the last part where you test:
"enter text --> dropdown is open --> clear text --> dropdown is hidden

Issue #984, opened 2017-10-15
DropdownLayout


lbelinsk:
- [ ] Fix scrollbar to be on top
![image](https://user-images.githubusercontent.com/18248989/31590289-73c82358-b216-11e7-9a03-8034219ca2c1.png)

shlomitc:
Fixed here
https://github.com/wix/wix-style-react/pull/2007/files#diff-5341231b2beab209794b09006bf55d53

PR #1031, opened 2017-10-23
Remove styles which customized native scrollbar

ArtemAstakhovWix:
Fix for https://github.com/wix/wix-style-react/issues/984

- Remove styles which customized native scrollbar
- After discussion with @lbelinsk we decided to remove custom styles and use native scrollbar

### DIFF
--- a/src/DropdownLayout/DropdownLayout.scss
+++ b/src/DropdownLayout/DropdownLayout.scss
@@ -70,31 +70,6 @@ $top-arrow-size: 8px;
   opacity: 1;
 }
 
-.options::-webkit-scrollbar {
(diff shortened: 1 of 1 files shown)


lbelinsk:
Let's try to see if we really need it tomorrow, or we can still do something

lbelinsk:
@AudriusP @ArtemAstakhovWix please give it another try, see maybe there is a solution after all. maybe only coloring the scrollbar?

lbelinsk:
Closing it for now as we are not ready to decide whether to use the Editors custom scrollbar or simply to use the browsers one.

Might reopen it in the close future according to the decision

PR #1034, opened 2017-10-23
Migrate Label component to new self-documenting system

nktssh:
- [x] Update documentation.

I'm not sure whether it will be better to remove example now, so just left them as-is.


lbelinsk:
Thanks again!!!
See what I wrote in here: #1049 
And here: https://github.com/wix/wix-style-react/pull/1035

PR #1035, opened 2017-10-23
Migrate Checkbox story to new "story()"

nktssh:
- [x] Update documentation.


lbelinsk:
Thanks @nktssh .
We actually don't need all the example below, just the dynamic one on the right.
It will look like:

image

So the `example` property can be removed, and all files can be erased.
What is blocking us:

1. The dynamic checkbox on the right is not controlled. It means if you click it wouldn't get checked. 
2. After it will be controlled, we need to modify the e2e tests so it will run on the component, and will check the story on the way.
3. We want it to sit under `4.2 Selection`, not under core.

Since you only improved documentation, I will merge the pr, and if you have time to contribute with the things I pointed out, it will be really helpful!!!

See #1048

Issue #988, opened 2017-10-15
TextField


lbelinsk:
- [ ] Support required and info icon props when a label was not provided

PR #1038, opened 2017-10-24
Add ability to show tooltip and required attributes without label com…

ArtemAstakhovWix:
- Support required and info icon props when a label was not provided
- Fix for https://github.com/wix/wix-style-react/issues/988


src/Composite/InputAreaWithLabelComposite/InputAreaWithLabelComposite.spec.js
```diff
@@ -52,10 +52,22 @@ describe('InputAreaWithLabelComposite', () => {
       expect(driver.hasFieldLabelAttributes()).toBe(true);
     });
 
+    it('should FieldLabelAttributes exists if required and with one child', () => {
```
lbelinsk: The description is not really english. It is not your fault, there is alerady an existing string which you probably cp pasted.

What I suggest:
```
 describe('FieldLabelAttributes', () => {
   it('should exists... 
   it('should 
 });
```

PR #1041, opened 2017-10-24
Introduced Search component

skyiea:
Basic version of Search component.

What is missing:
- "icony" appearance (only magnifying glass in a circle, which expands into an input on click)
- highlighting matched text in dropdown options
- consider localizing `"Search"` placeholder


src/Search/Search.testkit-usage-e2e.js
```diff
@@ -0,0 +1,21 @@
+// Example used in Testkit README
```
lbelinsk: This file will not run. see in protractor.conf:
```
module.exports.config = {
  specs: ['test/**/*.e2e.js', 'src/**/*.e2e.js'],
```
It doesn't end with .e2e.js

src/Search/Search.testkit-usage-spec.js
```diff
@@ -0,0 +1,59 @@
+// Example used in Testkit README
```
lbelinsk: This file won't run. see in yoshi:
https://github.com/wix/yoshi/blob/master/yoshi/package.json#L13

stories/Search/index.js
```diff
@@ -0,0 +1,46 @@
...
+      />
+  </div>
+);
+
+story({
+  category,
+  storyName,
+  name: 'Search',
+  source,
+  readmeTestkit,
```
lbelinsk: Although you are passing it, the page is blank:
image

src/Search/Search.js
```diff
@@ -0,0 +1,100 @@
...
+      this.setState({
+        inputValue: nextProps.value
+      });
+    }
+  }
+
+  render() {
+    return (
+      <InputWithOptions
+        placeholder="Search"  // TODO: Localization?
```
lbelinsk: No need at this point. If someone needs to support localization, he will pass his own translated placeholder. You can remove this comment

src/Search/Search.js
```diff
@@ -0,0 +1,100 @@
...
+
+    const searchText = this._isControlled ? value : this.state.inputValue;
+
+    return options.filter(option => (searchText && searchText.length) ?
+      new RegExp(searchText.trim(), 'i').test(option.value) :
+      true
+    );
+  }
+
+  _onChange = e => {
```
lbelinsk: This is one approach, you support either state or prop.
The alternative could be supporting both. It will look like:
```
_onChange = e => {
  this.setState({inputValue: e.target.value});
  this.props.onChange && this.props.onChange(e);      
};
```

lbelinsk: wdyt? Will let you decide. If you do take the second approach, it means you would have to update also the constructor and other places.

lbelinsk:
About your comments, 
- highlighting matched text in dropdown options: GW takes care of it this week.
- consider localizing "Search" placeholder: See my comments in the cr

src/Search/Search.testkit-usage-e2e.js
skyiea: Yes, this is intentional. My comment means "This is an example of a test (not real component test), which is used within Search Testkit README", which implies I don't want it to be run as a part of `npm t`.
I found that almost all code examples are broken in some way (authors just wrote "pseudo-code", never bothering to check whether it is valid). Thus I decided it would be great *to be able to test it* before releasing README (by renaming it to `.e2e.js`/`.spec.js`, and then back).

stories/Search/index.js
skyiea: Strange. For me it's filled.
Are you sure you've built storybook and everything else?

src/Search/Search.js
skyiea: Hmm. Why would I need to support both?

src/Search/Search.testkit-usage-e2e.js
lbelinsk: I think that it is nice that you thought of that, but maybe this is a one time thing. You ran the test, saw it is not just a pseudo code but an actual working test, and then just copy it there, no?

If you prefer to leave it it is fine as well, just a bit confusing.

stories/Search/index.js
lbelinsk: I saw the rest of the tabs
If it works for you then fine, maybe i did something wrong, but just make sure to look after we merge that it all looks fine

src/Search/Search.js
lbelinsk: just simplifies the code, and keeping the state and the prop synched, just like you do in `componentWillReceiveProps`
We can keep it like that if you wish

skyiea: I would rather remove `componentWillReceiveProps` completely ;)
Thanks for the tip.

src/Search/Search.testkit-usage-e2e.js
skyiea: Yes, it is one time thing. Just thought it might give the future developers one more reason to actually test what they wrote in Testkit code examples before pushing changes.
I thought about some better naming, but nothing better came up in my mind.
It is confusing, though. Agree. Will remove it.

PR #1042, opened 2017-10-24
feat: introduced locale prop to DatePicker

audriuss:
Introduced locale prop to DatePicker and added readme.


lbelinsk:
close #846

PR #1044, opened 2017-10-24
add pushable prop to Slider component

ipanasenko:
same prop as here: https://www.npmjs.com/package/rc-slider#range


src/Slider/Slider.js
```diff
@@ -99,6 +100,10 @@ Slider.propTypes = {
   /** The slider's step */
   step: PropTypes.number,
 
+  /** Allow pushing of surrounding handles when moving a handle */
+  /** Number means a minimum distance between handles */
```
argshook: @lbelinsk fyi, only the second line of comment will be added to documentation, first one is ignored

lbelinsk: @ipanasenko can you take care of it maybe? small fix

ipanasenko: @argshook @lbelinsk https://github.com/wix/wix-style-react/pull/1050

PR #1052, opened 2017-10-26
feat: introduced Highlighter component

audriuss:
New component introduced in Core: The Highlighter. It highlights string type children by wrapping match with ```<strong>``` DOM element and adds provided color as background. It remains children elements structure. Mat
This component is added to the InputWithOptions. It allows to highlight the search term on options. Highlighting can be enabled by prop 'highlight'. It supports highlight color and case sensitiveness  as a prop. 

This PR is not complete, we still need to add the documentation. and story of Highlighter. Please check the idea, we are working documentation.

These changes are part of guild week tasks from @lbelinsk


src/Highlighter/HighlightedItem.js
```diff
@@ -0,0 +1,102 @@
+import React from 'react';
+import WixComponent from '../BaseComponents/WixComponent';
+import PropTypes from 'prop-types';
+import escapeRegExp from 'lodash/escapeRegExp';
+
+class HighlightedItem extends WixComponent {
+
+  static propTypes = {
```
lbelinsk: Use AutoDoc comments. Read about it on storybook

src/Highlighter/HighlightedItem.js
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  renderPlain(plainString, key) {
+    return <span key={key}>{plainString}</span>;
+  }
+
+  renderHighlight(matchString, key) {
+    return React.DOM.strong({
+      key,
+      style: {backgroundColor: this.props.color}
```
lbelinsk: Didn't check it in action, but is it backgroundColor that we want?
In any case, we don't want it to be too configurable, it should follow the typography (like we have for Text component).

src/Highlighter/HighlightedItem.js
```diff
@@ -0,0 +1,102 @@
...
+      return this.highlightChildren(children, matchRegExp);
+    }
+
+    return children;
+  }
+
+  hasMatchTerm() {
+    const {match} = this.props;
+
+    return !!((typeof match !== 'undefined') && match);
```
lbelinsk: default prop can simplify it here. no need for this function that way

src/Highlighter/Highlighter.driver.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
```
lbelinsk: Add README_TESTKIT

src/Highlighter/Highlighter.js
```diff
@@ -0,0 +1,121 @@
...
+
+  isReactElement(element) {
+    return element.props !== undefined;
+  }
+
+  renderHighlightItem(element, reactKey) {
+    const {match, color, caseSensitive} = this.props;
+
+    return (
+      <HighlightedItem
```
lbelinsk: looks like HighlitedItem and Highlighter has the same props at the moment, so maybe use {...props} instead of specifying each one.

src/Highlighter/Highlighter.spec.js
```diff
@@ -0,0 +1,120 @@
...
+    const wrapper = createDriver(
+      <Highlighter match="p" color={HIGHLIGHT_COLOR}>
+        <aside>Optionp</aside>
+      </Highlighter>
+    );
+
+    expect(wrapper.html()).toEqual(expectedResult);
+  });
+
+  it('should show highlighted text on with case sensitive match', () => {
```
lbelinsk: I would modify this to 'should match caseSensitive by default' test (without supplying any prop).

src/Highlighter/Highlighter.spec.js
```diff
@@ -0,0 +1,120 @@
...
+      expect(isTestkitExists(<Highlighter/>, highlighterTestkitFactory)).toBe(true);
+    });
+  });
+
+  describe('enzyme testkit', () => {
+    it('should exist', () => {
+      expect(isEnzymeTestkitExists(<Highlighter/>, enzymeHighlighterTestkitFactory)).toBe(true);
+    });
+  });
+
```
lbelinsk: There is no test for color, but actually I think it will be better to not support colors at this point. let's keep it simple, if we would want in the future we can extend it to support colors as well

src/InputWithOptions/InputWithOptions.js
```diff
@@ -77,12 +78,34 @@ class InputWithOptions extends WixComponent {
...
+  _processOptions(options) {
+    const {highlightMatch, highlight, highlightCaseSensitive, highlightColor} = this.props;
+
+    return !highlight ? options : (
+      options.map(option => {
+        return {
+          ...option,
+          value: (
+            <Highlighter
+              match={highlightMatch}
```
lbelinsk: highlightMatch should simply be the input value

stories/InputWithOptions/ExampleControlled.js
```diff
@@ -62,6 +62,8 @@ class ControlledInputWithOptions extends React.Component {
         onChange={onChange}
         onSelect={onSelect}
         onManuallyInput={onManuallyInput}
+        highlightMatch={this.state.value}
```
lbelinsk: Pay attention that value and highlightMatch is the equal, so it is not a needed prop (see my previous comment)

src/DropdownLayout/DropdownLayout.driver.js
```diff
@@ -57,12 +57,16 @@ const dropdownLayoutDriverFactory = ({element, wrapper, component}) => {
       ReactDOM.render(<div ref={r => element = r}>{ClonedWithProps}</div>, wrapper);
     },
     hasTopArrow: () => !!element.querySelector(`.${styles.arrow}`),
+    optionById(optionId) {
```
lbelinsk: Every function here is an arrow function, so I prefer we will stick to the same pattern here

src/Highlighter/HighlightedItem.js
```diff
@@ -6,15 +6,18 @@ import escapeRegExp from 'lodash/escapeRegExp';
 class HighlightedItem extends WixComponent {
 
   static propTypes = {
-    color: PropTypes.string,
     match: PropTypes.string,
     caseSensitive: PropTypes.bool
```
lbelinsk: Why not to drop caseSensitive? Let's just leave it as true without any prop.
Maybe I wasn't clear enough, but I meant that we probably always want to support case sensitive in WIX

src/Highlighter/README.TESTKIT.md
```diff
@@ -0,0 +1,70 @@
+# Highlighter Testkit
```
lbelinsk: Need to document both e2e and component drivers

src/InputWithOptions/InputWithOptions.js
```diff
@@ -271,7 +266,8 @@ InputWithOptions.defaultProps = {
   valueParser: option => option.value,
   dropdownWidth: null,
   dropdownOffsetLeft: '0',
-  showOptionsIfEmptyInput: true
+  showOptionsIfEmptyInput: true,
+  highlight: false
```
lbelinsk: no need for this line

src/Highlighter/HighlightedItem.js
audriuss: yeah, but, it's only on internal component 'HighlightedItem', which is not exposed. We thought that would be nice to leave if it will be needed later. It could be passed from 'Highlighter'

src/DropdownLayout/DropdownLayout.driver.js
audriuss: Okay, but how to refer to 'optionByHook' in an arrow function?

stories/InputWithOptions/ExampleControlled.js
audriuss: Oh, this is a leftover. Sorry, will remove

src/Highlighter/HighlightedItem.js
lbelinsk: hmm, ok, as long as the default is true it is fine

src/DropdownLayout/DropdownLayout.driver.js
lbelinsk: Oh, you can extract optionByHook to the top, and then to do:
```
const optionByHook = {...}
...

return {

optionById: optionId => optionByHook(`dropdown-item-${optionId}`),
optionByHook
```

lbelinsk: But you can leave it as you suggested, I didn't think about the lose of "this", so both solutions are fine

PR #1055, opened 2017-10-27
improve InputWithOptions Accessibility

ArtemAstakhovWix:
improve InputWithOptions Accessibility

- enter/arrowUp/arrowDown - opens options (focus)
- tab - moves to next element
- escape close dropdown or unfocus
- type any other key - nothing happens


lbelinsk:
@varzager This is the pr that takes care of accessibility issue. I will go over it next week, but you can take a look as well if you are interested

PR #1056, opened 2017-10-27
feat(oneOf): add oneOf children validator

Jefoss:
Added a 'oneOf' children validator, one of the specified types has to exist.


lbelinsk:
So we decided only to add the oneOf for now.
Will update about the dropdownComposite

PR #1066, opened 2017-10-30
correct proptypes

lbelinsk:


lbelinsk:
@shlomitc I verified with Ben, the closeButotn is optional. So I am merging this

PR #1069, opened 2017-10-30
update doc

lbelinsk:
Not sure why we suggest using babel loader because when users import from wix-style-react, they actually consume the transpiled files from dist.

I added wix-style-react to some project wich doesn't use yoshi, and I updated the doc on the fly.


argshook:
frankly, i only moved these explanations from older `readme.md` to `usage-without-yoshi.md` without really checking. if it works for you without some configs, then they're no longer needed.

PR #1078, opened 2017-11-02
WixStyleProvider as theme provider with HOC

lbelinsk:
One thing that I am not sure of is the React.PureComponent which we are using.
This was recommended by Eyal from biz manager, and Nir took his advise.
From going over the doc, I am not sure it is good for us:

https://reactjs.org/docs/react-api.html#reactpurecomponent.
I will talk to Eyal just in case.

Also, about the name, open to suggestions


src/providers/WixStyleProvider/index.js
```diff
@@ -0,0 +1,36 @@
...
+  children: PropTypes.any,
+  theme: PropTypes.string.isRequired
+};
+
+WixStyleProvider.childContextTypes = {
+  theme: PropTypes.string.isRequired
+};
+
+export function withTheme(WrappedComponent) {
+  class ThemeWrapper extends React.Component {
```
shlomitc: Why isn't this a `PureComponent`?

src/providers/WixStyleProvider/index.js
```diff
@@ -0,0 +1,36 @@
...
+    render() {
+      const extraProps = {theme: this.context.theme};
+      return React.createElement(WrappedComponent, {...this.props, ...extraProps});
```
shlomitc: Why did you need to use `createElement`? This could also work:
```
const newProps = {...this.props, theme: this.context.theme};
return <WrappedComponent {...newProps}/>;
```

And if you're a one-liner guy
```
const ThemeWrapper = (props, context) => <WrappedComponent {...{...props, theme: context.theme}}/>;
```

lbelinsk: Agree. I will push this tomorrow:
image

src/providers/index.js
```diff
@@ -0,0 +1,3 @@
+import WixStyleProvider from './WixStyleProvider';
+
+export {WixStyleProvider};
```
argshook: what's the purpose for this file? `import WixStyleProvider from 'wix-style-react/providers/WixStyleProvider';` would work without this, no?

src/providers/WixStyleProvider/index.js
```diff
@@ -0,0 +1,36 @@
...
+WixStyleProvider.propTypes = {
+  children: PropTypes.any,
+  theme: PropTypes.string.isRequired
+};
+
+WixStyleProvider.childContextTypes = {
+  theme: PropTypes.string.isRequired
+};
+
+export function withTheme(WrappedComponent) {
```
argshook: @lbelinsk i would separate this. `WixStyleProvider` is for consumers, `withTheme` is helper for our internal use, right?

src/providers/WixStyleProvider/index.js
```diff
@@ -0,0 +1,36 @@
+import React, {Children} from 'react';
+import PropTypes from 'prop-types';
+
+export default class WixStyleProvider extends React.PureComponent {
+  render() {
+    return Children.only(this.props.children);
```
argshook: @lbelinsk what's the reason for `only`?
i would just
```js
render() {
  return this.props.children;
}
```

src/providers/WixStyleProvider/index.js
argshook: i would go without jsx, there's really no need for it

```js
export default component => {
  const wrapper = (props, {theme}) =>
    React.createElement(component, {...props, theme});

  wrapper.contextTypes = {
    theme: PropTypes.object
  };

  return wrapper;
};
```

also, `theme` should probably be object but i guess as this is poc we start simple?

src/providers/index.js
lbelinsk: I just thought that we might have more providers in the future, so it will be a nice syntax to import from ./providers.

Since we don't have any other provider at the moment, i don't care if we remove it.

The only tradeoff is that users will do now:
```
import WixStyleProvider from 'wix-style-react/providers/WixStyleProvider';
```

Instead of:
```
import {WixStyleProvider} from 'wix-style-react/providers';
```

So maybe we can configure your export-script to allow nicer import api?

src/providers/WixStyleProvider/index.js
lbelinsk: From what I saw, It is a best practice to check that the provider wraps one element (the app).
It will encourage our users to actually wrap the whole App, and not from a lower point I guess.

src/providers/WixStyleProvider/index.js
lbelinsk: correct. yes, I can split them, just thought they are going hand to hand so they can live in the same file.
I will extract it to a different file

src/providers/WixStyleProvider/index.js
argshook: but do we need to encourage to wrap only whole app? wrapping sections can be valid use case, why not allow it

```js
export default () =>
  <WixStyleProvider theme={backoffice}>
    <Header>Im backoffice app</Header>

    <Content>
      <h1>I preview components</h1>

      <WixStyleProvider theme={mobile}>
        <MobileButton/>
        <MobileInput/>
      </WixStyleProvider>
    </Content>
  </WixStyleProvider>;

```

src/providers/index.js
argshook: even though this is longer, it should be the one we use:
```js
import WixStyleProvider from 'wix-style-react/providers/WixStyleProvider';
```
otherwise user imports all providers but might use just one, so bundle size, modules count blahblah

---
regarding export script, we can improve it, now it's pretty stupid.

all of this is anyway minor things im talking about :)

lbelinsk: Agree. removing this file then

src/providers/WixStyleProvider/index.js
lbelinsk: Well, I don't really care, just look at the famous providers like redux, translations, ...
Usually you will have an App that renders the whole structure you just showed,
then you do:

```
 <WixStyleProvider theme={backoffice}>
     <App/>
  </WixStyleProvider>
```

Then in tests it is easier to mock.

src/providers/WixStyleProvider/index.js
lbelinsk: I am not sure according to the docs whether we need to do that only for the root, or to all of the components in the app. According to Eyal I think that he told Nir that we don't need to, but from the last sentence (which is confusing) I think we might need to use the pureComponent for all the components:

image

PR #1082, opened 2017-11-05
#EE-10895|feat(tooltip): added min width to tooltip

danielagreen:
- added min width prop to tooltip
- need it to fix bug in tooltip component when its rendered from 'TextField' component


src/Tooltip/Tooltip.driver.js
```diff
@@ -40,6 +40,11 @@ const tooltipDriverFactory = ({element, wrapper}) => {
       const values = content.style._values;
       return values['max-width'];
     },
+    getMinWidth: () => {
+      const content = document.body.querySelector('.tooltip');
```
lbelinsk: Any chance to extract it to a function. It is already being repeated too much, and this is another one :)
Talking about the 
```
const content = document.body.querySelector('.tooltip');
```

To 
```
const getTooltipContent: () => document.body.querySelector('.tooltip');

```

src/Tooltip/Tooltip.spec.js
```diff
@@ -255,6 +255,26 @@ describe('Tooltip', () => {
     });
   });
 
+
+  describe('min-width attribute', () => {
```
lbelinsk: minWidth.
I see that above there is max-width, so if you can update it on the fly it will be nice

PR #1085, opened 2017-11-05
RadioGroup: fixed lineHeight prop

Roeia1:
When you put vAlign=top, the radio button isn't aligned with the text.
The lineHeight prop didn't work at all,
changed it so it now affects only the first line- the label,
so it can be leveled with the radio button.
(when you enter lineHeight 17px, it's good)

Can change the default value of the lineHeight prop now to 17px, instead of 24px,
but it's a breaking change so I didn't push it.


lbelinsk:
@Roeia1 There were 2 problems which I took care of:
1. You passed a string instead of an object to `style` so the code was broken.
2. There was a failing test since the driver needed to change a little bit.

Since the radioGroup is not looking good at the moment with the vAlign="top" (before the change), I am entering the breaking change.

I will announce it on slack.

Thanks!

close #1085

Issue #1090, opened 2017-11-06
`<Text/>` with appearance h0 - h4 adds unnecessary margins


argshook:
for now, as we don't yet have themes, we decided to simply remove margins added by browser when using `<Text/>` with `apperance`s `'H0', 'H1', 'H2', 'H2.1', 'H3', 'H4'`

lbelinsk:
Closing it as we have the new core text component with the updated typography.
@argshook if you see this issue returning in the new implementation please open an issue for it there as well in wix-ui-core

argshook:
@lbelinsk the `<Text/>` & `<UIText/>` does not have heading styles, so no more heading tags created by `<Text/>`? if so, then there's nothing to open new issue for :)

hepiyellow:
Related #1092

PR #1092, opened 2017-11-06
chore(Text): set margin 0 for heading tags

argshook:
also small refactor for classnames

closes: #1090


argshook:
im hesitant to merge since there might be projects that rely on those margins and so merging could break some layouts.

semver, come closer

PR #1102, opened 2017-11-09
added shouldUpdatePosition prop

ChenAlmagor:


src/Tooltip/Tooltip.driver.js
```diff
@@ -52,6 +52,12 @@ const tooltipDriverFactory = ({element, wrapper}) => {
       const values = content.style._values;
       return values.padding;
     },
+    getTooltipPosition: () => {
+      const content = getTooltipContent();
+      //const curRect = content.getBoundingClientRect();
```
lbelinsk: Is this here on purpose, or did you just forget these comments?

src/Tooltip/Tooltip.spec.js
```diff
@@ -292,6 +292,33 @@ describe('Tooltip', () => {
     });
   });
 
+  describe.skip('shouldUpdatePosition prop', () => {
```
lbelinsk: If you don't have time to solve this, please just open an issue for this so it won't get forgotten

stories/Tooltip_New/Composite/ExampleTooltip.js
```diff
@@ -18,7 +19,9 @@ class ExampleTooltip extends Component {
     onShow: () => console.log('onShow triggered'),
     onHide: () => console.log('onHide triggered'),
     onShowText: 'onShow triggered',
-    onHideText: 'onHide triggered'
+    onHideText: 'onHide triggered',
+    moveBy: {x: 0, y: 0},
+    shouldUpdatePosition: false
```
lbelinsk: no need

stories/Tooltip_New/Composite/ExampleTooltip.js
```diff
@@ -97,6 +100,38 @@ class ExampleTooltip extends Component {
                 />
             </div>
           </div>
+
```
lbelinsk: The story doesn't really demonstrate the new `shouldUpdatePosition` prop. I don't have an idea of how we can demonstrate it nicely, but at the moment I think it is better not to display it instead of what we have now...

src/Tooltip/Tooltip.spec.js
argshook: it's fine, i asked to commit these so that i can see what @ChenAlmagor wasn't able to test and hopefully help

ChenAlmagor: I think that if we won't find a solution for getting the position (in the component test) I should remove getPosition function from the driver, to avoid misleadings. e2e can cover this properly, but there is no e2e for the tooltip. I'll open an issue regarding this prop tests.

stories/Tooltip_New/Composite/ExampleTooltip.js
ChenAlmagor: maybe we can create another tooltip that will be always active, and add an explanation to change the moveBy values. but I'm fine with remove those files and open an issue to add a demonstration for this prop.

lbelinsk: On a second thought, just leave it like this because we are going to change all examples to use some automated documentation tool...

PR #1118, opened 2017-11-13
[ToggleBox] Added support for colors, in order to add component to Wi…

3LOK:
Added support for colors, in order to add component to Wix Viewer


src/ToggleSwitch/ToggleSwitch.js
```diff
@@ -68,14 +78,22 @@ ToggleSwitch.propTypes = {
   disabled: PropTypes.bool,
 
   /** Specifies toggle size  */
-  size: PropTypes.oneOf(['x-small', 'small', 'large'])
+  size: PropTypes.oneOf(['x-small', 'small', 'large']),
+
+  /** Specifies toggle colors */
+  colorUnchecked: PropTypes.string,
```
argshook: @3LOK would be nice to document all props, not only `colorUnchecked`

src/ToggleSwitch/ToggleSwitch.js
```diff
@@ -68,14 +78,22 @@ ToggleSwitch.propTypes = {
...
+  colorDisabled: PropTypes.string
 };
 
 ToggleSwitch.defaultProps = {
   checked: false,
   onChange: () => {},
   disabled: false,
-  size: 'large'
+  size: 'large',
+  colorUnchecked: null,
```
argshook: @3LOK if `PropType.string`, then `defaultProp` should be `''` not `null`?

stories/ToggleSwitch/ExampleColors.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import ToggleSwitch from 'wix-style-react/ToggleSwitch';
+
+const style = {
+  display: 'inline-block',
+  padding: '0 25px 0 5px',
+  lineHeight: '10px'
+};
+
+export default () =>
```
argshook: @3LOK no need for this file, `ToggleSwitch` has automated story where consumers can play with props directly

stories/ToggleSwitch/index.js
```diff
@@ -31,6 +34,10 @@ story({
       <CodeExample title="Controlled input" code={ExampleControlledRaw}>
         <ExampleControlled/>
       </CodeExample>
+
+      <CodeExample title="Colors" code={ExampleColorsRaw}>
```
argshook: @3LOK since `ExampleColors.js` is not needed, this is not needed also

src/ToggleSwitch/ToggleSwitch.js
3LOK: @argshook well, the comment was to document all three, but ok - added specific comment to each one.

src/ToggleSwitch/ToggleSwitch.js
3LOK: @argshook canceled the defaultProp - they are actually not needed.

stories/ToggleSwitch/ExampleColors.js
3LOK: @argshook not sure i understand - in that case, why do you have ExampleStandard & ExampleControlled? Isn't it to show case easily all the different features?

stories/ToggleSwitch/index.js
3LOK: @argshook same question as above.

stories/ToggleSwitch/ExampleColors.js
argshook: those are to not break current e2e tests but eventually will be gone as well.

3LOK:
Done.

argshook:
Thanks, merged

PR #1127, opened 2017-11-14
Create sticky page

amiryonatan:

- [ ] Added tests.
- [ ] Update documentation.


src/StickyPage/components/Header.js
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+
+const Header = props => React.cloneElement(props.children, {minimized: props.minimized});
```
shlomitc: I think we're also cloning inside the StickyPage render function, why do we need both here and there?

src/StickyPage/StickyPage.js
```diff
@@ -0,0 +1,105 @@
+import s from './StickyPage.scss';
+import React from 'react';
+import PropTypes from 'prop-types';
+import WixComponent from '../BaseComponents/WixComponent';
+import Header from './components/Header';
+import Content from './components/Content';
+
+const SCROLL_TOP_THRESHOLD = 24;
```
shlomitc: does this mean that always after 24px the header becomes minimized? If so - is this the required behavior?

src/StickyPage/StickyPage.scss
```diff
@@ -0,0 +1,34 @@
...
+    position: relative;
+    width: 100%;
+    height: 100%;
+}
+
+.static-background {
+    position: absolute;
+    width: 100%;
+    height: 84px;
+    background: rgba($D70,.94);
```
shlomitc: why do we need this static-background?

src/StickyPage/StickyPage.scss
```diff
@@ -0,0 +1,34 @@
...
+
+.scrollable-content {
+    overflow: scroll;
+    width: 100%;
+    height: 100%;
+    position: relative;
+}
+
+.content-background {
+    background: rgba($D70,.94);
```
shlomitc: is this background opacity according to the zeplin spec?

src/StickyPage/StickyPage.js
amiryonatan: Yes, 24 is the padding-bottom, which means were getting to the actual element.

src/StickyPage/StickyPage.js
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  componentWillUnmount() {
+    super.componentWillUnmount();
+
+    this._getScrollContainer().removeEventListener('scroll', this._handleScroll);
+  }
+
+  _getScrollContainer() {
+    return document.querySelector(`.${s.scrollableContent}`);
```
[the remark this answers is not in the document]
amiryonatan: 👍

src/StickyPage/components/Header.js
amiryonatan: 👍

src/StickyPage/StickyPage.js
shlomitc: OK, although It looks a bit fragile. Maybe have this padding inline in order not to break the UI by accident

src/StickyPage/StickyPage.driver.js
```diff
@@ -1,3 +1,16 @@
+import ReactTestUtils from 'react-dom/test-utils';
+
 export default ({element}) => ({
-  exists: () => !!element
+  when: {
```
shlomitc: Please take a look how other drivers are written. we're not using the given/when/is conventions here but prefer direct calls to functions.

src/StickyPage/StickyPage.spec.js
```diff
@@ -3,30 +3,35 @@ import StickyPage from './StickyPage';
...
-    );
+    const driver = createDriver(stickyPage);
+    expect(driver.is.exists()).toBeTruthy();
+  });
 
+  it('should be minimized', () => {
     const driver = createDriver(stickyPage);
-    expect(driver.exists()).toBeTruthy();
+    driver.when.scrollDown();
+    expect(driver.is.classExists('minimized')).toBeTruthy();
```
shlomitc: you should also test what happens when you scroll less than the threshold - then it should not be minimized

package.json
```diff
@@ -73,7 +73,7 @@
     "jquery": "^1.12.4",
     "jsdom": "^9.2.1",
     "react": "^15.5.4",
-    "react-docgen": "^2.15.0",
+    "react-docgen": "^2.20.0",
```
shlomitc: unless you needed to change `package.json` in order to add/fix something, it's better not to push it.

src/StickyPage/StickyPage.scss
amiryonatan: 👍

src/StickyPage/StickyPage.scss
amiryonatan: It's a work around that helps me get the experience of the sticky background - after the scroll. without having to change values on the scroll events.
Before I had to change an element height on scroll, but this allows me to keep the same template unchanged - so that the scroll events doesnt cause re-render of the component (only when minimized changes) which gives us the best performance

src/LanguagePicker/LanguagePicker.spec.js
```diff
@@ -46,6 +46,7 @@ describe('LanguagePicker', () => {
     createDriver(badLanguagePicker());
 
     expect(stub.calledWithMatch(`Invalid Prop children was given. Validation failed on child number 0`)).toBeTruthy();
+    console.error.restore();
```
shlomitc: how is this change related?

src/StickyPage/StickyPage.js
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  render() {
+    const {headerHeight, minimized} = this.state;
+    const [headerElement, contentElement] = React.Children.toArray(this.props.children);
+
+    return (
+      <div className={s.stickyPage}>
+        <div className={s.staticBackground}/>
+        <div className={s.stickyHeader} ref="stickyHeader" style={{paddingBottom: `${SCROLL_TOP_THRESHOLD}px`}}>
```
shlomitc: I think that the best practice is to add a ref in a callback function `ref={r => this.stickyHeaderRef = r}`

src/LanguagePicker/LanguagePicker.spec.js
amiryonatan: It's something we came across while writing UT, this is mandatory since we are changing the actual console.error.
It didn't affect us because there was only 1 test, but it's important to do it t avoid weird behaviours in the future.

src/StickyPage/StickyPage.badFormats.spec.js
```diff
@@ -0,0 +1,127 @@
+import React from 'react';
```
lbelinsk: For compatibility reasons, please create only one file StickyPage.spec.js and inside have an inner describe block for the badFormats

src/StickyPage/StickyPage.js
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    if (propValue[1].type !== StickyPage.Content) {
+      return new Error(`StickyPage: Invalid Prop children, second child must be StickyPage.Content`);
+    }
+
+    if (!propValue[1].props.children) {
+      return new Error(`StickyPage: Invalid Prop children, StickyPage.Content must have children`);
+    }
+  })
```
lbelinsk: isReqired is needed here, so in case the user will not give children he will get the default react prop-types warning

src/StickyPage/StickyPage.badFormats.spec.js
```diff
@@ -0,0 +1,127 @@
...
+  beforeEach(() => {
+    React = require('react');
+  });
+
+  afterEach(() => {
+    jest.resetModules();
+    stub.mockReset();
+  });
+
+  it('should not initialize component without children', () => {
```
lbelinsk: You don't need to check this if you add the "isRequired" in the propTypes validation (see the comment there). By doing that, you would pnly test react, and we don't need to do that.

Instead, check that no warning is being reported when you provide legal children, to make sure we are not warning when it is not needed.

src/StickyPage/StickyPage.badFormats.spec.js
```diff
@@ -0,0 +1,127 @@
...
+    const stickyPage = (
+      <StickyPage>
+        <div/>
+        <div/>
+        <div/>
+      </StickyPage>
+    );
+
+    createDriver(stickyPage);
+    expect(stub).toHaveBeenCalledWith('Warning: Failed prop type: StickyPage: Invalid Prop children, first child must be StickyPage.Header, and second child must be StickyPage.Content\n    in StickyPage');
```
lbelinsk: maybe add a constant for the duplicated string?

stories/StickyPage/Header.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+import s from './StickyPage.scss';
+
+export default p => (
```
lbelinsk: better to rename p to props

src/StickyPage/StickyPage.js
```diff
@@ -0,0 +1,115 @@
...
+
+  render() {
+    const {headerHeight, minimized} = this.state;
+    const [headerElement, contentElement] = React.Children.toArray(this.props.children);
+
+    return (
+      <div className={s.stickyPage}>
+        <div className={s.staticBackground}/>
+        <div className={s.stickyHeader} ref={r => this.stickyHeaderRef = r} style={{paddingBottom: `${SCROLL_TOP_THRESHOLD}px`}}>
+          {React.Children.map(this._safeGetChildren(headerElement), child => React.cloneElement(child, {minimized}))}
```
lbelinsk: we have 2 options:
1. Add defaultProps: { children: [] }
2. isRequired in propTypes

Then you won't have to have this safety function

src/StickyPage/StickyPage.badFormats.spec.js
amiryonatan: 👍

stories/StickyPage/Header.js
amiryonatan: 👍

PR #1136, opened 2017-11-16
Dropdown layout selected highlight prop

Roeia1:
Added a selectedHightlight prop to DropdownLayout, meaning after selecting an option if it should highlight it when dropdown is reopened. this behavior needed for the GoogleAddressInput, it shouldn't highlight the selected option because it's a dynamic dropdown.
Also, I found a problem after removing the highlight, that the hover highlight (the light blue color) isn't resetting after selecting an option. It was less noticed when the selectedHighlight is enabled (the current state).


src/DropdownLayout/DropdownLayout.driver.js
```diff
@@ -58,6 +58,7 @@ const dropdownLayoutDriverFactory = ({element, wrapper, component}) => {
       ReactDOM.render(<div ref={r => element = r}>{ClonedWithProps}</div>, wrapper);
     },
     hasTopArrow: () => !!element.querySelector(`.${styles.arrow}`),
+    isSelectedHighlight: () => component.props.selectedHighlight,
```
lbelinsk: 1. I guess you aren't you using this anywhere?
2. please document it in the readme testkit:
https://github.com/wix/wix-style-react/blob/master/src/DropdownLayout/README.TESTKIT.md

src/DropdownLayout/DropdownLayout.spec.js
```diff
@@ -258,6 +258,27 @@ describe('DropdownLayout', () => {
     });
   });
 
+  describe('selectedHighlight prop', () => {
+    const selectedId = 0;
+
+    it('should be true by default', () => {
+      const driver = createDriver(<DropdownLayout visible options={options}/>);
+      expect(driver.getSelectedHighlightProp()).toBe(true);
```
lbelinsk: what is this function? did you mean to use isSelectedHighlight?

src/GoogleAddressInput/GoogleAddressInput.js
```diff
@@ -72,6 +72,7 @@ class GoogleAddressInput extends React.Component {
           value={value}
           options={options}
           fixedFooter={(suggestions.length && this.props.poweredByGoogle) ? GoogleAddressInput.getGoogleFooter() : null}
+          selectedHighlight={false}
```
lbelinsk: For compatibility reasons, 
this should be added also in MultiSelect here:
https://github.com/wix/wix-style-react/blob/master/src/MultiSelect/MultiSelect.js#L31

The problem is that the current solution takes care only for the selected options, but there is a problem with the hover index.
In order to reproduce, use the multiSelect, add some options, remove one of them, click outside to blur, click on the input to focus and you will see that the option that you removed from the tags is hovered for no reason.

If you don't have time to invest in this then it is really ok because this is only 1/2 connected to what you added and I don't want to exploit your contribution, so it is up to you :)

src/DropdownLayout/DropdownLayout.spec.js
Roeia1: yeah i renamed and forgot to change here

src/DropdownLayout/DropdownLayout.driver.js
Roeia1: will add to readme :)

src/GoogleAddressInput/GoogleAddressInput.js
Roeia1: fixed :)

PR #1145, opened 2017-11-20
Add header impl

amiryonatan:

- [ ] Added tests.
- [ ] Update documentation.


src/StickyPage/StickyHeader/StickyHeader.driver.js
```diff
@@ -0,0 +1,17 @@
+import s from './StickyHeader.scss';
+
+export default ({element}) => {
+  const driver = {
+    element: () => element,
+    titleElement: () => element.querySelector('[data-hook="sticky-header-title"]'),
+    titleText: () => driver.titleElement().innerHTML,
+    isTitleExists: () => !!driver.titleElement(),
+    subtitleElement: () => element.querySelector('[data-hook="sticky-header-subtitle"]'),
```
shlomitc: You probably don't want to expose it as a driver function since it's kind of a private function.
Just take it outside of the json

src/StickyPage/StickyHeader/StickyHeader.scss
```diff
@@ -0,0 +1,48 @@
...
+    @include FontThin();
+    color: $D10;
+    line-height: 48px;
+    font-size: 36px;
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+}
+
+.subtitle {
```
shlomitc: Maybe instead of styling it specifically for the header, we can use the `<Label>` component with the right `appearance` prop? same would probably be for `title`

src/StickyPage/StickyHeader/StickyHeader.spec.js
```diff
@@ -0,0 +1,103 @@
...
+    const subtitle = 'This is a subtitle';
+    const stickyHeader = <StickyHeader minimized breadcrumbs={generateDefaultBreadcrumbs(title)} title={title} subtitle={subtitle}/>;
+    const driver = createDriver(stickyHeader);
+    expect(driver.isTitleExists()).toBeFalsy();
+    expect(driver.isSubtitleExists()).toBeFalsy();
+    expect(driver.isBreadcrumbsExists()).toBeTruthy();
+  });
+});
+
+function resolveIn(timeout, callback) {
```
shlomitc: I'd go for a mechanism that does polling instead if possible. waiting an explicit amount of time is ok for one test, but once it repeats, you'd prefer being more precise and wait less time.

I'd suggest using something like [`wix-eventually`](https://github.com/wix-platform/wix-node-platform/blob/6cb9e52421ddad159db7c931975ae5e5b9b8b319/private/testing/wix-eventually/index.js
), but since it's a private resource, maybe prefer just copy-paste it as a helper function that everyone can use.

src/StickyPage/StickyHeader/StickyHeaderAnimator.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Animator} from 'wix-animations';
+
+export default class StickyHeaderAnimator extends React.Component {
```
shlomitc: looks to me like it should be a helper function inside the Sticky header instead of being a standalone component.
Also, it looks too coupled to the breadcrumbs implementation. Try to simplify it maybe.

src/StickyPage/StickyPage.e2e.js
```diff
@@ -3,18 +3,18 @@ import {stickyPageTestkitFactory, getStoryUrl, waitForVisibilityOf} from '../../
...
 
-  eyes.it('should change the class on scroll', async () => {
+  eyes.it('should change the title opacity on scroll', async () => {
     const driver = stickyPageTestkitFactory({dataHook});
 
     browser.get(storyUrl);
 
     await waitForVisibilityOf(driver.element(), 'Cannot find StickyPage')
       .then(async () => {
-        await expect(driver.isHeaderMinimized()).toBeFalsy();
```
shlomitc: why not test also the minimization?

src/StickyPage/StickyHeader/StickyHeader.driver.js
amiryonatan: 👍

src/StickyPage/StickyHeader/StickyHeaderAnimator.js
amiryonatan: 👍

src/StickyPage/StickyHeader/StickyHeader.spec.js
amiryonatan: 👍

src/StickyPage/StickyHeader/StickyHeader.scss
amiryonatan: 👍

PR #1153, opened 2017-11-21
fixed missing separators in data table

adimara:
When creating a datatable with no animated expansion, the separating lines where missing.


argshook:
Hello, thanks for contributing!

The PR build fails due to now visible separators which appears in screenshot diffs, that's expected.

Thanks, merged

PR #1162, opened 2017-11-22
chore: build wix-style-react as umd

roymiloh:
This PR aims to add another build step for wix-style-react, as umd. In addition, it runs css modules over the compiled scss files and also over js files (transform `*.scss` imports to `Map[class name -> obfuscated value]`). It means that projects can consume it now without using Webpack at all.

It's still a POC, hence it's not in Yoshi, but it will be once we figure out that it actually works.


lbelinsk:
Hey, what about this one?
Still relevant?
@roymiloh @giladartzi

roymiloh:
No, closing.

PR #1167, opened 2017-11-22
added StatsWidget component for using in App Widget in Business Manager

mariafeldman:

- [ ] Added tests.
- [ ] Update documentation.


src/StatsWidget/StatsWidget.js
```diff
@@ -0,0 +1,82 @@
...
+class StatsWidget extends WixComponent {
+
+  static propTypes = {
+    /** Widget title */
+    title: PropTypes.string.isRequired,
+    /** Statistics to display */
+    statistics: PropTypes.arrayOf(PropTypes.shape({
+      title: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
+      subtitle: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
+      persent: PropTypes.number
```
argshook: typo in `persent`, should be `percent`. it is repeated throughout this PR. Please fix, api shouldn't have typos :)

src/StatsWidget/StatsWidget.js
```diff
@@ -0,0 +1,82 @@
...
+
+      if (propValue[key].type !== StatsWidget.Filter) {
+        return new Error(`StatsWidget: Invalid Prop children, only StatsWidget.Filter is allowed`);
+      }
+    })
+  };
+
+  renderPersentage(persent) {
+    return (<Text appearance="H3">
+      <div className={classnames(styles.persents, {[styles.isNegative]: persent < 0}, {[styles.isPositive]: persent > 0})} data-hook="persent-wrapper">
```
argshook: `classnames(styles.percents, percent > 0 ? styles.isPositive : styles.isNegative)`

src/StatsWidget/StatsWidget.js
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Card} from '../Grid';
+import WixComponent from '../BaseComponents/WixComponent';
+import styles from './StatsWidget.scss';
+import Text from '../Text';
+import classnames from 'classnames';
+import * as Icons from '../Icons';
```
argshook: dont import all if you use only one `import {ArrowVertical} from 'wix-style-react/Icons';`

src/StatsWidget/StatsWidget.js
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Card} from '../Grid';
```
argshook: `import {Card} from 'wix-style-react/Grid';`

src/StatsWidget/StatsWidget.js
```diff
@@ -0,0 +1,82 @@
...
+import WixComponent from '../BaseComponents/WixComponent';
+import styles from './StatsWidget.scss';
+import Text from '../Text';
+import classnames from 'classnames';
+import * as Icons from '../Icons';
+import Dropdown from '../Dropdown';
```
argshook: `import Dropdown from 'wix-style-react/Dropdown';`

src/StatsWidget/StatsWidget.js
```diff
@@ -0,0 +1,82 @@
...
+      if (propValue[key].type !== StatsWidget.Filter) {
+        return new Error(`StatsWidget: Invalid Prop children, only StatsWidget.Filter is allowed`);
+      }
+    })
+  };
+
+  renderPersentage(persent) {
+    return (<Text appearance="H3">
+      <div className={classnames(styles.persents, {[styles.isNegative]: persent < 0}, {[styles.isPositive]: persent > 0})} data-hook="persent-wrapper">
+        <span className={classnames(styles.persentArrow)}>{React.createElement(Icons.ArrowVertical)}</span>
```
argshook: no need for `classnames` when its not dynamic and only one.

also no real need for `createElement`, simply `<ArrowVertical/>`

src/StatsWidget/StatsWidget.js
```diff
@@ -0,0 +1,82 @@
...
+        <span data-hook="persent-value">{Math.abs(persent)}%</span>
+      </div>
+    </Text>);
+  }
+
+  renderColumn(statistics, index) {
+    return (<div className={styles.statsColumn} key={index} data-hook="statistics-item">
+      <div>
```
argshook: this `<div>` should not be required

src/StatsWidget/StatsWidget.js
```diff
@@ -0,0 +1,82 @@
...
+        <Text dataHook="statistics-item-title" appearance="H1">{statistics.title}</Text>
+        <Text dataHook="statistics-item-subtitle" appearance="H3">
+          {statistics.subtitle}
+        </Text>
+        { (typeof (statistics.persent) !== 'undefined') ? this.renderPersentage(statistics.persent) : null}
```
argshook: `{ statistics.percent && this.renderPercentage(statistics.percent) }`

src/StatsWidget/StatsWidget.js
```diff
@@ -0,0 +1,82 @@
...
+
+  render() {
+    const {title, statistics, children} = this.props;
+
+    return (
+      <Card data-hook="stats-widget">
+        <Card.Header dataHook="stats-widget-title" title={title} suffix={this.renderFilters(children)}/>
+        <Card.Content>
+          <div className={styles.statsColumnWrapper} data-hook="stats-widget-content-wrapper">
+            {statistics.map((stat, index) => this.renderColumn(stat, index))}
```
argshook: i think i would just add components here directly instead of `renderColumn`, but up to you

src/StatsWidget/StatsWidget.scss
```diff
@@ -0,0 +1,51 @@
+@import '../common.scss';
+
+.filters-wrapper {
+	& > * {
```
argshook: such selector always looks fishy to me but if it's reaaally needed, then why not `.filter-wrapper > * {  ...`

also, please 2 spaces for css for consistency :)

src/StatsWidget/StatsWidget.scss
```diff
@@ -0,0 +1,51 @@
...
+}
+
+.stats-column-wrapper {
+	display: flex;
+
+	h1,h2,h3,h4,h5,h6 {
+		margin: 0;
+	}
+
+	.stats-column {
```
argshook: you should not need to nest, css modules does not require that

src/StatsWidget/StatsWidget.spec.js
```diff
@@ -0,0 +1,149 @@
...
+  const stub = console.error = jest.fn();
+
+  function createComponent(props) {
+    driver = createDriver(<StatsWidget {...props}/>).driver;
+  }
+
+  let React;
+
+  beforeEach(() => {
+    React = require('react');
```
argshook: why? `import React from 'react';` should work. Does it not?

src/StatsWidget/StatsWidget.spec.js
```diff
@@ -0,0 +1,149 @@
...
+  });
+
+  afterEach(() => {
+    jest.resetModules();
```
argshook: you should not need this, i don't see any modules mocked

src/StatsWidget/StatsWidget.scss
```diff
@@ -0,0 +1,51 @@
...
+	& > * {
+		display: inline-block;
+		margin-left: 10px;
+	}
+}
+
+.stats-column-wrapper {
+	display: flex;
+
+	h1,h2,h3,h4,h5,h6 {
```
shlomitc: This is dangerous! we are using css modules and

.storybook/stories.js
```diff
@@ -112,3 +112,5 @@ import '../stories/Modal';
 
 // 11. Pickers and Selectors
 import '../stories/ColorPicker'; // 11.5 Color Picker
+
+import '../stories/StatsWidget';
```
shlomitc: might want to locate it according to the ABC so it will appear in the right place

src/StatsWidget/StatsWidget.driver.js
```diff
@@ -0,0 +1,45 @@
...
+    setProps: props => {
+      const ClonedWithProps = React.cloneElement(component, Object.assign({}, component.props, props), ...(component.props.children || []));
+      ReactDOM.render(<div ref={r => element = r}>{ClonedWithProps}</div>, wrapper);
+    }
+  };
+
+  return {
+    exists: () => driver.exists(),
+    driver,
+    headerDriver
```
shlomitc: why do you need to expose it?

src/StatsWidget/StatsWidget.js
```diff
@@ -0,0 +1,82 @@
...
+
+  renderFilters(filters) {
+    return (<div className={styles.filtersWrapper}>{filters}</div>);
+  }
+
+  render() {
+    const {title, statistics, children} = this.props;
+
+    return (
+      <Card data-hook="stats-widget">
```
shlomitc: We don't allow adding `data-hook`s to the root component. `WixComponent` adds it for us.
Also, this specific hook doesn't work since it's a prop.

src/StatsWidget/StatsWidget.js
```diff
@@ -0,0 +1,82 @@
...
+          <div className={styles.statsColumnWrapper} data-hook="stats-widget-content-wrapper">
+            {statistics.map((stat, index) => this.renderColumn(stat, index))}
+          </div>
+        </Card.Content>
+      </Card>
+    );
+  }
+}
+
+StatsWidget.Filter = Dropdown;
```
shlomitc: I prefer that `Filter` will be a wrapper around `Dropdown` instead of directly proxying `Dropdown`'.
It can simply mean that it expects a child component of type `Dropdown`. It's also easier to test the `Dropdown` component without need the driver to know anything about it

src/StatsWidget/StatsWidget.protractor.driver.js
```diff
@@ -0,0 +1,6 @@
+const statsWidgetDriverFactory = component => ({
+  element: () => component,
+  statisticsLength: () => component.$$('[data-hook="statistics-item"]').count()
```
shlomitc: maybe just rename to numberOfStatistics

src/StatsWidget/StatsWidget.driver.js
```diff
@@ -0,0 +1,45 @@
...
+    },
+
+    setProps: props => {
+      const ClonedWithProps = React.cloneElement(component, Object.assign({}, component.props, props), ...(component.props.children || []));
+      ReactDOM.render(<div ref={r => element = r}>{ClonedWithProps}</div>, wrapper);
+    }
+  };
+
+  return {
+    exists: () => driver.exists(),
```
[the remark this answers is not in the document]
lbelinsk: we want to support that every driver (even the complex ones) has exist function. nice convention and with that we can on the way also use the `isTestkitExists` test-util function.

src/StatsWidget/StatsWidget.driver.js
lbelinsk: It is to be able to use the Filter (dropdown --> textlinLayout).
Although it is not in use in the internal unit tests, it might be useful to expose it to the users.
https://github.com/wix/wix-style-react/blob/929a5e32b97091a5174cacaeb69294fd5fb68fcb/src/Card/Header/Header.driver.js

There is a good (horrible) reason why Mariya didn't use it, and it is something that I will open an issue for.

src/StatsWidget/StatsWidget.js
lbelinsk: If you do decide to leave this implementation, change it to:
```
{statistics.map(this.renderColum)}
```

src/StatsWidget/StatsWidget.js
lbelinsk: Other then testing, I don't see any pros to change it actually...
If we would want to change it in the future, we can still do that without breaking the users, so it works imo

src/StatsWidget/StatsWidget.scss
lbelinsk: @argshook we probably should add css linter :)

src/StatsWidget/StatsWidget.scss
lbelinsk: @shlomitc this is something we should discuss tomorrow because it brings a bigger issue.
But is this really dangerous if it is scoped under `.stats-column-wrapper {` ?

src/StatsWidget/StatsWidget.spec.js
lbelinsk: resetModules is needed to clean the cache (remove react from cache)
stub.mockReset is being called since console.error is mocked.

stories/StatsWidget/ExampleStatsWidget.scss
```diff
@@ -0,0 +1,6 @@
+@import '../../src/common.scss';
+
+.statsWidgetWrapper {
```
lbelinsk: 2 lines spacing please :)

src/StatsWidget/StatsWidget.scss
shlomitc: hmm yes, you're correct, it is probably not a problem if nested.
sorry about that 😅

src/StatsWidget/StatsWidget.scss
shlomitc: It makes the structure more clear in some cases, so I do nesting as well when it makes sense

src/StatsWidget/StatsWidget.driver.js
shlomitc: The filter is not related to the header's driver AFAIU.

src/StatsWidget/StatsWidget.js
mariafeldman: There is a special case for percents = 0, so I need two conditons

src/StatsWidget/StatsWidget.js
mariafeldman: statistics.percent could be 0 and it is not the same with statistics.percent is undefined
if statistics.percent = 0 I want to render it, if statistics.percent is undefined
statistics.percent returns false if , statistics.percent = 0

src/StatsWidget/StatsWidget.js
argshook: what is the special case? from CSS it looks like different color? If so, simply add it under `.percents` class

src/StatsWidget/StatsWidget.js
argshook: oh, that's right. maybe then handle what to render in `this.renderPercentage` then?

src/StatsWidget/StatsWidget.scss
argshook: i also nest sometimes but this case really does not require it at all. In addition, it unnecessarily creates longer classnames when bundled

src/StatsWidget/StatsWidget.js
argshook: also, i notice that `styles.persents` is empty so you can remove it?

src/StatsWidget/StatsWidget.js
argshook: tl;dr; use `dataHook` instead of `data-hook` on non native dom elements (so, here :) )

src/StatsWidget/StatsWidget.js
mariafeldman: But this way function this.renderColum would be not bound to "this" obj. I see you don't use library 'react-autobind', what is recommended way to bind function?

src/StatsWidget/StatsWidget.scss
argshook: @lbelinsk is the bigger issue this #1090? There's PR #1092 but it's potentially breaking change

src/StatsWidget/StatsWidget.js
argshook: implement class methods with arrow functions and you will not need to bind anything:
`renderColumn = (statistic, index) => { /* ...implementation */ }`

src/StatsWidget/StatsWidget.js
```diff
@@ -34,24 +34,22 @@ class StatsWidget extends WixComponent {
...
-        <Text dataHook="statistics-item-subtitle" appearance="H3">
-          {statistics.subtitle}
-        </Text>
-        { (typeof (statistics.persent) !== 'undefined') ? this.renderPersentage(statistics.persent) : null}
-      </div>
+      <Text dataHook="statistics-item-title" appearance="H1">{statistics.title}</Text>
+      <Text dataHook="statistics-item-subtitle" appearance="H3">
+        {statistics.subtitle}
+      </Text>
+      { (typeof (statistics.percent) !== 'undefined') ? this.renderPercentage(statistics.percent) : null}
```
lbelinsk: (typeof (statistics.percent) !== 'undefined') && this.renderPercentage(statistics.percent)

src/StatsWidget/StatsWidget.scss
```diff
@@ -25,16 +23,16 @@
...
+		.percent-arrow {
 			display: none;
 			font-size: 11px;
 		}
 
 		&.isNegative {
 			color: $R20;
 
-			.persent-arrow {
+			.percent-arrow {
```
lbelinsk: double check 2 spaces

src/StatsWidget/StatsWidget.driver.js
```diff
@@ -19,14 +19,18 @@ const statsWidgetDriverFactory = ({element, wrapper, component}) => {
...
-    selectFilter: dataHook => {
+    getFilterDriver: dataHook => {
       const dropdownElement = element.querySelector(`[data-hook="${dataHook}"]`);
-      const dropdownDriver = dropdownDriverFactory({wrapper: element, element: dropdownElement});
-      dropdownDriver.dropdownLayoutDriver.clickAtOption(0);
+      return dropdownDriverFactory({wrapper: element, element: dropdownElement});
+    },
+
+    selectFilter: (dataHook, index) => {
+      const dropdownDriver = driver.getFilterDriver(dataHook);
```
lbelinsk: wat?? driver = {....... driver in use inside } ?

PR #1174, opened 2017-11-23
ToggleSwitch should not be clickable when disabled

lbelinsk:
ToggleSwitch disabled mode still allowed to click on the toggle, it was just grey but fully functional.


lbelinsk:
@gonengar @giladartzi

PR #1180, opened 2017-11-23
fix(Notification.TextLabel): use Text with ellipsis instead of Label

### LINKED ISSUE #1171: notification with ellipsis
Replace `<Label>` With `<Text ellipsis>`
https://github.com/wix/wix-style-react/blob/929a5e32b97091a5174cacaeb69294fd5fb68fcb/src/Notification/TextLabel/TextLabel.js#L6

argshook:
closes #1171


lbelinsk:
@argshook yes, but only with test :)
Also maybe we need to change the width of the notification content, to make sure the "..." starts when we expect them to start

src/Notification/Notification.scss
```diff
@@ -1,153 +1,116 @@
...
+  justify-content: center;
+  overflow: hidden;
+}
 
-      .label-wrapper {
-        margin-right: 12px;
-      }
-    }
+.label {
+  max-width: 70%;
```
argshook: @lbelinsk here i set max width for ellipsis to show up. I pulled `70%` out of my ~a~... hat

argshook:
@lbelinsk i started by adding test, then noticed more things so refactored a little.

looks, behaves the same, just code is simpler imo. I can tell you all about this but really it's kinda boring.

added `max-width: 70%` to label so that ellipsis would show up. Maybe it should be something other than 70%?

src/Notification/Notification.scss
lbelinsk: out of the hat is good :)
We will get immediate from the bizManager guys, they need it for next week. nice :)

lbelinsk:
sorry, missed your comment. we will get feedback from the blitz after we release

argshook:
no you didn't, you answered me [here](https://github.com/wix/wix-style-react/pull/1180#discussion_r153028494) :)

PR #1196, opened 2017-11-27
Add page background

amiryonatan:

- [ ] Added tests.
- [ ] Update documentation.


stories/Page/Breadcrumbs.js
```diff
@@ -1,15 +1,10 @@
...
-        items={[{id: '1', value: 'First item'}, {id: '2', value: 'Linked item', link: 'http://www.wix.com'}, {id: '3', value: 'Third item'}]}
-        activeId="3"
-        size="medium"
-        theme="onGrayBackground"
-        onClick={() => { }}
-        />);
-  }
-}
+export default <Breadcrumbs
+  items={[1, 2, 3].map(i => ({id: `${i}`, value: `#${i} item`}))}
```
shlomitc: you removed the link element, are you sure about it?

src/PageHeader/PageHeader.spec.js
```diff
@@ -90,54 +86,64 @@ describe('PageHeader', () => {
...
+      subtitle={subtitle}
+      hasBackgroundImage
+      breadcrumbs={breadcrumbs}
+      actionsBar={actionsBar}
+      showBackButton
+      onBackClicked={onBackClicked}
+      />);
+
+    const driver = createDriver(pageHeader);
+    driver.titleContainsClass('h1_1');
```
shlomitc: I'd prefer wrapping this behavior. I don't want the user of this component to test it like this.
Maybe something more explicit like `driver.isTitleDark()`

Also, instead of looking for the className directly, prefer using the `Text`'s  testkit `getClassName()`
https://github.com/wix/wix-style-react/blob/master/src/Text/Text.driver.js
The same approach should be applied to the subtitle text, back button and the rest.

src/PageHeader/PageHeader.js
```diff
@@ -19,60 +19,89 @@ const animateComponent = (show, useEnterDelay, content) => {
...
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      themedBreadcrumbs: this.generateThemedBreadcrumbs(props)
+    };
+  }
+
+  componentWillReceiveProps(nextProps) {
+    this.setState({
```
shlomitc: how does it affect the performance that you cloneElement and render on every props change?

src/PageHeader/PageHeader.scss
```diff
@@ -49,6 +49,16 @@
             height: 42px;
         }
     }
+
+    &.dark-theme {
+        button {
```
argshook: @amiryonatan i think the majority of css is 2 spaces, please adhere

PR #1202, opened 2017-11-28
Update MessageBoxFunctionalLayout buttons type

malcolmmadsheep:
We use [react-intl](https://github.com/yahoo/react-intl) in our project for i18n and want to use `FormattedMessage` component it provides, but `string` type doesn't allow us to do it like for `title` property.


lbelinsk:
@malcolmmadsheep it is a reasonable, but this is a breaking change.
any objection doing `oneOfType([node, string])` ?

malcolmmadsheep:
@lbelinsk No objections. Just a question why not only `node` if it works also for strings. I firstly read [this article](https://reactjs.org/docs/typechecking-with-proptypes.html) about type checking with proptypes.

argshook:
@lbelinsk @malcolmmadsheep i'd say `node` is perfectly fine and even preferable. shouldn't break current users as `string` is also `node`

malcolmmadsheep:
@lbelinsk @argshook Can we merge it?

argshook:
yep, i was waiting for build to pass, cause you never know.

do mind this will reach you after we release new version (sometime today)

malcolmmadsheep:
Great. Thank you!

PR #1208, opened 2017-11-29
Input autofocus with value cursor position fix

Roeia1:
When Input is with autofocus and value attributes, cursor location was at 0 instead of at the end of the value.
Fixed and added a test.


argshook:
thanks!

PR #1215, opened 2017-11-30
chore(storybook): use wix-storybook-utils

argshook:
Don't merge yet please
I'm taking care of it


stories/Backoffice/Button/index.js
```diff
@@ -6,12 +6,14 @@ import ExampleControlledRaw from '!raw-loader!./ExampleControlled';
...
 
-import story from '../../utils/Components/Story';
+import {story} from 'wix-storybook-utils';
+import {storiesOf} from '@storybook/react';
 import * as Icons from 'wix-style-react/Icons';
 
 const icons = Object.values(Icons).map(icon => React.createElement(icon));
 
 story({
+  storiesOf,
```
lbelinsk: This is ok for now, and we will remove this after wsr will be deprecated?
If so, let's create a organized document of documentation mechanism deprecation so we won't have unneeded left overs in our clean ui.

lbelinsk:
Also, what about the ".storybook" folder? Today each theme has almost the same thing in wix-ui.

stories/Backoffice/Button/index.js
argshook: frankly, i don't know yet, i want this to be as effortless as possible and am experimenting. right now there aren't thaaaat many `story` calls so it's fairly easy to do changes like this.

Issue #993, opened 2017-10-17
rethinking dropdowns


argshook:
currently we have many "dropdown" like components: `<Dropdown/>`, `<InputWithOptions/>`, `<AutoComplete/>`, `<GoogleAddressInput/>`, `<Multiselect/>`.

if you squint a little, not much difference is with `<PopoverMenu/>`, `<ButtonWithOptions/>`, `<IconWithOptions/>` and even `<Tooltip/>`

what do they all have in common?

all of them have something visible at all times.
that "something", when interacted with, displays another box nearby.

my proposal is to have one general approach for such behaviour which is
extendable enough to allow any behaviour (i guess looking similar to good 'ol bootstrap):

```js
import Dropdown from 'wix-style-react/Dropdown';

const MyDropdown = () =>
  <Dropdown>
    <Dropdown.Toggle>
      <ToggleComponent/>
    </Dropdown.Toggle>

    <Dropdown.Content>
      <DropdownMenu/>
    </Dropdown.Content>
  </Dropdown>;

// or even like this
const MyDropdown2 = () =>
  <Dropdown
    toggle={ToggleComponent}
    content={DropdownMenu}
  />
```

where:
 * `<ToggleComponent/>` is any component that receives, say, `toggle` function. Call it anytime you need to show/hide `<Dropdown.Content/>`.
 * `<DropdownMenu/>` is also any component that receives something similar as `<ToggleComponent/>` and so control of `<Dropdown>` is possible

of course more than just `toggle` is required. however, with this approach it's only one place to add/remove more functionality, so it's easy to implement and maintain.

all `Dropdown` needs to take care of is to show/hide whatever component at appropriate times and in correct location. `popper` can be used here as we already use this dependency in tooltip.

this issues is to brainstorm, at the moment there is no code for any of this. i came up to this point when thinking how to simplify `InputWithOptions`.

shlomitc:
I like the first approach. Very declarative.

lbelinsk:
I also like the first one better.
Instead of `toggle` we need to think of it as `onToggled` because this is more the react way (dataFlow is top to bottom, we pass a callback to be called on toggle, not a function that opens the options).

Let's add to this family the `IconWithOptions` and `LanguagePicker` which are similar. The only difference is the "trigger" event which is onMouseEnter instead of onClick.

[...]

Issue #668, opened 2017-07-28
InputWithOptions onBlur bug


lbelinsk:
onBlur is being called not only when loosing focus, but each time the mouse is being clicked, never mind where on the screen.

This prevents me from implementing some new feature in my app which is based on the fact that onblur should be called only when it is expected of it.

I took some time to investigate, and although I found it while using multiselect, It is happening in InputWithOptions.

When you click on the screen, onBlur will be called with undefined event, while when it is called when it should have been called, the event is not undefined.

@nirhart can this issue get hight priority? 
I will try to make some time to investigate more but I am not sure I will get to it.

In order to reproduce, go to storybook and pass onBlur with console log in the InputWithOptions.

Thanks!

Varnius:
What is the exact expected behavior? Fire onBlur only when nested text input loses focus OR also including such cases when you open up the dropdown and either click outside or pick an option there? In the latter case the actual focused element atm is not text input but rather whole container div.
@lbelinsk

lbelinsk:
Hmm I think that if you have `shouldCloseOnSelect` prop set to true than you should loose focus (call onBlur).

`shouldCloseOnSelect` = true only in multiSelect (can reproduce it in guineapig add experiment form).

It should call onBlur every time you are focused (the dropdownLayout options are open) and then the dropdownLayout get closed, either if it is by pressing `ESC`, clicking outside or pressing tab twice..

lbelinsk:
@Varnius

Varnius:
@lbelinsk Ok, I'm assigning this one to myself.

lbelinsk:
@Varnius thanks!

PR #1225, opened 2017-12-04
Input with options remove forced blur

liorur:
forced blur is messing a11y functionality on this component


lbelinsk:
@argshook the thing here is that Lior found a bug in Safari that when you press tab on InputWithOptions, the next focused element is the first one in the page.
This pr fixes it, it doesn't break any test but I am pretty sure we had a good reason for it.
Maybe you recall something about this?
If not I will dive in and try to refresh my memory

argshook:
im playing with this and it works fine but there's a long story, staring from #668, then #841 and might even be related to #1079.

so yeah, i remember some cases but not everything.
i vote to merge this and start working on #993. we have too many issues and too many ducttape fixes

PR #1227, opened 2017-12-05
added padding to dropdown layout option

naamagr:
added padding-top and padding-bottom to dropdown layout options to fit the zeplin requirement in https://app.zeplin.io/project/5864e02695b5754a69f56150/screen/59709362833c377e3a5ac90a.
This only adds one pixel to normal options but it's very significant for multiline (long) options.


argshook:
Thank you!

Issue #1074, opened 2017-10-31
DropdownComposite


lbelinsk:
We have a plan to the future for the composites, but at the moment what was decided is to add a new component `DropdownComposite`, just like `TextField` only with Dropdown instead of Input.

PR #1238, opened 2017-12-08
Added dropdown composite component

Jefoss:
Added the dropdown composite component as described on #1074

The component is a composite of an optional Label component and a Dropdown component.

### DIFF
--- a/src/DropdownComposite/DropdownComposite.driver.js
+++ b/src/DropdownComposite/DropdownComposite.driver.js
@@ -1,0 +1,13 @@
+import inputAreaWithLabelCompositeDriverFactory from '../Composite/InputAreaWithLabelComposite/InputAreaWithLabelComposite.driver';
+import dropdownDriverFactory from '../Dropdown/Dropdown.driver';
+
+const dropdownCompositeDriverFactory = ({element, wrapper}) => {
+  const dropdown = element && Array.from(element.childNodes).slice(-1)[0];
+
+  return {
+    ...inputAreaWithLabelCompositeDriverFactory({element, wrapper}),
+    dropdownLayoutDriver: dropdownDriverFactory({element: dropdown, wrapper: dropdown}).dropdownLayoutDriver
+  };
+};
+
+export default dropdownCompositeDriverFactory;
--- a/src/DropdownComposite/README.md
+++ b/src/DropdownComposite/README.md
@@ -1,0 +1,12 @@
+# DropdownComposite component
+
+
+## Properties
+
+| propName | propType | defaultValue | isRequired | description |
+|----------|----------|--------------|------------|-------------|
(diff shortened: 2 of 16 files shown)


src/DropdownComposite/DropdownComposite.driver.js
```diff
@@ -0,0 +1,13 @@
+import inputAreaWithLabelCompositeDriverFactory from '../Composite/InputAreaWithLabelComposite/InputAreaWithLabelComposite.driver';
+import dropdownDriverFactory from '../Dropdown/Dropdown.driver';
+
+const dropdownCompositeDriverFactory = ({element, wrapper}) => {
+  const dropdown = element && Array.from(element.childNodes).slice(-1)[0];
+
+  return {
+    ...inputAreaWithLabelCompositeDriverFactory({element, wrapper}),
+    dropdownDriver: dropdownDriverFactory({element: dropdown, wrapper: dropdown}).dropdownLayoutDriver
```
lbelinsk: either call this dropdownLayoutDriver or just return the whole dropdown.
It doesn't really matter since when you render:

```
<DropdownComposite>
   <Label/>
   <Dropdown dataHook="bla"/>
</DropdownComposite>
```

then you have access to the dropdown as you passed the dataHook for it

src/DropdownComposite/README.md
```diff
@@ -0,0 +1,8 @@
+# DropdownComposite component
+
+
+## Properties
+
+| propName | propType | defaultValue | isRequired | description |
+|----------|----------|--------------|------------|-------------|
+| children | element(s) | -          | true       | Label(optional), Dropdown(required) |
```
lbelinsk: Maybe we want to support `required` and `tooltip` like `<TextField>` to stay compatible.
@wuwa is this required for Dropdown composite with Label as well?

image

Jefoss: they are supported already, as they are part of the Dropdown's props.
if i missed something please tell me what did you mean by that

src/DropdownComposite/DropdownComposite.driver.js
Jefoss: renamed it to dropdownLayoutDriver.
you can access the label the same way, i feel like the composite driver should provide a way to interact with its components, otherwise it becomes redundant

src/DropdownComposite/README.md
lbelinsk: So the readme is not complete. see here about the required for example:
https://github.com/wix/wix-style-react/tree/master/src/TextField

Just update the doc and we will merge.
Thanks

Jefoss: Did you look at the updated readme file? or am i missing something?

lbelinsk: I didin't look at the updated one :)
looks good

PR #1242, opened 2017-12-09
use wix-ui-backoffice/ToggleSwitch

lbelinsk:
Do not merge. wix-ui-backoffice is not in npm-registry for some reason. opened a jira.
Need to make sure before it is merged:

1. wix-ui-backoffice is available on public npm registry
2. all tests are passing with the new toggleSwitch


lbelinsk:
@argshook this pr is basically ready, but the story function is failing to work with the wix-ui-backoffice component. I can't see the component in the storybook, there is infinite loader...

When I render a story in the vanilla way with storyOf it works.
For this reason I havea a failing test as well.

How would you suggest solving it? We want to be able to start moving components and to point them like I am doing for ToggleSwitch, so we must make sure we can document them so in wix-style-react everything will look the same as before.

argshook:
well, #1215 is not merged yet and so `story` doesn't yet know what `wix-ui-backoffice` is, hence infinite loader.

i'll try to finish up and merge #1215 today

argshook:
@lbelinsk im assigning to myself and will merge after im done with #1215 

unless you tell me not to :)

lbelinsk:
@argshook keep me updated what is with this pr. no rush

PR #1252, opened 2017-12-11
fix(tooltip-testkit): support append to parent

### LINKED ISSUE #1214: Tooltip testkit famous issue
We all know that the testkit for the tooltip has a lot of problems, so this issue should unify all of the issues.

The issue I am familiar with is that when you use enzyme, mount a component with Tooltip without the `appendToParent` prop (apend to body by default), the the testkit methods are working.

If the `appendToParent` is given then the testkit methods are not working at all.

@hesher @eddierl feel free to elaborate.

eddierl:
in the testkit, query first the document and it nothing found query the wrapper
this fix #1214


eddierl:
I haven't change this line(tl;dr: getTooltipWrapper)
https://github.com/wix/wix-style-react/blob/00d92dac4ada87a6002fe85ebb29069730ffc343/src/Tooltip/Tooltip.driver.js#L34
Because i'm not sure of the use of it, it's the same implementation as(tl;dr: getTooltipContent)
https://github.com/wix/wix-style-react/blob/00d92dac4ada87a6002fe85ebb29069730ffc343/src/Tooltip/Tooltip.driver.js#L15

src/Tooltip/Tooltip.driver.js
```diff
@@ -12,22 +12,29 @@ const arrowDirectionToPlacement = {
...
-    hasDarkTheme: () => !!document.body.querySelector('.dark'),
-    hasLightTheme: () => !!document.body.querySelector('.light'),
+    hasErrorTheme: () => !!documentOrWrapper.querySelector('.error'),
+    hasDarkTheme: () => !!documentOrWrapper.querySelector('.dark'),
+    hasLightTheme: () => !!documentOrWrapper.querySelector('.light'),
     getTooltipWrapper: () => document.body.querySelector('.tooltip'),
     getChildren: () => element.innerHTML,
     getPlacement: () => {
-      const arrowDirection = last(document.querySelectorAll('.arrow')).className.split(' ')[1];
+      const arrowDirection = last(documentOrWrapper.querySelectorAll('.arrow')).className.split(' ')[1];
```
lbelinsk: Basically, `documentOrWrapper === document.body`, right?
So here you are switching `document` (not `document.body`) with `documentOrWrapper`.
Legit?

src/Tooltip/Tooltip.spec.js
```diff
@@ -303,15 +303,49 @@ describe('Tooltip', () => {
   });
 
   describe('testkit', () => {
-    it('should exist', () => {
+
+    const createTooltipTestkitDriver = props => {
       const div = document.createElement('div');
       const dataHook = 'myDataHook';
-      const wrapper = div.appendChild(ReactTestUtils.renderIntoDocument(<div><Tooltip dataHook={dataHook} {..._props}>{children}</Tooltip></div>));
+      const wrapper = div.appendChild(ReactTestUtils.renderIntoDocument(<div><Tooltip dataHook={dataHook} {..._props} {...props}>{children}</Tooltip></div>));
```
lbelinsk: Instead of `{..._props} {...props}`, maybe we can do `{..._props, ...props}`?

lbelinsk:
closes #1214 

@hesher your review is also welcome as you are one of the prime victims :)

src/Tooltip/Tooltip.driver.js
```diff
@@ -12,22 +12,29 @@ const arrowDirectionToPlacement = {
 };
 
 const tooltipDriverFactory = ({element, wrapper}) => {
-  const getTooltipContent = () => document.body.querySelector('.tooltip');
+  const documentOrWrapper = {
+    querySelector: query => (document.body.querySelector(query) || (wrapper.querySelector && wrapper.querySelector(query))),
```
hesher: If we search the body first and then the wrapper it means we’re trying the slow query first and the fast one later. Also, if something is in the wrapper it usually means it’s also on the body (depending on the query)

Wouldn’t it be sensible to switch the order?
I probably don’t see the big picture cuz I just jumped into the changes without context

src/Tooltip/Tooltip.spec.js
```diff
@@ -303,15 +303,49 @@ describe('Tooltip', () => {
...
       const driver = tooltipTestkitFactory({wrapper, dataHook});
+      return driver;
+    };
+
+    it('should exist', () => {
+      const driver = createTooltipTestkitDriver();
+      driver.mouseEnter();
+      expect(driver.isShown()).toBeFalsy();
+      return resolveIn(30).then(() => {
```
hesher: Much better to use something like eventually (a asynchronous polling) then just wait 30. 
Doing this from cell. I can show example tomorrow.

src/Tooltip/Tooltip.driver.js
eddierl: True, I haven't notice it .Sounds legit, the content should be in the body.

src/Tooltip/Tooltip.spec.js
eddierl: yeah! for some reason eslint didn't like it, so I did it outside of the jsx

src/Tooltip/Tooltip.driver.js
eddierl: good point, I did this order since the default behaviour for tooltip to appending to body.
in a test environment the body contains mainly the tooltip, it's not so deep, so I not sure how much it will influence the performance. wdyt?

src/Tooltip/Tooltip.spec.js
lbelinsk: eventually is private (inside the node-platform). Already asked Vilius if we can make it open source and there is no problem to do that.
In the meantime, in tooltip this is how the tests are written unfortunately.

hesher: Use mine, it's just a function :)
```
const delay = <T>(cb: () => T, timeout: number) =>
  new Promise(resolve => setTimeout(() => resolve(cb()), timeout));

function signalTrue(signal: () => boolean) {
  let ret;
  try {
    ret = signal();
  } catch (e) {
    ret = false;
  }

  return ret;
}

const poll = ({
  signal,
  timeout
}: {
  signal: () => boolean;
  timeout: number;
}): Promise<boolean> =>
  typeof signal !== 'function'
    ? Promise.reject(
        Error('Passed non-function as a signal to polling function')
      )
    : timeout <= 0
      ? Promise.resolve(false)
      : signalTrue(signal)
        ? Promise.resolve(true)
        : delay(() => signal, 50).then((sig: () => boolean) =>
            poll({signal: sig, timeout: timeout - 50})
          );
export const until = (signal: () => boolean) => poll({signal, timeout: 2000});
```

src/Tooltip/Tooltip.driver.js
eddierl: maybe i'll change it to "bodyOrWrapper"?

src/Tooltip/Tooltip.spec.js
hesher: typical usage is:
```typescript
expect(await until( () => x === y )).to.be.true;

src/Tooltip/Tooltip.driver.js
hesher: Not just performance - what happens if there is more than one tooltip in the test dom?

eddierl: I'm convinced :)

eddierl: wait, the driver accepts a data-hook.
I think something is messed up if you have 2 tooltips with same datahook, one on the parent and one on the body.

hesher: I think more than 1 data hook of the same value can occur - example - 
you have a component that has a Tooltip embedded - if more than 1 such components is rendered, there will be more than 1 tooltips with the same data hook.

That's one of the reasons `dataHook` is bad. Seriously, people should listen to me.
@lbelinsk Next wix styles react!!!

src/Tooltip/Tooltip.driver.js
lbelinsk: yes, I think it is a good idea. what about the last one which is "document". we will leave it like this?

src/Tooltip/Tooltip.driver.js
lbelinsk: @hesher I imagine that if you create a pr for allowing to pass a predicate as well, we will merge it :) We already dicussed it in the weekly

PR #1258, opened 2017-12-12
removed padding in fieldWithSelection dropdown

naamagr:
This fix removes the padding of dropdown in fieldWithSelection composite component (see images below)
before the fix:
image
after the fix:
image


argshook:
thanks!

PR #1275, opened 2017-12-18
Info icon with tooltip

iuriikomarov:
`InfoIconWithTooltip` component


src/InfoIconWithTooltip/InfoIconWithTooltip.driver.js
```diff
@@ -0,0 +1,18 @@
+import ReactTestUtils from 'react-dom/test-utils';
```
lbelinsk: why not just to use the driver of the Tooltip?
```
export {default} from '../Tooltip/Tooltip.driver';
```

src/InfoIconWithTooltip/README.TESTKIT.md
```diff
@@ -0,0 +1,3 @@
+# InfoIconWithTooltip component
+
+TODO
```
lbelinsk: Write something like:
InfoIconWithTooltip Enzyme/ReactTestUtils testkit is identical to Tooltip test kit

stories/Tooltip/Composite/CompositeStory.js
```diff
@@ -33,4 +34,12 @@ storiesOf('7. Tooltips', module)
       <ExamplePopoverMenu/>
     </InteractiveCodeExample>
   </div>
+))
+.add('7.4. Info Icon with tooltip', () => (
```
lbelinsk: I would like to see and example of a real use case. for example, the icon should be aligned to some content properly:
image

Maybe we can use the existing HBox component (we can discuss it tomorrow).

lbelinsk: The goal is to prevent people of using css. they shouldn't wrap this component at all, not for alignment and not for setting the distance of the tooltip with respect to the icon.

lbelinsk: @wuwa please help us out with the ux design:
1. what is the distance between the tooltip & the icon.
2. what is the default placement (top, right, ...)

src/InfoIconWithTooltip/InfoIconWithTooltip.js
```diff
@@ -0,0 +1,32 @@
+import _ from 'lodash';
```
lbelinsk: import omit from 'lodash/omit';
import extend from 'extend/omit';


Don't bring the whole lodash package

src/InfoIconWithTooltip/InfoIconWithTooltip.js
```diff
@@ -0,0 +1,32 @@
...
+import WixComponent from '../BaseComponents/WixComponent';
+
+import Tooltip from '../Tooltip';
+import {Info} from 'wix-style-react/Icons';
+import Icon from '../Icons/dist/Icon';
+
+const exposeApi = key => _.extend({}, Tooltip[key], {iconSize: Icon[key].size});
+
+class InfoIconWithTooltip extends WixComponent {
+  static displayName = 'InfoIconWithTooltip';
```
lbelinsk: Missing autodocs comments

src/InfoIconWithTooltip/InfoIconWithTooltip.js
```diff
@@ -0,0 +1,32 @@
...
+
+  static propTypes = exposeApi('propTypes');
+
+  static defaultProps = exposeApi('defaultProps');
+
+  render() {
+    const tooltipProps = _.omit(this.props, ['iconSize']);
```
lbelinsk: const {iconSize, ...tooltipProps} = this.props.
No need for omit

src/InfoIconWithTooltip/InfoIconWithTooltip.js
```diff
@@ -0,0 +1,32 @@
...
+import Tooltip from '../Tooltip';
+import {Info} from 'wix-style-react/Icons';
+import Icon from '../Icons/dist/Icon';
+
+const exposeApi = key => _.extend({}, Tooltip[key], {iconSize: Icon[key].size});
+
+class InfoIconWithTooltip extends WixComponent {
+  static displayName = 'InfoIconWithTooltip';
+
+  static propTypes = exposeApi('propTypes');
```
lbelinsk: I feel like exposeApi is not really needed.
```
static proptypes = {
   ...Tooltip.propTypes,
    iconSize
};
```

no?

lbelinsk:
@iuriikomarov After we talked last week we thought that by trying to use this Composite, the thing that reduced your css usage is using HBox, and not the composite itself.
This is valuable input I got from you.

This was important for us to know, and I am glad I could see this in your story example.
I am closing this branch now and from my point of view, this small research ended successfully. 

Thanks a lot!

PR #1277, opened 2017-12-19
Add gradient to page

amiryonatan:

- [ ] Added tests.
- [ ] Update documentation.


src/Page/Page.js
```diff
@@ -154,6 +161,8 @@ Page.propTypes = {
   backgroundImageUrl: PropTypes.string,
   /** Max width of the content */
   maxWidth: PropTypes.number,
+  /** Header gradient class name */
```
lbelinsk: Header gradient class name is too low level. maybe write what it will actually do, like affect the gradient of the background color

src/Page/Page.js
```diff
@@ -77,9 +77,10 @@ class Page extends WixComponent {
   }
 
   render() {
-    const {backgroundImageUrl, maxWidth, children} = this.props;
+    const {backgroundImageUrl, maxWidth, gradientClassName, children} = this.props;
```
shlomitc: are we sure we want users to configure it by themselves? why not define a set of configurable backgrounds?

amiryonatan: Yes this is what Ben asked @shlomitc

PR #1279, opened 2017-12-19
Rename Row to Columns, set default span value for Col

iuriikomarov:


src/Grid/index.js
```diff
@@ -1,3 +1,17 @@
-import {Container, RawContainer, Row, AutoAdjustedRow, Col, Card} from './Grid';
+import {
```
lbelinsk: actually we can do this like:
```
export {Container, RawContainer, Columns, Columns as Row,...} from './Grid';
```

src/Grid/Grid.js
```diff
@@ -136,4 +140,4 @@ class Col extends Component {
   }
 }
 
-export {Container, RawContainer, Row, AutoAdjustedRow, Col, Card};
+export {Container, RawContainer, Columns, AutoAdjustedColumns, Col, Card};
```
lbelinsk: You should also expose Row and AutoAdjustedRow here since people might be doing:
```
import {Row} from 'wix-style-react/dist/src/Grid/Grid';
```

I am not saying there is a reason to do it, but it might happen.
You can search across github to see if such import path is happening.

src/Grid/Grid.spec.js
```diff
@@ -112,21 +112,21 @@ describe('Grid', () => {
...
+    it('for Col', () => {
       const element = mount(<Col dataHook="test"/>);
       expect(element.find('[data-hook="test"]').length).toEqual(1);
     });
 
-    it('for Col', () => {
-      const element = mount(<Row dataHook="test"/>);
+    it('for Columns', () => {
+      const element = mount(<Columns dataHook="test"/>);
       expect(element.find('[data-hook="test"]').length).toEqual(1);
```
lbelinsk: Please add a test for the default span = 12

lbelinsk:
@iuriikomarov we also need to update the docs
https://wix.github.io/wix-style-react/?selectedKind=Common&selectedStory=Grid%20with%20card%20layout&full=0&down=0&left=1&panelRight=0&downPanel=kadirahq%2Fstorybook-addon-actions%2Factions-panel

src/Grid/Grid.js
```diff
@@ -100,6 +100,10 @@ class Col extends Component {
     dataHook: PropTypes.string
   };
 
+  static defaultProps = {
+    span: 12
```
lbelinsk: The pr is ready, looks good. The thing is that this change should be considered as an breaking change as this might affect the UI of the users. @shlomitc @argshook I am thinking this can cause initial left padding go away in  the applications which might have already hack their way around it, and now it might harm them.

Let's maybe create a major version with all of the changes in the Modal Selector?

shlomitc: yes, a major version is a good idea.

PR #1291, opened 2017-12-22
feature(tabs): supporting a new 'compactSide' type with maxWidth for items and a dynamically showed tooltip

TomasRup:
- Tabs now have a new `compactSide` type, which has a dynamically calculated maximum with according to the maximum possible container widht
- Implemented a `sideContent: PropTypes.node` property so one can add any content to the right side of the tab
- A tooltip is showed in `compactSide` type when an item is ellipsed
- Refactored the `src/Tabs` heavily: added HOCs for tooltip/maxwidth logic, covered with unit tests
- Added a simple e2e test
- Added an `npm run lint` command
- Removed divider

`compactSide` was requested by WiX events team, the logic was discussed with Ben.

screen shot 2017-12-22 at 12 55 47

screen shot 2017-12-22 at 12 58 03


TomasRup:
FYI: the `eyes.it Page should display the title and after scroll should hide it` tests will fail for eyes.it, the difference is the thickness of the border below active tab item.

argshook:
that's okay, i'll approve when merging and it'll be fine :)

aarnoldaas:
@argshook Is there any progress with this feature?

argshook:
we spoke, you know, but not others: i'm on it, planning to finish today/tomorrow

PR #1312, opened 2017-12-31
Icon button height fix

illBeRoy:
Added support for large height (`height={'large'}`) for icon buttons.

I needed that change since my project's design requires an icon button to be of that size.

I did that by changing the `.heightlarge` class in `ButtonLayout.scss` to address `icon-button` as well, similarly to how it was done with `.heightsmall`.

Since the change was purely in css, no new tests were added.


illBeRoy:
closed since it mistakely contains unwanted changes I've experimented with a couple of months back. will reopen soon

PR #1322, opened 2018-01-04
InputArea remove hasCounter class

aarnoldaas:
- Removed has-counter class that has weird margin
- It breaks grid layout
- Fixed tests. It just tested that class is present that class had only margin-bottom :/

This is breaking change I think


aarnoldaas:
Any news?

src/InputArea/InputArea.driver.js
```diff
@@ -20,7 +21,8 @@ const inputAreaDriverFactory = ({element, wrapper, component}) => {
     getTabIndex: () => textArea.tabIndex,
     getReadOnly: () => textArea.readOnly,
     getResizable: () => textAreaElement.classList.contains(styles.resizable),
-    getHasCounter: () => textAreaElement.classList.contains(styles.hasCounter),
+    getHasCounter: () => !!$component.find(counterSelector).length,
+    getCounterValue: () => $component.find(counterSelector).text(),
```
shlomitc: can you please add this to the `README.TESTKIT.md` file?

aarnoldaas:
I do not understand what examples should I add.
I think it's a bug. How inputArea can have margin bottom 30px?
This is a breaking change because probably everybody applied hack around this weird behavior. 
If I'm wrong and inputArea should have margin, just close pull request.

argshook:
i think this change is fine, i find only handful of projects that use it.
ill rebase this branch on master, add automated storybook for `<InputArea/>` and merge

Issue #1343, opened 2018-01-09
drop down use pointer cursor


zeevgg:
### Summary:
drop down use pointer cursor instead of text caret

### Priority:
Urgent

### Steps to reproduce:
 1. hover on drop down

### Expected behavior:
mouse pointer will show 'hand'

PR #1346, opened 2018-01-10
dropdown - use cursor pointer (fix bug)

zeevgg:
issue: https://github.com/wix/wix-style-react/issues/1343
I have fixed the typo in the css usage


src/Dropdown/Dropdown.spec.js
```diff
@@ -56,8 +56,9 @@ describe('Dropdown', () => {
   });
 
   it('should be read only', () => {
-    const {inputDriver} = createDriver(<Dropdown options={getOptions()}/>);
+    const {inputDriver, driver} = createDriver(<Dropdown options={getOptions()}/>);
     expect(inputDriver.getReadOnly()).toBeTruthy();
+    expect(driver.hasClass('readonly')).toBeTruthy();
```
argshook: i think consumers of testkit should not know or care about internal classnames, so `driver.isReadOnly()` would be better imo. It could also call `inputDriver.getReadOnly()` inside

zeevgg: ok, will try

src/InputWithOptions/InputWithOptions.driver.js
```diff
@@ -12,6 +12,8 @@ const inputWithOptionsDriverFactory = ({element, wrapper, component}) => {
 
   const driver = {
     exists: () => !!element,
+    isReadOnly: () => inputDriver.getReadOnly() &&  driver.hasClass('readonly'),
+    hasClass: className => inputWrapper.className.includes(className),
```
argshook: Please don't expose `hasClass`

zeevgg:
build passed on second time. how to run again?

src/InputWithOptions/InputWithOptions.driver.js
zeevgg: why? it is useful

zeevgg: fine

argshook: testkit is public api, if you add something, we must support and maintain it. This specific method encourages consumers to work with internals and this is not what testkit should do. I know we do it in a lot of places but it's not a good thing at all and so adding more isn't the way.

If you disagree, prove me wrong, im may be totally incorrect.

src/InputWithOptions/InputWithOptions.driver.js
```diff
@@ -12,6 +12,7 @@ const inputWithOptionsDriverFactory = ({element, wrapper, component}) => {
 
   const driver = {
     exists: () => !!element,
+    isReadOnly: () => inputDriver.getReadOnly() &&  inputWrapper.className.includes('readonly'),
```
argshook: the build is failing due to eslint error, `15:53  error  Multiple spaces found before 'inputWrapper'  no-multi-spaces`

argshook:
thanks for having patience with me :)

PR #1347, opened 2018-01-10
indexOf returns -1 if not found and not zero. in any case we should us…

zeevgg:
indexOf returns -1 if not found and not zero. in any case we should use includes


argshook:
thanks!

PR #1360, opened 2018-01-14
<SectionHelper /> - Title

liorzisman:


src/SectionHelper/SectionHelper.scss
```diff
@@ -11,6 +11,10 @@
   padding: 30px 18px;
 }
 
+.title {
```
shlomitc: Do you need any styling?

src/SectionHelper/SectionHelper.js
```diff
@@ -23,6 +23,7 @@ class SectionHelper extends WixComponent {
     return (
       <div className={rootClasses}>
         <div className={css.content}>
+          {this.props.title.length > 0 ? <h3 data-hook="title">{this.props.title}</h3> : null}
```
shlomitc: why check length? just check if it exists.

src/SectionHelper/SectionHelper.js
```diff
@@ -32,11 +33,14 @@ class SectionHelper extends WixComponent {
 
 SectionHelper.propTypes = {
   /** Sets the look and feel of the component */
-  appearance: PropTypes.oneOf(Object.keys(HELPER_APPEARANCE))
+  appearance: PropTypes.oneOf(Object.keys(HELPER_APPEARANCE)),
+  /** Adds text as the title */
+  title: PropTypes.string
```
shlomitc: should it be a string or a dom element? Because the currently the title element doesn't have any styling. do you want the user to pass it in or do you want to style it yourself?

liorzisman: I was sure i added that :) fixing. I believe the title should only be a string indeed.

src/SectionHelper/SectionHelper.scss
liorzisman: added

src/SectionHelper/SectionHelper.js
argshook: i'd say we should use `node` for strings, as at some point we'll need to change it anyway

src/SectionHelper/SectionHelper.js
liorzisman: right. fixed

src/SectionHelper/SectionHelper.js
liorzisman: you mean custom elements as the title? wish we had a PM :)

argshook: eventually people will need some translation hoc, or some experiment hoc, or some tooltip, or similar, it happened many times already

liorzisman: after discussion we decided to indeed implement the `node` type, but the autodoc does not seem to be able to handle it. after putting `exampleProps` as well we only got a radio button. what do we expect there for propType `node`?

argshook: when you have `node`, in `exampleProps` give array of nodes you wish the user to choose from, e.g.: `exampleProps: { title: ['', 'hello', <span>hello again</span>] } }`

i left one empty so that user could also remove title, if that's what you want them to do :)

liorzisman: yes we are aware of the radio options - but we would like so the user would insert whatever text (normal strings or HTML or JSX) and it would render accordingly. is that possible?

argshook: no unfortunately not at the moment. im aware of this just didnt implement yet. this is somewhat related https://github.com/wix/wix-ui/issues/107

PR #1362, opened 2018-01-15
chore(ButtonSelection): deprecate

argshook:
@lbelinsk i remember you said some time ago, that this can easily be removed, so here's a PR.

I'll leave for you to decide though :)


argshook:
🍾

PR #1371, opened 2018-01-17
add generic data-hook to Tag and TooltipContent components

alexk0ch:

2 new data-hooks added
For automation testing, we want to change css-modules className generation pattern.


argshook:
hey, thanks for contributing!

however, i don't think you need this, since all component of wix-style-react support `dataHook` prop which means you can control it in your project

alexk0ch:
@argshook - hey, thanks for fast reply. Not these ones, these are not rendered by us directly, but rendered through other components from WixStyleReact - Tooltip -> TooltipContent, InputWithTags -> Tag.

argshook:
i see. i'd suggest to use protractor driver but both are missing what you need. convinced, merging :)

alexk0ch:
Thanks!

PR #1398, opened 2018-01-29
add missing getRadioAtIndex to testkit

iknow213:

We need to check radio buttons labels in tests, so we need to get actual buttons from a radiogroup. That can be done with getRadioAtIndex, and it present in docs, but available only in protractor driver


argshook:
thanks!

iknow213:
@argshook thanks) what is the process of releasing, e.g. when can we install updated version with changes?

argshook:
usually we release simply after we collect a sane amount of changes. i'll release today and update you on slack

PR #1399, opened 2018-01-30
be able to handle different mouse events

eyaleizenberg:

- [x] Explain why the change is needed.
- [ ] Added tests.
- [x] Update documentation.
- [x] Make sure the build is passing on ci.

### DIFF
--- a/src/SideMenu/DrillView/DrillView.js
+++ b/src/SideMenu/DrillView/DrillView.js
@@ -81,6 +81,11 @@ class SideMenuDrill extends WixComponent {
       return true;
     }
 
+    if (item.props.onMouseDown) {
+      item.props.onMouseDown(event);
+      return true;
+    }
+
     Children.forEach(item.props.children, child => {
       if (!found && child.props) {
         found = this.clickFirstClickableChild(child, event);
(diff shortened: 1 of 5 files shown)


argshook:
@eyaleizenberg please cover at least few of the `Before you request a review, please make sure to check the following boxes`

why is this needed, what is it you try to do?

eyaleizenberg:
@argshook 
This PR allows the Drill Menu to response to `onMouseDown` events instead of just `onClick`, shaving about 200ms in perceived performance. By default, the events are still `onClick`, but you can pass the `onMouseDown` instead.

eyaleizenberg:
@argshook We don't need any of this 'accessibility' in the sidebar drill. We much rather gains the performance and start navigation when the mouse is down and not up.

argshook:
we do care about accessibility.
currently it's far from perfect but this change makes it even worse.
this also adds more overhead than it solves.

if there are real performance issues, they should be solved where bottleneck is, surely time between `mouseDown` and `mouseUp` is not it. this is a workaround, not a solution.

sorry, no i can't allow this.

PR #1418, opened 2018-01-31
Automatic Storybook Deployment Upon PR

omeb:
**What has changed?**
 We added a script that builds and deploys a new storybook upon every PR or PR push. The script is executed by Travis CI build.

Once the build is finished, the script deploys storybook statics and then posts a comment in the relevant PR with a link to the generated storybook. This link is hosted on a free static web publishing service called 'surge.sh'.

Other Changes:
	1.	Removed all tests from Travis-CI build
	2.	Added additional devDependency: surge

**Why is this change needed?**
PR changes don't trigger a new storybook build until they are merged into master. This makes it difficult to test and verify them for non technical roles, such as designers or project managers. With this solution they could review changes and approve them before they are merged.


.travis.yml
```diff
@@ -11,9 +11,11 @@ before_install:
  - wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
  - sudo dpkg -i google-chrome*.deb
 
-script: >
-    npm run build &&
-    npm run test
+script:
```
argshook: if we're skipping tests on travis, then whole `before_install` can be removed.

maybe you saw what are the issues on travis regarding tests? is it too much effort to make them run?

.travis.yml
```diff
@@ -11,9 +11,11 @@ before_install:
...
+ - npm run build
+ - npm install surge
+ - 'if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then bash ./scripts/storybook-deploy.sh; fi'
```
argshook: i suppose you don't need `bash` since script is shebang'ed with it

scripts/storybook-deploy.sh
```diff
@@ -0,0 +1,28 @@
...
+  DEPLOY_DOMAIN=https://${REPO_OWNER}-${REPO_NAME}-${DEPLOY_SUBDOMAIN}.surge.sh
+  surge --project ${DEPLOY_PATH} --domain $DEPLOY_DOMAIN;
+  
+if [ "$TRAVIS_PULL_REQUEST" != "false" ]
+then
+  GITHUB_COMMENTS=https://api.github.com/repos/${TRAVIS_REPO_SLUG}/issues/${TRAVIS_PULL_REQUEST}/comments
+else
+  GITHUB_COMMENTS=https://api.github.com/repos/${TRAVIS_REPO_SLUG}/commits/${TRAVIS_COMMIT}/comments
+fi
+  curl -H "Authorization: token ${GITHUB_API_TOKEN}" --request POST ${GITHUB_COMMENTS} --data '{"body":"Travis automatic deployment completed successfully. Share storybook using this link: '${DEPLOY_DOMAIN}'"}'
```
argshook: i say `View storybook at '${DEPLOY_DOMAIN}'` or similar is enough for comment

scripts/storybook-deploy.sh
```diff
@@ -0,0 +1,28 @@
...
+then
+  DEPLOY_SUBDOMAIN_UNFORMATTED_LIST+=(pr-${TRAVIS_PULL_REQUEST})
+else
+  DEPLOY_SUBDOMAIN_UNFORMATTED_LIST+=(pr-${TRAVIS_COMMIT})
+fi
+
+for DEPLOY_SUBDOMAIN_UNFORMATTED in "${DEPLOY_SUBDOMAIN_UNFORMATTED_LIST[@]}"
+do
+  DEPLOY_SUBDOMAIN=`echo "$DEPLOY_SUBDOMAIN_UNFORMATTED" | sed -r 's/[\/|\.]+/\-/g'`
+  DEPLOY_DOMAIN=https://${REPO_OWNER}-${REPO_NAME}-${DEPLOY_SUBDOMAIN}.surge.sh
```
argshook: perhaps a comment with url example for future generations? :)

shlomitc: I think it's explicit enough without a comment

.travis.yml
shlomitc: +1

omeb: You're right, removed.

Regarding tests - e2e tests are failing for some reason and this build hasn't run for 5 months now. We can try removing specific tests that are failing, but what is the need for running tests in Travis if they are running on TC?

.travis.yml
omeb: correct

scripts/storybook-deploy.sh
omeb: np, changed.

.travis.yml
argshook: ah, i thought maybe there's something in general, like eyes failing to connect or smthg.

maybe one day we wont be running on TC so it would be good to know if travis works for us.
either way, figuring that out is not relevant to this PR

shlomitc:
👏

PR #1428, opened 2018-02-06
bugfix(PopoverMenuItem): don't render icon element when not exists

matveyok:
don't render icon span element when icon doesn't exist.


src/PopoverMenu/PopoverMenu.driver.js
```diff
@@ -32,7 +32,10 @@ const popoverMenuDriverFactory = ({element}) => {
     menu: {
       isShown: protect(() => itemsArray().length > 0),
       itemsLength: protect(() => itemsArray().length),
-      itemContentAt: protect(index => itemsArray()[index].querySelectorAll('button >span')[1].innerHTML),
+      itemContentAt: protect(index => {
+        const innerSpans = itemsArray()[index].querySelectorAll('button > span');
+        return innerSpans[innerSpans.length - 1].innerHTML;
```
argshook: hm, i think we should better add internal datahook and target that as content. cause with this change we might break some builds where tests have popover menu with icons

PR #1437, opened 2018-02-07
feat(PopoverMenuItem): add size prop

matveyok:
Add size prop to `PopoverMenuItem`.


matveyok:
@argshook, please review :) I'm not using `Text` component because the size should influence the entire `li` element (+icon) and not only the text. This behavior achieved as a side effect from the typography styling.

Example: 
screen shot 2018-02-07 at 17 15 30

PR #1441, opened 2018-02-08
add puppeteer 'label' driver

yanivefraim:


src/Label/Label.puppeteer.driver.js
```diff
@@ -0,0 +1,7 @@
+const labelDriverFactory = (component, page) => ({
+  click: () => component.click(),
+  getLabelText: () => page.eval(_component => _component.innerText, component),
+  element: () => component
```
shlomitc: why returning the element is needed?

PR #1448, opened 2018-02-11
Notification remove size

alex5il:
Remove size prop - to use the auto height, add 12 px padding


shlomitc:
please remove all irrelevant code to notification

shlomitc:
with regular button - 54px.
with textLink - 48px

Issue #1405, opened 2018-01-30
12.4 Full Text View


milkyfruit:
Every component, that has text content which overflows its container must have a tooltip on hover which displays the full text content.

Why not to make a separate component-function would do this job for all components? 

Initial index:
zpl.io/a7GjPPq

shlomitc:
Agree.
1. we already have a text component with ellipsis behavior, so it should be a relatively easy task to replace texts with this component.
2. we need to add the tooltip behavior to this component.

shlomitc:
The component will soon be implemented (WIP https://github.com/wix/wix-style-react/pull/2004) but refactoring existing components will be made separately.

anyway, closing it as no further action is needed here

PR #1453, opened 2018-02-13
fix (Input) pass title prop to input element

xmityaz:
The title prop exists in Input component props and in component spec in storybook, but the property didn't actually work before.


argshook:
hey, thanks for contribution!

`title` is weird, it's used when `theme="amaterial"` only: https://github.com/wix/wix-style-react/blob/3c8973da46793b55cbc219b4501a49fd3cb2b628/src/Input/ThemedInput.js#L49-L50

hence, its not for the purpose you may think it is, it's shown as label.

until input is used from `wix-ui-backoffice`, i suppose a simple comment explanation would be enough [here](https://github.com/wix/wix-style-react/blob/3c8973da46793b55cbc219b4501a49fd3cb2b628/src/Input/Input.js#L369). would you like to add it?

xmityaz:
@argshook Thanks for explanation. It really a bit miss-leading. So yes I will add it. Probably with different PR.

But still, do we have something for such cases? The dropdown value is much bigger then input width. It not a problem to do a fix from our side, just wondering maybe we have some common solution

image

argshook:
our design guidelines cover this case exactly like in your screenshot: just with simple ellipsis.

there is also this idea https://github.com/wix/wix-style-react/issues/1405 (just idea, no code yet). But im not 100% it should be applied on dropdowns.

@milkyfruit say something :)

milkyfruit:
The dropdown behaviour with long texts is described here: zpl.io/a87Zo6V
It shows the full value, once the dropdown is clicked and the list with options is opened.

If the full value must be always visible, your product designer should make the dropdown wider or use a different component, which serves better for this use case. He can always contact me, I would be glad to help :)

xmityaz:
@milkyfruit 
Thanks. Could you please invite me to the zeplin project?

Closing this PR.
Will add a new PR with comment to the `title` prop.

argshook:
@xmityaz sent invitation

PR #1457, opened 2018-02-13
Fix radio button width

aarnoldaas:
Fix issue when radio button breaks

Current:
screen shot 2018-02-13 at 15 15 57

Previous:
screen shot 2018-02-13 at 15 16 31


argshook:
thanks for fixing but i hope you are not using this component like this?

actually `type="button"` should not exist at all, it's autodocs (yay) that exposed this broken `button` & `display="vertical"` posibility

aarnoldaas:
We use this with `button` and `horizontal`. Should we remove it?
screen shot 2018-02-13 at 16 28 11

argshook:
you can use it now just beware some time in future `type="button"` will be gone and `4.5a Advanced Toggle` will take its place. under major release, of course, i wont remove it on a whim :)

aarnoldaas:
@shlomitc I think you got wrong story book build:
https://wix-wix-style-react-pr-1457.surge.sh
Or check on real one, works for me
https://wix.github.io/wix-style-react/?selectedKind=4.%20Selection&selectedStory=4.3%20Radio%20Button%20Group&full=0&addons=0&stories=1&panelRight=0

shlomitc:
fresh from now

image

PR #1458, opened 2018-02-13
Update Page header gradient height

aarnoldaas:
Update Page header gradient height. 
Zeplin: http://zpl.io/VxK85Rw

Current:
screen shot 2018-02-13 at 15 25 13

Previous:
screen shot 2018-02-13 at 15 25 46


amiryonatan:
@aarnoldaas 
hey, thanks!
I'm waiting for a confirmation from my designer, the zepelin is actually conflicting the one I was working with and mine had specifically that height for gradient.
I will update u once I have an answer from him

aarnoldaas:
My designer approved this with @milkyfruit

amiryonatan:
Ok, my designer says we need to support both cases, the existing one, and the one u changed.
Can you change it please ? (add a prop, etc .. )

argshook:
@amiryonatan your designer is @wuwa, right?
if we need to support both cases, they should both be in zeplin. maybe i can't find but i don't see there a specification for the one @aarnoldaas needs.

amiryonatan:
Yeah @wuwa is my designer.
The zepelin I was working with:
https://zpl.io/bPrWyrk
The zepelin @aarnoldaas was working with:
https://zpl.io/VxK85Rw

aarnoldaas:
On it

aarnoldaas:
@amiryonatan Added prop, let me know if you have idea for better naming 👍

milkyfruit:
Updated the Zeplin, so specs demonstrate full and trimmed gradient versions: zpl.io/VxK85Rw

amiryonatan:
@aarnoldaas Hey, looks good, can you just update the README file so the documentation will be clear ?

aarnoldaas:
@amiryonatan where should I update readme? I added toggle to example. I think it is enough

amiryonatan:
@aarnoldaas 
In the table:
https://github.com/wix/wix-style-react/blob/master/src/Page/README.md

aarnoldaas:
@amiryonatan updated

PR #1461, opened 2018-02-14
Fix modal selector overflow issues

aarnoldaas:
Fix modal selector overflow issues. 
Change description text style.
Did not add test for textContainer with calc property because of:
https://github.com/jsdom/jsdom/issues/1332

Previous:
screen shot 2018-02-13 at 22 30 07

Current
screen shot 2018-02-13 at 22 30 29


aarnoldaas:
Do not merge

argshook:
@milkyfruit how should modal selector handle long texts? wrap or ellipsis? i either missed or there is no design in zeplin for that.

milkyfruit:
Ellipsis are ok. Let's keep them single line blocks.

One thing to notice - the extra area on the right might expand up to 180px. In this case title and subtitle shouldn't break.

image
image

argshook:
for transparency, cause we spoke in private: let's not merge this one yet and try to find css-only solution

PR #1462, opened 2018-02-14
Migrate autocomplete

amiryonatan:

- [ ] Added tests.
- [ ] Update documentation.


src/AutoComplete/AutoComplete.driver.js
```diff
@@ -1,4 +1,4 @@
-import inputWithOptionsDriverFactory from '../InputWithOptions/InputWithOptions.driver';
+import inputWithOptionsDriverFactory from 'wix-ui-backoffice//dist/src/components/Autocomplete/Autocomplete.driver.js';
```
argshook: double slash

src/AutoComplete/AutoComplete.js
```diff
@@ -1,22 +1,42 @@
...
-    predicate: () => true
-  }
+const AutoComplete = ({options, disabled, selectedId, onSelect, onManualInput, placeholder, trasformOptions}) => {
+  return (
+    <Autocomplete
+      initialSelectedId={selectedId || selectedId === 0 ? [selectedId] : []}
+      disabled={disabled}
+      onSelect={onSelect}
+      onManualInput={onManualInput}
+      options={trasformOptions ? transformOptions(options) : options}
```
argshook: why do you need `transformOptions` prop? should consumer ever do `transformOptions={false}`?

also, since this will be breaking change anyway, maybe it makes sense to use new `options` shape and not transform anything here?

src/AutoComplete/AutoComplete.js
```diff
@@ -1,22 +1,42 @@
...
-  static propTypes = {
-    ...InputWithOptions.propTypes,
-    predicate: func
-  }
+const transformOptions = options => {
+  return (options || [])
+    .map(x =>
+      x.value === '-' ?
+      Autocomplete.createDivider() :
+      Autocomplete.createOption(x.id, !!x.disabled, true, x.value, () => x.value));
```
argshook: this api is sad :( for example, what does `true` stand for?
also i think we have eslint warning for > 4 arguments, so this line will warn.
shouldn't `createOption` accept object instead?

src/AutoComplete/AutoComplete.js
```diff
@@ -1,22 +1,42 @@
-import {func} from 'prop-types';
-import InputWithOptions from '../InputWithOptions/InputWithOptions';
+import React from 'react';
+import DropdownLayout from '../DropdownLayout';
+import {bool, oneOfType, number, string, func} from 'prop-types';
+import {Autocomplete} from 'wix-ui-backoffice/Autocomplete';
 
-class AutoComplete extends InputWithOptions {
```
argshook: the current thing `extends InputWithOptions` as well as `...InputWithOptions.propTypes` is really bad, but changing them is a breaking change, so please consider how to handle it

PR #1452, opened 2018-02-13
Multiple fixes - TextLink grayscale, hover css

shilomangam:
## Before you request a review, please make sure to check the following boxes:

- [x] Explain what has changed.
- [x] Explain why the change is needed.
- [x] Added tests.
- [x] Update documentation.
- [x] Merge master with no conflicts.
- [x] Make sure the build is passing on ci.
- [x] If this pr closes any issue, please add a github link to close it.
- [x] If you changed something in the product design/UX, you got the approval of [Ben Benhorin](https://wix.slack.com/messages/@benb).

This PR contains the following changes: 
* Text Link component: Added theme proptype which contains one of - normal, darkbackground, greyscale. Tests were added accordingly, *testkit was updated*.
* Button Card component padding was changed to 30px
* Multiselect - added hover background color which behave like @milkyfruit wanted :)

https://github.com/wix/wix-style-react/issues/1376

PR #1466, opened 2018-02-15
feat(Row) - fix row issue which has both padding and margin now by re…

alex5il:
fix row issue which has both padding and margin now by removing last child margin


shlomitc:
@argshook - this PR is passing, only eyes test is failing. Domas also reviewed it, so it is ok.

This PR should be merged with the following one and then we can merge
https://github.com/wix/wix-style-react/pull/1466

shlomitc:
merged with https://github.com/wix/wix-style-react/pull/1452

Issue #1474, opened 2018-02-20
Search and Modal | Prevents composing characters


aarnoldaas:
### Summary:
Impossible to type Korean and Japanese when input is inside Modal
### Priority:
- [x] Blocker

### Steps to reproduce:
 1. Put input inside Modal
 2. Try to compose character (Japanese or Korean input)

### Expected behavior:
Should let compose characters
### Relevant Components:
Modal, Search

Without modal:
![screen shot 2018-02-20 at 09 24 13](https://user-images.githubusercontent.com/9351975/36411786-edad2494-161f-11e8-95da-95998aa684c7.png)
Inside modal:
![screen shot 2018-02-20 at 09 24 33](https://user-images.githubusercontent.com/9351975/36411791-f645438e-161f-11e8-99b5-a50a6bcf143e.png)

argshook:
Could you isolate where the issue actually is? Is modal somehow related? I see you use search, so is`<Search/>` component not working or `<Input/>` alone?

aarnoldaas:
Add example in storybook: https://github.com/wix/wix-style-react/pull/1475
Works as expected when using this function for settings state: https://github.com/wix/wix-style-react/pull/1475/files#diff-4536457f5f29509ac12cd31ff1215d9aL18

shlomitc:
I can copy and paste Korean characters to the Search component, is it caused only when typing?

argshook:
copy-paste is non issue, it happens when typing, since then composition window appears and we don't handle its events properly

shlomitc:
low priority, won't be fixed

PR #1475, opened 2018-02-20
Modal and Search Korean and Japanese

aarnoldaas:
Do not merge


shlomitc:
can we close this PR?

argshook:
yes, this is just an example of how to reproduce issue described in #1474

PR #1476, opened 2018-02-20
removed component instance

amiryonatan:
### What changed

...

### Why it changed

...

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


amiryonatan:
@shlomitc let me know what u think.
I think it's an ugly hack, but unfortunately I couldn't find any other way to do this.

amiryonatan:
@shlomitc we forgot about this pr, let me know what u think

PR #1477, opened 2018-02-20
DataTable: Fix column width not set when header is hidden

### LINKED ISSUE #1420: DataTable: when hideHeader={true}, then column widths are ignored
when hideHeader={true}, then column widths are ignored,
and all columns have the same width.

Happens with:
wix-style-react@2.0.16

This is code I've been using:
```
<DataTable dataHook="security-table"
  hideHeader={true}
  data={rowItems}
  columns={[
    {title: 'Icon', render: (row: RowTemplate) => row.icon,
      width: '10%'},
    {title: 'Text', render: (row: RowTemplate) => {
        return (
          <div>
            <div data-hook="main-text" className={s.mainText}>
              {row.mainText}
            </div>
            <div data-hook="secondary-text" className={s.secondaryText}>
              {row.secondaryText}
            </div>
          </div>
        );
      }, width: '60%'},
    {title: 'Status_and_Action',
      render: (row: RowTemplate) => row.actionElement,
      width: '30%'}
]}
```

Where rowItems are:
```
const rowItems: Array<RowTemplate> = [
  {
    mainText: 'Email authentication',
    secondaryText: 'Make sure that you are using @wix.com as primary email address in your github profile.',
    icon: 'mail',
    actionElement: (
      <div>{`${props.email}  ${
        props.validEmail ?
          'is valid' :
          'is invalid'
        }`}</div>
      )
  },
  {
    mainText: 'Two-factor authentication',
    secondaryText: 'Make sure that 2FA is On. Otherwise your accout will be automatically removed from Wix organizations!',
    icon: 'shield',
    actionElement: (
      <div>{`${props.email}  ${
        props.twoFactorAuthEnabled ?
          'is ON' :
          'is OFF'
        }`}</div>)
  }
];
```

With Header:
screen shot 2018-02-01 at 13 05 16

Without header:
screen shot 2018-02-01 at 13 04 53

hepiyellow:
Fixes #1420 

Please see comments in the 2nd commit: 
https://github.com/wix/wix-style-react/commit/a0d8477e2d61bf68fd53f227d1afdeaccd1ef8c7


hepiyellow:
Build fails due to eyes. Because I added an example to the e2e. I will accept it once this PR is reviewed, so I can merge it right after.

src/DataTable/DataTable.spec.js
```diff
@@ -10,16 +10,20 @@ import {mount} from 'enzyme';
...
-      {title: 'Row Num', render: (row, rowNum) => rowNum},
-      {title: 'A', render: row => row.a},
-      {title: 'B', render: row => row.b}
-    ],
-    rowClass: 'class-name'
-  };
+  let defaultProps;
+
+  beforeEach(() => {
+    defaultProps = {
```
argshook: i would prefer to not mutate `defaultProps`, it's a place of hard to track issues

src/DataTable/DataTable.spec.js
```diff
@@ -92,6 +96,40 @@ describe('Table', () => {
     expect(driver.getTitles()).toEqual(defaultProps.columns.map(col => col.title));
   });
 
+  describe('column width', () => {
+    const widths = ['30px', '40%', '50%'];
```
argshook: perhaps just `const columns` with all necessary data? Easier to understand when you see whole fixture

src/DataTable/DataTable.js
```diff
@@ -183,8 +183,12 @@ class DataTable extends WixComponent {
 
   renderCell = (rowData, column, rowNum, colNum) => {
     const classes = classNames({[s.important]: column.important});
+    const width = rowNum === 0 && this.props.hideHeader ? column.width : undefined;
```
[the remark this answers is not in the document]
hepiyellow: Yep. The column has no shape definition.
I can do that in a separate PR.

src/DataTable/DataTable.spec.js
```diff
@@ -92,6 +96,40 @@ describe('Table', () => {
...
+        defaultProps.columns[colIndex].width = width;
+      });
+    });
+
+    it('should apply column.width', () => {
+      const driver = createDriver(<DataTable {...defaultProps}/>);
+      widths.forEach((width, colIndex) => {
+        expect(driver.getHeaderCellWidth(colIndex), 'header width').toEqual(width);
+        for (let rowIndex = 0; rowIndex < driver.getRowsCount(); rowIndex++) {
+          expect(driver.getCellWidth(rowIndex, colIndex), 'cell width').toBeFalsy();
```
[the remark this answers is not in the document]
hepiyellow: 1. `cell width` is the message for a the `expect` assertion. it describes what I am "expecting" on. I feel it is needed when a test has more then one `expect` assertion. I'll change it to `non-header cell width`, and change the `header width` to `header cell width`.
2. The actual value is an empty string, so I'll change it to `toBe('')`.

src/DataTable/DataTable.spec.js
hepiyellow: I know. Because it is hard to track (and I made myself a bug that way). I decided to create a protection from that happening in the future. There are 3 ways to do it: 
1)  to make it immutable by freezing it. 
(That requires full cloning in every test which needs something a little bit different)
2) to make it mutable (which is convenient), but make sure it is recreated every test, so one test can not affect another test.
3) to make `defaultProps` a function which creates fresh props, and call it in every test.

I prefer either solution 2 or 3, I chose 2 only because it involves less code change.

src/DataTable/DataTable.spec.js
hepiyellow: Well I would like an array of the widths to make the assertions easier.
But I'll change it to:
```
const widths = ['30px', '40%', '50%'];

    beforeEach(() => {
      defaultProps.columns = [
        {title: 'Row Num', width: widths[0], render: (row, rowNum) => rowNum},
        {title: 'A', width: widths[1], render: row => row.a},
        {title: 'B', width: widths[2], render: row => row.b}
      ];
    });
```

hepiyellow: nice. I like it.

hepiyellow:
@argshook Please re-review my comments and fixes.

src/DataTable/DataTable.js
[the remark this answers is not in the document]
argshook: i think it's this shape, no? https://github.com/wix/wix-style-react/blob/b8adb09e2b7330da84b3c19be61046b201a1cb6a/src/DataTable/DataTable.js#L306-L314

`width` is a different prop, it's for table width, not individual column widths. Unless i'm misunderstanding something?

src/DataTable/DataTable.spec.js
argshook: what about
```
const defaultProps = { // not in beforeEach
  id: 'id',
  data: [1,2,3]
  // etc etc
}
```

and then spread it like butter

```
const props = {  // inside some `it` or `describe` where you need it
   ...defaultProps,
   columns: ['something', 'new']
}

<DataTable {...props} />
```
i think you don't need deep properties, so this should be enough

hepiyellow: Your suggestion would work. Using `Object.assign()` or spread operator is equivalent.
But it doesn't make it "safe". It still is possible to mutate defaultProps in one test, and affect other tests. 
When that happens (by mistake of course), other tests might fail and it is hard to understand why, or even worse they might pass.
The 3 options I wrote in my previous comment, make it "safer". Or at least follow the principal of stopping propagation of errors as soon as possible.

I guess it is implicitly agreed that `defaultProps` should not be mutated between tests, so why not make it explicit?

PR #1481, opened 2018-02-22
expand Tag component functionality

malcolmmadsheep:
### What changed

1. Add `onClick` method.
2. Add `background-color` change on hovering.
3. Add `maxWidth` property;

### Why it changed

1. Additional usability for component;
2. According to design;
3. Allow users control maximum possible width that tag can occupy.

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green


malcolmmadsheep:
@lbelinsk @argshook can you take a look, please?

stories/Tag.story.js
```diff
@@ -7,9 +7,11 @@ export default {
   component: Tag,
   componentPath: '../src/Tag',
   componentProps: {
-    children: 'Hello World'
+    children: 'Hello World',
+    maxWidth: null
```
argshook: why null?

stories/Tag.story.js
```diff
@@ -7,9 +7,11 @@ export default {
...
   },
   exampleProps: {
-    onRemove: () => 'Removed!'
+    onRemove: () => 'Removed!',
+    onClick: () => 'Clicked!'
```
argshook: since both `onRemove` & `onClick` return `id`, you can show it here:
```
onRemove: id => `ID: ${id} removed!`
```
 or something like that

stories/Tag.story.js
malcolmmadsheep: First time prop didn't appear in story, so I thought that we need to provide it somehow

argshook: stories are rerendered when you trigger save in `Tag.story.js` but not if you trigger save in files under `src`, maybe that's the reason.

although yes, it should rerender, know issue, somewhere on my TODO :)

PR #1488, opened 2018-02-26
Adds Chromatic regression tests

zol:
### What changed/Why it changed

Hi,

I’m one of the founders at [Chromatic](http://chromaticqa.com), a visual regression testing tool. We’ve been using Wix's Design System as a test case for Chromatic itself (we really like your design aesthetic too).  

As a token of our appreciation, I’d like to offer the design system project team Chromatic for free. We’ll also give you hands on support to make sure you have a smooth ride and can get the most out of the tool without taking up much of your time. I've made a small PR that adds visual regression tests. 

Chromatic will quickly and easily let you review any visual changes introduced by new code -- it’s especially handy for open source projects where it can be time consuming to evaluate PRs being opened out of the blue by unknown authors.

I’ve already run Chromatic on the project to establish a baseline for future tests (you can view the results [here](https://www.chromaticqa.com/build?appId=5a85fb601341780020097b55&number=4)). You can run additional tests anytime by calling `npm chromatic`. You could try running `npm chromatic` in this branch after changing a component to see an example of the review flow.

Chromatic really shines when you run the above command from within your CI system as it will then tag PRs with the result of the regression tests, like in the image below.

screen shot 2018-02-23 at 4 42 52 pm

Let me know if you have any questions, once this PR is merged I can help get chromatic running in your CI builds too if you'd like.

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -1,5 +1,6 @@
 import {configure, storiesOf} from '@storybook/react';
 import {setOptions} from '@storybook/addon-options';
+import 'react-chromatic/storybook-addon';
 
 function loadStories() {
   if (global.self === global.top) {
--- a/package.json
+++ b/package.json
@@ -22,7 +22,8 @@
     "storybook": "start-storybook -p 6006",
     "release": "node scripts/npm-release.js",
     "export-components": "import-path --path src",
-    "lint": "haste lint"
+    "lint": "haste lint",
+    "chromatic": "chromatic test --storybook-addon  --script-name=start --port=6006 --app-code=f9lbfi0h7x"
   },
   "yoshi": {
     "features": {
@@ -74,6 +75,7 @@
     "import-path": "^1.0.0",
     "jquery": "^1.12.4",
     "react": "^15.5.4",
+    "react-chromatic": "^0.7.10",
     "react-docgen": "^2.15.0",
     "react-dom": "^15.5.4",
     "react-element-to-jsx-string": "^13.0.0",


argshook:
Hello! Thanks for reaching out to us

chromatic looks really interesting, although i can't say i 100% understand the way it works and if it fits us well enough.

today we use applitools for visual tests. It works although we miss support for branches. It seems chromatic would help us in this regard.

if i understand correctly, chromatic simply hooks into story pages of storybook and inspects/saves them for later diffing. i suppose this means one snapshot per one story page?

by "story page" i mean `storiesOf('category').add(() => /* Me! */)`

with applitools we are able to do screenshots through e2e tests, meaning that we have full control. one story page can have multiple screenshot diffs and we can control what components with what props are tested.

is this something chromatic supports also + ability to have separate "pipes" of visual regression for each PR branch?

i'd love to play more with chromatic if questions above yield true :) 

by the way, following a [link you posted](https://www.chromaticqa.com/build?appId=5a85fb601341780020097b55&number=4), im seeing an error page (did log out/in, with same error)

zol:
Hi @argshook . We're looking into exactly why you couldn't log in, I'll report back. You're correct in that  Chromatic will take a snapshot of every one of your stories.

With your Applitools example, how do you control what props are passed in when you're testing a story page with multiple props? The way Chromatic is designed to solve the above use case is by creating multiple stories and passing in a different set of props each time. If however you wanted to render the same component multiple times in a story with different props, that works fine too. So I think the answer to your question on having different 'pipes' of visual regression on each PR branch is yes.

A picture tells a thousand words though so I will work on fixing your login problem so you can see for yourself!

zol:
@argshook got it, thanks for the update!

PR #1489, opened 2018-02-27
feat(PopoverMenu): change show delay to 0

matveyok:
Changed popover show delay from 200ms to 0ms.
Looks better and faster in tests :)

200ms
200-delay

0ms
no-delay


argshook:
even with example, neato

Issue #1495, opened 2018-03-01
New Focus styles


hepiyellow:
Need to update these components with new Focus styles from the zeplin specs:
(Applies only to `normal` theme)

For example: (The shadow-box is the new thing)
![screen shot 2018-03-01 at 13 52 18](https://user-images.githubusercontent.com/6338722/36843504-d4b7eae8-1d57-11e8-99d6-8c9381e17183.png)
(Also border color changed from B20 to B10)

And for Error state & Focus: (A red shadow box)
![screen shot 2018-03-01 at 13 53 20](https://user-images.githubusercontent.com/6338722/36843542-f51efd1c-1d57-11e8-8c7e-f4ccd5dcc1c9.png)

- [x] Input
- [x] Text Area
- [x] Input Area
- [x] RichTextComposite
- [x] RichTextArea
- [x] FieldWithSelection
- [x] TimePicker
- [x] DatePicker
- [x] Range
- [x] Search
- [x] Dropdown
- [x] Autocomplete
- [x] InputWithOptions
- [x] Dropdown Composite
- [x] Radio Button Group
- [x] ToggleSwitch
- [x] Button
- [ ] EditableSelector
- [ ] ModalSelectorLayout
- [x] Tag

Domas (@milkyfruit) is UI owner

shlomitc:
@hepiyellow - is this issue still relevant?
what are the required changes for `EditableSelector` and `ModalSelectorLayout`?

milkyfruit:
I just checked both `EditableSelector seems` and `ModalSelectorLayout`.
They all use updated focus styles.

Although, `EditableSelectorLayout` list should be [selected with arrow keys](https://zpl.io/a7wkJkp), not with tabs, as described in [Wix a11y guides.](https://v9yfnh.axshare.com/#g=1&p=radio_buttons) WSR [radio group](https://wix-wix-style-react.surge.sh/?selectedKind=4.%20Selection&selectedStory=4.3%20Radio%20Button%20Group&full=0&addons=0&stories=1&panelRight=0) also doesn't work like wix guides and [Zeplin specs](https://zpl.io/blGmR5a).

Discussed it with @hepiyellow, probably it should be a separate task that would be related to all types of list items.

shlomitc:
OK, so closing this issue and will open a separate accessibility feature request for the selection groups and specifically for the `RadioGroup` and `EditableSelector`

PR #1496, opened 2018-03-01
<Input>: update Focus style (shadow-box)

hepiyellow:
Implements partially issue  #1495 

- Only changing <Input> (as first step)
- Add new color (F00)

### DIFF
--- a/src/Input/Input.scss
+++ b/src/Input/Input.scss
@@ -103,7 +103,7 @@ $material-gap: 8px;
   &-normal {
     @include ThemeDefault($color: $D10, $borderColor: $B30, $backgroundColor: $D80);
     @include ThemeHover($backgroundColor: $B50);
-    @include ThemeFocus($borderColor: $B20, $boxShadow: inset 0px 0px 5px 0px fade($B10, 60%));
+    @include ThemeFocus;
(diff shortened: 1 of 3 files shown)


argshook:
@hepiyellow just a thought, im not too keen on adding more mixins into `common.scss`. I'd put them maybe in `styles/mixins.scss`.

i'm well aware of people importing `common.scss` into their projects and using it while in reality `common.scss` shouldn't be used at all.

anyway, this is no biggie for now so mergin.

hepiyellow:
Just a couple of thoughts:
1) how come eyes didn’t fail on this change. I’ll have to add tests probably. 
2) i thought maybe its better to wait with the merge until i finish adding the new focus style to all components, so users don’t see inconsistent visuals. 

3) Which brings me to the question, do we have the ability to use experiments in a library (wsr) and add a “feature toggle”?

argshook:
> 1) how come eyes didn’t fail on this change. I’ll have to add tests probably.

eyes did fail and i approved changes before merging, allisgood

> 2) i thought maybe its better to wait with the merge until i finish adding the new focus style to all components, so users don’t see inconsistent visuals.

i thinks it's okay, we'll move gradually and wont hold back for long. this already covers many components

> do we have the ability to use experiments in a library

nope and never under my surveillance :) we did have some cases with "feature toggle" but those were really horrible. we rely on semver for these things

PR #1503, opened 2018-03-04
Input/InputArea: (scss) Remove redundant&invalid fade() function call

hepiyellow:
There a a few calls to a non-existing function `fade()` in the `scss` of `Input` and `InputArea`.
It might be leftovers of the porting from `less` (since `less` has a `fade` function).

This inset shadow-box which used the `fade` function, was not visible (because of this error),
And catually, the visual specs did not include an `inset shadow-box`.
so this change does not change anything.
Only removing redundant code.


hepiyellow:
Replaced by #1554

PR #1506, opened 2018-03-04
New Focus styles: InputArea and RichTextArea

hepiyellow:
Partially implements #1495 

These components do not have e2e tests.
I'll add them later this week, with auto-example.

There is 1 commit here, `Remove editor withError erronous style`, which is unrelated.
I just noticed that the removed style, adds 1 strong red pixel which seems to be an error.


hepiyellow:
Will add e2e tests

PR #1515, opened 2018-03-06
fix(<RadioGroup />) in the testkit get selected value after change to…

david-pe:
… value

currently the testkit stores the initial selected value so tests cannot see updates to the radio group

### What changed

...

### Why it changed

...

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


argshook:
squashing one more bug, thanks!

hepiyellow:
Why merge so fast? I was looking at that.

I think we are missing a test here.
Where the selected radio button is changed, and the we call `getSelectedValue()`.

argshook:
well you are still able to look at that

we are not using a lot of driver methods in our own `RadioGroup.spec`. of course they should be tested but there's no reason to hold this PR till then

PR #1521, opened 2018-03-07
Checkbox - Add new focus styles

hepiyellow:
- Make checkbox focusable
- Add focus state
- Add focus styles


src/Checkbox/Checkbox.driver.js
```diff
@@ -1,21 +1,26 @@
...
 import labelDriverFactory from '../Label/Label.driver';
 import {testkitFactoryCreator} from '../test-common';
 
 const labelTestkitFactory = testkitFactoryCreator(labelDriverFactory);
 
-const checkboxDriverFactory = ({element, wrapper, component}) => {
+const checkboxDriverFactory = ({element, wrapper, component, eventTrigger}) => {
 
-  const checkbox = $(element).find('input')[0];
+  const input = $(element).find('input')[0];
```
liorzisman: the factory provides you with the `element`, probably better to use `querySelector`on it to get the sub elements.

src/Checkbox/Checkbox.e2e.js
```diff
@@ -6,18 +6,45 @@ describe('Checkbox', () => {
...
   });
 
-  eyes.it('should toggle state when clicked', () => {
-    waitForVisibilityOf(checkboxDriver.element(), 'Cannot find Checkbox')
-      .then(() => {
-        autoExampleDriver.setProps({checked: true});
-        expect(checkboxDriver.isChecked()).toBe(true);
+  // This test doesn't need eyes, it is only to verify that
+  // checkboxDriver.isChecked() works.
+  it('should set unchecked when clicked', async () => {
```
liorzisman: `isChecked` basically invokes protractor's `isSelected`. assuming we trust protractor, there is no need for this test in my opinion but not a must

src/Checkbox/Checkbox.e2e.js
```diff
@@ -6,18 +6,45 @@ describe('Checkbox', () => {
...
+  });
 
-        autoExampleDriver.setProps({checked: false});
-        expect(checkboxDriver.isChecked()).toBe(false);
-      });
+  eyes.it('should show focused styles', async () => {
+    await clickTab();
   });
+
+  describe('has error', () => {
```
liorzisman: i would assert the error/focus in a way @shlomitc

src/Checkbox/Checkbox.js
```diff
@@ -38,6 +44,14 @@ class Checkbox extends WixComponent {
 
   _id = `${Checkbox.displayName}-${uniqueId()}`;
 
+  handleInputBlur = () => {
+    this.state.isFocused && this.setState({isFocused: false});
```
liorzisman: probably not saving much with the `this.state.isFocused` before the `setState`

src/Checkbox/Checkbox.e2e.js
hepiyellow: It is not testing that `isSelect` is working. It is testing that `isChecked` is implemented correctly, and not for example, returns always `true`. I'll fix the comment.

hepiyellow:
Failed tests are e2e snapshots, which can be approved.

src/Checkbox/Checkbox.js
```diff
@@ -77,8 +93,13 @@ class Checkbox extends WixComponent {
           />
 
         <Label for={id} appearance={disabled ? 'T1.4' : 'T1.1'} dataHook="checkbox-label">
-          <div className={classNames(styles.checkbox, styles[size])}>
-            <div className={styles.inner}>
+          <div
+            className={classNames(styles.checkbox, styles[size])}
+            tabIndex={disabled ? -1 : 0}
```
alisey: Should be null instead of -1

src/Checkbox/Checkbox.driver.js
argshook: 3.0.0 release includes removed jquery from drivers, so it would be better not to add $ here

PR #1522, opened 2018-03-07
ToggleSwitch - New focus styles + protractor isFocused util

hepiyellow:


hepiyellow:
Previous applitools showed errors for unrelated tests.
Waiting for new results.

PR #1524, opened 2018-03-08
Add eslint rules for browser.sleep and eyes.fit

hepiyellow:
- protractor's `browser.pause()`is not working with node 8, so temporarily we use a workaround by using `browser.sleep()`. We should be careful not to commit it.
- eyes.fit also should not be committed


argshook:
@hepiyellow i guess those eslint rules would be a good addition to https://github.com/wix/eslint-config-wix

PR #1525, opened 2018-03-08
fixing the hover prop on no-border theme

Roeia1:
added `.hover` class to `no-border` css.
prop `hover` didn't work at all in `no-border` theme.


Roeia1:
@argshook

PR #1529, opened 2018-03-10
fix(Notification): include the padding as part of the width

Jefoss:
added border-box for the notification wrapper to count the notification inner padding
as part of it's width, since it was exceeding the container width

Example:
image

After fix:
image


argshook:
thanks!

PR #1530, opened 2018-03-11
Remove padding from Button no-border theme

Roeia1:
### What changed
remove padding from the button `no-border` theme
...

### Why it changed
useless padding for text only button, making blank place be clickable
...

### DIFF
--- a/src/ButtonLayout/ButtonLayout.scss
+++ b/src/ButtonLayout/ButtonLayout.scss
@@ -861,6 +861,7 @@ $base-transition: all 100ms linear;
   background: transparent;
   color: $D10;
   border: none;
+  padding: 0;
   @extend .t1_2;
 
   &:hover {


Roeia1:
@argshook

argshook:
why do you need this? `no-border` and `dark-no-border` are used within buttonWithOptions and in there `padding` is not totally useless. what's your usecase?

Roeia1:
@argshook 
```
<Button theme='no-border'>
<Text appearance="T1">abc</Text>
</Button>
```
no reason for padding here.
in this example, there is 23px padding clickable on each side of the text.
why is this needed?
the padding exists in the first place to enable the border of other themes buttons, it's not the case here.

argshook:
as i said, `no-border` & `dark-no-border` are primarily used within buttonWithOptions where this padding is not totally useless. and even if elsewhere, `no-border` is still considered as button and so dimensions are not any different.

i don't know what design you are trying to achieve but it looks like you want to disguise button as text and i'm not sure if it's okay UX wise. @milkyfruit could you chirp in?

milkyfruit:
@Roeia1 @argshook we have Zeplin specs for a dropdown without a border:
https://zpl.io/aMdwdNr

It uses the same proportions as a regular dropdown:
image
image

Paddings help to keep the right spacing between the objects, as well it shares the same click-ability as a regular dropdown.

I'm for keeping the paddings, unless we have a good ux reason why to remove them.

Roeia1:
@argshook @milkyfruit 
well, I had several places where indeed plain text should be as a button and not as an <a> (TextLink).
(really common UX requirement, with the same colors of TextLink- blue and lighter blue when hovered)
I understand you need the padding for someplace else cause it's needed, why not adding the padding and height and w/e else needed there and only there?
the actual theme should be useful without any wrappers, and right now it's `padding` and `height` isn't required at all.
the dimensions of this theme button should be the dimensions of the text it contains, the other buttons need bigger dimensions cause of the border of the circle around the text.
I can hack the TextLink and use the onClick prop e.preventDefault() but it still gonna add <a> to the dom with a useless link instead of adding <button> to the dom.

argshook:
>  plain text should be as a button
> really common UX requirement

can you show an example?

currently we have too many button themes, i would really prefer not to add another one.

Roeia1:
@argshook 
https://ibb.co/gA8ziS
`Settings`, `More Info`
https://ibb.co/bBq6V7
`Add`, `Analyze My Keywords`
https://ibb.co/isK0q7
`Copy`
https://ibb.co/gwMfq7
`Create New`

argshook:
all of these look like TextLink. the problem with that is that you need to `preventDefault`. I suppose we should fix that part instead of fiddling with button themes.

if textlink didn't require `preventDefault`, would it work well for you?

Roeia1:
@argshook 
If it's fixed yeah it will be the same, but there is still 2 differences that IMO matters.
1. when you will hover on this TextLink,
it will still be an `<a>` element in the dom instead of a `<button>` element.
So you will see a link some sort of URL in the left bottom corner: https://ibb.co/e4u5sx
2. in terms of accessibility I think there is a difference between the elements.

both solutions work for me, just be aware of the differences. your call :)

argshook:
> So you will see a link some sort of URL in the left bottom corner

that's no biggie for the time being imo. from wsr side we can make it a `button` if you give `onClick` but no `link`. so in the end you don't need to bother about this

> in terms of accessibility I think there is a difference between the elements.

both `a` and `button` are accessible, no prob here.

i'll try to patch it up today so that `TextLink` works well for you.

argshook:
#1551 has been merged and will be available in next release. This means you wont need to `preventDefault` in `onClick` and you wont need to add empty `link`, it's no longer required.o

please reopen if there is something else

PR #1534, opened 2018-03-12
<ToggleSwitch> : Add Hover styles

hepiyellow:
Update styles for hover state. independent from focus.


shlomitc:
closing due to irrelevance

PR #1535, opened 2018-03-12
Added more height to content on minimize

amiryonatan:
### What changed

...

### Why it changed

...

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


src/Page/Page.js
```diff
@@ -207,6 +207,9 @@ class Page extends WixComponent {
             <div className={classNames(s.content, {[s.contentFullScreen]: contentFullScreen})} style={contentFullScreen ? null : pageDimensionsStyle}>
               {this._safeGetChildren(PageContent)}
             </div>
+            {
+              minimized && <div style={{height: PageTail ? 78 : 54}}/>
```
argshook: @amiryonatan why not `className` instead of additional div?

amiryonatan: @argshook still underwork the final version will be nicer :)

PR #1538, opened 2018-03-12
MultiSelect Composite

Wo1v3r:
### What changed
Added Composite MultiSelect 
...

### Why it changed
Because it is required for Smart Actions.
We should invest the time when possible in a more general way of composing input + label as it is a very common requirement.
...

---

- [ X] Change is tested
- [X ] Change is documented
- [ ] Build is green


lbelinsk:
@Wo1v3r We would prefer having one generic composite instead of various ad-hoc ones, but since you created it I assume you need it so we probably shouldn't slow you down

testkit/enzyme.js
```diff
@@ -78,6 +78,9 @@ export const dropdownCompositeTestkitFactory = enzymeTestkitFactoryCreator(dropd
 import autoCompleteCompositeDriverFactory from '../src/AutoCompleteComposite/AutoCompleteComposite.driver';
 export const autoCompleteCompositeTestkitFactory = enzymeTestkitFactoryCreator(autoCompleteCompositeDriverFactory);
 
+import multiSelectCompositeDriverFactory from '../src/MultiSelectComposite/MultiSelectComposite.driver';
+export const multiSelectCompositeTestkitFactory = enzymeTestkitFactoryCreator(multiSelectCompositeDriverFactory);
```
lbelinsk: What about the 3r'd testkit (the `testkit/index.js` file)

stories/MultiSelectComposite/index.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import {storiesOf} from '@storybook/react';
+import InteractiveCodeExample from 'wix-storybook-utils/InteractiveCodeExample';
+import Markdown from 'wix-storybook-utils/Markdown';
+
+import Readme from '../../src/MultiSelectComposite/README.md';
+import ExampleStandard from './ExampleStandard';
+
+storiesOf('3. Inputs', module).add('3.8 + MultiSelectComposite', () => (
```
lbelinsk: We already have 3.8. @argshook what do you say?

argshook: well i would prefer to not add this component at all and instead focus on more generic way for label + input, as @Wo1v3r said.

if it's not what we're gonna do, then 3.8 + MultiSelectComposite` i guess is okay

Wo1v3r: When we'll cool down with Smart Actions I can PR a general Composer, I'd love to pr one based on component types or maybe one that takes an order of arranging the components.
We do need it for the end of the week so let me know if there's anything additional to promote this current  PR ^_^

testkit/enzyme.js
Wo1v3r: I'll add and tag for review ^^

lbelinsk:
I suggest to merge this, and then to remove it after we will have the generic composite..

hepiyellow:
@lbelinsk @argshook 
 I think we should have separately a generic CompositeLabel, and a generic CompositeContainer.

For a CompositeLabel, we might want to create a component called <LabeledField> (or even use the existing <Label>) which could receiving **children or any kind.**


The CompositeContainer is required only in components like <Range> and <FieldWithSelection> which (for the moment) will support only HBox layout of 2 items, and will manage the borders and focus borders behaviour. 

We don't have to implement this CompositeContainer now. I suggest we first release a breaking change which  separates components like <DropdownComposite>  into <LabeledField><Dropdown/></LabeledField>. 

What say you?

PR #1541, opened 2018-03-13
DataTable - render newly received data on infinite scroll properly

Jefoss:
What changed:
the current page is updated when the new data is received instead when only trying to load more data (on scroll). this renders the new data instantly instead of waiting for additional scroll to invoke the re-render.

Why it changed:
when working with infinite scroll, the table only displays the data up to the current page. the current page was updated only when trying to load more data. after receiving more data, to actually see it in the page you would have to scroll a bit just to trigger another loading, that in turn updates the current page to display the data you already have.


hepiyellow:
Need to write pagination tests.
So merge deferred.
See issue: #1547

Jefoss:
theres a new PR: #1618 (old one after force push to that branch)
with an e2e test for this scenario as @shlomitc requested

PR #1550, opened 2018-03-14
feat(ModalSelectorLayout)| support disabled items

aryelu:
### What changed
added support for disabled items in ModalSelectorLayout
...

### Why it changed
needed to support for disabled items in ModalSelectorLayout
...

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green


src/ModalSelectorLayout/ModalSelectorLayout.js
```diff
@@ -148,7 +149,8 @@ export default class ModalSelectorLayout extends WixComponent {
     searchValue: '',
     selectedItems: [],
     shouldShowNoResultsFoundState: false,
-    isEmpty: false
+    isEmpty: false,
+    disabledCount: 0
```
argshook: is this not used?

src/Selector/Selector.js
```diff
@@ -62,7 +64,7 @@ class Selector extends WixComponent {
         onClick={this._onClick(id)}
         >
         { toggleType === 'checkbox' ?
-          <Checkbox dataHook="toggle" checked={isSelected}/> :
+          <Checkbox dataHook="toggle" checked={isSelected} disabled={isDisabled}/> :
```
argshook: i suppose same should go for radio button as well

src/ModalSelectorLayout/ModalSelectorLayout.js
```diff
@@ -305,6 +309,7 @@ export default class ModalSelectorLayout extends WixComponent {
       .then(({items: itemsFromNextPage, totalCount}) => {
         if (this.state.searchValue === searchValue) { // react only to the resolve of the relevant search
           const newItems = [...items, ...itemsFromNextPage];
+          const newDisabledCount = newItems.filter(item => item.disabled).length;
```
argshook: this is used in `setState` but where is that number used from state? did i miss something?

src/ModalSelectorLayout/ModalSelectorLayout.js
```diff
@@ -247,11 +249,13 @@ export default class ModalSelectorLayout extends WixComponent {
     );
   }
 
+  _isDisabled = item => this.props.multiple && item.disabled;
```
argshook: does `this.props.multiple` really need to be here?

src/ModalSelectorLayout/ModalSelectorLayout.js
```diff
@@ -247,11 +249,13 @@ export default class ModalSelectorLayout extends WixComponent {
...
+
   _renderItems() {
     const {items, selectedItems} = this.state;
     const {imageSize, imageShape, multiple} = this.props;
 
-    const isSelected = item => !!selectedItems.find(({id}) => item.id === id);
+    const isSelected = item => !!selectedItems.find(({id}) => item.id === id) || this._isDisabled(item);
```
argshook: maybe
`const isSelected = item => !item.disabled && !!selectedItems.find(({id}) => item.id === id)`

src/ModalSelectorLayout/ModalSelectorLayout.js
aryelu: I update it in https://github.com/wix/wix-style-react/pull/1550/files/d96c32871f364f48bc868c8089700488d95f7f5d#diff-c262611a545d555687acab8089e548e3R323

src/ModalSelectorLayout/ModalSelectorLayout.js
aryelu: I did this only from `multiple` mode, I don't know if this specific implementation (disable and set as selected) can work with radio button

src/ModalSelectorLayout/ModalSelectorLayout.js
aryelu: I did it only for `multiple` mode

src/ModalSelectorLayout/ModalSelectorLayout.js
aryelu: I just override the number, we don't have access to `items` in the construction time

PR #1561, opened 2018-03-18
<Range> : Fix missing dataHook 

hepiyellow:


hepiyellow:
Got a better fix, with missing e2e

PR #1562, opened 2018-03-18
Fix missing dataHook - add Range e2e test for DatePicker

hepiyellow:
Fixes a bug in <RangeInputWithLabelComposite> which blocked the user's dataHook from being rendered. 
This bug was introduced in WSR 2.0.30.
Reported by @mariafeldman .

- Added e2e test to cover Range with DatePicker
- Moved related functions from Range.protractor.driver to the e2e test.


hepiyellow:
Build errors are eyes - and can be approved

PR #1565, opened 2018-03-19
fix(PopoverMenu): pass additional Tooltip props

Jefoss:
What changed:
passes missing tooltip props like `appendTo`, `appendToParent`, `zIndex` , etc..

Why it changed:
Make the popover configurable like the tooltip, and mainly support scrolling properly.

### DIFF
--- a/src/PopoverMenu/PopoverMenu.js
+++ b/src/PopoverMenu/PopoverMenu.js
@@ -13,7 +13,16 @@ class PopoverMenu extends WixComponent {
     placement: Tooltip.propTypes.placement,
     buttonTheme: Button.propTypes.theme,
     buttonHeight: Button.propTypes.height,
-    maxWidth: Tooltip.propTypes.maxWidth
+    maxWidth: Tooltip.propTypes.maxWidth,
+    showTrigger: Tooltip.propTypes.showTrigger,
+    hideTrigger: Tooltip.propTypes.hideTrigger,
+    active: Tooltip.propTypes.active,
+    onShow: Tooltip.propTypes.onShow,
+    onHide: Tooltip.propTypes.onHide,
+    zIndex: Tooltip.propTypes.zIndex,
+    appendToParent: Tooltip.propTypes.appendToParent,
+    appendTo: Tooltip.propTypes.appendTo,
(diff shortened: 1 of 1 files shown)


src/PopoverMenu/PopoverMenu.js
```diff
@@ -69,28 +78,24 @@ class PopoverMenu extends WixComponent {
 
   render() {
     const {
-      placement,
-      size,
-      maxWidth,
       buttonHeight,
-      buttonTheme
+      buttonTheme,
+      ...tooltipProps
```
hepiyellow: `size, `maxWidth` and `placement` could also go in `tooltipProps`

src/PopoverMenu/PopoverMenu.js
```diff
@@ -13,7 +13,16 @@ class PopoverMenu extends WixComponent {
     placement: Tooltip.propTypes.placement,
     buttonTheme: Button.propTypes.theme,
     buttonHeight: Button.propTypes.height,
-    maxWidth: Tooltip.propTypes.maxWidth
+    maxWidth: Tooltip.propTypes.maxWidth,
```
shlomitc: maybe just spread the entire `Tooltip.propTypes` instead of duplicating?

hepiyellow:
I think its better to take Tooltip.propTypes, and omit the ones that PopoverMenu is fixing:
which are these ones:
```
        ref={tooltip => this.tooltip = tooltip}
        alignment="center"
        content={this.menu()}
        showTrigger="click"
        hideTrigger="click"
        showDelay={0}
        hideDelay={0}
        theme="light"
        padding={0}
        shouldCloseOnClickOutside
```

src/PopoverMenu/PopoverMenu.js
Jefoss: i'm just not sure if other props are relevant, my first intent was to add only the `appendTo` and `appendToParent`
as these are blocking my code right now

Jefoss:
I think those fixed props are what defines the PopoverMenu and not just wrapper that does what the tooltip does already

hepiyellow:
Replaced by #1568

PR #1573, opened 2018-03-20
fix Tooltip active and disabled new props transition

Roeia1:
### What changed
when checking new active prop added checking disabled as well.
...

### Why it changed
in showTrigger="custom"
when moving
from
active=false disabled=false
to
active=true disabled=true
the Tooltip will be shown.
now it won't :)


shlomitc:
hey. can you please create a test for this feature? thanks!

Roeia1:
Forgot to mention I'm out of time to test it, but I can't say no to you @shlomitc :)
I'm on it

PR #1574, opened 2018-03-21
[WIP] Update yoshi to 2.0 and allow to ship es modules.

yavorsky:
### What changed
👋
[We've updated yoshi to 2.0](https://github.com/wix-private/fed-handbook/wiki/Yoshi-2.0). This update includes webpack@4 and eslint@4 support.
We didn't want to spread breaking changes, so it shouldn't affect most of the projects.

But to take advantage of webpack 4 migration, which supports `sideEffects ` option, I've created this PR which actually adds `sideEffects` and `module` fields to `package.json`.
This means that yoshi [will create `es` directory](https://github.com/wix-private/wix-haste/pull/178) with ES modules.

So instead of
```js
import TextField from 'wix-style-react/TextField';
import DatePicker from 'wix-style-react/DatePicker';
```
we could just use:
```js
import { Button, TextField, DatePicker } from 'wix-style-react';
```
and the bundle size will be the same.

I've tested it with `npm link` for [guineapig2](https://github.com/wix-private/guineapig2):
screen shot 2018-03-20 at 8 55 00 pm
and it works as expected.
The only thing consumers should do is to exclude `commonjs` transforms.

### Why it's WIP?
`@react/storybook` isn't supported webpack 4 yet, but they have [almost finished PR](https://github.com/storybooks/storybook/pull/3148) and it should be merged in the near future.

### Why so many changes?
We also updated eslint@4. It updates some of the rules, mostly indentation. It's auto-fixable and w/o conflicts, so we could just apply it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -188,7 +188,11 @@
         "error",
         "fdescribe",
         "fit"
-      ]
+      ],
+      "indent" : 0,
+      "indent-legacy": [2, 2, {
+        "SwitchCase": 1
+      }]
     }
   }
 }
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "description": "wix-style-react",
   "version": "2.0.30",
   "main": "./dist/src/index.js",
+  "module": "./dist/es/src/index.js",
   "files": [
     "dist",
     "src",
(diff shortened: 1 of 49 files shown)


shlomitc:
Few things need to be taken care of:
1. The storybook docs should explain how to import components. For components using`AutoDocs` it should be [very simple](https://github.com/wix/wix-ui/blob/master/packages/wix-storybook-utils/src/StoryNew/index.js#L67). Other components stories should be changed manually.
2. As `stylable` is not yet supporting webpack 4, we can't merge this change, so we'll need to wait for it

lbelinsk:
closing it in favor of  https://github.com/wix/wix-style-react/pull/1780

package.json
```diff
@@ -3,9 +3,12 @@
   "description": "wix-style-react",
   "version": "2.0.30",
   "main": "./dist/src/index.js",
+  "module": "./dist/es/src/index.js",
```
hepiyellow: Remind me to ask you about this. Just for general knowledge.

package.json
```diff
@@ -84,7 +86,8 @@
     "sinon": "^1.17.6",
     "svg2react-icon": "^1.0.0",
     "wait-for-cond": "^1.0.3",
-    "wix-storybook-utils": "^1.0.0"
+    "wix-storybook-utils": "^1.0.0",
+    "yoshi": "^2.0.0-beta.2"
```
hepiyellow: Since we've seen that yoshi is sometimes breaking. Shell we put `~2.7.0`?

PR #1575, opened 2018-03-21
DO NOT MERGE - Drillview wix animations

ronenst:
### What changed

Replaced internal `SlideAnimation` with `wix-animations` 

### Why it changed

If a sidebar transition is invoked while the sidebar is already in transition the sidebar can reach a state where the menu disappears completely.

In addition a story + fixes for RTL were added, the `rtl` prop is now required for the `DrillView` to slide in the correct direction.

---

- [ ] Change is tested
- [x] Change is documented
- [x] Build is green

My pleasure ;)


src/SideMenu/core/styles.scss
```diff
@@ -13,6 +13,10 @@
   flex-grow: 1;
 }
 
+.root:global(.rtl) {
```
argshook: select attribute `.root:global([dir="rtl"])`

stories/SideMenu/ExampleSideMenuDrillRtl.js
```diff
@@ -0,0 +1,202 @@
...
+        <SideMenu.FooterLink
+          href="https://support.wix.com/"
+          target="_blank"
+          icon={<HelpIcon size="1em"/>}
+          >
+          Help Me!
+        </SideMenu.FooterLink>
+
+        <SideMenu.FooterTinyLink
+          href="https://support.wix.com/en/article/wix-seo-wiz-suggestions-and-feedback"
```
argshook: maybe `https://support.wix.com/` here also

stories/SideMenu/ExampleSideMenuDrillRtl.js
```diff
@@ -0,0 +1,202 @@
...
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      items
+    };
+  }
+
+  componentWillMount() {
+    items[2].items[1].isActive = true;
```
argshook: i this section really needed for storybook example?

src/SideMenu/core/navigation/styles.scss
```diff
@@ -160,6 +160,11 @@ $padding-vertical: 14px;
   padding: 17px 26px;
   top: 0;
   right:0;
+}
 
-
+:global(.rtl) {
```
argshook: select attribute `:global([dir="rtl"])`

ronenst:
@argshook Thanks for the review, I've attempted to address the issues.

I only now noticed there is an issue where the animation is reversed the first time a component is mounted, I guess I didn't see it before because I was relying on hot reloading.

Please don't merge yet until I can fix it.

ronenst:
Closing this and we'll try another solution, this is very hard to achieve `wix-animations` without major changes. In `wix-animations` the "exit" direction of the animation must be specified when the component is mounted but for submenus which have submenus we don't know if the direction until the user decides to click either the back button or the parent item, therefore the direction is sometimes wrong.

PR #1579, opened 2018-03-21
fix(ButtonWithOptions): doesn't invoke onSelect when same option is s…

matveyok:
when no selectedId presented on ButtonWithOptions onSelect invokes on same option.


shlomitc:
Looks like the core issue is in the `DropdownLayout` component and should be handled there.
https://github.com/wix/wix-style-react/blob/master/src/DropdownLayout/DropdownLayout.js#L53

The `selectedId` prop should be handled as an initial value.

In the case above, we compare the newly selected id with the current prop (the initial value).
The right implementation will be comparing it with the current state:
```
options[index] && onSelect && onSelect(options[index], options[index].id === this.state.selectedId);
```

I pushed my change (will also see that it passes on CI).
Any thoughts about it?

matveyok:
@argshook can you check this please?

matveyok:
@argshook can you check this please?

matveyok:
@shlomitc can you merge this :) ?

src/DropdownLayout/DropdownLayout.js
```diff
@@ -45,12 +45,12 @@ class DropdownLayout extends WixComponent {
...
 
   _onSelect(index) {
-    const {options, onSelect, selectedId} = this.props;
+    const {options, onSelect} = this.props;
     this.setState({
       selectedId: options[index] ? options[index].id : undefined,
       hovered: NOT_HOVERED_INDEX
     });
-    options[index] && onSelect && onSelect(options[index], options[index].id === selectedId);
+    options[index] && onSelect && onSelect(options[index], options[index].id === this.state.selectedId);
```
argshook: it is possible that by the time line 53 is evaluated, the value of `this.state.selectedId` could be wrong.

probably should be using `this.setState({}, () => /* this callback */)`

shlomitc: Actually, it should be the other way around. it should check for the `selectedId` state value before the state change

shlomitc:
@argshook - I did a small refactor for code and test, please review

PR #1586, opened 2018-03-23
Dropdown Composite driver fix

Wo1v3r:
### What changed
Expose dropdown driver in composite driver
...

### Why it changed
Before, dropdown composite exposed only dropdownlayout driver, to test the selection on the input you'd need to also use inputWithOptions \ dropdown drivers seperately
...

---

- [ ] Change is tested
- [ ] Change is documented
- [x] Build is green

### DIFF
--- a/src/DropdownComposite/DropdownComposite.driver.js
+++ b/src/DropdownComposite/DropdownComposite.driver.js
@@ -6,7 +6,7 @@ const dropdownCompositeDriverFactory = ({element, wrapper}) => {
 
   return {
     ...inputAreaWithLabelCompositeDriverFactory({element, wrapper}),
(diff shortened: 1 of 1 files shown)


shlomitc:
Besides the current implemenation being ugly, anything you need from the dropdown driver that you need but don't have?

because this will apply other some nasty nesting like `dropdownCompositeTestkit.driver. isReadOnly()`, which is not quite clear since `driver` refers to `InputWithOptions` driver. so I find it confusing in general.

Wo1v3r:
Well I need the input itself, so I need to instantiate an InputWithOptions testkit in addition for that.
But yeah, it's more of a refactor then a fix 👍 

I understand what you say about the nesting, I'll close the PR ^_^

shlomitc:
Cool. I plan on doing that soon. I’ll ping you when it becomes relevant

PR #1588, opened 2018-03-25
right margin only if more then one button

shahafan:
### What changed

MessageBox/Footer/Layout.scss:28 button class

### Why it changed

When having two buttons there should be a margin on the right of the first button, but when cancelText is empty and there is only one button on the message - there shouldn't be a space on the right..

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green

### DIFF
--- a/src/MessageBox/FooterLayout.scss
+++ b/src/MessageBox/FooterLayout.scss
@@ -25,6 +25,6 @@
   margin-left: auto;
 }
 
-.footerbuttons > button:first-of-type:not(:last-of-type) {
+.footerbuttons > :first-of-type:not(:last-of-type) {
   margin-right: 12px;
 }
--- a/src/MessageBox/FooterLayout.scss
+++ b/src/MessageBox/FooterLayout.scss
@@ -25,6 +25,6 @@
   margin-left: auto;
 }
 
-.footerbuttons > button:first-of-type {
+.footerbuttons > button:first-of-type:not(:last-of-type) {
   margin-right: 12px;
 }


shlomitc:
Thanks for the fix!
Do you mind removing the reference to the `button` element?
Something like this: `.footerbuttons > :not(:last-child)`.
The implementation shouldn't care about the button type, it's enough for it to know that there are two elements there.

shahafan:
It was there, I just added the not part...
I think it shouldn't be removed though... feels like if you remove it this may affect some other tags in the footer. I can take a look tomorrow

shahafan:
So, i'm afraid it will affect the wrapping divs without the button tag.. better to do it that way imo.. wdyt?

shlomitc:
which other wrapping `div`s?
https://github.com/wix/wix-style-react/blob/master/src/MessageBox/FooterLayout.js#L27
If you target `footerbuttons` children, then it should affect only the buttons.

shahafan:
Those button components contains divs.., it's not an html button tag

shlomitc:
I might be missing something.
`.footerbuttons > :not(:last-child)` targets the `.fotterbuttons` direct children, meaning it can be a native `button` / `div` / anything else.

The current rule targets only `button`, which does the work, but it means the component knows the internal implementation of `<Button/>` (it knows that`<Button/>` is using `button`).

Did you try applying the above rule and it failed?

shahafan:
Yes, conceptually you are totally right.
The thing is that I didn't add that button tag, it was there so might be there for a reason, you are more than welcome to test it without that tag... I didn't wanna touch it..

shahafan:
@shlomitc I've updated like you said

PR #1593, opened 2018-03-25
fixed tooltip testkit documentation

Roeia1:
### What changed
import TooltipTestkitFactory instead of tooltipTestkitFactory.
...

### Why it changed
wrong import
...


shlomitc:
👑

PR #1601, opened 2018-03-28
FocusableHOC - Add FocusableHOC with tests and Use it with Button.

hepiyellow:
Open issues:
- `attachHandlersAndState()` - should attach the handlers explicitly (`onFocus, onBlur,...`)


src/Backoffice/Button/Button.e2e.js
```diff
@@ -12,6 +14,13 @@ describe('Backoffice Button', () => {
   beforeAll(() => browser.get(storyUrl));
   afterEach(() => autoExampleDriver.reset());
 
+  eyes.it('should not be disabled by default', () =>
```
danielstro: This seems like an odd test, I think that this should be a unit test.
Also, was there a reason to remove the disabled test?

src/common/Focusable/Focusable.driver.js
```diff
@@ -0,0 +1,46 @@
+import {LEFT_MOUSE_BUTTON} from './FocusableHOC';
+
+const focusableDriverFactory = ({element, getFocusableElement, clickableElementGetters, eventTrigger}) => {
+  const clickLeftMouse = elm => {
```
danielstro: This doesn't call click events, so if an element does anything special during onClick, it might not be called in the tests.
This may work in e2e but could cause problems in unit tests.

src/common/Focusable/FocusableTests.js
```diff
@@ -0,0 +1,67 @@
...
+      const driver = createDriver();
+      driver.focus();
+      expect(driver.hasFocusState()).toBeTruthy();
+      expect(driver.hasFocusVisibleState()).toBeTruthy();
+    });
+
+    it('should switch to non-focus state when blured', () => {
+      const driver = createDriver();
+      driver.focus();
+      expect(driver.hasFocusState()).toBeTruthy();
```
danielstro: There is no need to test that it has focus state again. you tested this in the previous test.

src/common/Focusable/FocusableTests.js
```diff
@@ -0,0 +1,67 @@
...
+      expect(driver.hasFocusVisibleState()).toBeTruthy();
+      driver.blur();
+      expect(driver.hasFocusState()).toBeFalsy();
+      expect(driver.hasFocusVisibleState()).toBeFalsy();
+    });
+
+    it('should not have focus-visible when clicked by mouse', () => {
```
danielstro: I think you should rename the test. "It should have focus but not focus-visible when clicked by mouse" or something like that.
You are testing for focus but it's not mentioned in the title.

src/common/Focusable/FocusableTests.js
```diff
@@ -0,0 +1,67 @@
...
+      driver.blur();
+      expect(driver.hasFocusState()).toBeFalsy();
+      expect(driver.hasFocusVisibleState()).toBeFalsy();
+
+      driver.focus();
+      expect(driver.hasFocusState()).toBeTruthy();
+      expect(driver.hasFocusVisibleState()).toBeTruthy();
+    });
+
+    it('should have focus-visible when any keyboard key pressed while in focus', () => {
```
danielstro: Same as mouse focus visible test. You should mention the regular focus. Also, here you are not testing blur, which I think you shouldn't do in the other test as well, but either way it isn't consistent.

src/common/Focusable/FocusableTestsE2E.js
```diff
@@ -0,0 +1,81 @@
+import {createEnhancedFocusableDriverFactory} from './Focusable.protractor.driver';
+
+/**
+ * IMPORTANT !!!
+ * It is the responsibility of the caller to reset the tested component's state beforeEach test.
+ * Reseting the state should include reseting the document's activeElement in such a way so that when pressing TAB one time,
+ * then the tested component would be focusable.
+ */
```
danielstro: I think we need to find a solution for resetting focus. I don't think it's reasonable to demand the user to reset the focus.

src/common/Focusable/FocusableTestsE2E.js
```diff
@@ -0,0 +1,81 @@
...
+export function runFocusTests(driver, internalDriverFactory) {
+  const enhancedFocusableDriverFactory = createEnhancedFocusableDriverFactory(internalDriverFactory);
+
+  runFocusTestsImpl(enhancedFocusableDriverFactory(driver.element()));
```
danielstro: is it really needed to extract a function here? it seems like all of the tests could be written in this function.

src/common/Focusable/FocusableTestsE2E.js
```diff
@@ -0,0 +1,81 @@
+import {createEnhancedFocusableDriverFactory} from './Focusable.protractor.driver';
+
```
danielstro: We need to figure out what to do with the duplication between e2e and unit test.

src/common/Focusable/focusable.scss
```diff
@@ -0,0 +1,18 @@
+$focus: "[data-focus=\"true\"]";
```
danielstro: is there a way to add an alias so that users can use: comp:focus in their css?

src/Backoffice/Button/Button.e2e.js
hepiyellow: Specifically, this test it adding a missing test for `drvier.isButtonDisabled()`, checking that it can actually return `false`.
Generally, I like to have one test case, where I test all the default values,  currently it has only the `disabled` check.

hepiyellow: The test that was removed, was a redunant copy&paste:
`eyes.it('should render disabled'...`

src/common/Focusable/Focusable.driver.js
hepiyellow: This driver is for testing the FocusableHOC. The FocusableHOC does NOT listen on `onClick` events, so this is sufficient in order to the the HOC.

src/common/Focusable/FocusableHOC.js
```diff
@@ -0,0 +1,97 @@
...
+      return (<Component
+        ref={isStatelessComponent(Component) ? undefined : ref => this.wrappedComponentRef = ref}
+        {...this.props}
+        focusableEventHandlers={this.focusableEventHandlers}
+        {...focusStates}
+        />);
+    }
+  }
+
+  return isStatelessComponent(Component) ?
```
hepiyellow: I need to remove this `isStateless` check

src/Backoffice/Button/Button.protractor.driver.js
```diff
@@ -1,7 +1,17 @@
+import {isAttributeExists} from '../../test-common';
+
+/**
+ * @return <T extends InternalFocusableDriver>
+ */
+export const internalDriverFactory = element => ({
+  focusableElement: element,
+  clickableGetters: [() => element]
```
alisey: Looks a bit confusing, can you please add a comment explaining the purpose of this func?

src/common/Focusable/FocusableHOC.spec.js
```diff
@@ -0,0 +1,43 @@
...
+  const createDriver = createDriverFactory(pureChildDriverFactory);
+  const FocusableComp = withFocusable(PureChildComponent);
+
+  describe('Pure components', () => {
+    it('should render the wrapped component', () => {
+      const wrapper = render(<FocusableComp/>);
+      expect(wrapper.children().instance()).toBeInstanceOf(PureChildComponent);
+    });
+  });
+
```
hepiyellow: I will also add here hoisting tests

src/common/Focusable/Focusable.driver.js
```diff
@@ -0,0 +1,48 @@
+import {LEFT_MOUSE_BUTTON} from './FocusableHOC';
```
alisey: I think this should be in a separate file, something like `mouse-constants.js`

src/common/Focusable/FocusableHOC.js
danielstro: maybe you also need to remove it from line 86?

src/common/Focusable/FocusableTests.js
hepiyellow: I adopted a style of defensive testing, I admit. Only because its very easy to break tests and leave them passing, when changing the `beforeEach`, and creating a situation where the state doesn't really reset between tests. 
I think it is actually easier to read and debug a defensive test, than looking-over other tests.

src/common/Focusable/Focusable.driver.js
alisey: Just want to reiterate the point that interaction tests that rely on jsdom cannot be trusted.

src/common/Focusable/FocusableTests.js
hepiyellow: agreed

src/common/Focusable/focusable.scss
alisey: Should be simply `[data-focus]`, the attribute should be added/removed based on focus, by analogy with `disabled`, `required`, `readonly` in HTML. It doesn't need any value.

src/test-common.js
```diff
@@ -6,6 +7,14 @@ export {
     isFocused,
     scrollToElement
 } from 'wix-ui-test-utils/protractor';
+/**
+ * isAttributeExists: Protractor helper since I've seen too many times
+ * somehting like this: isDisabled: ()=> !!component.getAttribute('disabled')
+ * which returns always true.
+ */
+export const isAttributeExists = (component, name) =>
```
alisey: There's standard `element.hasAttribute('attrname')`, does it work in protractor? In any case seems to be a more fitting name.

src/common/Focusable/FocusableTestsE2E.js
```diff
@@ -0,0 +1,81 @@
...
+  const pressTab = () => browser.actions().sendKeys(protractor.Key.TAB).perform();
+  const pressShiftTab = () => browser.actions().sendKeys(protractor.Key.chord(protractor.Key.SHIFT, protractor.Key.TAB)).perform();
+  const pressSpace = () => browser.actions().sendKeys(protractor.Key.SPACE).perform();
+
+  it('should be initialy not in focus state', async () => {
+    expect(await driver.hasFocusState()).toBeFalsy();
+    expect(await driver.hasFocusVisibleState()).toBeFalsy();
+  });
+
+  xit('should switch to focus state when focused programatically', async () => {
```
alisey: Not sure if `xit` is accidental. If this test shouldn't be included remove it.

src/common/Focusable/FocusableHOC.js
```diff
@@ -0,0 +1,97 @@
...
+export function attachHandlersAndState(props) {
+  return {...props.focusableEventHandlers, ...pick(props, FOCUS_ATTR_NAME, FOCUS_VISIBLE_ATTR_NAME)};
+}
+
+export function getOnMouseDownHandler(props) {
+  return props.focusableEventHandlers.onMouseDown;
+}
+
+/*
+ * TODO: Consider adding 'disabled' state to this HOC, since:
```
alisey: I think you need to figure out how what to do with `disabled` before adding this HOC.

src/common/Focusable/FocusableTests.js
```diff
@@ -0,0 +1,67 @@
+export function runFocusTests(createDriver) {
```
hepiyellow: This test is not to be re-usable, I'm moving it into the `spec`

src/common/Focusable/FocusableTests.js
hepiyellow: will relate to focus in message.
This test is not about blur.

src/common/Focusable/FocusableHOC.js
```diff
@@ -0,0 +1,97 @@
+import React from 'react';
+import hoistNonReactMethods from 'hoist-non-react-methods';
+import pick from 'lodash/pick';
+import omit from 'lodash/omit';
+import {wrapDisplayName, getDisplayName, isStatelessComponent} from '../hocUtils';
+
+export const LEFT_MOUSE_BUTTON = 0;
+
+const FOCUS_ATTR_NAME = 'data-focus';
```
alisey: Not really compatible with stylable components. Does it mean we cannot use this in the core?

src/common/Focusable/FocusableHOC.js
```diff
@@ -0,0 +1,97 @@
...
+    };
+
+    render() {
+      const focusStates = {
+        [FOCUS_ATTR_NAME]: this.state.focus,
+        [FOCUS_VISIBLE_ATTR_NAME]: this.state.focusVisible
+      };
+
+      return (<Component
+        ref={isStatelessComponent(Component) ? undefined : ref => this.wrappedComponentRef = ref}
```
alisey: Since it doesn't work with stateless components, can we skip this?

src/common/Focusable/FocusableHOC.js
```diff
@@ -0,0 +1,97 @@
...
+        {...this.props}
+        focusableEventHandlers={this.focusableEventHandlers}
+        {...focusStates}
```
alisey: A bit confusing to call them states, since in fact they're props.

src/common/Focusable/FocusableHOC.js
```diff
@@ -0,0 +1,97 @@
...
+    render() {
+      const focusStates = {
+        [FOCUS_ATTR_NAME]: this.state.focus,
+        [FOCUS_VISIBLE_ATTR_NAME]: this.state.focusVisible
+      };
+
+      return (<Component
+        ref={isStatelessComponent(Component) ? undefined : ref => this.wrappedComponentRef = ref}
+        {...this.props}
+        focusableEventHandlers={this.focusableEventHandlers}
```
alisey: Defining a const, then hardcoding it anyway. I'd just ditch the consts, they're local and the component is so small, imo they're just adding complexity and making the code harder to read.

src/common/Focusable/FocusableTestsE2E.js
hepiyellow: We will find a solution. but for now the only robust solution is to get the page again (`browser.get(url)`).

src/common/Focusable/FocusableTestsE2E.js
hepiyellow: I don't want to have both the original driver, and the `enhancedFocusableDriverFactory` driver in the same function. It is an opening for confusion and bugs.

src/common/Focusable/FocusableTestsE2E.js
hepiyellow: I changed the unit test, so that it still tests the same cases, but it is more specific to the tested component (inside FocusableHOC.spec), and it test only the events that are specific for the FocusableHOC, and doesn't try to be generic and reusable. Meaning that it tests the FocusableHOC itself, it doesn't test how the FOcusableHOC is integrated in other components.

src/common/Focusable/focusable.scss
hepiyellow: As far as I know, no way to alias. I want to keep it as a variable, so we can easily change it in the future. For example we might want to change it to a class.

I will fix it so that the data-focus will not appear if it is false.

src/Backoffice/Button/Button.protractor.driver.js
hepiyellow: I refactored this. change the name to `createButtonInternals`, and added documentaiton in `Focusable.protrsactor.driver`.

src/test-common.js
hepiyellow: non in protractor.

src/common/Focusable/FocusableTestsE2E.js
hepiyellow: added comments

src/common/Focusable/FocusableHOC.js
hepiyellow: In the core I'll have to make some changes, and this FocusableHOC will be deprecated once all components in WSR (that use it) will use the Backopffice components.
If WSR consumers  would use the `data-focus` to style WSR focus states, then when we change to WUB components, they will break. 
Which makes me think, that maybe its better to make it a class instead of a data attribute, that way it is namespaced, and WSR consumers can not style it. They would have to wait for stylable.

What do you think?

src/common/Focusable/FocusableHOC.js
hepiyellow: fixed. thanks.

src/common/Focusable/FocusableHOC.js
hepiyellow: renamed to focusProps

src/common/Focusable/FocusableHOC.js
hepiyellow: I added use of the const instead of hard-coding `focusableEventHandlers`. The consts are used in order to use them to omit the reserved props from the wrapped Compnents propTypes.

src/common/Focusable/FocusableHOC.js
hepiyellow: I think that `disabled` is not an issue. the only scenario I was concerned about is that a focused component would re-render (update) with a `disable=true` property, and then it would be disabled & with `data-focus-visible`. But, 
I assuem that when a component is focused it wouldn't sponteniously re-render, nor if you click on it. In order to make is disabled, you would probably do something else in the page (focusout first), that will make it disabled.

src/Backoffice/Button/Button.e2e.js
```diff
@@ -2,58 +2,64 @@ import eyes from 'eyes.it';
...
+    beforeAll(() => browser.get(storyUrl));
+    beforeEach(() => waitForVisibilityOf(driver.element(), 'Cannot find Button'));
+    afterEach(() => autoExampleDriver.reset());
+
+    eyes.it('should be in initial state when renders with default', async () => {
+      expect(await driver.isButtonDisabled()).toBe(false, 'isButtonDisabled');
+      expect(await driver.isFocused()).toBe(false, 'isFocused');
+    });
+
+    eyes.it('should alert on click', async () => {
```
alisey: Not sure we need this test. Certainly don't need an alert, you can use sinon.spy or a simple closure that updates a variable.

src/Backoffice/Button/Button.e2e.js
```diff
@@ -2,58 +2,64 @@ import eyes from 'eyes.it';
...
+
+    describe('FocusableDriver', () => {
+      it('should be focused when clicked', async () => {
+        await driver.click();
+        expect(await driver.isFocused()).toBe(true);
+      });
+    });
+  });
+
+  describe('Generic', () => {
```
alisey: Why 'generic' and not 'focus'?

src/Backoffice/Button/Button.js
```diff
@@ -67,6 +68,7 @@ class Button extends WixComponent {
           type={type}
           onMouseEnter={onMouseEnter}
           onMouseLeave={onMouseLeave}
+          {...attachHandlersAndState(this.props)}
```
alisey: I would prefer a more explicit API, don't mind it being more verbose. Too much magic right now imo, if I want to understand how the component works or want to change it I have no choice but to go and see implementation of `attachHandlersAndState`.

src/Backoffice/Button/Button.protractor.driver.js
```diff
@@ -1,10 +1,25 @@
...
+ */
+export const createButtonInternals = element => ({
+  nativeFocusableElement: element,
+  clickableElements: [() => element]
+});
+
+const focusableDriverFactory = createFocusableDriverFactory(createButtonInternals);
+
+const buttonDriverFactory = element => ({
+  ...focusableDriverFactory(element),
```
alisey: Can we have a function here that receives button internals and avoid the intermediate step above that creates `focusableDriverFactory`?

src/common/Focusable/Focusable.protractor.driver.js
```diff
@@ -0,0 +1,52 @@
+import {isFocused, hasAttribute} from '../../test-common';
+
+/** TypeScript description of FocusableInternals:
+
+An inteface that describes the functionality of a component's internal elements,
```
alisey: The comment is too abstract. What I really need to know here is what `createFocusableDriverFactory` is and how am I supposed to use it.

But ideally we would just have a mixin that I add to my driver without all the layers of indirection.

src/test-common.js
```diff
@@ -6,6 +7,14 @@ export {
     isFocused,
     scrollToElement
 } from 'wix-ui-test-utils/protractor';
+/**
+ * hasAttribute: Protractor helper since I've seen too many times
+ * somehting like this: isDisabled: ()=> !!component.getAttribute('disabled')
+ * which returns always true.
+ */
+export const hasAttribute = (component, attributeName) =>
```
alisey: `component` is actually an element finder, right? Would prefer the parameter name to indicate that.

And the comment could be more to the point:

`// Protractor helper for checking the presence of an attribute`

src/common/Focusable/FocusableHOC.js
hepiyellow: Decided with @shlomitc to put this in the backlog.

src/Backoffice/Button/Button.e2e.js
hepiyellow: The describe('Focus') is inside `runFocusTests()`. I just want a placeholder for generic re-usable test, as opposite to tests which are specific to the component. It is important to have the Generic tests in a separate `describe` block, in order to separate the `before / after` behavior.

src/Backoffice/Button/Button.js
hepiyellow: Added this to open issues.

src/Backoffice/Button/Button.protractor.driver.js
hepiyellow: Well, right now we are still under the assumption that there are some driver methods which we do NOT want to expose to WSR consumers. so the `ButtonInternals` is used both for the "public" driver, and for the "Enhanced" driver (for E2E).  so in order to re-use the code, there has to be an intermediate step.
We can remove it later, if we decide to expose everything.

src/common/Focusable/Focusable.protractor.driver.js
hepiyellow: I added this comment:
"From the FocusableInternals, we can derive the "public" FocusableDriver (which is exposed to WSR consumers),
or we can derive an "Enhanced" driver which is used for e2e tests only."

again, if we decide to expose all mothods , then we can simplify this.

src/test-common.js
hepiyellow: done

hepiyellow:
Replaced by #1614 which has an InputMethod singleton.

PR #1614, opened 2018-04-02
`<Button>` - Make it focusable by keyboard only - using new FocusableHOC (#1614)

hepiyellow:
Replaces PR #1601 

- Added FocusableHOC with unit-test - including an InputMethod singleton.
- Added reusable FocusableE2E tests and driver. (Driver include hidden INTERNAL methods)
- Use FocusableHOC with Button
- Use Focusable E2E test and driver in Button E2E test.

Open Issues:
 - Storybook for Button Standard 5.2 is broken. AutoExample error.
 - Can you please check if it gets blurred automatically when it becomes disabled?


src/Backoffice/Button/Button.e2e.js
```diff
@@ -2,58 +2,63 @@ import eyes from 'eyes.it';
...
+    describe('FocusableDriver', () => {
+      it('should be focused when clicked', async () => {
+        await driver.click();
+        expect(await driver.isFocused()).toBe(true);
+      });
+    });
+  });
+
+  describe('Generic', () => {
+    runFocusTests(driver, storyUrl);
```
alisey: Like this

src/Backoffice/Button/Button.js
```diff
@@ -16,6 +17,7 @@ class Button extends WixComponent {
 
   static propTypes = {
     ...ButtonLayout.propTypes,
+    ...FocusablePropTypes,
```
alisey: Are these props going to be visible in the storybook?

src/common/Focusable/FocusableHOC.js
```diff
@@ -0,0 +1,168 @@
...
+  [FOCUS_VISIBLE_ATTR_NAME]: PropTypes.bool
+};
+
+/**
+ * Singleton for managing current input method (keyboard or mouse).
+ *
+ * TODO: What should be the input method if a coponent is focused proramatically?
+ */
+class InputMethod {
+  method = 'keyboard';
```
alisey: Would be nice to add a comment explaining that issue you discovered. I'm just imagining someone could run into other issue with would require this to be `mouse` by default, and then they would fix one bug and introduce another.

src/common/Focusable/FocusableHOC.js
```diff
@@ -0,0 +1,168 @@
...
+  focusableOnFocus: PropTypes.func,
+  focusableOnBlur: PropTypes.func,
+  [FOCUS_ATTR_NAME]: PropTypes.bool,
+  [FOCUS_VISIBLE_ATTR_NAME]: PropTypes.bool
+};
+
+/**
+ * Singleton for managing current input method (keyboard or mouse).
+ *
+ * TODO: What should be the input method if a coponent is focused proramatically?
```
alisey: it should be keyboard

src/common/Focusable/FocusableHOC.js
```diff
@@ -0,0 +1,168 @@
...
+
+  setMethod(method) {
+    if (method !== this.method) {
+      this.method = method;
+      this.subscribers.forEach(f => f());
+    }
+  }
+}
+
+export const INPUT_METHOD = new InputMethod();
```
alisey: Don't like this kind of naming for instances. Shouldn't this be private anyway?

src/common/Focusable/FocusableHOC.js
```diff
@@ -0,0 +1,168 @@
...
+
+/*
+ * TODO: Consider adding 'disabled' state to this HOC, since:
+ * 1) When component is disabled the onMouseDown should not be called.
+ * 2) When component is focused and then it becomes disabled, then the focus needs to be blured.
```
alisey: Can you please check if it gets blurred automatically when it becomes disabled?

src/common/Focusable/FocusableHOC.js
```diff
@@ -0,0 +1,168 @@
...
+      this.subscribers.forEach(f => f());
+    }
+  }
+}
+
+export const INPUT_METHOD = new InputMethod();
+
+/*
+ * TODO: Consider adding 'disabled' state to this HOC, since:
+ * 1) When component is disabled the onMouseDown should not be called.
```
alisey: Which onMouseDown? Don't understand this part.

src/common/Focusable/Focusable.protractor.driver.js
```diff
@@ -0,0 +1,21 @@
...
+      rootElement,
+      hasFocusState: () => hasAttribute(rootElement, 'data-focus'),
+      hasFocusVisibleState: () => hasAttribute(rootElement, 'data-focus-visible'),
+      clickRoot: () => rootElement.click(),
+      clickableElements
+    }
+  };
+};
+
+export default focusableDriverFactory;
```
danielstro: Maybe I'm missing something but why not use "export default function () {...}?

src/common/Focusable/FocusableHOC.js
```diff
@@ -0,0 +1,168 @@
...
+ */
+export function focusableStates(props) {
+  if (!props) {
+    throw new Error('FocusableHOC.focusableStates(props): props must be defined');
+  }
+  return {...pick(props, FOCUS_ATTR_NAME, FOCUS_VISIBLE_ATTR_NAME)};
+}
+
+export const FocusablePropTypes = {
+  focusableOnFocus: PropTypes.func,
```
danielstro: supposed to be [ON_FOCUS_PROP_NAME]?

src/common/Focusable/FocusableHOC.js
```diff
@@ -0,0 +1,168 @@
...
+ *
+ * TODO: What should be the input method if a coponent is focused proramatically?
+ */
+class InputMethod {
+  method = 'keyboard';
+  subscribers = new Map();
+
+  constructor() {
+    if (typeof window !== 'undefined') {
+      window.addEventListener('mousedown', () => this.setMethod('mouse'));
```
danielstro: Not sure how big of an issue this is, but we never remove these listeners. Maybe add them after mount and remove them on unmount?

src/common/Focusable/FocusableHOC.js
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  /**
+   * Is the current input method `keyboard`. if `false` is means it is `mouse`
+   */
+  isKeyboard() {
+    return this.method === 'keyboard';
+  }
+
+  setMethod(method) {
```
danielstro: This name is a bit confusing. It sets the method but also runs all the callbacks.

src/common/Focusable/FocusableHOC.js
```diff
@@ -0,0 +1,168 @@
...
+
+  if (isStatelessComponent(Component)) {
+    throw new Error(`FocusableHOC does not support stateless components. ${getDisplayName(Component)} is stateless.`);
+  }
+
+  class FocusableHOC extends React.PureComponent {
+    wrappedComponentRef = null;
+    focusedByMouse = false;
+
+    constructor(props) {
```
danielstro: I don't think you need a constructor. I think you can just write
state = {...}

src/Backoffice/Button/Button.js
hepiyellow: I hope not (-:
Currently I noticed that the story is broken. I added it to Open Issues.

src/common/Focusable/FocusableHOC.js
hepiyellow: fixed

src/common/Focusable/FocusableHOC.js
hepiyellow: You are right, it should be private.

src/common/Focusable/FocusableHOC.js
hepiyellow: Later

src/common/Focusable/FocusableHOC.js
hepiyellow: This like is a mistake, leftover from previous solution.

src/common/Focusable/Focusable.protractor.driver.js
hepiyellow: same thing, I don't have a preference in this case.

src/common/Focusable/FocusableHOC.js
hepiyellow: thanks.

src/common/Focusable/FocusableHOC.js
hepiyellow: Not a real issue. You are technically right. And it helps testing.
But it also is room for more bugs, and if it isn't really necessary so I am ok with leaving it like this.
I think @alisey agrees.

src/common/Focusable/FocusableHOC.js
hepiyellow: I am ok with it. It's a small enough class to get oriented easily. and it's a "private" method.

src/common/Focusable/FocusableHOC.js
hepiyellow: you are right

src/common/Focusable/FocusableHOC.js
alisey: @danielstro these listeners are added globally on page load. And we need them the entire time, including before the components are mounted and after they're unmounted.

src/common/Focusable/FocusableHOC.js
alisey: We could rename to `updateMethod()`, but on the other hand `setState()` in React does more than set state too.

src/Backoffice/Button/Button.js
argshook: yes, they will be visible

src/ButtonLayout/ButtonLayout.scss
```diff
@@ -45,6 +45,10 @@ $base-transition: all 100ms linear;
   outline: none;
 
   transition: $base-transition;
+
+  &#{$focusVisible} {
```
argshook: https://github.com/wix/wix-style-react/blob/focus/26b-FocusableHOC2/src/common.scss#L1 not sure about this, i'd be explicit and import `focusable.scss` where its needed

src/common/Focusable/FocusableHOC.js
```diff
@@ -29,25 +29,26 @@ export function focusableStates(props) {
...
   [FOCUS_VISIBLE_ATTR_NAME]: PropTypes.bool
 };
 
 /**
  * Singleton for managing current input method (keyboard or mouse).
- *
- * TODO: What should be the input method if a coponent is focused proramatically?
  */
 class InputMethod {
+  // Default is keyboard in case an element is focus programatically.
```
alisey: 'is focus' -> 'is focused', 'programatically' -> 'programmatically'. Also I think you should mention tabbing into the page from the address bar.

src/common/Focusable/FocusableHOC.js
```diff
@@ -84,12 +85,11 @@ class InputMethod {
...
 }
 
-export const INPUT_METHOD = new InputMethod();
+const INPUT_METHOD = new InputMethod();
 
 /*
  * TODO: Consider adding 'disabled' state to this HOC, since:
- * 1) When component is disabled the onMouseDown should not be called.
- * 2) When component is focused and then it becomes disabled, then the focus needs to be blured.
+ * - When component is focused and then it becomes disabled, then the focus needs to be blured.
```
alisey: Not sure what this means.

src/common/Focusable/FocusableHOC.js
```diff
@@ -84,12 +85,11 @@ class InputMethod {
   }
 }
 
-export const INPUT_METHOD = new InputMethod();
+const INPUT_METHOD = new InputMethod();
```
alisey: All caps makes it look like a configuration constant of a primitive type, call it like a normal variable `inputMethod`.

Or even, since it's supposed to be a singleton anyway, just do:
```
const inputMethod = new class { ... }
```

src/common/Focusable/FocusableHOC.js
```diff
@@ -29,25 +29,26 @@ export function focusableStates(props) {
...
+  // Default is keyboard in case an element is focus programatically.
   method = 'keyboard';
   subscribers = new Map();
 
   constructor() {
     if (typeof window !== 'undefined') {
       window.addEventListener('mousedown', () => this.setMethod('mouse'));
       window.addEventListener('keydown', () => this.setMethod('keyboard'));
+      // We need to listen on keyUp, in case a TAB is made from the browser's URL input,
+      // so the keyDown is not fired, only the keyUp.
```
alisey: Since the keyboard is default now, do we still need keyUp listener?

src/common/Focusable/FocusableHOC.spec.js
```diff
@@ -0,0 +1,246 @@
...
+
+    it('should have focus and focus-visible [when] focused [given] initial render', () => {
+      const driver = createDriver(<WithFocusableComp2/>);
+
+      driver.focus();
+      // Default input is keyboard
+      expectKeyboardFocused(driver);
+    });
+
+    it('should have focus and focus-visible [when] tabbed in [given] mouseDown', () => {
```
danielstro: Not sure I understand the mousedown part of this test. It seems like you only use tabin

src/common/Focusable/FocusableHOC.spec.js
```diff
@@ -0,0 +1,246 @@
...
+      jest.resetModules();
+      focusableModule = require('./FocusableHOC');
+      WithFocusableComp2 = focusableModule.withFocusable(PureChildComponent);
+      // TODO: find a way to reset the eventHandlers which are added to the window.
+    });
+
+    it('should not have focus nor focus-visible [given] initial render', () => {
+      const driver = createDriver(<WithFocusableComp2/>);
+
+      expectNotFocused(driver);
```
danielstro: You are missing msg?

src/common/Focusable/FocusableHOC.spec.js
```diff
@@ -0,0 +1,246 @@
...
+    });
+
+    it('should have focus and focus-visible [when] focused [given] initial render', () => {
+      const driver = createDriver(<WithFocusableComp2/>);
+
+      driver.focus();
+      // Default input is keyboard
+      expectKeyboardFocused(driver);
```
danielstro: same

src/common/Focusable/FocusableHOC.spec.js
hepiyellow: msg is important only when there is more than one expect in thew test case.

src/common/Focusable/FocusableHOC.spec.js
```diff
@@ -0,0 +1,246 @@
...
+    });
+
+    it('should have focus and focus-visible [when] tabbed in withot keyDown', () => {
+      // This test case checks a scenario when the focus is on the browser's
+      // url input, and we press tab. The keyDown is not fired.
+      const driver = createDriver(<WithFocusableComp2/>);
+
+      driver.focus();
+      driver.fireKeyUp();
+      expectKeyboardFocused(driver, 'after focus');
```
danielstro: Maybe a more detailed message?

src/common/Focusable/FocusableHOC.spec.js
```diff
@@ -0,0 +1,246 @@
...
+      // This test must be here , after a test which finishes with the input method = mouse.
+      // this test is only to test that we are reseting the FocusableHOC global state proparely,
+      // usin the jest.resetModules() call.
+      const driver = createDriver(<WithFocusableComp2/>);
+
+      driver.focus();
+      expectKeyboardFocused(driver, 'after click');
+    });
+
+    it('should have focus and focus-visible [when] focused [given] mouseDown and blur', () => {
```
danielstro: This is still a very weird test. Are you sure there is a need for it? This test is a red flag for me.

src/common/Focusable/FocusableTestsE2E.js
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * IMPORTANT !!!
+ * It is the responsibility of the caller to reset the tested component's state beforeEach test.
+ * Reseting the state should include reseting the document's activeElement in such a way so that when pressing TAB one time,
+ * then the tested component would be focusable.
+ *
+ * Adds a `describe` block.
+ */
+export function runFocusTests(driver, storyUrl) {
+  runFocusTestsImpl(flattenInternalDriver(driver), storyUrl);
```
danielstro: I don't think extracting this to a function is needed, it's pretty straightforward.

src/common/Focusable/FocusableHOC.spec.js
alisey: I think the description is confusing. Also putting words in brackets - how does it help? For me it just makes the message harder to read.

src/common/Focusable/FocusableHOC.spec.js
danielstro: but like this it will print "undefined - hasFocus....", maybe don't print anything if there is no msg?

src/common/Focusable/FocusableHOC.spec.js
hepiyellow: Wrong desc message. thanks. fixed.

src/common/Focusable/FocusableHOC.js
hepiyellow: Yes, if the user clicks (input = mouse) then clicks on URL input, and then TAB.

src/Backoffice/Button/Button.js
hepiyellow: I decided to remove these props from propTypes for now, since AutoDocs can't parse it, and also we are not really exposing these props to the consumer.

src/ButtonLayout/ButtonLayout.scss
hepiyellow: I also put in `focusable.scss` the `FocusBox` and `FocusErrorBox` mixins, which are used by components which don't use (yet or ever) the FocusableHOC.
I put them there as a "separation of concerns" to not pollute `common.scss` with to much stuff.
Its true that the mixins and the focus state selectors might not be directly related (yet), but I think it's good to have them in one file.
I am ok with it as it is now. anyways `common.scss` is a mix of things (fonts, colors).
We can have a refactor task of splitting `common.scss`.

src/common/Focusable/FocusableHOC.js
hepiyellow: fixed.
I mention the address-bar in the comment before the mouseUp handler.

src/common/Focusable/FocusableHOC.js
hepiyellow: removed (1)

src/common/Focusable/FocusableHOC.js
hepiyellow: accepted

src/common/Focusable/FocusableHOC.spec.js
hepiyellow: If you are suggesting that the test would end after `blur`, assuming that blur clears the focus states, and we are at "initial state", then it IS more complicated then that, since there is an input-method state which is global. This test tests both the component's onFocus/Blur handers and the global InputMethod state.
It's true that we could have a unit-test for the InputMethod, and unit-test for the focus/blur handlers. But I prefer to have both in this case.

src/common/Focusable/FocusableTestsE2E.js
hepiyellow: It's not a matter of readability,
it's a matter of avoiding bugs. Without typescript it is easy to confuse the driver from the flattened driver. Its a simple solution which might help avoid a little head-ache some day.

src/common/Focusable/FocusableHOC.spec.js
danielstro: Then I think you should add a comment explaining that, or at least giving more information on why this is happening because it looks like an oversight.

PR #1615, opened 2018-04-02
Fix render new state while in transition casues menu to disappear

ronenst:
### What changed

Added locking state to menu transitions, animations will now have to complete before a new one is started. 

### Why it changed

If the active submenu changed while in transition, either manually by the end user or automatically, the menu disappeared completely.

---

- [x] Change is tested
- [ ] Change is documented
- [x] Build is green


src/SideMenu/DrillView/DrillView.e2e.js
```diff
@@ -0,0 +1,23 @@
...
+
+    waitForVisibilityOf(driver.element(), 'Cannot find DrillView');
+    driver.clickSubMenu(1);
+
+    waitForVisibilityOf(driver.getBackLink(), 'Cannot find BackLink');
+    driver.clickBackLink();
+
+    await browser.wait(() => driver.hasSingleDrillViewPanel());
+    const classNames = await driver.getDrillViewPanelClasses();
+    expect(classNames.split(' ').length).toBe(1);
```
shlomitc: what exactly is being asserted here? should it be a driver function?

ronenst: Yes, I thought about it might be better, it could be:
`driver.hasNoTransitionClassesInDrillView()`?

I could either check the count, or check the class names don't include `Appear`, but checking count is safer (could break this fix and change the animation class names and the test wouldn't fail).

PR #1623, opened 2018-04-04
fix(DataTable): add fallback to text-align:start

skyiea:
`text-align: start` [is experimental](https://developer.mozilla.org/en-US/docs/Web/CSS/text-align) and it not supported in IE11 & Edge.
Adding fallback to `left` value.


lbelinsk:
@skyiea can you please add a comment for it in the code?
In the fiture it might look like a bug

PR #1624, opened 2018-04-04
Make keyboard-only focusable for: `<Checkbox> <RadioButton> <RichTextAreaButton> <ToggleSwitch>`

hepiyellow:


src/Backoffice/Button/Button.e2e.js
```diff
@@ -4,12 +4,14 @@ import autoExampleDriver from 'wix-storybook-utils/AutoExampleDriver';
...
 
+const NO_DESCRIPTION = '';
+
 describe('Backoffice Button', () => {
   const storyUrl = getStoryUrl('5. Buttons', '5.1 Standard');
   const driver = buttonTestkitFactory({dataHook: 'storybook-button'});
 
   // Specific as opposed to 'Generic' tests like the Focusable tests.
-  describe('Specific', () => {
+  describe(NO_DESCRIPTION, () => {
```
argshook: huh? why?

src/Checkbox/Checkbox.protractor.driver.js
```diff
@@ -1,15 +1,43 @@
...
+    clickableElements: [checkboxElement, childrenElement]
+  });
+
+  const publicDriver = {
+    element: () => element,
+    click: () => element.click(),
+    getLabel: () => element.$(`label`),
+    getInput: () => element.$(`input`),
+    isChecked: () => element.$(`input`).isSelected(),
+    isDisabled: async () => await element.$(`input`).getAttribute('disabled') !== null,
```
argshook: i suppose if we stop relying on protractor control flow and do `async`/`await` then need to do it everywhere? like, in `click`, `getLabel`... methods also

src/Checkbox/Checkbox.scss
```diff
@@ -2,13 +2,27 @@
...
+  outline: none;
+  
   label {
     display: flex;
     cursor: pointer;
     justify-content: center;
     align-items: center;
   }
+
+  &#{$focusable-focus-visible} {
```
argshook: i reeeally vote for explicit import for this one

src/Checkbox/Checkbox.protractor.driver.js
```diff
@@ -1,15 +1,43 @@
-import {isFocused} from '../test-common';
 import styles from './Checkbox.scss';
+import focusableDriverFactory from '../common/Focusable/Focusable.protractor.driver';
+import {mergeDrivers} from '../test-common';
```
argshook: i don't know... i get the idea but it feels as unnecesarry complication. if we want to make driver which is not exposed... then let's just not expose it. `merge`, `flatten`, symbol.. wont it make harder to maintain without much benefit? maybe i miss something, but for what i understand, wouldn't do any of that

src/RadioGroup/RadioGroup.e2e.js
```diff
@@ -37,29 +39,71 @@ describe('RadioGroup', () => {
...
-        expect(radioGroupDriver.isRadioFocused(0)).toBe(false);
-        await pressTab();
-        expect(radioGroupDriver.isRadioFocused(0)).toBe(true);
+        // TODO: replace with forEachAsync
+        for (let index = 0; index < NUM_OF_BUTTONS_IN_EXAMPLE; index++) {
+          const driver = flattenInternalDriver(groupDriver.getButtonDriver(index));
+          await expectNotFocused(`button ${index} - before`, driver);
+          await pressTab();
+          await expectFocusedByKeyboard(`button ${index} - after`, driver);
+          index === 0 && await eyes.checkWindow(`button ${index} with focus-visible`);
```
argshook: move out from loop and no need for `index === 0`

src/Backoffice/Button/Button.e2e.js
hepiyellow: I need to separate the Generic tests so they run in a separate `describe` block, so I have control over the `beforeEach`.
So the non-generic tests should be wrapped with a `describe` block.
I used to call the `Specific` and `Generic`, but it looks bad in the combined test names "Button Specific ....", so I decided to have the "Specific" tests with no description, and only have "Generic" for the generic test suites.

src/Checkbox/Checkbox.protractor.driver.js
hepiyellow: The driver doesn't need to do async/await for 1 liner methods. this `isDisabled` is a mistake. not mine. I didn't fix it. I can fix it.

src/Checkbox/Checkbox.scss
hepiyellow: I can do that.

src/Checkbox/Checkbox.protractor.driver.js
hepiyellow: I see your pain. 
We'll discuss it in the next weekly.

PR #1627, opened 2018-04-05
Icons: Restore and EmptyTrash added

kelyar:
### What changed
These are two new icons we need in Media Manager G6
EmptyTrash
Restore

...

### Why it changed

...

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


argshook:
@milkyfruit pretty please
@alisey fyi, since you work on icons

alisey:
So the "Restore" icon is already in `wix-ui-icons-common` under the name "Revert". Not sure how to re-export it from WSR since the index is auto-generated.

Let's just add it to WSR as "Revert" and figure out re-export later.

milkyfruit:
I approve. @argshook let's add both icons to WSR.

About Wix-UI:
As Alexey mentioned, there's already "Revert", so no need to add this one. While "empty trash" I'll tweak it to be in 24x24 canvas and will send it directly to @alisey.

argshook:
wait im lost, help

1. no need to add `Restore`, cause we should use `Revert` from wix-ui-icons-common (?)
1. no need to add `EmptyTrash`, cause the correct one will be in wix-ui-icons-common (?)

so nothing to be done in this PR and we should focus on exporting icons from wix-ui-icons-common?

@alisey

alisey:
@argshook, I discussed this with Shlomi, what he said is that until we release the new version of wix-style-react that doesn't include icons, we don't want to tell the consumers to start using wix-ui-icons-common. So let's keep everything as usual and add these two icons to WSR.

milkyfruit:
@argshook all Wix-UI general icons use 24x24px canvas size and are fixed size.
Wix Style React icons are resizable (due to different component code) and use _em_ measure.

Let's keep these icons separate, while the names should be the same in both repositories.
We need to rename `Restore` to `Revert` in this pull request.

PR #1654, opened 2018-04-09
add getInnerContent method to ModalDriver

eranshabi:
we need to get the portal of the modal, so I added a function that let us do it.

### DIFF
--- a/src/Modal/Modal.driver.js
+++ b/src/Modal/Modal.driver.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
+import ReactModal from 'react-modal';
 import ReactTestUtils from 'react-dom/test-utils';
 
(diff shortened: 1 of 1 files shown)


src/Modal/Modal.driver.js
```diff
@@ -15,6 +16,7 @@ const modalDriverFactory = ({element, wrapper, component}) => {
     isThemeExist: theme => !!getPortal().querySelector(`.${theme}`),
     getChildBySelector: selector => getPortal().querySelector(selector),
     isScrollable: () => !getPortal().classList.contains('portalNonScrollable'),
+    getInnerContent: () => wrapper.find(ReactModal).node.portal,
```
lbelinsk: Why are you importing reactModal?
You can take it from the dom, it is the root element:
https://github.com/wix/wix-style-react/blob/master/src/Modal/Modal.js#L117

PR #1655, opened 2018-04-10
Style/input font smoothing

illBeRoy:
### What changed

Added font smoothing to input

### Why it changed

To match the appearance of the input component to the one that shows in the storyboard.

How it looks in Storybook:
Storybook

How it looks in production (notice that the placeholder font is "bolder"):
Production

This PR ought to fix that.
---

- [ ] Change is tested
- [X] Change is documented
- [ ] Build is green


argshook:
hello, thank you for contributing!

currently in our readme, aka, [homepage](https://wix.github.io/wix-style-react/?selectedKind=Introduction&selectedStory=Getting%20started&full=0&addons=0&stories=1&panelRight=0) we have a "Tips" section with exactly the same font-smoothing properties but for the whole `html`.

im not sure if we should add font smoothing properties per components, cause then all of them will need that.

@shlomitc @lbelinsk wdyt?

shlomitc:
not sure what is the best practice here...
Anyway, I find it useful to be added to the entire page as this is a really common use case for almost every component.

shlomitc:
As we have a documented solution, I'm closing this PR for now.
If any other thoughts are raised, please re-open.

Issue #1612, opened 2018-03-29
datepicker position is off when opened inside modal on safari


liorur:
### Priority:
- [ ] Blocker
- [ ] Urgent
- [ ] Normal
- [X] Minor

### Steps to reproduce:
 1. place datepicker inside modal
 2. open on safari
@argshook

liorur:
![screen shot 2018-03-29 at 15 45 14](https://user-images.githubusercontent.com/1437073/38091914-832ef15c-336f-11e8-8892-9c809d529f51.jpg)

shlomitc:
Fixed in #1656. will be released in 3.1.4

PR #1656, opened 2018-04-10
fix(wix-style-react): #1612: fixed incorrect offset for safari

cTux:
### What changed

ReactModal_Content's position changed to 'relative'.

### Why it changed

Fixing #1612
`element.offsetParent` in Safari skips ReactModal_Content as parent because of missing position property (should be not static).

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


shlomitc:
Is there a way to see the fix in the storybook?

cTux:
Now you can find it at Modals -> Modal -> Open with DatePicker.

Issue #1641, opened 2018-04-08
Card - Card Header without a divider gets wrong CSS


shlomitc:
![image](https://user-images.githubusercontent.com/6093192/37778642-21395afe-2df3-11e8-95b3-7f90488012db.png)

cTux:
Should be:

![image](https://user-images.githubusercontent.com/873220/38545059-d56b19a0-3cb1-11e8-843c-8f85cbb21c42.png)

shlomitc:
more or less, there is still a +- 1px on the right card that's missing

PR #1662, opened 2018-04-10
fix(wix-style-react): #1641: fixed padding of next element related to header

cTux:
### What changed

Element next to Card.Header (it's a Card.Content) now has zero padding-top if there is no divider between header and content.

### Why it changed

Fixing https://github.com/wix/wix-style-react/issues/1641
There is should be no double padding between header and content.

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green

### DIFF
--- a/src/Card/ButtonHeader/ButtonHeader.js
+++ b/src/Card/ButtonHeader/ButtonHeader.js
@@ -37,7 +37,8 @@ class ButtonHeader extends WixComponent {
     const headerClasses = classNames({
       [styles.headerOnlyTitle]: !subtitle,
       [styles.headerTitleSubtitle]: subtitle,
-      [styles.withDivider]: !withoutDivider
+      [styles.withDivider]: !withoutDivider,
+      [styles.withoutDivider]: !!withoutDivider
     });
 
     const buttonClass = classNames({
--- a/src/Card/Header/Header.scss
+++ b/src/Card/Header/Header.scss
@@ -11,6 +11,10 @@ $header-padding: 27px 30px 26px;
     border-bottom: 1px solid $D60;
   }
 
+  &:not(.with-divider) + * {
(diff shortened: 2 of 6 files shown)


src/Card/Header/Header.scss
```diff
@@ -11,6 +11,10 @@ $header-padding: 27px 30px 26px;
     border-bottom: 1px solid $D60;
   }
 
+  &:not(.with-divider) + * {
```
argshook: i would prefer this to happen as separate class, which is toggled from component. since there is `withoutDivider` prop, it can easily be used to adjust classnames accordingly.

maintaining selectors like this is no fun

cTux: in this case we need to update 4 more similar files (different headers for Card component)
logic with adding new class will be the same

src/Card/ButtonHeader/ButtonHeader.js
```diff
@@ -37,7 +37,8 @@ class ButtonHeader extends WixComponent {
     const headerClasses = classNames({
       [styles.headerOnlyTitle]: !subtitle,
       [styles.headerTitleSubtitle]: subtitle,
-      [styles.withDivider]: !withoutDivider
+      [styles.withDivider]: !withoutDivider,
+      [styles.withoutDivider]: !!withoutDivider
```
argshook: no need for `!!` as `withoutDivider` propType is bool and there's even default prop set to `false`

argshook:
just for the record, the `padding-top` fix is not really fixing all issues. for example, having `<ToggleSwitch/>` in header is different height from card with link in header.

we should make a separate task to go over Card component and align it to zeplin https://app.zeplin.io/project/5864e02695b5754a69f56150/screen/58becb51a8a68f975297d222

Issue #1630, opened 2018-04-05
MultiSelect height is not correct


andriuss:
### Summary:
MultiSelectComposite heigh is 36px
https://wix.github.io/wix-style-react/?selectedKind=3.%20Inputs&selectedStory=3.8%20%2B%20MultiSelectComposite&full=0&addons=0&stories=1&panelRight=0

It should be 34px normal height as other input elements.
Example - https://wix.github.io/wix-style-react/?selectedKind=3.%20Inputs&selectedStory=3.1%20%2B%20Input&full=0&addons=0&stories=1&panelRight=0

### Priority:
- [ ] Minor

### Steps to reproduce:
 1. Check Storybook

### Expected behavior:
Height is consistent with input elements

### Relevant Components:
MultiSelect

shlomitc:
Are you sure that MultiSelect is the problematic one? or maybe the Input should be 36px and not 34px.
@wuwa - can you please assist here?

wuwa:
all should be 36px height (including the border). btw - i see multiselect as 38 right now.

andriuss:
I am sure all the inputs should be the same height. You pick a correct height.

argshook:
@shlomitc yes it's multiselect. it has `input` inside with `border: 0`.

that input has height 36px. multiselect has border which is 2px, and so we end up with 36 + 2 = 38

wuwa:
why is the input with border 0? can't it be used as-is inside the multi?

argshook:
nope. Multiselect is using `<Input/>` inside, and by default `<Input/>` has a border. if it weren't somehow set to 0, we would see blue border around multiselect and also blue border around input, so two borders.

im describing current implementation, not discussing how it should be done

shlomitc:
Fixed in #1663 and will be released in 3.1.4

shlomitc:
Fix failed in CI. reopening

argshook:
fixed by #1683

PR #1663, opened 2018-04-10
fix(wix-style-react): #1630: Fix Multiselect size

ihork:
### What changed

Wrapper for input and tags in multiselect was added. It provides a negative margin which decrease a size of inputWithTags wrapper, which draws border around multiline.

### Why it changed

It was made because multiline contains input 36px (with hidden border) and tags, so this content influences on multiline height. inputWithTags wrapper draws a border around component, so component receives 36 + 2 = 38px. Simple override of content's height in multiline style is not an option because we need to cover all use cases with input size (normal, small, large) and in case that sizes will be changed, we will need to change them in multiline style too. So, this fix provides an independent from other styles solution which handle all use cases and don't influence on calculations of height.

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green

### DIFF
--- a/src/MultiSelect/InputWithTags.js
+++ b/src/MultiSelect/InputWithTags.js
@@ -92,29 +92,31 @@ class InputWithTags extends React.Component {
         onMouseOut={() => this.handleHover()}
         data-hook={this.props.dataHook}
         >
-        {tags.map(({label, ...rest}) => <Tag key={rest.id} disabled={disabled} onRemove={onRemoveTag} {...rest}>{label}</Tag>)}
-        <span className={styles.input} data-hook="inner-input-with-tags">
-          <div className={styles.hiddenDiv} style={{fontSize}}>
-            {this.state.inputValue}
-          </div>
+        <div className={styles.contentWrapper}>
+          {tags.map(({label, ...rest}) => <Tag key={rest.id} disabled={disabled} onRemove={onRemoveTag} {...rest}>{label}</Tag>)}
+          <span className={styles.input} data-hook="inner-input-with-tags">
+            <div className={styles.hiddenDiv} style={{fontSize}}>
+              {this.state.inputValue}
+            </div>
 
-          <Input
-            width={this.props.width}
-            ref={input => this.input = input}
-            onFocus={() => this.handleInputFocus()}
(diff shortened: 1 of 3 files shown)


src/MultiSelect/InputWithTags.js
```diff
@@ -92,29 +92,31 @@ class InputWithTags extends React.Component {
...
-              if (!delimiters.includes(e.target.value)) {
-                this.setState({inputValue: e.target.value});
-                desiredProps.onChange && desiredProps.onChange(e);
-              }
-            }}
-            />
-        </span>
+        <div className={styles.contentWrapper}>
+          {tags.map(({label, ...rest}) => <Tag key={rest.id} disabled={disabled} onRemove={onRemoveTag} {...rest}>{label}</Tag>)}
+          <span className={styles.input} data-hook="inner-input-with-tags">
```
argshook: instead of adding wrapper and adjusting testkit, it's enough to `.input { margin-top: -1px; margin-bottom: -1px; }`

argshook:
btw, build is failing because of screenshot diff, which is a-ok

src/MultiSelect/InputWithTags.js
ihork: @argshook Want to notice that your idea also requires to reduce paddings of Tag components, which are also inside multiline. So, when we have few lines, there will be a case - each line will become 34px, so 2 lines multiline will be 34px * 2 + 2px = 70px. And in my solution it will be 36px * 2 = 72px. That's why I chose a wrapper, but If you're okay with that, I'll change solution.

argshook: i see. it's possible to go without wrapper but then i see you'd need to also adjust tags padding, but it's separate component, so you'd need to tweak that one too, but it's used as standalone component so changing its dimensions will have more impact and yaddayadda...

so unless you work on this more and see it's not too much work, please do it without wrapper.
otherwise meh, wrapper it is

ihork: @argshook we have a list of tasks on this guild week, so we'll consider your approach if there will be time at the end of the week. Thanks

PR #1665, opened 2018-04-11
add option to pass render props function as title prop to pageHeader

liorur:
### What changed

added option to pass render props function as title prop to pageHeader

### Why it changed

if title is a component - it needs to be aware of the minimize status

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


src/PageHeader/PageHeader.driver.js
```diff
@@ -14,6 +14,7 @@ export default ({element}) => ({
   subtitleText: () => subtitleElement(element).textContent,
   isSubtitleExists: () => !!subtitleElement(element),
   isBreadcrumbsExists: () => !!breadcrumbsElement(element),
+  breadcrumbsText: () => breadcrumbsElement(element).textContent,
```
amiryonatan: Won't it be easier to find the breadcrumbs element and then return the breadcrumbs driver ? so u have all the functionality u need ?

liorur: possibly, there are very few actions here, so i thought it redundent

Issue #1666, opened 2018-04-11
<Tabs> - hasDivider prop is ignored. 


hepiyellow:
No divider is displayed.

I think it was removed by accident in this commit:
https://github.com/wix/wix-style-react/commit/260b91cd953be4fd44321d906f30fbdc233ff890

argshook:
closed by #1676

PR #1676, opened 2018-04-17
Adding accidentally removed hasDivider prop support

nickg-wix:
### What changed
Adding accidentally removed hasDivider prop support for Tabs component. Turning it on by default (as documented)

### Why it changed
Issue - https://github.com/wix/wix-style-react/issues/1666

---

- [x] Change is tested
- [x] Change is documented
- [ ] Build is green


src/Tabs/Tabs.js
```diff
@@ -5,18 +5,26 @@ import PropTypes from 'prop-types';
...
 
+  static defaultProps = {
+    hasDivider: true
+  }
+
   render() {
     const {sideContent} = this.props;
     const tabItemsProps = omit(['sideContent'], this.props);
+    const className = classNames(styles.container, {
+      [styles.hasDivider]: this.props.hasDivider
```
argshook: there's already props destructuring going on on line 20, you can use that

nickg-wix: You are right, will fix.

Issue #1647, opened 2018-04-08
onClickOustide - opened popovers don't close properly


shlomitc:
The issue was revealed when tapping on the `Range` component first `DatePicker` component but relevant to many components with a Popover behavior like `Autocomplete`.

The suspect: `onClickOutside` feature in the `WixComponent`
https://github.com/wix/wix-style-react/blob/4c59e6c4f95f8008568cfe540178050cd9a8302d/src/BaseComponents/WixComponent.js#L43

**First way - tapping**
Tap on the first Date Picker
Tap on the second Date Picker
both to stay open and the first one didn't close.

(The issue doesn't happen when clicking the mouse, only tap)

**Second way - Clicking and dragging:**
Click on the first DatePicker
Click on the second DatePicker but don't release.
Drage the mouse outside and release.

Range:
![image](https://user-images.githubusercontent.com/6093192/38472722-ce886a0a-3b8c-11e8-8655-e7e9bb5c2ca3.png)

Autocomplete:
![image](https://user-images.githubusercontent.com/6093192/38472748-5007f33e-3b8d-11e8-91df-e2c98dc5547b.png)

argshook:
closed by #1647

PR #1680, opened 2018-04-17
#1647 onClickOustide - opened popovers don't close properly

manishma:
### What changed

handle 'mouseup' and 'touchend' instead of 'click' when detect onClickOutside

### Why it changed

#1647

---

- [v] Change is tested
- [ ] Change is documented
- [v] Build is green


shlomitc:
Why `mouseUp` and not `mouseDown`?

argshook:
i think `mouseUp` is better as it more closely resembles behaviour of `click`. i want action to happen when i "finish" click and not immediatelly.

Issue #1675, opened 2018-04-16
Tooltip within Page does not follow element when Page state changes to minimize


amiryonatan:
### Summary:
When a Tooltip opens with a click trigger, and then scrolling down the Page which causes the Header to minimize, the Tooltip does not follow the opening element.

### Priority:
- [ ] Blocker
- [x] Urgent
- [ ] Normal
- [ ] Minor

### Steps to reproduce:
 1. Open https://wix.github.io/wix-style-react/?selectedKind=2.%20Layout&selectedStory=2.6%20%2B%20Page%20Example
 2. Click on `Tooltip`
 3. Scroll down

### Expected behavior:
Tooltip will follow the element once the Header minimizes

### Relevant Components:
`Page`

### Additional notes:
@Jefoss found the issue.

PR #1681, opened 2018-04-17
attempt to fix tooltip position during scroll

shlomitc:
This is an attempt to fix https://github.com/wix/wix-style-react/issues/1675
It means `Page` will trigger render also for the content of it when changing the state (minimize <=> not minimized).
As a side effect, `Tooltip` will trigger it's `componentWillReceiveProps` which can recalculate the position of it.

Not sure how it can be tested as this is affected by dom scrolling.
Any suggestions?


amiryonatan:
Looks ok to me, only way to test it is with e2e .. 
are u not worried of performance implications for cloning the entire content ?

argshook:
if this is an issue with tooltip, why try solve it with side effect in page?

btw, how do you reproduce? I think i know, but would like to know the exact reproduction. #1675 says `click on Tooltip` in [here](https://wix.github.io/wix-style-react/?selectedKind=2.%20Layout&selectedStory=2.6%20%2B%20Page%20Example) but there is no tooltip there

shlomitc:
@amiryonatan - I'm not sure that there will be much of performance hit as cloning it will just call render function of the content again when minimizing. so it's hard to tell if there will be a major effect...

@argshook - it relates to both. The Page needs to notify when it changes its size due to header changes, so that needs to happen anyway and I managed to make it only by cloning the content. Tooltip, on the other hand, was missing the logic of re-positionings.

The example is there - it just was not updated until you released the new version
image

amiryonatan:
I prefer the previous solution :)

src/Tooltip/Tooltip.js
```diff
@@ -6,9 +6,13 @@ import TooltipContent from './TooltipContent';
 import position from './TooltipPosition';
 import styles from './TooltipContent.scss';
 import {TooltipContainerStrategy} from './TooltipContainerStrategy';
+import throttle from 'lodash/throttle';
 
 const renderSubtreeIntoContainer = ReactDOM.unstable_renderSubtreeIntoContainer;
 
+//maintain a 60fps rendering
+const createAThrottledOptimizedFunction = cb => () => window.requestAnimationFrame(throttle(cb, 16));
```
argshook: `throttle` here does more harm than good, this line creates throttled function for each scroll event but doesn't reuse it, making `throttle` practically useless.

each scroll event calls `() => window.requestAnimationFrame(throttle(cb, 16))`. this is a function that creates throttled function inside which is called immediatelly. The next time `scroll` happens, it creates throttled function again.

you can try changing `16` to `9999999` and it would still behave the same, throttling doesn't bring any benefit.

instead it should use previously created function which is throttled.

i would do it simply like this:

```js
this._containerScrollHandler = throttle(() => this._updatePosition(this.tooltipContent), 16);
container.addEventListener('scroll', this._containerScrollHandler);
```

not sure what good `requestAnimationFrame` gives us

argshook: here's a quick demo on jsbin of three examples: no throttling, correct throttling and throttling done same as in this PR http://jsbin.com/nazorixoyu/edit?js,output

scroll up and down, you'll see numbers change - these represent times `draw` function was called. notice how `no throttle` and `wrong throttle` are same numbers, whereas `correct throttle` is slightly smaller number. you should also see `correct throttle` block to visually jump - a sign that it is actually throttled.

Issue #1679, opened 2018-04-17
`DatePicker` has locales hardcoded into component


argshook:
`<DatePicker/>` supports `locale` prop and currently it can be one of ``'en', 'es', 'pt', 'fr', 'de', 'pl', 'it', 'ru', 'ja', 'ko', 'tr', 'sv', 'no', 'nl', 'da'``.

these locales are imported into component https://github.com/wix/wix-style-react/blob/master/src/DatePicker/LocaleUtils.js

this means that even if consumer uses only `en`, component will still contain other locales.
moreover, we should support a way for consumer to supply more locales.

1. [ ] add ability to control what locales end up in a bundle
2. [x] add ability to provide additional, custom locales
3. [ ] this page displays dropdown with more locales than current `DatePicker` supports  https://www.wix.com/dashboard/[metasiteId]/settings need to make `DatePicker` so that it can support all of them

nickg-wix:
### Idea - Refactor DataPicker localization ot use `Intl` API

#### Pros
* remove dependency for data-fns library
* native browser support
* lightweight bundle size
* various languages support

#### Cons
* node has limited `Intl` API support (only English), meaning need to use polyfill in tests
* more logic on our end to construct locales for DayPicker
* `Intl` API support optionally 1 or 3 symbols labels for week days, while current implementation based on 2 symbols.

#### Polyfill

Use [Intl.js](https://github.com/andyearnshaw/Intl.js/) only in Unit tests, also package [intl-locales-supported](https://github.com/yahoo/intl-locales-supported) is needed to properly include the polyfill.

shlomitc:
Soon, we're going to support `dynamic Imports` in our components, so maybe we could just import  requested language, while having a fallback to english.
@argshook - wdyt?

argshook:
> so maybe we could just 

yes, maybe we could just but need to investigate. Last time we investigated it was not possible (hence this issue)

shlomitc:
We plan to rework on all date and time components in the upcoming months, including doing performance optimizations and internationalization.
I'm closing the issue to clean old stuff from the backlog.

PR #1684, opened 2018-04-19
<DatePicker/> - Add ability to use external locales in [WIP]

nickg-wix:
### What changed
Added new DatePicker prop - `locales`. By using it customers could pass additional translations from [date-fns](https://github.com/date-fns/date-fns) library to DatePicker.
...

### Why it changed
Issue: [#1679](https://github.com/wix/wix-style-react/issues/1679)

---

- [x] Change is tested
- [ ] Change is documented
- [ ] Build is green


src/DatePicker/DatePicker.js
```diff
@@ -53,23 +53,10 @@ export default class DatePicker extends WixComponent {
...
-      'ru',
-      'ja',
-      'ko',
-      'tr',
-      'sv',
-      'no',
-      'nl',
-      'da'
-    ]),
+    locale: PropTypes.string,
```
nickg-wix: Don't really like this change, but seems thats the price we need to pay if want custom translations

argshook: @nickg-wix what if we do not add `locales` but instead extend `locale` prop to be `PropTypes.oneOf(['en', 'es' ..., PropTypes.shape(/* shape of locale object */)])`.

this way API remains the same but consumer is able to pass custom locale
if think consumer doesn't need to provide multiple locales at once. if they need to change it dynamically, they can pass different locale to `locale`

wdyt? would it be possible to do without changing too much of what you already did?

nickg-wix: Nice suggestion! Will change to this approach

src/DatePicker/DatePicker.js
```diff
@@ -53,22 +54,27 @@ export default class DatePicker extends WixComponent {
...
-      'ru',
-      'ja',
-      'ko',
-      'tr',
-      'sv',
-      'no',
-      'nl',
-      'da'
+    locale: PropTypes.oneOfType([
+      PropTypes.oneOf([
```
nickg-wix: `oneOf` seems to be limited to simple literal enum, it doesn't support mixed values.

nickg-wix:
On hold because of idea to use `Intl` API.

Varnius:
Investigated Intl API with a conclusion that it's not really an option atm because browser implementation of Intl is inconsistent with Node Intl polyfill (eg. produces different month names compared to browser under certain locales). This can lead to problems with SSR and tests run on node (jsdom). Also, Intl options of formatting dates are much more limited compared to date-fsn.
We should also update the version of date-fns we use, after they release a stable date-fsn v2.

shlomitc:
@argshook - what should we do with this PR?

argshook:
@shlomitc i have small tweaks and will merge

Issue #927, opened 2017-10-11
Remove margin from the last element of RadioGroup


joncys:
## Thanks for reporting an issue :)
You betcha!

### Summary:
Oh boy, I guess, you can't really change this anymore, because most probably everyone just relies on the margin being there and subtracts `12px` from the desired spacing, but having this just makes me a very very _very_ sad panda. 

Also, the OCD rage is barely bearable.

### Priority:
- [ ] Blocker
- [ ] Urgent
- [ ] Normal
- [x] Minor

### Steps to reproduce:
<img width="828" alt="screen shot 2017-10-11 at 15 52 27" src="https://user-images.githubusercontent.com/373939/31441634-6892184a-ae9c-11e7-9f09-565368604b27.png">

### Expected behavior:
I am to control my margins as I control my own destiny.

### Relevant Components:
`RadioGroup`

argshook:
could you please clarify? you wish for `radiogroup:last-child { margin-bottom: 0 }`?

joncys:
@argshook Yes. Because, there's no top margin for the `RadioGroup`, it is somewhat strange, that there's bottom margin (that results from the last item having that bottom margin, but still).

argshook:
closed by #1687

PR #1687, opened 2018-04-20
#927 Remove margin from the last element of RadioGroup

manishma:
### What changed

remove bottom margin from last item in vertical radio button group

### Why it changed

#927 

---

- [ ] Change is tested
- [ ] Change is documented
- [v] Build is green


src/RadioGroup/RadioGroup.js
```diff
@@ -20,28 +20,29 @@ class RadioGroup extends WixComponent {
...
+            <RadioGroup.Radio
+              dataHook={radio.props.dataHook}
+              value={radio.props.value}
+              name={this.name}
+              onChange={onChange}
+              vAlign={vAlign}
+              type={type}
+              disabled={disabled || disabledRadios.indexOf(radio.props.value) !== -1}
+              checked={radio.props.value === value}
+              style={style}
```
argshook: i would do it all here, to avoid nesting, `return`, `{` & `}`

`style={{marginTop: index > 0 && display === 'vertical' ? spacing : 0}}`

src/RadioGroup/RadioGroup.scss
```diff
@@ -19,7 +19,10 @@
 }
 
 .vertical .radio-wrapper {
-  margin-bottom: 12px;
+  margin-top: 12px;
```
argshook: if spacing is passed in `style` attribute, is it still needed in css?

src/RadioGroup/RadioGroup.driver.js
```diff
@@ -22,7 +22,8 @@ const radioGroupDriverFactory = ({element, wrapper, component}) => {
     isVerticalDisplay: () => isClassExists(element, 'vertical'),
     isHorizontalDisplay: () => isClassExists(element, 'horizontal'),
     isButtonType: () => isClassExists(element, 'buttonType'),
-    spacing: () => radios[0].style._values['margin-bottom'],
+    spacing: () => radios[1].style._values['margin-top'],
+    topMargin: () => radios[0].style._values['margin-top'],
```
argshook: let's not add testkit methods to check css, we have screenshot diffs, i think those are enough

src/RadioGroup/RadioGroup.js
manishma: ok, but prefer to remain the expression 
display === 'vertical' && index > 0 ? {marginTop: spacing} : {}

keep other the 'vertical' case untouched. and dont want to set marginTop: 0 explicit

src/RadioGroup/RadioGroup.driver.js
argshook: let's not add new testkit methods to check css, we have screenshot diffs, i think those are enough

src/RadioGroup/RadioGroup.scss
manishma: There is possible to set an empty  string value, then CSS plays the game

argshook: and it's ok, isn't it? if consumer passes empty string as `spacing`, let's use it as is,why not

manishma: You are dragging me into grey zone.  What documentations says about this. What expected behaviour by the product? I just made a fix without changing current behaviour.

argshook: grey zone is fun :)

product does not know about `spacing` prop.
there is already `defaultProps = { spacing: '12px' }` so most users don't set it by hand.

those who set `spacing='100px"` or whatever, wont see difference.
those who set `spacing="''"` will now get no spacing and i think it's correct behaviour. it's a small chance that anyone is doing this but possible, so we can mention this in changelog.

given all this, css seems redundant to me. expecially `&:first-child` part, since this is covered in JS now.

we're not only fixing issue but also cleaning code, it's all good

manishma: Ok, will remove this part from css.

argshook:
@manishma thanks for your patience :)

Issue #1669, opened 2018-04-12
<DatePicker /> No way to set the empty value and see the placeholderText in action


xmityaz:
### Summary:
Can't set the empty value for datePicker. If you render a <DatePicker /> without the value it will set `new Date()` by default. This makes `placeholderText`  useless (although it is an actual API prop). We have the placeholder text by our desing

### Priority:
Normal

### Steps to reproduce:
 1. render the code <DatePicker placeholderText="placeholder text" onChange={(value) => value} />

### Expected behavior:
Can see the "placeholder text". But instead you see the `new Date()` value
### Relevant Components:
DatePicker

PR #1690, opened 2018-04-20
fix(DatePicker): Add empty string as option to set DatePicker value in proptypes

xmityaz:
### What changed

Add possibility to set empty string as value to DatePicker without a console error of violating the proptypes

### Why it changed

When I've tried to reset the datepicker value with null it throws the error:

> prop on `input` should not be null. Consider using the empty string to clear the component or `undefined` for uncontrolled components.

If I'm setting it as undefined it tells me to not mess with controlled/uncontrolled `value` prop.

So the only way to see placehoder with controlled value is to set it to empty string, but it throws an errror to console with invalid prop:

> jsWarning: Failed prop type: Invalid prop `value` of type `string` supplied to `DatePicker`, expected `object`.


argshook:
hm, not sure if this is really an issue. `placeholderText` is displayed if you don't set `value` prop at all.

setting `value: undefined` also works ok for me.
`value` for datepicker should be object, empty string is a workaround.

looks like #1669 fixed this earlier (already released).

closing, please reopen if there's something i missed.

argshook:
@xmityaz 
> And then it impossible to clean up the value back to empty.
this was fixed in latest release, please verify.

argshook:
well im a little slow sometimes. turns out i just fixed locally and forgot to push :)

https://github.com/wix/wix-style-react/pull/1711

should become part of next release

argshook:
i merged https://github.com/wix/wix-style-react/pull/1711 with a fix.

closing this again, feel free to reopen (perhaps as issue) if it still persists.

the reason im not merging this PR because it is a workaround and can even be misleading. DatePicker would look like it supports empty string for no apparent reason, thus bringing confusion, i want to avoid that.

xmityaz:
@argshook works with `undefined`, thanks

Issue #1691, opened 2018-04-22
Input Prefix is cut in IE 11 on small inputs


mariafeldman:
### Priority:
- [ ] Blocker
- [ ] Urgent
- [V] Normal
- [ ] Minor

### Steps to reproduce:
 1. Add input with width less then 180px with prefix
```
      <div style="width: 180px">
        <TextField>
          <Label>
            {'blablabla'}
          </Label>
          <Input prefix={<Input.Unit value={'$$$'}/>}
                 error={error}
                 errorMessage={errorMessage}
                 value={this.props.value}/>
        </TextField>
      </div>
```

The input looks broken, like on attachment

### Relevant Components:
Input
#### Thanks a lot!

![dashboard](https://user-images.githubusercontent.com/2092879/39095987-02325f30-4652-11e8-908f-76b0898f36ce.png)

PR #1692, opened 2018-04-23
fix issue https://github.com/wix/wix-style-react/issues/1691 (Input P…

mariafeldman:
…refix is cut in IE 11 on small inputs)

### What changed
Fixed css of unit for prefix in input 

### Why it changed
There is a bug in ie for small inputs with prefix https://github.com/wix/wix-style-react/issues/1691
---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


argshook:
hmm, i wonder how this fixes anything, as `.unit` is `div` and `div`s are `display: block` by default. bug in IE or something? what do i not see?

mariafeldman:
@argshook yep, bug in ie 11 https://github.com/wix/wix-style-react/issues/1691

mariafeldman:
@argshook ie11 doesn't work well with flex

PR #1696, opened 2018-04-23
feat(notification): global notification can be hide after timeout

matveyok:
### What changed
Global notification now accepts timeout prop.

### Why it changed
Product need.


shlomitc:
1. What happens if you don't supply timeout? this will cause the notification to close even though it should stay open.
2. Please add this configuration also to the storybook example when the Global type is selected:
https://wix-wix-style-react-pr-1696.surge.sh/?selectedKind=8.%20Notification%20Bars&selectedStory=8.1%20Standard&full=0&addons=0&stories=1&panelRight=0

src/Notification/Notification.js
```diff
@@ -61,13 +61,11 @@ class Notification extends WixComponent {
...
   }
 
-  startCloseTimer({type, timeout}) {
-    if (type !== GLOBAL_NOTIFICATION) {
-      this.closeTimeout = setTimeout(
-        () => this.hideNotificationOnTimeout(),
-        timeout || DEFAULT_TIMEOUT
-      );
-    }
+  startCloseTimer({timeout}) {
```
argshook: @matveyok copy-pasting from slack so it doesn't fade away

Your PR makes a change in behaviour - all types of notifications will hide after default (6s) timeout, even if consumer didn't set one. I know for sure there are cases when notification should hide only after user action and not a timeout.

matveyok:
what about this change? still compatible with the previous behavior but now doesn't ignore the timeout if exists.

shlomitc:
@matveyok.
Since there is a default timeout for notifications, global notification will always get a timeout, which is not the wanted behavior.

matveyok:
@shlomitc please review my change, the default timeout is not part of the default props so it is the default only when I use it.

`<Notification ... type="global" />` - won't make the notification to disapear.
`<Notification ... type="global" timeout={2000}/>` - will make the notification to disapear after 2 seconds.

src/Notification/Notification.spec.js
```diff
@@ -194,43 +194,44 @@ describe('Notification', () => {
...
-    ['local', 'sticky'].forEach(type => {
+    ['local', 'sticky', 'global'].forEach(type => {
       describe(`Closing after timeout for ${type} Notification`, () => {
-        const defaultTimeout = 6000;
+        const someTimeout = 132;
 
-        it('should close after default timeout (6s)', () => {
-          driver = createDriver(renderNotificationWithProps({show: true, type}));
-          jest.runAllTimers();
+        if (type !== 'global') {
```
shlomitc: then you should add a test that makes sure that global stills appear after default timeout when not passing a specific timeout

matveyok:
updated

shlomitc:
@matveyok 
Please add the timeout configuration also to the storybook example when the Global type is selected:
https://wix-wix-style-react-pr-1696.surge.sh/?selectedKind=8.%20Notification%20Bars&selectedStory=8.1%20Standard&full=0&addons=0&stories=1&panelRight=0

matveyok:
something else?

shlomitc:
Thanks, looks good. maybe let's remove the `6000` ms and leave it as an empty fied, as we don't want to encourage timeout by default.
That's it

matveyok:
@shlomitc let's merge :) ?

PR #1697, opened 2018-04-24
Moving shadow

shypeleg:
### What changed

I've moved the shadow from the modal to the modal content layouts.

### Why it changed

because i'm using the modal without the provided layouts and i dont need the border. 

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


hepiyellow:
-  Why are you using the Modal without shadow?
-  Is this something that is coming from UX?
- Is there a spec for it?
- Would you like to contribute you new component to Wix-Style-React?

and...

Can't you simply override the box-shadow of the modal?

Also, I see that we don't have e2e tests with eyes for these components, so there is no way to track if they haven't been broken by this change.

PR #1705, opened 2018-04-29
add disableAnimations prop to disable the fadeIn animation

jonathanadler:
### What changed
Added a prop: `disableAnimations` which disables the addition of the `.fadeIn` class which adds fade in animation to the tooltip content.
...

### Why it changed
The animation was not configurable, and we needed to remove it for our use.
...

---

- [ v ] Change is tested
- [ v ] Change is documented - in storybook
- [ v ] Build is green

Thank you!


src/Tooltip/Tooltip.js
```diff
@@ -90,7 +90,10 @@ class Tooltip extends WixComponent {
     padding: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
 
     /** Allows updating the tooltip position **/
-    shouldUpdatePosition: PropTypes.bool
+    shouldUpdatePosition: PropTypes.bool,
+
+    /* Disable fade in animation */
```
shlomitc: Add another `*` for it to be displayed in the API tab of the component
`/** Disable fade in animation */`

jonathanadler:
Thanks! Updated 👍

shlomitc:
Failed for linting
```
  25:29  error  Unexpected trailing comma  comma-dangle
```

jonathanadler:
Fixed linting.
Seems to be fine now.

shlomitc:
looks good. My only UX-wise question is whether `showDelay` should be applied in such case.
It does make sense to be able to control it, however when you don't want animation, why not always get an instant feedback?

shlomitc:
@jonathanadler - consulting with @wuwa, the behavior should be:
if disable animation -> do not use showDelay

jonathanadler:
ok, but I have to say that as an outside user, it's not that intuitive.
If I was a developer (and even more a developer outside of wix) using this prop, I wouldn't have expected the `showDelay` to not work.
If so, maybe it's better to go back to the `showDelay=0` solution.

shlomitc:
if `showDelay` would have been called `animationShowDelay`, that would make more sense that `disableAnimations` also disables the delay.
So I think it's ok. maybe we can add a short explanation to this prop, just to be on the safe side

jonathanadler:
@shlomitc 
Sorry, i'm not sure I understand.
I'm just saying that the way I see it, the delay and the animations independent from each other.

jonathanadler:
@shlomitc 
Updated 👍

PR #1719, opened 2018-05-07
Always call props.onSelect when selecting an option in ButtonWithOptions

mennyWix:
### What changed
Behavior of preventing selection of the same option in ButtonWithOption is removed to support backwards compability
...

### Why it changed

...

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


src/ButtonWithOptions/ButtonWithOptions.js
```diff
@@ -116,12 +116,9 @@ class ButtonWithOptions extends WixComponent {
...
   showOptions = () => this.setState({showOptions: true});
 
-  onSelect = (option, sameOptionSelected) => {
+  onSelect = option => {
     this.hideOptions();
-
-    if (!sameOptionSelected) {
-      this.props.onSelect(option);
-    }
+    this.props.onSelect(option);
```
shlomitc: 1. pass the `sameOptionSelected` as a prop to the `this.props.onSelect()` function.
2. add this to the [docs](https://github.com/wix/wix-style-react/blob/master/src/ButtonWithOptions/README.md)
3. fix failing tests

lbelinsk:
@mennyWix in case we do need to allow reselecting the selected option, this looks good.
But can you explain the use case where you need that?

mennyWix:
@lbelinsk In code embed, I'm using ButtonWithOptions as the main button for selecting to create a script. The use should be able to select 'custom' script from the dropdown and then select to create another 'custom'... This used to be working fine till now... Perhaps we should prevent from selecting same option twice depending on a new props

lbelinsk:
I think if this is something you need it is fine to add it.
We blocked this option just for optimization reasons.
Looks fine to me

mennyWix:
@lbelinsk I don't need the extra props, thought you might need it :)
If this looks OK, please merge and let me know when wix-style-react is released.
Thanks

PR #1720, opened 2018-05-07
update position after dom is updated

jonathanadler:
### What changed
Tooltip - add a timeout before updating position in componentDidUpdate
...

### Why it changed
Position wasn't calculated correctly because the position was changed before the content of the tooltip was changed 
...

---

- [ v ] Change is tested - with e2e
- [ x ] Change is documented - I don't think this is needed, no?
- [ v ] Build is green
- [ ? ] Change of UX is approved by [wuwa](https://github.com/wuwa) or [milkyfruit] (https://github.com/milkyfruit) - do we need to on this fix? 

Thanks you!


jonathanadler:
@argshook What do you say?
It's kind of a hacky solution, but it works.
The issue is that the `_updatePosition` happens before the tooltip's content is updated.
The timeout fixes this issue.

argshook:
@jonathanadler i think it's ok since sooner or later we're going to have a new tooltip from wix-ui-core, the current one is already full of hacks.

do you think an e2e is doable for such a thing? without a test we might miss this feature in new tooltip. in e2e we use eyes, maybe that would be enough, wdyt?

jonathanadler:
Yeah, e2e with eyes is a great idea.
I'll add that

jonathanadler:
@argshook 
Updated finally!

argshook:
awesome, thanks! I see you had to fight this a little, congrats on the win :)

PR #1733, opened 2018-05-14
Allow react elements in Loader's 'text' prop

yevhen-kutishchev:
### What changed

Added `PropTypes.element` type to Loader's `text` prop.

### Why it changed

We want to be able to translate Loader's text with **react-intl**'s `<FormattedMessage />` message component.

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


src/Loader/Loader.js
```diff
@@ -39,7 +39,7 @@ export default class Loader extends WixComponent {
     color: PropTypes.oneOf(['blue', 'white']),
 
     /** Text to be shown below the loader */
-    text: PropTypes.string
+    text: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
```
argshook: @yevhen-kutishchev please do `text: PropTypes.node`
also trailing comma is breaking build during linting

argshook: what about `PropTypes.node`? :)
maybe there's a reason you want to be explicit?

ipanasenko: Yes, just `node`. Thanks, @argshook

PR #1742, opened 2018-05-17
<Breadcrumbs/> - documentation fixes:

Faradey27:
### What changed

 - Add automated story for Breadcrumbs
 - Take the second story of Breadcrumbs as example page in the same indexing.
 - fixed typos

### Why it changed

 - Because of migration to auto stories(was decided in guild)

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green


shlomitc:
https://github.com/wix/wix-style-react/pull/1727/files
See this example how to create new baseline image for the eyes tests

Faradey27:
nice, thank you @shlomitc

src/Breadcrumbs/Breadcrumbs.e2e.js
```diff
@@ -25,7 +25,7 @@ describe('Breadcrumbs', () => {
         expect(driverWithLinks.breadcrumbContentAt(idx)).toBe(item)
       );
     });
-  });
+  }, {version: '1.1.2'});
```
hepiyellow: Where do you take this version number from ?
This thing is new to me. 
Is this version number, per test case?

argshook: summoning @shlomitc

shlomitc: Effectively what this version variable is doing is just adding some prefix to the test name.
So you can just write the change that was done. it will also help understanding why the new baseline was required.

argshook: @Faradey27 yes https://github.com/wix/eyes.it/blob/3db59f9ac9209c0737fef68372a6d07712136f95/index.js#L39

shlomitc: 👍

shlomitc:
0. The items prop is not very useful as it is. should either be in exampleProps.
image

1. It displays the number 1 `activeId` twice
image

2. this example can be removed as there is an RTL in the playground
image


@wuwa - can you please review it?

src/Breadcrumbs/Breadcrumbs.e2e.js
```diff
@@ -42,7 +42,7 @@ describe('Breadcrumbs', () => {
       driver.clickBreadcrumbAt(itemToSelect);
       expect(driver.getActiveItemId()).toBe(itemToSelect);
     });
-  }, {version: '1.1.2'});
+  }, {version: '<Breadcrumbs/> - documentation fixes'});
```
hepiyellow: How does that explain the snapshot difference?

hepiyellow: And what about the `version` in line 28?

wuwa:
what exactly needs review? it looks different on my end, and also a different issue - but most props need to be omitted (no size, no white bg, no item selection).

shlomitc:
@wuwa - this task was just moving the stories. Previously it was split into two files without. Changing the API is a different task

Faradey27:
@shlomitc I was lost in all the comments, this is how it looks in my story http://prntscr.com/jo0kx9

should i close pull request as not relevant?

hepiyellow:
@faradey27 you wrote: "Looks like to resolve [object, object] i require some changes from autodocs, so merge is blocked currently",
As SHlomi said, you can add:
```
exampleProps: {
  items: [
   {label: 'single item', value:[item1]},
   {label: 'multiple items', value:[item1, item2, item3]},
]
}
```
And it should look very pretty (-:

hepiyellow:
About (1) (duplicate first item), I think it is resolved in autodocs, if not, you can keep it that way, and it will be resolved later.

About your local storybook, see if you have the latest node_modules with latest wix-storybook-utils.
I use:
```
alias npmref="echo removing node_modules && rm -rf node_modules && echo done removing node_modules && rm -f package-lock.json && echo done removing package-lock.json && npm i"
```

PR #1746, opened 2018-05-18
RTE link without protocol leads to relative path

Faradey27:
### What changed

- added new prop that give ability to RTE to automatically transform relative links to absolute after user insert

### Why it changed

- our customers asked for this, because most of them enter links like "wix.com" and expect to go to wix.com, but with current behivier they will go to http://somesite.com/wix.com - which is wrong for them

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green

### DIFF
--- a/src/RichTextArea/README.md
+++ b/src/RichTextArea/README.md
@@ -5,7 +5,7 @@
 
 | propName | propType | defaultValue | isRequired | description |
 |----------|----------|--------------|------------|-------------|
-| absoluteLinks | bool | false | - | Is the rich text area automatically transform relative links to absolute after user insert |
+| absoluteLinks | bool | false | - | Is the rich text area automatically transforming relative links to absolute after user insert |
 | disabled | bool | false | - | Is the rich text area disabled |
 | error  | bool | false | - | Is input value erroneous |
 | errorMessage | string | - | - | The error message to display when hovering the error icon, if not given or empty there will be no tooltip |
--- a/src/RichTextArea/README.md
+++ b/src/RichTextArea/README.md
@@ -5,6 +5,7 @@
 
 | propName | propType | defaultValue | isRequired | description |
 |----------|----------|--------------|------------|-------------|
+| absoluteLinks | bool | false | - | Is the rich text area automatically transform relative links to absolute after user insert |
 | disabled | bool | false | - | Is the rich text area disabled |
 | error  | bool | false | - | Is input value erroneous |
 | errorMessage | string | - | - | The error message to display when hovering the error icon, if not given or empty there will be no tooltip |
--- a/src/RichTextArea/RichTextArea.js
+++ b/src/RichTextArea/RichTextArea.js
@@ -20,7 +20,7 @@ const defaultBlock = {
   key: 'defaultBlock'
 };
 
-const makeHrefAbsolute = href => href.indexOf('://') === -1 ? `//${href}` : href;
+const makeHrefAbsolute = href => /^(https?:)?\/\//.test(href) ? href : `//${href}`;
 
 class RichTextArea extends WixComponent {
   static propTypes = {
@@ -39,6 +39,7 @@ class RichTextArea extends WixComponent {
   }
 
   static defaultProps = {
+    absoluteLinks: false,
     errorMessage: '',
     value: '<p></p>'
   }
--- a/src/RichTextArea/RichTextArea.js
+++ b/src/RichTextArea/RichTextArea.js
@@ -20,8 +20,11 @@ const defaultBlock = {
   key: 'defaultBlock'
 };
 
+const makeHrefAbsolute = href => href.indexOf('://') === -1 ? `//${href}` : href;
+
 class RichTextArea extends WixComponent {
   static propTypes = {
+    absoluteLinks: PropTypes.bool,
     buttons: PropTypes.arrayOf(PropTypes.string), // TODO: use PropTypes.oneOf(),
     dataHook: PropTypes.string,
     disabled: PropTypes.bool,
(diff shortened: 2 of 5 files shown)


stories/RichTextArea/RichTextArea.story.js
```diff
@@ -12,6 +12,7 @@ export default {
   component: RichTextArea,
   componentPath: '../../src/RichTextArea',
   componentProps: setProps => ({
+    absoluteLinks: false,
```
argshook: perhaps set `defaultProps` instead of one in story? autodocs will catch default prop

src/RichTextArea/RichTextArea.js
```diff
@@ -20,8 +20,11 @@ const defaultBlock = {
   key: 'defaultBlock'
 };
 
+const makeHrefAbsolute = href => href.indexOf('://') === -1 ? `//${href}` : href;
```
argshook: i think you need to be a bit more careful, cause there could be a link like `https://www.google.com/search?q=http://` and you'd end up with `//https://www.google.com/search?q=http://` which is no longer valid

perhaps `/^(https?:)?\/\//.test(href)`

Faradey27: good catch, will try, thank you
@argshook

hepiyellow:
What about if the text needs to include both relative an absolute links?
I mean, how will the consumer of this component decide if to set `absoluteLinks` to true or false?

src/RichTextArea/README.md
```diff
@@ -5,6 +5,7 @@
 
 | propName | propType | defaultValue | isRequired | description |
 |----------|----------|--------------|------------|-------------|
+| absoluteLinks | bool | false | - | Is the rich text area automatically transform relative links to absolute after user insert |
```
hepiyellow: 1) Fix english: change "transform" to "transforming".
2) I think an example would be very useful for consumers.

src/RichTextArea/RichTextArea.js
```diff
@@ -20,8 +20,11 @@ const defaultBlock = {
   key: 'defaultBlock'
 };
 
+const makeHrefAbsolute = href => /^(https?:)?\/\//.test(href) ? href : `//${href}`;
```
hepiyellow: I would be really glad if we had a comment here describing what this line does: like "add `//` prefix if href has no protocol prefix"

Faradey27:
@hepiyellow 
1) From my vision and vision of my managers, absolute links - should be default behavior and current relative links behavior looks really strange.

2) the user need to decide, if he want to have absolute links or relative

If you think that this is wrong behavior then let just close pull request and related jiras(my team will just create some hack over wix-style-react or fork it)

hepiyellow:
I don't have concrete use-cases in my head, so I trust your judgment. You can go ahead with the absoluteLinks prop.

Faradey27:
@hepiyellow so can you approve pull request?

hepiyellow:
I thought that maybe we don't need `absoluteLinks` prop, and we should always convert to absolute.
On the other hand., @shlomitc  mentioned that he thinks this "auto-correct" should work only when typing e.g. "wix.com" in the text input area, while if the user open the addLink popover, then we should not manipulate what the use is entering.

Could you verify with @wuwa about the this?

If you ask me, I would leave the prop, but set it's default value to `true` (convert to absolute).

wuwa:
i would say the opposite - using the link input should convert to absolute, typing in textarea changes nothing (someone might want to write site.com as non-linked content).
also an idea - have the placeholder in the input be "http://"

shlomitc:
@wuwa I think there is a mixture here in terminology and functionality
When generating the link, the text and the link value are separate things.
When speaking about absolute paths, we talk about the value of the link.

when I enter wix.com as a link value:
absolute link: https://wix.com
relative link: http://my-site.com/wix.com, which is the issue to solve here.

@Faradey27 suggests that in the input dialog, it will auto-correct to absolute paths. it solves the bug but raises a question regarding the predictability of this behavior.
adding a placeholder is nice but is it clear enough? because it will disappear once typing

wuwa:
yes, i understood the case, but you wrote also about links in the textArea itself.
regarding the popover - i think placeholder+autocorrect is the way to go.

wuwa:
i think should say: www.site.com, maybe UX writers know better? and yes, should convert to absolute.

Faradey27:
@wuwa @shlomitc 
Can you please give me a contact of person with whom i can disscuss this change and merge it?
as i understood, latest thing that you want is to change placeholder from http://prntscr.com/jsetfh to "https://www.site.com"

Cashier has this bug for more then 5 months and a lot of users contact support because of this.
Let do a call maybe to disscuss problem more detail, because this issue open for too long time and we need to solve it or to tell to use another solution for people that has such problem

p.s. and reminder this pull request did not provide any breaking changes, it just give ability to use new prop to solve available problem

src/RichTextArea/RichTextArea.js
```diff
@@ -20,8 +20,16 @@ const defaultBlock = {
   key: 'defaultBlock'
 };
 
+/*
+  here we are checking is link relative(if it contain 'https' or http or '//')
+  and if it not relative, then we add '//' at the beginning of it,
+  to make link absolute(instead of relative)
+*/
+const makeHrefAbsolute = href => /^(https?:)?\/\//.test(href) ? href : `//${href}`;
```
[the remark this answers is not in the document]
Faradey27: done @hepiyellow

src/RichTextArea/RichTextArea.js
```diff
@@ -20,8 +20,16 @@ const defaultBlock = {
   key: 'defaultBlock'
 };
 
+/*
+  here we are checking is link relative(if it contain 'https' or http or '//')
```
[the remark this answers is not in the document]
Faradey27: done, @hepiyellow

argshook:
wow this is a long discussion over a rather minor change.
i say let's go ahead and merge.

`absoluteLink` prop may not be necessary,  there's an `if` anyway. But since it makes this behavior opt-in, we're on safe side and so i don't see why this shouldn't be merged

@hepiyellow if you don't have any other changes in mind, let's just merge.

shlomitc:
mergine it. we can treat it as a feature flag for now. if we find it the right behavior, let's add it as a default behavior

PR #1748, opened 2018-05-21
<DateInput/> - should focus on the selected year/month when opening dropdown

Faradey27:
### What changed

- added autoscroll to selected option functionality

### Why it changed

- to improve user expiarence

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green

### DIFF
--- a/src/DatePicker/DatePickerDropdown/index.js
+++ b/src/DatePicker/DatePickerDropdown/index.js
@@ -71,6 +71,7 @@ export default class DropdownPicker extends React.Component {
           <div className={styles.dropdown}>
             <DropdownLayout
               dataHook={`${dataHook}-menu`}
+              focusOnSelectedOption
               visible={isOpen}
               options={options}
               onSelect={this.onSelect}
--- a/src/DropdownLayout/DropdownLayout.js
+++ b/src/DropdownLayout/DropdownLayout.js
@@ -32,6 +32,25 @@ class DropdownLayout extends WixComponent {
     this.onClickOutside = this.onClickOutside.bind(this);
   }
 
+  componentDidMount() {
+    super.componentDidMount();
+    if (this.props.focusOnSelectedOption) {
+      this.focusOnSelectedOption();
+    }
+  }
+
+  focusOnSelectedOption() {
+    if (this.selectedOption) {
+      this.options.scrollTop = Math.max(this.selectedOption.offsetTop - this.selectedOption.offsetHeight, 0);
+    }
+  }
+
+  setSelectedOptionNode(optionNode, option) {
+    if (option.id === this.state.selectedId) {
+      this.selectedOption = optionNode;
+    }
+  }
+
   isLegalOption(option) {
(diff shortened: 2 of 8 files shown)


src/DropdownLayout/DropdownLayout.js
```diff
@@ -32,6 +32,25 @@ class DropdownLayout extends WixComponent {
     this.onClickOutside = this.onClickOutside.bind(this);
   }
 
+  componentDidMount() {
+    super.componentDidMount();
+    if (this.props.focusOnSelectedOption) {
```
argshook: @Faradey27 i think it should work like this by default, no prop needed. could there be a case where consumer would wish to disable it?

src/DatePicker/DatePickerDropdown/index.js
```diff
@@ -71,6 +71,7 @@ export default class DropdownPicker extends React.Component {
           <div className={styles.dropdown}>
             <DropdownLayout
               dataHook={`${dataHook}-menu`}
+              focusOnSelectedOption
```
argshook: if it's default, then no prop here, then it just works

src/DropdownLayout/DropdownLayout.spec.js
```diff
@@ -30,6 +30,18 @@ describe('DropdownLayout', () => {
...
+  it('should focus on selected option', () => {
+    const driver = createDriver(
+      <DropdownLayout
+        focusOnSelectedOption
+        visible
+        options={options}
+        selectedId={3}
+        />
+    );
+    expect(driver.optionsScrollTop()).toBe(0);
```
argshook: @Faradey27 shouldn't you be checking for something other than 0?

hepiyellow:
Please mention if there is a related issue.

src/DropdownLayout/DropdownLayout.js
Faradey27: @argshook i am usually scary to change logic in code that widely used, so my way is before doing to do this as option and enable by default just for small part(like DatePicker) see that everything is fine, no bugs, UX is nice and just then change default behevier

src/DropdownLayout/DropdownLayout.spec.js
Faradey27: @argshook in general yes, but on practice looks like scrolls on jsdom do not work(real css do not applied, offsetTop and scrollTop always 0, etc.) and to do more interesting test it require a lot of tricks(and it will be bad test with to much emulation), i was thinking about e2e test, but i am worry about flakines(with scrolls i usually have 80-90% success rate for e2e tests, which may be bad for big lib)

what do you think about this?

argshook:
@hepiyellow there is, it's already mentioned https://github.com/wix-private/wsr-issues/issues/110

src/DropdownLayout/DropdownLayout.spec.js
Faradey27: why no, will try to do this

PR #1753, opened 2018-05-22
DataTable design update

liorur:
### What changed
some design and layout parameters

### Why it changed
new spec
---

- [N/A] Change is tested
- [N/A] Change is documented
- [ V] Build is green


src/DataTable/DataTable.js
```diff
@@ -333,6 +335,10 @@ DataTable.propTypes = {
   loader: PropTypes.node,
   useWindow: PropTypes.bool,
   scrollElement: PropTypes.object,
+  tdVerticalPadding: PropTypes.oneOf([
```
hepiyellow: check autodocs, wether it shows these options (if not, write manual docs)

src/DataTable/DataTable.scss
```diff
@@ -19,8 +19,8 @@
 .table td {
   font-size: 14px;
```
hepiyellow: make like Typography T1 appearance, default for TD

src/DataTable/DataTable.scss
```diff
@@ -66,7 +65,7 @@
   td {
     border-bottom: none medium;
```
hepiyellow: Seems like border was replaced with a box-shadow, so is this 'border-bottom' here needs an update also?

src/DataTable/README.md
```diff
@@ -25,11 +25,12 @@
...
-| thPadding | string | '5px' | - | Table headers padding |
-| thHeight | string | '36px' | - | Table headers height |
-| thFontSize | string | '12px' | - | Table headers font size |
-| thBorder | string | '1px' | - | Table headers border |
-| thColor | string | '#123456' | - | Table headers color |
+| tdVerticalPadding | string | 'small' | - | Table cell vertical padding. should be 'small' or 'large' |
+| thPadding | string | '12px' | - | Table headers padding |
+| thHeight | string | '18px' | - | Table headers height |
+| thFontSize | string | - | - | Table headers font size |
+| thBorder | string | - | - | Table headers border |
```
hepiyellow: Its kind of weird that we replaced the border with box-shadow, but we keep this thBorder prop. This PR is breaking anyways, so I think we should remove the thBorder props. Anyways, it would probably not look well with ALSO a box-shadow, so the box-shadow needs to be overwritten somehow.

Let's remove it. @shlomitc ok ?

You can start a MIGRATION.ms file that collects all the breaking changes.

BTW: why don't understand why we have all these th-props. we should simply have a `headerClass` props.

src/DataTable/DataTable.scss
```diff
@@ -45,33 +43,33 @@
...
   transform: rotate(180deg);
 }
 
 .row-details:not(:last-child) {
-  border-bottom: 1px solid $D70;
+  box-shadow: inset 0 -1px 0 0 $D70;
 
   .row-details-inner {
-    border-top: 1px solid $D70;
+    box-shadow: inset 0 -1px 0 0 $D70;
```
hepiyellow: I think this should be `box-shadow: inset 0 1px 0 0 $D70;`
v-offset `1` not `-1`

src/DataTable/DataTable.scss
```diff
@@ -45,33 +43,33 @@
...
+    box-shadow: none;
   }
-  
+
   cursor: pointer;
 
   &:nth-last-child(2) {
     td {
-      border-bottom: 1px solid $D70;
+      border-top: 1px solid $D70;
```
hepiyellow: Is this staying a border and not a shadow, on purpose?

src/DataTable/DataTable.js
```diff
@@ -298,9 +302,9 @@ DataTable.defaultProps = {
   loader: <div className="loader">Loading ...</div>,
   scrollElement: null,
   useWindow: true,
-  thPadding: '5px',
-  thHeight: '36px',
-  thFontSize: '12px'
+  thPadding: '12px',
```
shlomitc: all thPadding and similar should remain as they were but add a deprecation message.
something like this

```
/** this prop is deprecated and should not be used
* @deprecated
*/
```

src/DataTable/DataTable.js
```diff
@@ -184,12 +185,13 @@ class DataTable extends WixComponent {
...
-  renderCell = (rowData, column, rowNum, colNum) => {
-    const classes = classNames({[s.important]: column.important});
+  renderCell = (rowData, column, rowNum, colNum, tdVerticalPadding) => {
+    const classes = classNames(typography.t1, {[s.important]: column.important});
     const width = rowNum === 0 && this.props.hideHeader ? column.width : undefined;
+    const verticalPadding = tdVerticalPadding === 'large' ? '24x' : '18px';
 
     return (<td
-      style={column.style}
+      style={{...column.style, paddingTop: verticalPadding, paddingBottom: verticalPadding}}
```
shlomitc: the inline-styles should be moved to css classes.

src/DataTable/README.md
shlomitc: let's maintain backward compatibility and add `@deprecated` for the relevant props

src/DataTable/DataTable.js
```diff
@@ -185,7 +190,11 @@ class DataTable extends WixComponent {
   };
 
   renderCell = (rowData, column, rowNum, colNum) => {
-    const classes = classNames({[s.important]: column.important});
+    const classes = classNames(
+      typography.t1,
```
hepiyellow: Shouldn't this be conditional by `this.props.newDesign`? (Like you do for the Header)

src/DataTable/DataTable.old.scss
```diff
@@ -0,0 +1,80 @@
...
+  background-color: white;
+}
+
+.table th {
+  background-color: rgba(240, 243, 247, 0.5);
+  text-align: left;
+  text-align: start;  // not supported in IE11 & Edge; 'left' as fallback
+  color: $paletteD3;
+  border-top: solid 1px $D60;
+  padding: 5px;
```
hepiyellow: These `padding` , `height`, `font-size` are not in master branch now. Are they here by mistake? meaning , do they belong only to the newDesign? or is it a fix to the old design?

src/DataTable/DataTable.scss
```diff
@@ -2,37 +2,37 @@
...
-  font-family:HelveticaNeueW01-55Roma,HelveticaNeueW02-55Roma,HelveticaNeueW10-55Roma,sans-serif;
+  font-family: HelveticaNeueW01-55Roma, HelveticaNeueW02-55Roma, HelveticaNeueW10-55Roma, sans-serif;
   table-layout: fixed;
   background-color: white;
 }
 
 .table th {
-  background-color: rgba(240, 243, 247, 0.5);
+  background-color: $B40;
+  padding: 12px 0 12px 24px;
```
hepiyellow: Why `padding-right=0`?
Why `padding-left=24`?
Why not all 12px like before?
Remember that the selection column is optional.

src/DataTable/DataTable.scss
```diff
@@ -9,14 +9,16 @@
...
+  height: 18px;
   text-align: left;
   text-align: start; // not supported in IE11 & Edge; 'left' as fallback
   font-weight: normal;
   box-shadow: inset 0 1px 0 0 $B30, inset 0 -1px 0 0 $B30;
 }
 
 .table td {
-  padding: 18px 12px;
+  padding:  18px 0 18px 24px;
```
hepiyellow: Again, why the 0, 24...?

src/DataTable/DataTable.scss
```diff
@@ -73,3 +75,13 @@
     }
   }
 }
+
+.largeVerticalPadding {
```
hepiyellow: I would put these `. largeVerticalPadding` and `. mediumVerticalPadding` inside the `.table td` selector, like this:
```css
.table td {
  padding:  18px 0 18px 24px;
  font-weight: normal;
  box-shadow: inset 0 -1px 0 0 $D60;

  &.largeVerticalPadding {
    padding-top: 24px;
    padding-bottom: 24px;
  }

  &.mediumVerticalPadding {
    padding-top: 18px;
    padding-bottom: 18px;
  }  
}
```

src/DataTable/DataTable.scss
```diff
@@ -15,13 +15,27 @@
   text-align: start; // not supported in IE11 & Edge; 'left' as fallback
   font-weight: normal;
   box-shadow: inset 0 1px 0 0 $B30, inset 0 -1px 0 0 $B30;
+
+  &:first-child{
```
hepiyellow: Thats nice, but I don't see how it solves the gutter issue.
The second column from the last (In the zeplin spec, it is the "Inventory" column), doesn't have a right padding, and neither the last column has a left padding, so there is no gutter.

hepiyellow:
@milkyfruit please review:

It should include the header (title row) and rows. Changes in header color, padding (gutter), changing the border between rows and in the header to use box-shadow, fonts, etc....
All of it should conform to the recent WSR spec for the Table.
This of course does not include yet the Header part (the one with the search boxes, and the bulk-action-header. And does not include actions column nor any row content restriction.

https://wix-wix-style-react-pr-1753.surge.sh/?selectedKind=10.%20Tables&selectedStory=10.1%20DataTable&full=0&addons=0&stories=1&panelRight=0

src/DataTable/DataTable.js
hepiyellow: What do you mean? The column.style is an open API for any customization. Do you want to remove it?

src/DataTable/DataTable.js
hepiyellow: done

src/DataTable/README.md
hepiyellow: ok

milkyfruit:
SortBy Arrow color needs to be updated to B10;
SortBy SVG needs to be changed to the new ones:

https://github.com/wix/wix-ui/blob/master/packages/wix-ui-icons-common/src/system/raw/SortByArrowDown.svg

https://github.com/wix/wix-ui/blob/master/packages/wix-ui-icons-common/src/system/raw/SortByArrowUp.svg

src/DataTable/DataTable.js
```diff
@@ -251,8 +252,8 @@ class TableHeader extends Component {
     if (sortDescending === undefined) {
       return null;
     }
-    const sortDirectionClassName = sortDescending ? this.style.sortArrowAsc : this.style.sortArrowDesc;
-    return <span data-hook={`${colNum}_title`} className={sortDirectionClassName}><ArrowVertical/></span>;
+    const Arrow = sortDescending ? SortByArrowUp : SortByArrowDown;
+    return <span data-hook={`${colNum}_title`} className={this.style.sortArrow}><Arrow/></span>;
```
hepiyellow: I think this is broken when newDesign=false. There is no `sortArrow` class in `DataTable.old.scss`.

Optional: It would be great if you add an example in the storybook, with newDesign=false for comparison.

hepiyellow:
Replaced by https://github.com/wix/wix-style-react/pull/1817

PR #1754, opened 2018-05-22
New TooltipContainerStrategy: appendByPredicate

illBeRoy:
### What changed

Added a new tooltip strategy: now you can pass a predicate to a prop named `appendByPredicate`, which receives an element and returns a boolean value.
The tooltip will be attached to the closest ancestor which matches the predicate.

Example:

```javascript
<div className="container">
  <div className="panel">
    <Tooltip content={...} appendByPredicate={el => el.className === 'container'}>
      <Icons.QuestionMark />
    </Tooltip>
  </div>
</div>
```

### Why it changed

The whole matter was discussed with Shlomi. If you want the long version, here it is 😅:

Real life story: I have the following element within a scrollable list. The element has a hover state and at the bottom: an icon which opens a tooltip with a clickable link:

example

Default behaviour is no good, as two things happen:
1. The tooltip scrolls along with the page
2. The element loses focus

Appending to parent is no good, as the direct parent is only a part of the element and is covered by other parts of that element, which cover the tooltip as well if we append it to parent.

Just using `append` introduces new complexity: it forces me to add state, render the component "temporarily" until the ref to the correct ancestor is available, and change my tests to await the component to truly be ready (as now mounting it is not enough, I have to wait for ref).

With `appendByPredicate`, I can easily let the tooltip attach itself to the nearest matching ancestor by just passing a simple predicate.

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green

### DIFF
--- a/src/Tooltip/TooltipContainerStrategy.js
+++ b/src/Tooltip/TooltipContainerStrategy.js
@@ -4,16 +4,19 @@ const predicates = [
   element => element === document.body
 ];
 
+const neverMatchAnyElement = () => false;
+
 export class TooltipContainerStrategy {
-  constructor(appendTo, appendToParent) {
+  constructor(appendTo, appendToParent, appendByPredicate) {
     this._ancestor = null;
     this._appendTo = appendTo;
     this._appendToParent = appendToParent;
+    this._appendByPredicate = appendByPredicate || neverMatchAnyElement;
   }
 
   _findAncestor(element) {
     while (element) {
-      if (predicates.some(pred => pred(element))) { // eslint-disable-line
+      if ([...predicates, this._appendByPredicate].some(pred => pred(element))) { // eslint-disable-line
         break;
       }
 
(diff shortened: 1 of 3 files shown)


src/Tooltip/TooltipContainerStrategy.js
```diff
@@ -4,16 +4,19 @@ const predicates = [
...
 
+const neverMatchAnyElement = () => false;
+
 export class TooltipContainerStrategy {
-  constructor(appendTo, appendToParent) {
+  constructor(appendTo, appendToParent, appendByPredicate) {
     this._ancestor = null;
     this._appendTo = appendTo;
     this._appendToParent = appendToParent;
+    this._appendByPredicate = appendByPredicate || neverMatchAnyElement;
```
shlomitc: It would be better to move the `predicates` array into the class and alter it instead of holding it as a different variable

src/Tooltip/TooltipContainerStrategy.js
```diff
@@ -4,16 +4,19 @@ const predicates = [
...
   }
 
   _findAncestor(element) {
     while (element) {
-      if (predicates.some(pred => pred(element))) { // eslint-disable-line
+      if ([...predicates, this._appendByPredicate].some(pred => pred(element))) { // eslint-disable-line
```
shlomitc: see comment above

PR #1764, opened 2018-05-28
<Search /> - keep the focus after pressing clear button

dotansimha:
### What changed

Change `<Search />` behaviour: after pressing clear button the input should remain focused.

### Why it changed

https://github.com/wix-private/wsr-issues/issues/127 

---

- [x] Change is tested
- [x] Change is documented
- [ ] Build is green


src/Search/Search.js
```diff
@@ -59,16 +59,6 @@ export default class Search extends WixComponent {
...
-    } = this.props;
-
-    this.refs.searchInput.input.blur();
-
-    onClear && onClear();
-  };
-
   render() {
     return (
       <InputWithOptions
```
hepiyellow: Remove also `ref="searchInput"`, since it is not needed anymore

dotansimha: Done! And also rebased to master.

PR #1765, opened 2018-05-28
<Input /> - should not change background color if hovered while focused

dotansimha:
### What changed

`<Input />` component should not change it's background color while hovered and focused. 

### Why it changed

https://github.com/wix-private/wsr-issues/issues/127

---

- [x] Change is tested
- [x] Change is documented
- [ ] Build is green


src/Input/InputMixins.scss
```diff
@@ -40,7 +40,7 @@
 }
 
 @mixin ThemeHover($backgroundColor) {
-    &:hover    { background-color: $backgroundColor; }
+    &:hover:not(.hasFocus)    { background-color: $backgroundColor; }
     &.hasHover { background-color: $backgroundColor; }
```
hepiyellow: Are you sure we don;t need this also with `hasHover`?

dotansimha: I'm not sure about the behaviour in this case. @shlomitc what do you think?

shlomitc: 1. When are we using `hasHover`? in the focusable HOC? if so, then let's add it.
2. `forceHover` is used for the examples only, didn't see it anywhere else in our consumers' code:
image
https://wix-wix-style-react.surge.sh/?selectedKind=3.%20Inputs&selectedStory=3.1%20%2B%20Input&full=0&addons=0&stories=1&panelRight=0

@dotansimha - do you think there is a way to simulate the hover state outside of the component without passing this prop? it will probably be an even cleaner solution.

dotansimha: @shlomitc If we are using it only for the examples page and it's relevant only there, maybe the example page can set the styling using the parent element (`div.hoverExample > div { background-color: '...' }` instead of adding it to the component itself.
We can also trigger `mouseenter mouseleave` event on the demo page.

shlomitc: @donatas - this css approce is overriding the component styles and not simulating the state, so it doesn't sound like a good solution.

maybe triggering the event's can be useful, want to try and see it it's a clean solution?

hepiyellow: We can also remove the example with the forceHover.

hepiyellow:
Components which should be breaking in eyes but are not:
 - TimePicker : missing test
 - MultiSelect (Tags): ??

src/Input/InputMixins.scss
dotansimha: @shlomitc @hepiyellow remove the example or try to adapt to the changes?

dotansimha:
@hepiyellow why it should break? it uses `hasFocus` and I didn't remove it. 
@shlomitc it seems like we are using `hasFocus` in `InputWithTags` so we can't remove it...

shlomitc:
@dotansimha - are you sure it's the hasFocus of the input? I think it's a different class as part of the multiSelect component
https://github.com/wix/wix-style-react/blob/master/src/MultiSelect/InputWithTags.scss#L54

dotansimha:
@shlomitc yeah you are right. so do you want me to remove the example?

shlomitc:
@dotansimha - let's not touch the example if it still works correctly

hepiyellow:
Let me try and summarize:

1. InputArea - still has hover when focused. Please find out why and fix it.
2. Tags (MultiSelect) 
  2.1 Tags Should not have hover when focused and empty
  2.2 InputWithTags - something you did made it so that it has hover style when selected tags is not empty. Needs a fix.
  2.3 Missing e2e snapshots for : a) hover b) hover+focus c) hover+non_empty d) hover+focus+non_empty
3. Missing e2e test with hover+focus snapshot: TimeInput
4. Input - Guess what? Input dies not have an e2e test. (Also TextField doesn't). Please add e2e with these snapshots: a) hover b) hover+focus

------
Minor: Regarding the Input's `forceHover` prop. I personally don't see any use-case for it, and I think it should be removed (along with the example). But I think it does not affect this PR, it is a separate issue. We can also forget about it for now.
-----

BTW:  I found a hover bug: see this issue https://github.com/wix/wix-style-react/issues/1786

dotansimha:
@shlomitc I fixed the `<InputArea />`, it has the same fix as `<Input />`. 
Storybook examples works fine :)

hepiyellow:
Example for hover test:
https://github.com/wix/wix-style-react/blob/310e5a28f7b573d6a1eeb126a5406261d2aef448/src/IconWithOptions/IconWithOptions.e2e.js#L14

dotansimha:
@shlomitc fixed the tags issue :)

PR #1773, opened 2018-05-30
<Input/> - On text click - select all

Faradey27:
### What changed

- added select on click on inpit

### Why it changed

- because of wsr-issue and to improve user experience 

---

- [x] Change is tested
- [x] Change is documented
- [ ] Build is green

### DIFF
--- a/src/Input/Input.js
+++ b/src/Input/Input.js
@@ -239,6 +239,7 @@ class Input extends Component {
 Input.displayName = 'Input';
 
 Input.defaultProps = {
+  autoSelect: true,
   size: 'normal',
   theme: 'normal',
   errorMessage: '',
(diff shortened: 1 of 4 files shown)


Faradey27:
@hepiyellow will create e2e test

hepiyellow:
@shlomitc This feature is missing in `wix-ui-core`, can we ask @Faradey27  to implement it there as well?
If so, then in the core, I would put this behaviour under a prop , `selectOnClickEnabled`.
We might want to have it as a prop also in WSR.

src/Input/Input.js
```diff
@@ -192,6 +192,9 @@ class Input extends Component {
   };
 
   _onClick = e => {
+    if (e.target.value && e.target.select) {
```
hepiyellow: Should this select all only when the click is also focusing the component?

shlomitc: makes sense that also on focus

hepiyellow:
Only on focus.

Faradey27:
@hepiyellow @shlomitc i m closing pull request, because we already have such feature(auto select on focus) https://github.com/Faradey27/wix-style-react/blob/aa5e9fe4ce92f995f919997a8873e5a5972e225d/src/Input/Input.js#L183

shlomitc:
We still need to make this behavior default

Faradey27:
@shlomitc @hepiyellow updated with default props

PR #1775, opened 2018-05-30
<Search /> - Added expandable

dotansimha:
### What changed

Added `expandable` behaviour to `<Search />` component.

Also, implemented a tests driver for `<Search />` that exposes test utils for the new feature.

### Why it changed

https://github.com/wix-private/wsr-issues/issues/105

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green


src/Search/Search.driver.js
```diff
@@ -1 +1,25 @@
...
+const searchDriverFactory = args => {
+  const inputDriver = inputDriverFactory({
+    ...args,
+    element: args.element ? args.element.childNodes[0] : null
+  });
+
+  const {element} = args;
+
+  return {
+    ...inputDriver,
```
lbelinsk: Why not to expand ...inputWithOptionsDriver?
clickToExtend is redundant, better to use driver.inputDriver.click()

The only one I will keep is isCollapsed

See for reference `Autocomplete`, `Dropdown` or `MultiSelect`

src/Search/Search.js
```diff
@@ -15,19 +17,22 @@ export default class Search extends WixComponent {
...
+    expandable: false
   };
 
   constructor(props) {
     super(props);
 
     this.state = {
-      inputValue: (!this._isControlled && props.defaultValue) || ''
+      inputValue: (!this._isControlled && props.defaultValue) || '',
+      collapsed: true
```
lbelinsk: are you sure it shouldn't be configurable?
Maybe worth consulting with @wuwa if there is a scenario where a user would want to start with a non-collapsed state.

There is also a possible bug, what if the Input has an initial value? in that case collapsed can't be set to true

src/Search/Search.js
```diff
@@ -66,25 +71,71 @@ export default class Search extends WixComponent {
...
+    }
+  };
+
   render() {
+    const wrapperClasses = classNames({
+      [styles.expandableStyles]: this.props.expandable,
+      [styles.collapsed]: this.state.collapsed && this.props.expandable,
+      [styles.expanded]: !this.state.collapsed && this.props.expandable
+    });
+
```
lbelinsk: Just a suggestion for a small refactor, in order not to use `omit` here.
```
const {dataHook, ...otherProps} = this.props;

return (
   <div data-hook={dataHook} ...>
       <InputWithOptions {...otherProps} ref=...
)
```

src/Search/Search.scss
```diff
@@ -1,12 +1,53 @@
...
-        height: 22px;
+        width: $iconSize;
+        height: $iconSize;
+    }
+}
+
+.expandableStyles {
+  display: inline-block;
+
+  > div {
```
lbelinsk: This doesn't feel too robust. You can place an internal data-hook on the relevant element in order to target it

src/Search/Search.scss
```diff
@@ -1,12 +1,53 @@
...
+        input {
+          min-width: 0 !important;
+        }
+      }
+    }
+  }
+}
+
+.expanded {
+  > div {
```
lbelinsk: same here

src/Search/Search.spec.js
```diff
@@ -194,6 +195,61 @@ describe('Search', () => {
...
     });
   });
+
+  describe('Expandable', () => {
+    it('should start as collapsed element by default when expndable=true', () => {
+      const driver = createDriver(
+        <Search options={options} expandable/>
+      );
+
+      expect(driver.isExpandable()).toBeTruthy();
```
lbelinsk: I feel like isExpandable is too unit test.  isCollapsed is covering it.

I am missing 2 tests:
`it('should  not be collapsed by default', () => {...})' 
here you check that the input is not collapsed, even after it was clicked (before and after) since you didn't provide the `expandanle prop`. 

`it('should have non-collapsed input when expandaple=true' and the input has initial value)`

src/Search/Search.spec.js
```diff
@@ -194,6 +195,61 @@ describe('Search', () => {
...
+      const driver = createDriver(
+        <Search options={options} expandable/>
+      );
+
+      expect(driver.inputDriver.isFocus()).toBeFalsy();
+      driver.clickToExtend();
+      expect(driver.inputDriver.isFocus()).toBeTruthy();
+    });
+
+    it('should not collapse the input if the input has not value and blurred', () => {
```
lbelinsk: has no value

src/Search/Search.spec.js
```diff
@@ -194,6 +195,61 @@ describe('Search', () => {
...
+      );
+
+      driver.clickToExtend();
+      driver.inputDriver.enterText('wix');
+      driver.inputDriver.blur();
+      expect(driver.isCollapsed()).toBeFalsy();
+      expect(driver.isExtended()).toBeTruthy();
+    });
+
+    it('should collapse the input if the input has not value and blurred', () => {
```
lbelinsk: has no value

src/Search/Search.spec.js
```diff
@@ -194,6 +195,61 @@ describe('Search', () => {
...
+
+    it('should not collapse the input if the input has not value and blurred', () => {
+      const driver = createDriver(
+        <Search options={options} expandable/>
+      );
+
+      driver.clickToExtend();
+      driver.inputDriver.enterText('wix');
+      driver.inputDriver.blur();
+      expect(driver.isCollapsed()).toBeFalsy();
```
lbelinsk: isCollapsed and isCollapsed are always opposite. I would remove one of them from the driver

src/Search/Search.js
```diff
@@ -66,25 +71,71 @@ export default class Search extends WixComponent {
...
-        menuArrow={false}
-        clearButton
-        closeOnSelect
-        showOptionsIfEmptyInput={false}
-        options={this._filteredOptions}
-        onClear={this._onClear}
-        onChange={this._onChange}
-        highlight
-        />
+      <div data-hook={this.props.dataHook} className={wrapperClasses} onClick={this._onWrapperClick}>
```
shlomitc: the `data-hook` is applied automatically on top of the component by extending `WixComponent`, so you don't need to pass it in manually.

src/Search/Search.scss
```diff
@@ -1,12 +1,53 @@
...
         display: block;
-        width: 22px;
-        height: 22px;
+        width: $iconSize;
+        height: $iconSize;
+    }
+}
+
+.expandableStyles {
+  display: inline-block;
```
shlomitc: this rule makes the input not to spread to 100%.
image

src/Search/Search.scss
```diff
@@ -1,12 +1,53 @@
...
+        transition: all .3s;
+      }
+    }
+  }
+}
+
+.collapsed {
+  cursor: pointer;
+
+  > div {
```
shlomitc: I guess there is no other way of customizing the Input other than overriding its styles like that.
@lbelinsk - do you think about a better approach than that?

src/Search/Search.scss
shlomitc: `data-hook` is used for testing, not for targeting css.

src/Search/Search.js
shlomitc: if it is configurable, then it means it should open focused, which is kind of weird in my opinion

lbelinsk: We have autoFocus prop, just need to make sure it is not colliding

src/Search/Search.scss
lbelinsk: sorry, classname

src/Search/Search.scss
lbelinsk: If it is targeting an internal element (not accessible directly within search component) then I gues not :(

src/Search/Search.scss
shlomitc: not sure it will work, because css-modules will create a different hash...

milkyfruit:
Nice! The animation looks great.
I didn't think about the auto collapse on focus-out, but I think it works great.

Few things I noticed:

The collapsed search button is not proportional, must be 36x36:
image

Mouse click when is open and click on the icon – triggers the animation, while it shouldn't:
clicking always triggers

Hover effect appears on active state, it should appear only on default state (applies to all inputs and selections):
image

Padding from left to search text should be 36px (including the 1px border):
image

The icon should shift to the right a bit as well:
image

src/Search/Search.driver.js
dotansimha: It does ;) just the variable name was incorrect. fixed :)

src/Search/Search.js
dotansimha: great idea!

src/Search/Search.scss
dotansimha: Yeah, css-modules mess up with the classname, and if it's not constant I can't use it in the CSS file. 
I can also add `data-` to the input parent and this way it will be easier to access it.

src/Search/Search.spec.js
dotansimha: thanks, fixed

src/Search/Search.spec.js
dotansimha: thanks, fixed

src/Search/Search.spec.js
dotansimha: Yeah the reason I had them both is because it toggles between two CSS classes, and I wanted to makes sure that one is removed and the other one is added. 
But yeah, you are right, I changed it ;) thanks

src/Search/Search.js
dotansimha: Got it, thanks

src/Search/Search.scss
dotansimha: fixed

src/Search/Search.scss
dotansimha: I can add `data-` attribute in `<Input />` so it will be easier to target it... what do you think?

src/Search/Search.spec.js
dotansimha: added. thanks

src/Search/Search.scss
shlomitc: Ok. ‘data-class’ maybe

dotansimha:
> The collapsed search button is not proportional, must be 36x36:
Fixed :)

> Hover effect appears on active state, it should appear only on default state (applies to all inputs and selections)
It's another issue, fixed it here: https://github.com/wix/wix-style-react/pull/1765 

> Padding from left to search text should be 36px (including the 1px border):
Fixed :)

> The icon should shift to the right a bit as well:
Fixed :)

> Mouse click when is open and click on the icon – triggers the animation, while it shouldn't:
Fixed :)

src/Search/Search.scss
dotansimha: done :)

src/Search/Search.scss
```diff
@@ -1,12 +1,47 @@
...
+    display: block;
+    width: $iconSize;
+    height: $iconSize;
+  }
+}
+
+.expandableStyles {
+  > div {
+    > div {
+      > div {
```
hepiyellow: Can't you put a `data-xxx` attribute and a selector to find it?
All these nested div is not readable.
We should try to make our future work of migrating to Stylable more easy, buy using pseudo class selectors as much as possible, instead of depending on DOM structure.

src/Search/Search.js
```diff
@@ -66,25 +72,89 @@ export default class Search extends WixComponent {
...
+      if (value === '') {
+        e.preventDefault();
+      }
+    }
+  };
+
   render() {
+    const wrapperClasses = classNames({
+      [styles.expandableStyles]: this.props.expandable,
+      [styles.collapsed]: this.state.collapsed && this.props.expandable,
```
hepiyellow: Why not simply use one class `collapsed`, and use in the css `not(.collapsed)`? (Or the other way around, using `expanded`)

src/Search/Search.js
```diff
@@ -15,19 +16,24 @@ export default class Search extends WixComponent {
...
     ...InputWithOptions.defaultProps,
-    placeholder: 'Search'
+    placeholder: 'Search',
+    expandable: false
   };
 
   constructor(props) {
     super(props);
 
+    const initalValue = (!this._isControlled && props.defaultValue) || '';
```
hepiyellow: typo - `initialValue`

src/Search/Search.js
```diff
@@ -66,25 +72,89 @@ export default class Search extends WixComponent {
 
     this.refs.searchInput.input.blur();
```
hepiyellow: Remove this when merging master, since it was removed in
https://github.com/wix/wix-style-react/pull/1764

src/Search/Search.js
dotansimha: There are other CSS rules that I need to apply, because it's not just true/false, it's a combination of `expandable` & `collapsed`. I think it's more readable this way... do you think I still need to change it?

src/Search/Search.js
dotansimha: fixed

src/Search/Search.js
dotansimha: Rebased and did some changes to integrate with the changes in #1764

lbelinsk:
@dotansimha before we merge this, I want to raise 3 final things:
1. I think I wasn't clear enough with the driver pattern.
All the family of the *WistOptions components have the unique driver structure of:
{inputDriver, dropdownLayoutDriver, driver}
and maybe also additional exists method.

This driver here is not following this pattern.
Instead of having isExpandable, I would put it under driver. isExpandable

2. How do we handle when `autoFocus` prop is provided?

3. Did you take care of the UI issue Domas mentioned?

lbelinsk:
About the autoFocus - yes, that was my concern.
About the driver, we should have {inputDriver, dropdownLayoutDriver, driver} only. the search driver should be driver. If you need additional methods for the search, they go inside the `driver` part.

About UX - the flickering behaviour was the concern Domas mentioned

dotansimha:
> About UX - the flickering behaviour was the concern Domas (milkyfruit) mentioned.

Yeah so it's fixes now.

Regarding the driver - I get what you are saying, but then how does the search test can have access to the `inputWithOptions` driver? It should be also part of `driver`? or as `{inputDriver, dropdownLayoutDriver, driver, inputWithOptionsDriver}`? 
Thanks!  
@lbelinsk

PR #1776, opened 2018-05-30
fix(<MultiSelectCheckbox/>) pass on inputElement prop

Wo1v3r:
### What changed

When inputElement prop is pass, pass it on to inputWithOptions
...

### Why it changed

Because that prop is exposed on the api and not working otherwise.
...

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green

### DIFF
--- a/src/MultiSelectCheckbox/MultiSelectCheckbox.js
+++ b/src/MultiSelectCheckbox/MultiSelectCheckbox.js
@@ -42,7 +42,6 @@ class MultiSelectCheckbox extends InputWithOptions {
 
   inputAdditionalProps() {
     return {
-      inputElement: <Input textOverflow="ellipsis" readOnly/>,
       value: this.selectedOptionsToText()
     };
   }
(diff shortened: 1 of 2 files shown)


lbelinsk:
@Wo1v3r what is your use case?
Do you need some other kind of input?

Wo1v3r:
@lbelinsk  I need the InputWithTags there similarly to original MultiSelect.
Also, I think that in general that an inputElement should passthrough for easier composition.

lbelinsk:
Ok, so just please add a test for it and I will merge

lbelinsk:
But Notice that the current implementation has a readonly input, so we don't have any support for keyboard support. we would have to check that everything is working as planned, and probably to add some additional code for it

Wo1v3r:
@lbelinsk Added test, Passed

lbelinsk:
@Wo1v3r I scanned wsr zeplin and I couldn't find any component that fits what you are trying to do.
Can you share your zeplin, we need to make sure it is approved even

lbelinsk:
@Wo1v3r I am closing this pr for now, but not deleting the branch.
Let's open it when we will have a decision.
Thanks!

PR #1780, opened 2018-06-03
Fix linting as a preparation for yoshi 2 and expose more components from index.js

lbelinsk:
Since we migrated to yoshi2, this pr is becoming relevant:
https://github.com/wix/wix-style-react/pull/1574/files

Before merging this:
- [x] I would check this locally with some consuming project.
- [x] Create one big exporting index file

After merging this:
- [ ] Announce on slack channel that we now support `import {Button, Input} from 'wix-style-react';`
- [ ] Change the autodoc to demonstrate the new import way.
- [ ] Stop using import-path


hepiyellow:
So this PR replaces #1574 ?

lbelinsk:
@hepiyellow yes, I closed that one

lbelinsk:
This pr is ready for merge. tested and seemed to work as expected. I will announce it on slack when I will release a new version next week. 

Adding leankit cards for the followups tasks from here

lbelinsk:
I have reverted all tree-shaking and yoshi2 changes in order to first merge all the lint changes that we would have to fox for yoshi 2, then I will create another pr only for yoshi2 migration. Then another one only for tree shaking.

PR #1783, opened 2018-06-03
fix <Heading /> font weight

david-pe:
Looks like font-weight was forgotten behind

### What changed
<Heading /> font-weight
...

### Why it changed
Upgrading to v4 broke the header style
...

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green

### DIFF
--- a/src/Typography/Typography.scss
+++ b/src/Typography/Typography.scss
@@ -5,6 +5,7 @@
   font-size: $font-size;
   color: $color;
   line-height: $line-height;
+  font-weight: normal;
 }
 
 @mixin generateVariations($list) {


shlomitc:
@david-pe - note that this will not affect the `Text`, `Label` and `Heading` components as they are imported from a different library (wix-ui-backoffice).

should it be added there as well?
https://github.com/wix/wix-ui-backoffice/blob/master/src/components/Text/Text.st.css#L18
https://github.com/wix/wix-ui-backoffice/blob/master/src/components/Label/Label.st.css#L16

david-pe:
I see this was a naive solution ... the intention of the PR was to point
out what broke in v4.
If I get a chance I'll have another look.

lbelinsk:
Closing this in favor of this pr:
https://github.com/wix/wix-style-react/pull/1815

This is how it is going too look like: 
https://wix-wix-style-react-pr-1815.surge.sh/?selectedKind=1.%20Foundation&selectedStory=1.3%20Heading&full=0&addons=0&stories=1&panelRight=0

PR #1784, opened 2018-06-04
<Input /> - Added `status` and `statusMessage` prop, and added Loader suffix

dotansimha:
### What changed

Changed `<Input />` to allow `status` and `statusMessage` props. It's values could be `error` or `loading`.
`error` will display the regular error, and `loading` will display a `<Loader />`.

Also:
- Added `tiny` size to `Loader` component.
- Deprecated `error` and `errorMessage` (they still works)
- Added example for `<Input>` with `Loader`.

@shlomitc 

### Why it changed

https://github.com/wix-private/wsr-issues/issues/106

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green


src/Input/Input.js
```diff
@@ -84,10 +89,13 @@ class Input extends Component {
       }
     };
 
-    const isClearButtonVisible = (!!clearButton || !!onClear) && !!value && !error && !disabled;
+    const hasErrors = error || status === Input.StatusError;
```
shlomitc: since we'd like to deprecate `error` and `errorMessage`, it might be better to split the conditions.
```
if(error || errorMessage){
...
}
```

src/Input/Input.js
```diff
@@ -84,10 +89,13 @@ class Input extends Component {
...
+    const isClearButtonVisible = (!!clearButton || !!onClear) && !!value && !hasErrors && !disabled;
```
shlomitc: if this is a new logic, let's add a test for it

src/Input/InputLoaderSuffix.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
```
shlomitc: is there a way to merge all `Input***Sufix`? they look to me all very similar and share same tooltip logic

src/Input/ThemedInput.js
```diff
@@ -28,7 +29,7 @@ class ThemedInput extends Input {
     const classes = {
       [styles.rtl]: !!rtl,
       [styles.disabled]: disabled,
-      [styles.hasError]: !!error,
+      [styles.hasError]: (!!error) || status === 'error',
```
shlomitc: again for the depreciation, use a separate condition

src/Input/Input.js
```diff
@@ -273,11 +282,23 @@ Input.propTypes = {
...
   error: PropTypes.bool,
 
-  /** The error message to display when hovering the error icon, if not given or empty there will be no tooltip */
+  /** Error message to display
+   * @deprecated
+   * @see statusMessage
+   */
   errorMessage: PropTypes.node,
+
+  /** The status (error/loading) message to display when hovering the error icon, if not given or empty there will be no tooltip */
```
shlomitc: hovering the status icon

src/Input/Input.js
```diff
@@ -273,11 +282,23 @@ Input.propTypes = {
   /** when set to true this component is disabled */
   disabled: PropTypes.bool,
 
-  /** Is input value erroneous */
+  /** Is input has a special status */
```
shlomitc: explain the different status please

src/Loader/Loader.js
```diff
@@ -33,7 +38,7 @@ export default class Loader extends WixComponent {
 
   static propTypes = {
     /** The size of the loader */
-    size: PropTypes.oneOf(['small', 'medium', 'large']),
+    size: PropTypes.oneOf(['tiny', 'small', 'medium', 'large']),
```
shlomitc: it doesn't spin when using the `tiny` size (see storybook)

dotansimha:
@shlomitc fixed your notes and rebased ;)

src/Input/InputLoaderSuffix.js
dotansimha: It makes sense, but I think that it's important to keep the `Suffix` as standalone components that provides the `Tooltip` by themselves, because this way it will be easier to add more `Suffix` types later. 
Maybe it worth adding HOC `withTooltip` that adds it, and wrap the `Suffix` components with it. what do you think @shlomitc ?

shlomitc: let's keep it as it is now

src/Input/Input.js
```diff
@@ -273,11 +290,23 @@ Input.propTypes = {
   /** when set to true this component is disabled */
   disabled: PropTypes.bool,
 
-  /** Is input value erroneous */
+  /** Input status - use to display an status indication for the user. for example: 'error' or 'loading' */
+  status: PropTypes.oneOf([Input.StatusError, Input.StatusLoading]),
```
shlomitc: really a minor issue, just put it next to the `statusMessage`

milkyfruit:
@dotansimha 
Few tweaks required:
- [ ] Tiny spinner's thickness should be `stroke-width: 3`
- [ ] `Margin-right` and `margin-left` should be 12px (if is set `box-sizing:border-box` otherwise it's 11px)
- [ ] Please update error icon to [new one](https://github.com/wix/wix-ui/blob/master/packages/wix-ui-icons-common/src/system/raw/FormFieldErrorFilled.svg)

dotansimha:
@milkyfruit fixed. regarding the margin - i'm using `$gap` which it `12px`, just like the error suffix.
@shlomitc fixed

milkyfruit:
@dotansimha Regarding the spacing, I just checked [the storybook](https://wix-wix-style-react-pr-1784.surge.sh/?selectedKind=3.%20Inputs&selectedStory=3.1%20%2B%20Input&full=0&addons=0&stories=1&panelRight=0) 
image

versus [the specs](zpl.io/2Egjola):
image

lbelinsk:
@dotansimha the tiny loader feature will be implemented by guild weekers in a separated pr, so please remove it form this pr...

lbelinsk:
tiny loader merged pr: https://github.com/wix/wix-style-react/pull/1911

dotansimha:
@lbelinsk done :)

dotansimha:
@shlomitc can we please merge it? :( everything is resolved...

shlomitc:
yes, sorry for the long time waiting. I wanted to review it as part of a bigger task related to Input family but it got delayed, so we'll just take it from here.
@milkyfruit - just verifying - is the 12px margin correct?

milkyfruit:
@shlomitc in the code, no. By the guidelines - yes, it should be 12px (when box-sizing: border-box).

shlomitc:
@milkyfruit not sure I understand. Is some change requires?

milkyfruit:
@shlomitc Yes, statusSuffixes should be 12px from the right side (now it's 6px).
Link to Zeplin specs: http://zpl.io/2Egjola

shlomitc:
@yardnsm - can you please fix the suffix as mentioned in the last comment?

PR #1791, opened 2018-06-05
<InputWithOptions />- fix bug where delimiters did not behave like En…

ChenRoth:
<InputWithOptions />- fix bug where delimiters did not behave like Enter/Tab

### What changed
Delimiters passed to InputWithOptions now trigger onManuallyInput similarly to Enter/Tab

### Why it changed
pressing a delimiter when manually entering an option would not behave the same as pressing Enter/Tab

approved by [milkyfruit]
---

- [X] Change is tested
- [ ] Change is documented
- [x] Build is green
- [X] Change of UX is approved by [wuwa](https://github.com/wuwa) or [milkyfruit](https://github.com/milkyfruit)

### DIFF
--- a/src/InputWithOptions/InputWithOptions.js
+++ b/src/InputWithOptions/InputWithOptions.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import WixComponent from '../BaseComponents/WixComponent';
 import Input from '../Input';
 import omit from 'omit';
+import defaultTo from 'lodash/defaultTo';
 import DropdownLayout from '../DropdownLayout/DropdownLayout';
 import Highlighter from '../Highlighter/Highlighter';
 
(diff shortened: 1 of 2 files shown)


hepiyellow:
I tried the current Tags component  and your PR's component:
I both when I press 'ala' and ',' then I get `alabama` slected, the options is open.

What should be the difference in behaviour?

lbelinsk:
The change introduced here is allowing to pass delimiters such as "," or ";" and then each time a user type one of those delimiters, it will behave exactly like an Enter was pressed.

I am not sure it makes sense to use Autocomplete, press "," and to get a selection.

A delimiter makes sense in a component like the MultiSelect where you need to separate the items with delimiters (also in order to allow copy-paste of multiple options).

Moreover, the default delimiters for the MultiSelect are different than the ones for InputWithOptions.

I am not sure this is the desired behavior. @ChenRoth what is the use case?

@wuwa @milkyfruit

lbelinsk:
Also, the reason why when you type "," which is a default delimiter for the multiSelection components is because you want the dropdownLayout to remain open in order to allow the user to continue selecting items.

ChenRoth:
this was already discussed and settled with @milkyfruit 
the expected behaviour is for delimiters to act exactly like Enter/Tab.
Granted, my change doesn't cover the dropdown selection - only manual inputs - which is better than what is available now.

If you want to see the actual problem, go to storybook, press "Enter" in Tags and watch the console.
now do the same for "," and see what happens... this breaks my use case and shouldn't be this way.

milkyfruit:
I approved this fix, because I don't see a reason to provide different behaviours when selecting predefined option or entering a custom one. In both cases the controls should be the same, otherwise it can be confusing for a user.

hepiyellow:
@ChenRoth Replaced by #1804

ChenRoth:
👍

PR #1792, opened 2018-06-05
[fix] remove box shadow when focusing Input with 'tags' theme

ChenRoth:
### What changed

refactored code to make an &lt;Input theme="tags" /&gt; *actually* remove the shadow box when focused

### Why it changed

the current code doesn't really work in production env
also, it uses `data-hook` in css

here's a screenshot showing the current box shadow:
screen shot 2018-06-05 at 18 05 27

---
this is a css bug fix, so not sure 1-2, 4 are relevant
- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


shlomitc:
@ChenRoth - how is the bug not visible in the storybook? is it caused by the `data-hook`?

ChenRoth:
i could only see it in production (minified class names)
the problem is that data-hook css rule precedence is not strong enough to override the default rule

PR #1793, opened 2018-06-05
<MultiSelect/>: Error indication and message, readOnly styles

Wo1v3r:
### What changed
- Added error indication <MultiSelect/>
- Added optional read only style + render menu arrow on readOnly
...

### Why it changed
- Need of this design in Wix Automations
- error indication was missing
- readonly style to be used when interacting with dropdown options \ tags without typing

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


Wo1v3r:
@shlomitc  Can you look at the surge?

one failing test here is due to applitools eyes.

The second one does not reproduce locally nor on surge.

The e2e test looks very peculiar being based off some magic number and will break on smaller resolutions. (tests the opposite of what it declares)

Wo1v3r:
@shlomitc 
Can you have a look in surge?
first e2e fails because of eyes
second e2e uses some magic number which is weird to me (multiselect will break to newline depending on screen size on storybook) . on surge looks fine...

src/MultiSelect/InputWithTags.js
```diff
@@ -107,6 +108,7 @@ class InputWithTags extends React.Component {
             {...desiredProps}
             dataHook="inputWithTags-input"
             disabled={disabled}
+            error={error}
```
shlomitc: isn't it something the `desiredProps` pass in?

src/MultiSelect/README.md
```diff
@@ -21,6 +21,7 @@
 | fixedFooter | node | - | - | A fixed footer to the list |
 | delimiters | array | [','] (also Enter and Tab) | - | delimiters that will be treated like enter press |
 | maxNumRows | number | - | - | Max number of visible lines |
+| readOnly | bool | - | - | Sets the input to be read-only, rendering a menuArrow suffix
```
shlomitc: it's kind of a side effect that `readOnly` causes it to look like a dropdown selection, but it's actually functioning differently.
I'd prefer calling it `select`. maybe as `mode="select"`.
@wuwa - is this a correct terminology for this mode?

src/MultiSelect/InputWithTags.js
```diff
@@ -65,7 +67,6 @@ class InputWithTags extends React.Component {
       'inputElement',
       'closeOnSelect',
       'predicate',
-      'menuArrow',
```
shlomitc: why do you omit it?

src/MultiSelect/InputWithTags.js
Wo1v3r: it is grabbed from the passed props so its not in desiredProps e.g;

{prop1,prop2,error,...desiredProps} = this.props

error not in desiredProps

It was like that before, just not passed to <Input/>, same as disabled you can have a look

src/MultiSelect/README.md
Wo1v3r: Cool LMK resolution and I'll change accordingly both in docs and src

src/MultiSelect/InputWithTags.js
Wo1v3r: you mean why I do not omit anymore;
InputWithOptions sets it to true by default, that's why it was omitted. I de-omit it to be able to pass it from <MultiSelect/> based on readOnly (or whatever will be decided to be called)

src/MultiSelect/InputWithTags.js
shlomitc: missed it, cool

src/MultiSelect/InputWithTags.js
shlomitc: but this means `menuArrow` will now become a valid prop in the props and can be applied by the consumer of the `MultiSelect` component.

Wo1v3r: Hm...
What do you suggest then ?

Wo1v3r:
@shlomitc  any updates regarding the e2e  & select \ readonly prop?

src/MultiSelect/InputWithTags.js
shlomitc: Pass it directly to the input and determine its value based on the ‘select’ flag

wuwa:
dropdown/error should not move down - however this read-only state is not correct in my opinion. needs to be without X and dropdown. can we see a screenshot of automations need?

Wo1v3r:
@shlomitc  passing mode='select' now instead LMK if this is what you wanted...

Wo1v3r:
@shlomitc  any upd8 ? :[]

shlomitc:
API looks good.
So we still miss the error and arrow position.
I see it also caused horizontal scrolling for the component

image

shlomitc:
Hey @Wo1v3r ! any updates regarding this PR?

Wo1v3r:
hey @shlomitc  I'll be back next week and refurbish this PR :)

yardnsm:
I've updated the PR a bit.

* Fixed lint errors
* Update the propType of `select` accordingly
* Fixed the scroll overflow issue (sort-of, I removed a [`margin` property](https://github.com/wix/wix-style-react/pull/1793/files#diff-b853af1151734ba031f3c0c87894aaffL5) that doesn't seems that important, correct me if I wrong)

What has been decided regarding the position of the arrow and the error icon?

shlomitc:
@yardnsm 
According to Ben - dropdown/error should not move down
Does that answer your question?

Can you fix conflicts as well?

yardnsm:
Fixed the position of the arrow and the error icon:

screen shot 2018-07-19 at 11 30 02
screen shot 2018-07-19 at 11 30 48
screen shot 2018-07-19 at 11 31 04

Regarding the conflicts, I'm not sure what was the reason for the changes in `MultiSelect.js`, but I can resolve the conflicts in `InputWithTags.scss`.

shlomitc:
@yardnsm - so please solve the conflicts

stories/MultiSelect/index.js
```diff
@@ -13,23 +13,38 @@ import ExampleWithLimitedNumRows from './ExampleWithLimitedNumRows';
...
         </CodeExample>
 
         <CodeExample title="Limited num rows" code={ExampleWithLimitedNumRowsRaw}>
-          <ExampleWithLimitedNumRows/>
+          <div style={{maxWidth: 720}}>
+            <ExampleWithLimitedNumRows/>
+          </div>
+        </CodeExample>
+
+        <CodeExample title="Read Only with Arrow" code={ExampleReadOnlyRaw}>
```
shlomitc: The arrow looks weird
image

yardnsm: Fixed in bb83d97

PR #1794, opened 2018-06-05
make ThemedInput noRight(Left)BorderRadius work

ChenRoth:
### What changed

Added css classes for making `<Input noRightBorderRadius noLeftBorderRadius />` work

### Why it changed

these props didn't work properly

---

- [x] Change is tested
- [x] Change is documented
- [ ] Build is green


src/Input/ThemedInput.js
```diff
@@ -32,11 +32,21 @@ class ThemedInput extends Input {
...
       [styles.roundInput]: roundInput,
-      [styles.hasValue]: (value && value.length) || (this.input && !!this.input.value),
-      [noRightBorderRadius]: noRightBorderRadius,
-      [noLeftBorderRadius]: noLeftBorderRadius
+      [styles.hasValue]: (value && value.length) || (this.input && !!this.input.value)
     };
 
+    if (noRightBorderRadius === true) {
+      classes[styles.noRightBorderRadius] = true;
+    } else if (noRightBorderRadius) {
```
ChenRoth: i had to preserve some hack which used `noRightBorderRadius` prop as a string to inject a classname

src/Input/ThemedInput.js
```diff
@@ -32,11 +32,21 @@ class ThemedInput extends Input {
       [styles.hasHover]: forceHover,
       [styles.hasFocus]: forceFocus || this.state.focus,
       [styles.roundInput]: roundInput,
-      [styles.hasValue]: (value && value.length) || (this.input && !!this.input.value),
-      [noRightBorderRadius]: noRightBorderRadius,
```
shlomitc: just do it here
`[noRightBorderRadius]: noRightBorderRadius`
should be:
`[styles.noRightBorderRadius]: noRightBorderRadius`

it should be only boolean and not a string (need to remove from the docs as well)

src/Input/ThemedInput.js
shlomitc: let's remove it

src/Input/ThemedInput.js
ChenRoth: well, it would introduce a breaking chance

ChenRoth: ok, i understand the motivation for that hack and i'll remove it

shlomitc:
So I compared this PR storybook and the regular WSR and there are some minor changes the range now need several adjustments:

1. InputRange
  1.1. the width of it was changed and the left input outline is overriden by the right input. maybe it's the right behvavior though. @milkyfruit, can you verify?
  1.2. One of the intersecting borders should disappear (it's double)
2. DateRange
  2.1. you actually fixed this one with rounded edges, cool :)
  2.2. same like 1.2


## Before
https://wix-wix-style-react.surge.sh/?selectedKind=3.%20Inputs&selectedStory=3.7%20Range&full=0&addons=0&stories=1&panelRight=0

### InputRange
image
image
image

### DateRange
image
image
image


## After
https://wix-wix-style-react-pr-1794.surge.sh/?selectedKind=3.%20Inputs&selectedStory=3.7%20Range&full=0&addons=0&stories=1&panelRight=0

### Input Range
image
image
image

### Date Range
image
image
image

milkyfruit:
@shlomitc both number and date range inputs must have 1px border line in the middle. I see two ways to achieve it:
A. overlap each over by 1px
B. The second input doesn't show the left border on default state

In both suggested options the active input must have a higher number of z-index, like range input in "before" example:
image

hepiyellow:
I pushed some fixes.
### `<Input>` className
I added className prop to <Input> : We should write in Migration that if someone used noRightBorderRadius as a hack to pass class name to the Input, then they should now use the `className` prop.


### `<DatePicker >`- inputProps
 I also fixed Range with DatePicker which was  broken before.
- I added `inputProps` to DatePicker

I think everything works well now.

Just need to add the migration doc.
I saw that there are 7 occurences of noRightBorderRadius in wix-private where they are passing class names and not boolean (which didn't work).
I didn't verify if some of them are already using WSR 4.
If all of them are NOT using WSR 4, then we can go ahead and release this in 4.1.1, and add this to the migration doc of 4.0.0 (while mentioning it is actually implemented in 4.1.1)

hepiyellow:
Failing tests are e2e snapshots.
1 )is a flaky test of unrelated EditableSelector
2) 2 other are expected fixes

PR #1795, opened 2018-06-06
`<Search>` - Fixed bug with mounted component and expandable

dotansimha:
### What changed

Added support for changing `expandable` value after the component has been mounted.

### Why it changed

To fix a bug ;)

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green


lbelinsk:
@dotansimha I don't think I would add this behaviour. makes sense that the expandable behaviour will be in the mounting (or will simply not be). Not worth the listening and the extra condition on each render.

I think that you added it due to a technical issue in the storybook or something.
i would suggest:

1. Open an issue about it (tag @argshook )
2. Add a custom example (the autodoc supports custom examples with `example` property.

shlomitc:
It looks like there is an issue in the clickable area of the component
image

dotansimha:
@lbelinsk I reverted the changes, and added storybook example for expandable.
@shlomitc Fixed the clickable area issue.

stories/Search/index.story.js
```diff
@@ -2,7 +2,7 @@ import Search from 'wix-style-react/Search';
 
 const settings = {
   category: '3. Inputs',
-  storyName: '3.9 Search',
+  storyName: '3.9a Search',
```
shlomitc: Should stay 3.9 without a

stories/Search/expandable.story.js
```diff
@@ -0,0 +1,37 @@
+import Search from 'wix-style-react/Search';
```
shlomitc: instead of having an automated story you can just add it as a simple example with code snippet as part of the `examples` section of `3.9 Search`

He'res an example how to do it:
https://github.com/wix/wix-style-react/blob/master/stories/FloatingHelper/FloatingHelper.story.js#L68-L97

https://wix-wix-style-react-pr-1765.surge.sh/?selectedKind=8.%20Notification%20Bars&selectedStory=8.6%20FloatingHelper&full=0&addons=0&stories=1&panelRight=0

dotansimha: Fixed! thanks :)

stories/Search/index.story.js
dotansimha: fixed

PR #1799, opened 2018-06-09
5.0.0 - `<ThemedInput/>` fix noRightBorderRadius

hepiyellow:
This is same as https://github.com/wix/wix-style-react/pull/1794,
but as a breaking change, which is not allowing passing string in the noRightBorderRadius / noLeftBorderRadius.

See new Input - Migration.md


src/Input/MIGRATION.md
```diff
@@ -0,0 +1,7 @@
+# `<Input/>` Migration Guide
+
+## 4.x.x -> WSR 5.x.x
+### noRightBorderRadius / noLeftBorderRadius
+These props used to accept a string and pass it onto the className of the Input's root element. Now the accept boolean, and apply the propper classes that affect the border radius.
+In some cases these were used as a hack to pass className to the Input. Either  in order to set the border-radius (since passing boolean didn't work), or to do some other styling.
```
lbelinsk: extra space `Either  in`

src/Input/Input.js
```diff
@@ -300,13 +300,16 @@ Input.propTypes = {
   /** Displays clear button (X) on a non-empty input */
   clearButton: PropTypes.bool,
 
+  /** A single CSS class name to be appended to ther Input's wrapper element. */
```
lbelinsk: typo `the`

src/Input/Input.js
```diff
@@ -300,13 +300,16 @@ Input.propTypes = {
...
+  className: PropTypes.string,
```
lbelinsk: If the user passes className, it will override the `inputClassNames. This is wrong to say:
"A single CSS class name to be appended to the Input's wrapper element" when we actually override some classes, and not only adding.

What is your intention? Are you overriding the inputClasses on purpose?

src/DatePicker/DatePicker.js
```diff
@@ -49,6 +49,9 @@ export default class DatePicker extends WixComponent {
     /** Can provide Input with your custom props */
     customInput: PropTypes.node,
 
+    /** Properties appended to the default Input component or the custom Input component. */
+    inputProps: PropTypes.object,
```
lbelinsk: I would add a test for this. see that your input gets the additional props this way

src/Composite/FieldWithSelectionComposite/FieldWithSelectionComposite.js
```diff
@@ -113,6 +113,7 @@ class FieldWithSelectionComposite extends WixComponent {
         {label}
         <div className={classNames(inputsWrapperClassNames)} data-hook="input-wrappers">
           {React.cloneElement(textInput, {
+            noRightBorderRadius: true,
```
lbelinsk: I trust you here that this what should be here

shlomitc:
Are you sure we want it merged?

hepiyellow:
That's what we talked about. I thought its what you want.
I don't mind keeping the hack until we use <Input> from the backoffice.

shlomitc:
I do want, but I mean you already commit for the v5 with migration.
Also - you merged something bad to the changelog
let's revert it, restore the branch and keep it on the side

PR #1802, opened 2018-06-10
Add item

shaibv:
AddItem component
Unbind this markup and logic from ImageViewer.

https://github.com/wix-private/wsr-issues/issues/61


lbelinsk:
@shaibv you have a bug with the "no tooltip mode" - the tooltip is visible:

image

src/AddItem/AddItem.driver.js
```diff
@@ -0,0 +1,27 @@
...
+  const addTooltip = () => byHook('add-tooltip');
+
+  return {
+    getElement: () => element,
+    getContainerStyles: () => element.getAttribute('style'),
+    isAddButtonVisible: () => !!addImageButton(),
+    isAddTooltipVisible: () => !!addTooltip(),
+    clickAdd: () => ReactTestUtils.Simulate.click(addImageButton()),
+    exists: () => !!element,
+    setProps: props => {
```
lbelinsk: We prefer not to have the `setProps` method

src/AddItem/AddItem.driver.js
```diff
@@ -0,0 +1,27 @@
...
+
+  const addImageButton = () => byHook('add-area');
+  const addTooltip = () => byHook('add-tooltip');
+
+  return {
+    getElement: () => element,
+    getContainerStyles: () => element.getAttribute('style'),
+    isAddButtonVisible: () => !!addImageButton(),
+    isAddTooltipVisible: () => !!addTooltip(),
+    clickAdd: () => ReactTestUtils.Simulate.click(addImageButton()),
```
lbelinsk: won't it be simpler API to have `click` instead of `clickAdd`?
You should add comments for the autodoc on top of each method here to explain what it does.

src/AddItem/AddItem.driver.js
```diff
@@ -0,0 +1,23 @@
...
+const addItemDriverFactory = ({wrapper, element}) => {
+
+  const byHook = dataHook =>
+    element.querySelector(`[data-hook="${dataHook}"]`);
+
+  const addImageButton = () => byHook('add-area');
+  const addTooltip = () => byHook('add-tooltip');
+
+  return {
+    getElement: () => element,
```
lbelinsk: document methods for autodoc.

src/AddItem/AddItem.js
```diff
@@ -0,0 +1,50 @@
...
+import Tooltip from '../Tooltip';
+import Plus2 from '../Icons/dist/components/Plus2';
+import WixComponent from '../BaseComponents/WixComponent';
+
+class AddItem extends WixComponent {
+
+  render() {
+
+    let {
+      onAddItem,
```
lbelinsk: I think onAddItem is confusing, because it is the callback for the click event, not for after the user added item.. onClick is more fitting imo

src/AddItem/AddItem.spec.js
```diff
@@ -0,0 +1,106 @@
...
+
+
+  beforeEach(() => {
+    document.body.innerHTML = '';
+  });
+
+  describe('when default scenario', () => {
+    beforeEach(() => {
+      props = {
+        onAddItem: addItem,
```
lbelinsk: I suggest removing this props from here, as you define it differently in each `it` suite

src/AddItem/AddItem.spec.js
```diff
@@ -0,0 +1,106 @@
...
+  describe('height and width', () => {
+
+    it('should be added to style attribute when item is not present', () => {
+      props = {
+        tooltipContent: TOOLTIP_CONTENT,
+        width: 300,
+        height: 300
+      };
+      driver = createDriver(<AddItem {...props}/>);
+      expect(driver.getContainerStyles()).toEqual('width: 300px; height: 300px;');
```
lbelinsk: I would remove this getContainerStyles method from the driver, and instead use 
`window.getComputedStyles(driver.getElement())`

src/index.js
```diff
@@ -2,6 +2,7 @@ import './assets/helvetica.scss';
 
 import {MessageBoxMarketerialLayout, MessageBoxFunctionalLayout, HeaderLayout, FooterLayout} from './MessageBox';
 
+export {default as AddItem} from './AddItem';
```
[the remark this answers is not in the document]
shaibv: By the user

src/AddItem/AddItem.driver.js
shaibv: renamed to element and now it's the same as e2e

src/ImageViewer/ImageViewer.spec.js
```diff
@@ -46,6 +42,19 @@ describe('ImageViewer', () => {
...
 
+    it('should trigger add image', () => {
+      props = {
+        imageUrl: '',
+        onAddImage: addImage
+      };
+      driver = createDriver(<ImageViewer {...props}/>);
+      const dataHook = driver.getAddItemDataHook();
+      const wrapper = driver.getElement();
+      const addItemDriver = addItemTestkitFactory({wrapper, dataHook});
```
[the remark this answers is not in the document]
shaibv: That just wont work, don't know why it breaks the testkit (cyclic dependency)

src/AddItem/AddItem.js
```diff
@@ -41,7 +41,7 @@ class AddItem extends WixComponent {
 }
 
 AddItem.propTypes = {
-  onAddItem: PropTypes.func,
```
lbelinsk: Please document the props for autodoc

src/AddItem/AddItem.driver.js
```diff
@@ -10,8 +10,9 @@ const addItemDriverFactory = ({wrapper, element}) => {
   const addTooltip = () => byHook('add-tooltip');
 
   return {
-    getElement: () => element,
-    getContainerStyles: () => element.getAttribute('style'),
+    element: () => element,
```
lbelinsk: Document for autodocs

src/ImageViewer/ImageViewer.spec.js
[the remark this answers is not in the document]
lbelinsk: We would have to fix this. Ohterwise, how would you expect a user to click the imageViewer?

PR #1805, opened 2018-06-10
SideMenu.Header: when onClick is not set change the header to be type of div

mozesoron:
### What changed

...

### Why it changed

...

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


src/SideMenu/README.md
```diff
@@ -16,7 +16,7 @@ Container for your header (or anything you want)
 
 | propName | propType | defaultValue | isRequired | description |
 | -        | -        | -            | -          | -           |
-| onClick  | func     | -            | -          | -           |
+| onClick  | func     | -            | -          | header will be of type: button|
```
lbelinsk: please elaborate a bit more here. something like:
`A callback that will be called when the header is clicked. when no callback is passed, the Header will be wrapped by a div, otherwise, it will be a clickable button.

src/SideMenu/core/Header.js
```diff
@@ -2,15 +2,21 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import styles from './styles.scss';
 
-const Logo = ({onClick, children}) =>
-  <button type="button" onClick={onClick} className={styles.logo} data-hook="menu-header">
-    {children}
-  </button>;
+const Logo = ({onClick, children}) => {
```
lbelinsk: Nice idea to handle your use case.
Please make sure it is approved by our UX. I see you handled the cursor nicely, so I don't see any reason why not to allow this.

src/SideMenu/core/Header.js
```diff
@@ -2,15 +2,21 @@ import React from 'react';
...
+  if (onClick) {
+    return (<button type="button" onClick={onClick} className={styles.logo} data-hook="menu-header">
+      {children}
+    </button>);
+  } else {
+    return (<div className={styles.simpleLogo} data-hook="menu-header">
+      {children}
+    </div>);
+  }
+};
```
lbelinsk: Missing a test for this

src/Tabs/Tabs.driver.js
```diff
@@ -17,6 +17,7 @@ const tabsDriverFactory = ({element, wrapper, component}) => {
     getTitles: () => getItems().map(item => item.textContent),
```
lbelinsk: For next time: This is not related to this pr, better to create a separate one in case we would need to revert

src/Tabs/Tabs.driver.js
```diff
@@ -17,6 +17,7 @@ const tabsDriverFactory = ({element, wrapper, component}) => {
...
     clickTabAt: index => ReactTestUtils.Simulate.click(getItems()[index]),
     getActiveTabIndex: () => getItems().findIndex(item => item.classList.contains(styles.active)),
+    getActiveTabId: () => getItems().filter(item => item.classList.contains(styles.active))[0].id,
```
lbelinsk: maybe it should be safer. if there is no active tab, this will throw: can't access id of undefined or something...

src/Tabs/core/TabItem/index.js
```diff
@@ -24,6 +24,7 @@ class TabItem extends React.Component {
 
     const containerProps = {
       key: item.id,
+      id: item.id,
```
lbelinsk: Need to update the proptypes here: https://github.com/mozesoron/wix-style-react/blob/e1a167aa7b3b2da09eb67aebf390bd1dc5413874/src/Tabs/core/constants/tab-prop-types.js

src/Tabs/Tabs.driver.js
mozesoron: you are right I thought that it will let me create another pr.

src/Tabs/core/TabItem/index.js
mozesoron: I already see an id is set under the item const: https://github.com/mozesoron/wix-style-react/blob/e1a167aa7b3b2da09eb67aebf390bd1dc5413874/src/Tabs/core/constants/tab-prop-types.js#L8
 do I need to change anything else?

src/SideMenu/core/Header.js
mozesoron: added 2 tests

src/Tabs/core/TabItem/index.js
lbelinsk: No, I missed that, all good then

lbelinsk:
There is an issue that eyes helped us catching here.
If a user didn't give onClick then we must make sure that the dom layout would still look the same, only the cursor should change.

Eyes shows us that there is a diff with the margins / paddings

image

lbelinsk:
@mozesoron  I am closing this in favor of https://github.com/wix/wix-style-react/pull/1900
Please feel free to say if you think this is not the way to go, and we could then reopen this

PR #1814, opened 2018-06-12
[React 16] Tooltip doesn't work

Faradey27:
### What changed

Way how we set Tooltip position

### Why it changed

because it is broken for react@16

---

- [x] Change is tested
- [ ] Change is documented
- [x] Build is green

### DIFF
--- a/src/Tooltip/Tooltip.js
+++ b/src/Tooltip/Tooltip.js
@@ -158,47 +158,8 @@ class Tooltip extends WixComponent {
     this.props.onClickOutside && this.props.onClickOutside(e);
   }
 
-
   componentDidUpdate() {
-    if (this._mountNode && this.state.visible) {
-      const arrowPlacement = {top: 'bottom', left: 'right', right: 'left', bottom: 'top'};
-      const position = this.props.relative ? 'relative' : 'absolute';
-      const tooltip = (
-        <TooltipContent
-          onMouseEnter={() => this._onTooltipContentEnter()}
-          onMouseLeave={() => this._onTooltipContentLeave()}
-          ref={ref => {
-            if (this.props.relative) {
-              this.tooltipContent = ref.tooltip;
-            } else {
-              this.tooltipContent = ref;
-            }
-          }}
-          showImmediately={this.props.showImmediately}
-          theme={this.props.theme}
-          bounce={this.props.bounce}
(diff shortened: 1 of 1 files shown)


src/Tooltip/Tooltip.js
```diff
@@ -159,7 +159,7 @@ class Tooltip extends WixComponent {
   }
 
 
-  componentDidUpdate() {
+  componentDidUpdate(prevProps, prevState, snapshot, onSubtreeRendered) {
```
shlomitc: Is this really the API for compnentDidUpdate?
I think `snapshot` is only in React 16 and I don't remember `onSubtreeRendered` as a parameter.
I remember having a `context` instead of `snapshot` which I also don't know what was used for.

The big question is whether we need to callit

src/Tooltip/Tooltip.js
```diff
@@ -305,11 +305,15 @@ class Tooltip extends WixComponent {
           let fw = 0;
           let sw = 0;
           do {
-            this.componentDidUpdate();
-            const tooltipNode = ReactDOM.findDOMNode(this.tooltipContent);
-            fw = this._getRect(tooltipNode).width;
-            this._updatePosition(this.tooltipContent);
-            sw = this._getRect(tooltipNode).width;
+            // TODO migrate to React Portals
+            this.componentDidUpdate(null, null, null, () => {// eslint-disable-line no-loop-func
```
shlomitc: I think it's better to refactor the logic outside of `componentDidUpdate` and just call the logical function instead of calling the lifecycle hook

src/Tooltip/Tooltip.js
Faradey27: in react 15 it is props, state, context
in react 16 it is props, state, snapshot
as it is react 16 fix, then react 16 names

onSubtreeRendered - custom parameter, here in tooltip we already have antipatern with calling componentDidUpdate from component, `this.componentDidUpdate()`

so i just added one more parameter - trick

src/Tooltip/Tooltip.js
Faradey27: agree, i thought about this at start, but as this tooltip will be replaced with wix-ui-core tooltip soon, decided to not spent time on this

but will update)

Faradey27:
@shlomitc updated

shlomitc:
Amazing. Thanks a lot!

PR #1818, opened 2018-06-13
<Tags/> - missing highlight when typing (autocomplete feature)

Faradey27:
### What changed

- fixed <Tags/> highlight when typing (autocomplete feature)

### Why it changed

- it was broken

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green

### DIFF
--- a/src/Highlighter/Highlighter.js
+++ b/src/Highlighter/Highlighter.js
@@ -16,8 +16,6 @@ const childKeyGenerator = () => {
   return () => `highlighted-child-${childKey++}`;
 };
 
-let nextChildKey = childKeyGenerator();
-
 const ELEM_TYPES = {
   STRING: 'string',
   ARRAY: 'array',
@@ -38,7 +36,7 @@ const getElementType = element => {
   return '';
 };
 
-const highlight = (element, match) => {
+const highlight = (element, match, nextChildKey) => {
   if (!element) {
     return null;
   }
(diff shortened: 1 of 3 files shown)


Faradey27:
@shlomitc

Faradey27:
@shlomitc by default highlight prop is setted to false, change it to true and re-check

without my changes - nothing
with them - highlight works

shlomitc:
I understand, so you mean passing `highlight={true}` prop will allow it. This is nice but there are two issues:
1. This should be the default behavior
2. When `highlight` is on -> the tags options become textual and lose any icon/tag they have (take a look at the Arizona option for example)

Faradey27:
@shlomitc 
1) fixed
2) good catch, re-worked algoritm to support this

src/Highlighter/Highlighter.js
```diff
@@ -3,6 +3,58 @@ import WixComponent from '../BaseComponents/WixComponent';
...
 import PropTypes from 'prop-types';
 
+const childKeyGenerator = () => {
+  let childKey = 0;
+  return () => `highlighted-child-${childKey++}`;
+};
+
+let nextChildKey = childKeyGenerator();
+
+const _isReactElement = element => element && element.props !== undefined;
```
hepiyellow: While we're at it. Please use `React.isValidElement()`.
(double check me if this is really the required validation you need)

src/Highlighter/Highlighter.js
```diff
@@ -16,91 +68,16 @@ class Highlighter extends WixComponent {
...
-      }
-      case this._CHILD_TYPES.ARRAY:
-        return this._renderArrayOfChildren(element, parent);
-      default:
-        return;
-    }
+  constructor() {
+    super();
+    // we want to drop counter on component creation
+    nextChildKey = childKeyGenerator();
```
hepiyellow: why don't we reset the counter on constructor?
Does it overflow nicely (if it will ever happen)...?

Faradey27:
@hepiyellow @shlomitc updated, build is green

shlomitc:
Looks good.
@wuwa - highlighting is now default for the Tags component, please verify so we can merge

wuwa:
cool, looks good

PR #1831, opened 2018-06-17
`<MultiSelect>` - (fix) fix onManuallyInput() called twice

hepiyellow:
- Needed to fix InputWithOption driver which triggered all the key presses on the inputWrapper rather than the input itself.
- Trying to make order in the chaos, I try to go towards a solution which will not involve extending InputWithOptions, try to avoid overriding InputWithOptions methods. To help with that I added a utility for EventHandler chaining.

I suggest reviewing this commit-by-commit.


src/InputWithOptions/InputWithOptions.js
```diff
@@ -56,13 +63,18 @@ class InputWithOptions extends WixComponent {
...
+    const inputAdditionalProps = this.inputAdditionalProps();
+    const inputProps = Object.assign(omit(Object.keys(DropdownLayout.propTypes).concat(['onChange', 'dataHook']), this.props), inputAdditionalProps);
 
     const {inputElement} = inputProps;
     return React.cloneElement(inputElement, {
       menuArrow: true,
       ref: input => this.input = input,
       ...inputProps,
+      onKeyDown: chainEventHandlers(
+        this.props.onKeyDown,
```
shlomitc: user handler

src/InputWithOptions/InputWithOptions.js
```diff
@@ -56,13 +63,18 @@ class InputWithOptions extends WixComponent {
...
+        inputAdditionalProps && inputAdditionalProps.onKeyDown,
```
shlomitc: extending handler (for tags / multi-select)

src/MultiSelect/MultiSelect.js
```diff
@@ -116,24 +135,6 @@ class MultiSelect extends InputWithOptions {
...
-        }
-
-      } else {
-        this.props.onSelect([{id: value.trim(), label: value.trim()}]);
-      }
-
-      this.clearInput();
-    }
-
     if (this.props.onKeyDown) {
```
shlomitc: this will be called twice

src/utils/EventUtils.spec.js
```diff
@@ -0,0 +1,25 @@
...
+    expect(fn2.mock.calls.length).toBe(1);
+  });
+
+  it('should skip undefined handler', () => {
+    const fn2 = jest.fn();
+    chainEventHandlers(undefined, fn2)(newEvent());
+    expect(fn2.mock.calls.length).toBe(1);
+  });
+
+  it('should skip undefined handler', () => {
```
argshook: remove, this is duplicate test.

maybe instead of this it would be good to check the order of handler calls

src/utils/EventUtils.js
```diff
@@ -0,0 +1,22 @@
...
+ * - If an exception is thrown in the a handler, then the rest of the handler will NOT be called.
+ * - returns the return value of the first (defined) handler
+ * @returns An event handler that chains all given handlers
+ */
+export function chainEventHandlers(...handlers) {
+  return event => {
+    let res;
+    handlers.forEach(h => {
+      if (h) {
+        const r = h && h(event);
```
argshook: no need to check for `h` twice

im also thinking perhaps the return value of handler is not needed? here
the last result is returned so the others are discarded anyway. if so, it can become simpler

```js
export const chainEventHandlers = (...handlers) =>
  event =>
    handlers
      .filter(i => i)
      .forEach(handler => handler(event))
```

i would also put this into dedicated file `chain-event-handlers.js` but that's not really important

src/InputWithOptions/InputWithOptions.js
```diff
@@ -216,22 +227,24 @@ class InputWithOptions extends WixComponent {
...
     if (!this.dropdownLayout._onKeyDown(event)) {
-      switch (event.key) {
-        case 'Enter':
-        case 'Tab': {
-          this._onManuallyInput(this.state.inputValue);
-          break;
-        }
-        default:
-          this.showOptions();
+      if (this.getManualSubmitKeys().includes(event.key)) {
```
argshook: need to check if `includes` would work in IE, cause i've been bitten by it already.

thing is, when component is imported as standalone (`import Component from 'wix-style-react/Component`) it doesn't include common polyfills and so no `include` polyfill. most browsers are okay but that one of them fails

src/InputWithOptions/InputWithOptions.js
```diff
@@ -13,6 +14,12 @@ class InputWithOptions extends WixComponent {
   dropdownClasses() {}
   dropdownAdditionalProps() {}
   inputAdditionalProps() {}
+  /**
+   * An array of key codes (default ['Enter','Tab']) that act as manual submit. Will be used with the onKeyDown(event), tha key codes are values of event.key. When a manual submit key is pressed then onManuallyInput will be called.
+   */
+  getManualSubmitKeys() {
```
argshook: maybe it doesn't need to be a method, just a value?

src/InputWithOptions/InputWithOptions.js
```diff
@@ -216,22 +227,24 @@ class InputWithOptions extends WixComponent {
...
-        case 'Tab': {
-          this._onManuallyInput(this.state.inputValue);
-          break;
-        }
-        default:
-          this.showOptions();
+      if (this.getManualSubmitKeys().includes(event.key)) {
+        this._onManuallyInput(this.state.inputValue);
+      } else {
+        this._onNonSubmitKeyPressed();
```
argshook: i'd just `this.showOptions()` right here

src/InputWithOptions/InputWithOptions.js
hepiyellow: I made it a function and put it on top, in the `//Abstractions` section,
 so it is easy to understand that it can be overridden. 
It is overriden in MutliSelect.
It makes sense to be a member, rather than a function, but then I need to overwrite it in MultiSelect constructor, and it will be less clear.
For someone reading the InputWithOptions code, they might ask themselves if this member is a constant read-only or not, and if it is a function, then it is obvious.

ANyway, I'll keep it.

src/InputWithOptions/InputWithOptions.js
hepiyellow: I think we have polyfills loaded by the clients in the `html`. like this:
```
<script src="https://${clientTopology.staticsDomain}/polyfill/v2/polyfill#if(!${debug}).min#{end}.js?features=default,es6,es7,es2017&flags=gated&unknown=polyfill&rum=0"></script>
```

But I changed it anyways, to `indexOf`

src/InputWithOptions/InputWithOptions.js
hepiyellow: Yeah...
I thought to make it a function with semantic meaning, so that classes that inherit from it can override it.
It turned out not to be in use.
ok, I remove it.

src/utils/EventUtils.spec.js
hepiyellow: removed. and add order test.

src/InputWithOptions/InputWithOptions.js
argshook: don't remember exactly, maybe when i had this issue there were no polyfill.io and so now it's ok

src/utils/EventUtils.js
argshook: i don't know. maybe if some handler fails, let it crash? otherwise you'd need some error message, explanation and stuff. people might argue why we don't do defenses everywhere then. but i don't know, we can flip a coin :)

PR #1836, opened 2018-06-18
WIP - POC for integrating wix-ui-backoffice into WSR

hepiyellow:
## exporting `d.ts` files
- [ ] `allowJs` + `decalaration` - not  allowed.  Solution - we don't offer d.ts files to our consumers yet. We could in the future run a separate `tsconfig` for `ts` files only (without allowJs).

## Linting
Currently I did not handle linting.
I think we should first upgrade WSR to Yoshi 3, which has it's own yoshi lint settings,
and only then introduce Typescript to WSR, and see if it works.

### `[...Array(n)]`
- Typescript is not ES6 compatible, it has a [bug with array spread](https://github.com/Microsoft/TypeScript/issues/8856) - so we should not use something like `[...Array(8)]`, since it transpiles to `Array(8).slice()`, which when iterated over `Array(8).slice().map(()=>'foo')` yields array of 8 undefined items, instead of 8 `foo` items.
  - [ ] We should check out solution for better compatibility, maybe  we can transpile to  target `es6`, and then transpile with babel to es6.
- [ ] We should consider adding an eslint (tslint) rule for not using `[...Array()]`
- [ ] In the components code there is no issue. Only in the `e2e` code. Why is that?


hepiyellow:
Failing tests is eyes of `<TextTSComp>`.
This PR seems to be ready.
But...
We should first:
- [ ]  upgrade WSR to yoshi 3
- [ ]  then rebase this PR onto master,
- [ ] Fix linting config for both js and ts

shlomitc:
Closing it for now.
We might have a different approach now for using core components, so moving to typescript is in lower priority at the moment.
I'll keep the branch so we can come back to it when relevant

Issue #1842, opened 2018-06-18
<MultiSelect> with maxNumRows always shows a scrollbar


ChenRoth:
### Summary:

component currently always shows a scrollbar when passed `maxNumRows`, even when not needed

### Priority:
- [ ] Blocker
- [ ] Urgent
- [X] Normal
- [ ] Minor

### Steps to reproduce:
 1. specify any number of `maxNumRows`
 2. see how the scrollbar is always there

### Expected behavior:

show a scrollbar only when exceeding `maxNumRows`

### Relevant Components:

InputWithTags

### Additional notes:

https://wix-wix-style-react.surge.sh/?selectedKind=3.%20Inputs&selectedStory=3.8%20Tags&full=0&addons=0&stories=1&panelRight=0

`Limited num rows` example demonstrates the issue

ChenRoth:
I can take this, I already know how to fix it

shlomitc:
can you add a screenshot? I don't see it on my machine (mac, chrome)

ChenRoth:
![screen shot 2018-06-18 at 18 22 41](https://user-images.githubusercontent.com/36155213/41545776-a4f7e132-7324-11e8-8a22-ba062ecf594a.png)
![screen shot 2018-06-18 at 18 23 29](https://user-images.githubusercontent.com/36155213/41545820-b9acebcc-7324-11e8-872f-d1d4375a3095.png)

ChenRoth:
this is how i think it should be
![screen shot 2018-06-18 at 18 23 58](https://user-images.githubusercontent.com/36155213/41545844-caba7286-7324-11e8-9dcc-fe92df578319.png)
![screen shot 2018-06-18 at 18 24 17](https://user-images.githubusercontent.com/36155213/41545872-d439e5b2-7324-11e8-8b75-f53386d133ae.png)

ChenRoth:
fixed in #1843

PR #1843, opened 2018-06-18
fix bug where <MultiSelect> always shows scrollbar (#1842)

ChenRoth:
### What changed

changed overflow-y to auto and removed a negative margin in .input, so that <InputWithTags> container is lte than its content and would show a scrollbar only when needed

### Why it changed

scrollbar would be visible regardless of component height, rendering it useless.

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green

### DIFF
--- a/src/MultiSelect/InputWithTags.scss
+++ b/src/MultiSelect/InputWithTags.scss
@@ -2,7 +2,6 @@
 
 .input {
   min-width: 30px;
-  margin: -1px;
 }
 
 .emptyInput {
@@ -31,7 +30,7 @@
 }
 
 .hasMaxHeight {
-  overflow-y: scroll;
+  overflow-y: auto;
 }
 
 // Override <Input>'s default focus style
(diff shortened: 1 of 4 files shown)


src/MultiSelect/InputWithTags.scss
```diff
@@ -2,7 +2,6 @@
 
 .input {
   min-width: 30px;
-  margin: -1px;
```
shlomitc: Removing it causes the input field to change from 36px to 38px which is not according to specs
image

ChenRoth: I can either add vertical padding to the InputWithTags container, or alternatively reduce the Input height by 2px...

shlomitc: whatever works

shlomitc:
looks good, I'll do another quick review tomorrow morning and will merge

PR #1847, opened 2018-06-19
<Input />(Input.js) - removed size prop

Faradey27:
### What changed

- remove unused prop

### Why it changed

- because prop was unused

---

- [x] Change is tested
- [x] Change is documented
- [ ] Build is green


ChenRoth:
it's actually been used for InputWithTags, don't remove it

Faradey27:
@ChenRoth but InputWithTags do not accept size prop and inside of Input size prop is not used

ChenRoth:
it can be used, and i'm actually using it in another PR to fix a `<MultiSelect>` bug

Faradey27:
@ChenRoth it used inside of ThemedInput.js and I removed it from Input.js, so i do not see a problem)

Faradey27:
@ChenRoth it was just removed from propTypes and defaultProps inside Input.js, where it not used

ChenRoth:
oh, I thought you were talking about ThemedInput.js, as `<Input>` refers to it inside other components
ok

PR #1849, opened 2018-06-19
<RichTextArea/> - Remove the raw component from the docs and use only composite

Faradey27:
### What changed

<RichTextArea/> - Remove the raw component from the docs and use only composite

### Why it changed

https://github.com/wix-private/wsr-issues/issues/39

---

- [x] Change is tested
- [x] Change is documented
- [ ] Build is green


shlomitc:
We will have a different way to fix these display it.
We will still have the `RichTextArea` component displayed in the docs, but not under the current mapping.
closing for now

PR #1853, opened 2018-06-19
WIP - `<Table>`- API changes

hepiyellow:
### Clean up API:
 - [ ] Remove selectionCounterRenderer
 - [ ] Remove selectionHeader (It will be implemented as a snippet inside the header renderProp)

### Extract BulkSelection inner component
- [ ] manages bulk selection state and logic
- provider of BulkSelection context

### selectedIds
- [ ] switch selections prop to `selectedIds` be an array of selected IDs, instead of boolean array. (assuming data.id exists, if not then rowNum is used instead).
- [ ] DataTable: Put id on the row as a key prop (instead of rowNumber)

### Controlled mode
- [ ] If `selectedIds` is supplied then Table is controlled, inner state is not used.
- [ ] expose an instance method `setSelectedIds`

### onSelectionChange
- [ ]`onSelectionChanged`: (change: {newSelectedIds: string[], isSelectAll?: boolean, isDeselectAll?: boolean}) - if isSelectAll==true then it means that the SelectAllCheckbox was clicked and it should select All. same logic for isDeselectAll.

### renderProp arguments (TableContext)
Arguments for header & footer renderProps:
  * selectedIds: string[]
  * numSelected: number
  * isAnySelected: ()=> bool - helper for doing selectedIds.length > 0
  * setSelectedIds: (string[])=>void
  * selectAll: (bool)=>void - Select/Deselect all (will call onSelectionChanged with the selectAll property)

### Add Table context (React context provider)
Use https://www.npmjs.com/package/create-react-context 
Add context provider inside table which provides the renderProps arguments as described above.
- The `header` & `footer` props can still be of type renderProp or ReactNode.

#### `<Table.HeaderSection>`
New <Table.HeaderSection> sub component is a consumer of Table context.
 - [  ] Has `align` prop:  `start`|`end`

#### `<Table.SelectedStatus/>` - renders "5 Selected". (Is a TableContextConsumer)
- [ ] All it does is apply the correct font styles

```js
  <Table.SelectedStatus>
    {numSelected=> `%{numSelected} Selected`}
  </Table.SelectedStatus>
```
#### Example:
1. We might need to split the API of the view from the Table's logic, so we could place the Table Header inside the Page Header.

```jsx
<Table>
    <Table.Header/>
    <Table.Content/>
</Table>


<Table data={}>
   <Page>
      <Page.Tail>
        <Table.Header>
           {
              (headerProps) => {}
           }
        </Table.Header>
      </Page.Tail>
    <Page.Content>
      <Table.Content/>
    </Page.Content>
   </Page>
</Table>
```

2. `Table.HeaderSection` -> I suggest it should be a different component (layout), not part of the `Table` component. we can call it a TableHeaderLayout. it will be easier to document and separate from the Table concerns.
The implementation of it should still be discussed  (alignment, spacing, etc...)


hepiyellow:
Replaced by diffferent branch name #1854

PR #1854, opened 2018-06-19
WIP - `<Table>`- API changes

hepiyellow:
### Clean up API:
 - [x] Remove selectionCounterRenderer
 - [x] Remove selectionHeader (It will be implemented as a snippet inside the header renderProp)

### Extract BulkSelection inner component
- [x] manages bulk selection state and logic
- provider of BulkSelection context

### selectedIds
- [x] switch selections prop to `selectedIds` be an array of selected IDs, instead of boolean array. (assuming data.id exists, if not then rowNum is used instead).
- [x] DataTable: Put id on the row as a key prop (instead of rowNumber)

### Controlled mode - Need discussion?
- [ ] If `selectedIds` is supplied then Table is controlled, inner state is not used.
- [ ] expose an instance method `setSelectedIds`
- [ ]  should we add also `initiallySelectedIds - to provide initial selected ids in Uncontrolled mode?

### onSelectionChange
- [x] `onSelectionChanged`: (change: {newSelectedIds: string[], isSelectAll?: boolean, isDeselectAll?: boolean}) - if isSelectAll==true then it means that the SelectAllCheckbox was clicked and it should select All. same logic for isDeselectAll.

### renderProp arguments (TableContext)
- [x] Arguments for header & footer renderProps:
  * selectedIds: string[]
  * numSelected: number
  * isAnySelected: ()=> bool - helper for doing selectedIds.length > 0
  * setSelectedIds: (string[])=>void
  * selectAll: (bool)=>void - Select/Deselect all (will call onSelectionChanged with the selectAll property)

### Add Table context (React context provider)
- [x] Use https://www.npmjs.com/package/create-react-context 
Add context provider inside table which provides the renderProps arguments as described above.
- [x] The `header` & `footer` props should be removed. Instead use <Table.Header> and <Table.Footer>

### Sorting API
- [ ] Allow only 1 sorting column
- [ ] might want to implement sort by selected rows

#### Header Position: Fixed: (inner scroll) 
This mode is used only when the Table is in a Page component.
In which case the `<Table.Content>` is placed separately in the `<Page.Content>`.
The table height should appear to fo down all the way down to viewport end.
We may use the `<Card/>`'s `streachVertically` prop.
Last row item bottom divider should appears only if header position is Fixed (with inner scroll)

- [ ] Add `lastRowDividerVisible` prop to `<Table.Content/>`.
- [ ] Add `roundCorners` prop to `<Card/>`.
- [ ] Add snippet example for usage within a Page.

#### `<Table.HeaderSection>`
New <Table.HeaderSection> sub component is a consumer of Table context.
 - [  ] Has `align` prop:  `start`|`end`

#### `<Table.SelectedStatus/>` - renders "5 Selected". (Is a TableContextConsumer)
- [ ] All it does is apply the correct font styles

```js
  <Table.SelectedStatus>
    {numSelected=> `%{numSelected} Selected`}
  </Table.SelectedStatus>
```
#### Example:
1. We might need to split the API of the view from the Table's logic, so we could place the Table Header inside the Page Header.

```jsx
<Table>
    <Table.Header/>
    <Table.Content/>
</Table>


<Table data={}>
   <Page>
      <Page.Tail>
        <Table.Header>
           {
              (headerProps) => {}
           }
        </Table.Header>
      </Page.Tail>
    <Page.Content>
      <Table.Content/>
    </Page.Content>
   </Page>
</Table>
```

2. `Table.HeaderSection` -> I suggest it should be a different component (layout), not part of the `Table` component. we can call it a TableHeaderLayout. it will be easier to document and separate from the Table concerns.
The implementation of it should still be discussed  (alignment, spacing, etc...)


stories/Table/Toolbar/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+| propName | propType | defaultValue | isRequired | description |
+|----------|----------|--------------|------------|-------------|
+| layout | 'button' | - | false | applys appropriate css styles like margins. There might be more layout values in the future. |
+| children | node | - | true | any node. If it is a `<Button/>` then you should set `layout` prop to `'button'` to get the desired look |
+
+## Styling guidelines
+
+- `<Button/>` components  should have `theme="whiteblueprimary"`
+- `<Dropdown/> : should have rounded corners (add `roundInput` prop). Same with other types of Input components.
```
shlomitc: no closing "\`"

stories/Table/Toolbar/README.md
```diff
@@ -0,0 +1,97 @@
+# `<TableToolbar/>`
```
shlomitc: in general, I would split it into tabs - one for the APIs and one for the example. it's a really big doc.
the tl;dr should be the component examples.

stories/Table/Toolbar/ToolbarStory.js
```diff
@@ -0,0 +1,64 @@
...
+        <div className={s.example}>
+          <CodeExample title="Typical BulkActions Toolbar" code={BulkActionsExampleRaw}>
+            <Card>
+              <BulkActionsExample/>
+            </Card>
+          </CodeExample>
+        </div>
+        <Markdown
+          source={`
+#### Important Note about Search!
```
shlomitc: remove, too detailed

stories/Table/Toolbar/README.md
```diff
@@ -0,0 +1,97 @@
...
+### Properties
+
+| propName | propType | defaultValue | isRequired | description |
+|----------|----------|--------------|------------|-------------|
+| position | 'start' &#124; 'end' | 'start' | false | Typically, in an Left-To-Right layout, a `start` group (and all it's items) would be aligned to the left, and an 'end' group (and all it's items)would be aligned to the right |
+| children | node | - | true | `<Item/>` or `<Divider/>` elements |
+
+## `<Item/>`
+
+The layout of a toolbar is composed of item groups (`<ItemGroups/>` ).
```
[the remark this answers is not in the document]
hepiyellow: what do you mean?

Issue #832, opened 2017-09-17
Tooltip won't work on a disabled button


shahafan:
### Summary:
When trying to add a tooltip to a disabled button it doesn't work.

    <Tooltip textAlign="left" placement="left" alignment="center" content="test" maxWidth="250px">
        <Button theme="fullblue" onClick={() => console.log('clicked')} disabled={true}>
            Text
       </Button>
    </Tooltip>

### Priority:
- [x] Normal

### Expected behavior:
tooltip should be displayed when hovering

### Relevant Components:
Tooltip , Button

lbelinsk:
@amiryonatan could you please check that in the new planned Tooltip we don't have this issue. If so, you can close this one

Faradey27:
@lbelinsk looks like the problem is related to Button and not to Tooltip

https://github.com/wix/wix-style-react/blob/master/src/ButtonLayout/ButtonLayout.scss#L823 - root of problem, i am working on fix

Faradey27:
https://github.com/wix/wix-style-react/pull/1855

PR #1855, opened 2018-06-19
<Button/> - Tooltip doesn't work on a disabled button

Faradey27:
### What changed

<Button/> - Tooltip doesn't work on a disabled button

### Why it changed

https://github.com/wix/wix-style-react/issues/832

---

- [x] Change is tested
- [x] Change is documented
- [ ] Build is green


Faradey27:
@lbelinsk updated with e2e and unit tests

src/Backoffice/Button/Button.driver.js
```diff
@@ -5,6 +5,10 @@ import ReactTestUtils from 'react-dom/test-utils';
 export default ({element, wrapper, component}) => ({
   exists: () => !!element,
   click: () => ReactTestUtils.Simulate.click(element),
+  focusableOnFocus: () => ReactTestUtils.Simulate.focus(element),
```
shlomitc: maybe just `focus`?

src/Backoffice/Button/Button.driver.js
```diff
@@ -5,6 +5,10 @@ import ReactTestUtils from 'react-dom/test-utils';
...
+  focusableOnBlur: () => ReactTestUtils.Simulate.blur(element),
```
shlomitc: `blur`?

src/Backoffice/Button/Button.driver.js
```diff
@@ -5,6 +5,10 @@ import ReactTestUtils from 'react-dom/test-utils';
...
+  onMouseEnter: () => ReactTestUtils.Simulate.mouseEnter(element),
```
shlomitc: `mouseEnter`

src/Backoffice/Button/Button.driver.js
```diff
@@ -5,6 +5,10 @@ import ReactTestUtils from 'react-dom/test-utils';
...
+  onMouseLeave: () => ReactTestUtils.Simulate.mouseLeave(element),
```
shlomitc: `mouseLeave`

src/Backoffice/Button/Button.js
```diff
@@ -63,13 +63,13 @@ class Button extends WixComponent {
...
         <button
-          onClick={onClick}
+          onClick={disabled ? null : onClick}
           disabled={disabled}
           type={type}
           onMouseEnter={onMouseEnter}
           onMouseLeave={onMouseLeave}
-          onFocus={this.props.focusableOnFocus} // For some reason eslint react/prop-types rule doesn't work here ?!#$
-          onBlur={this.props.focusableOnBlur}
+          onFocus={disabled ? null : this.props.focusableOnFocus} // For some reason eslint react/prop-types rule doesn't work here ?!#$
```
shlomitc: is this being currently called even when the button is disabled?

src/Tooltip/Tooltip.e2e.js
```diff
@@ -11,6 +11,16 @@ describe('Tooltip', () => {
     browser.get(storyUrl);
   });
 
+  eyes.it('should show tooltip when button disabled', async () => {
+    browser.executeScript('window.setComponentState({disabledButtonWithTooltip: true})');
```
shlomitc: @argshook - have a better suggestion for that?

stories/Tooltip/Composite/TooltipForEyesOnly.js
```diff
@@ -15,28 +15,58 @@ export default class TooltipForEyesOnly extends Component {
     super();
     this.state = {
       content: SHORT_CONTENT,
+      disabledButtonWithTooltip: false,
```
shlomitc: this is more of a button issue than tooltip, so maybe it should be actually a button events issue. I think that testing it in a unit-test is enough

src/Backoffice/Button/Button.driver.js
Faradey27: in component api it called focusableOnFocus, so i decided to left same)
@shlomitc

src/Backoffice/Button/Button.driver.js
Faradey27: in component api it called focusableOnBlur, so i decided to left same)
@shlomitc

src/Backoffice/Button/Button.driver.js
Faradey27: in component api it called onMouseEnter, so i decided to left same)
@shlomitc

src/Backoffice/Button/Button.driver.js
Faradey27: in component api it called mouseLeave, so i decided to left same)
@shlomitc

src/Backoffice/Button/Button.driver.js
shlomitc: the driver should be an abstraction for the functionality, meaning the consumer will want to trigger a focus event on the component, so it will make it more declarative to call it `focus`

stories/Tooltip/Composite/TooltipForEyesOnly.js
Faradey27: @shlomitc agree about button related

but not agree about e2e test, because main change was done in css, by removing pointer-events: none, and such case could not be handled by jsdom(currently)

src/Backoffice/Button/Button.js
Faradey27: no, it was not

Faradey27:
@shlomitc could you approve eyes and merge?

src/Backoffice/Button/Button.js
shlomitc: so why do you need to add the check for the disabled state? doesn't the native button disables it?

src/Backoffice/Button/Button.protractor.driver.js
```diff
@@ -10,6 +11,10 @@ const buttonDriverFactory = element => {
 
   const publicDriver = {
     click: () => element.click(),
+    hover: element => mouseEnter(element),
+    tooltipElement: () => $('[data-hook="tooltip-e2e-wrapper"]'),
```
shlomitc: please don't! this is not part of the Button component, I really don't want to have such functionality in the driver.
The driver should be a simple abstraction to let consumer interact with the component in tests

same for the next functions related to the tooltip

stories/Button/standard.story.js
```diff
@@ -24,5 +28,11 @@ export default {
     onMouseLeave: () => 'Mouse Leave!',
     prefixIcon: icons,
     suffixIcon: icons
-  }
+  },
+
+  examples: (
+    <CodeExample title="Disabled button with tooltip" code={ExampleDisabledButtonWithTooltipRaw}>
+      <ExampleDisabledButtonWithTooltip/>
```
shlomitc: @argshook - What do you think about this kind of visual edge cases E2E tests?

The fix is very small and is CSS solution so It does require a browser test.

In my opinion, testing the hover functionality (even without a tooltip) will be enough but anyway, this means we pollute our API page with test-only examples.

I'll be happy to hear another voice in this issue

Faradey27:
@shlomitc i removed e2e test

src/Backoffice/Button/Button.js
shlomitc: I mean that when the native `button` gets `disabled` prop, it automatically stops triggering the focus and blur events.
for example (see the console logs):
http://jsbin.com/toqapigoga/1/edit?html,js,console,output

stories/Button/standard.story.js
argshook: i suppose such examples shouldn't be visible to end user and so we need to hide them. currently we don't have mechanism for that specificly

some thoughts
1. create separate story page for e2e only examples and somehow hide it (skimmed through storybook api, didn't see a way to hide story from sidebar, but im pretty sure it's possible, need to dig more)
2. improve wix-storybook-utils to support something like `examplesE2e` or similar to live nearby `examples`. These would be displayed only in e2e (and so current `withExamples` in url wouldn't be needed anymore)

src/Backoffice/Button/Button.js
Faradey27: yea, you are completely right

Faradey27:
@shlomitc updated

Issue #1762, opened 2018-05-24
DropdownLayout - pressing spacebar should select the highlighted option


tzvikaf:
need to take into consideration the other components that use DropdownLayout
(maybe all components with prop closeOnSelect=false should allow space to choose - just a thought).

for input fields, it should delegate the spacebar click to the input.

shlomitc:
should be fixed now. closing

PR #1860, opened 2018-06-20
DropdownLayout - pressing spacebar should select the highlighted opti…

romabelka:
…on (#1762)

### What changed
Pressing Selectbar selects highlighted options, disabled for InputWithOtions.
...

### Why it changed

...

---

- [x] Change is tested
- [ ] Change is documented
- [ ] Build is green

### DIFF
--- a/src/InputWithOptions/InputWithOptions.js
+++ b/src/InputWithOptions/InputWithOptions.js
@@ -5,6 +5,7 @@ import Input from '../Input';
 import omit from 'omit';
 import DropdownLayout from '../DropdownLayout/DropdownLayout';
 import Highlighter from '../Highlighter/Highlighter';
+import {chainEventHandlers} from '../utils/ChainEventHandlers';
 
 class InputWithOptions extends WixComponent {
 
@@ -13,6 +14,12 @@ class InputWithOptions extends WixComponent {
   dropdownClasses() {}
   dropdownAdditionalProps() {}
   inputAdditionalProps() {}
+  /**
+   * An array of key codes (default ['Enter','Tab']) that act as manual submit. Will be used with the onKeyDown(event), tha key codes are values of event.key. When a manual submit key is pressed then onManuallyInput will be called.
+   */
+  getManualSubmitKeys() {
+    return ['Enter', 'Tab'];
+  }
 
   constructor(props) {
     super(props);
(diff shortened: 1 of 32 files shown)


hepiyellow:
Doesn't seem like it is working for Dropdown:
https://wix-wix-style-react-pr-1860.surge.sh/?selectedKind=4.%20Selection&selectedStory=4.1%20Dropdown&full=0&addons=0&stories=1&panelRight=0

src/InputWithOptions/InputWithOptions.js
```diff
@@ -217,7 +222,7 @@ class InputWithOptions extends WixComponent {
   }
 
   _onKeyDown(event) {
-    if (this.props.disabled) {
+    if (this.props.disabled || !this._shouldDelegateEvent(event)) {
```
hepiyellow: The `_shouldDelegateEvent` check should only affect if `dropdownLayout._onKeyDown` is called.
If you dod it here, the we won't be updating the `isEditing` state, and we won't be doing `this.showOptions();`.

src/InputWithOptions/InputWithOptions.js
```diff
@@ -236,6 +241,10 @@ class InputWithOptions extends WixComponent {
     }
   }
 
+  _shouldDelegateEvent(event) {
```
hepiyellow: Please inline this method into `_onKeyDown` since it is relevant only to keyDown event.

Anyways, the name `_shouldDelegateEvent` would be better as `_shouldDelegatedKeyDownEvent`, but if you inline it then you don't need a name for it (=:

src/InputWithOptions/InputWithOptions.spec.js
```diff
@@ -368,6 +368,35 @@ const runInputWithOptionsTest = driverFactory => {
...
+
+      it('should not call onSelect on space key press', () => {
+        const onSelect = jest.fn();
+        const {driver} = createDriver(<InputWithOptions options={options} onSelect={onSelect}/>);
+        driver.focus();
+        driver.pressDownKey();
+        driver.pressSpaceKey();
+        expect(onSelect).not.toHaveBeenCalled();
+      });
+    });
```
hepiyellow: Please add test for pressing `Spacebar` when `readOnly` is true.

src/InputWithOptions/InputWithOptions.js
romabelka: As far as I understood we should not call `dropdownLayout._onKeyDown` if we don't want to delegate, am I wrong?

hepiyellow:
REpalced by #1885 and merged.

PR #1862, opened 2018-06-20
WIP - `<Table/>` Add react context and Compound Components

hepiyellow:
Add React context.
Add Compound components: Header, Footer, TitleBar, Content.
This allows splitting the rendering of the Table parts (required in order to render in a Page component).

- By default the `<Table>` renders a `<DataTable>` with header (aka TitleBar).
- The `<Table.Content>` is rendering the `<DataTable>`
- You can render the title bar separately by doing this:
```
<Table>
  <Table.TitleBar/>
  <Table.Content titleBarVisible={false}/>
</Table>
```


src/Table/BulkSelection/BulkSelection.js
```diff
@@ -42,6 +45,7 @@ export class BulkSelection extends React.Component {
     return isEqual(props.selections, state.selections) ? null : {selections: props.selections.slice()};
   }
 
+  componentWillReceiveProps
```
shlomitc: bug

src/Table/Table.js
```diff
@@ -28,82 +33,166 @@ export default class Table extends WixComponent {
...
-    return (
-      <div className={s.header} data-hook="table-header">
-        {typeof header === 'function' ? header(bulkSelectionContext) : header}
-      </div>);
+  return tableProps.showSelection ? [createCheckboxColumn(bulkSelectionContext), ...tableProps.columns] : tableProps.columns;
+}
+
+const TableHeader = props => {
+  return (
+    <div data-hook="table-header">
```
shlomitc: shouldn't it be more internal as a wrapper of the {props.children}?

hepiyellow: No. The child of `<BulkSelectionConsumer>` must be function.
The `<BulkSelectionConsumer>` is only a logical component, it does not render anything to the DOM. So practically this `<div>` is a direct wrapper of the `{props.children}`.

PR #1866, opened 2018-06-21
WIP - <Table/> -[1] switch from selection boolean array to  selectedIds

hepiyellow:
expect the `data` to include an `id` property in each row's data.
If it doesn't exist then use the rowIndex as id.

The selection state, holds selectedIds instead of an array or booleans:
So instead of `[true,false]`, now this selection is represented with `[0]`.

### DIFF
--- a/src/Table/BulkSelection/BulkSelection.js
+++ b/src/Table/BulkSelection/BulkSelection.js
@@ -1,7 +1,6 @@
 import React from 'react';
 import {string, number, arrayOf, oneOfType, func, any} from 'prop-types';
 import isEqual from 'lodash/isEqual';
-import without from 'lodash/without';
 import createReactContext from 'create-react-context';
 export const BulkSelectionContext = createReactContext();
 
@@ -55,9 +54,9 @@ export class BulkSelection extends React.Component {
 
   toggleAll = enable => {
     if (enable) {
-      this.setSelectedIds({selectedIds: this.props.allIds});
+      this.setSelectedIds(this.props.allIds);
     } else {
-      this.setSelectedIds({selectedIds: []});
+      this.setSelectedIds([]);
     }
   }
 
@@ -77,16 +76,17 @@ export class BulkSelection extends React.Component {
   }
 
   toggleSelectionById = id => {
-    let newSelectedIds;
-    if (this.isSelected(id)) {
-      newSelectedIds = without(this.state.selectedIds, id);
-    } else {
-      newSelectedIds = [...this.state.selectedIds, id];
-    }
-    this.setSelectedIds({selectedIds: newSelectedIds});
+    this.setSelectedIds(
+      this.isSelected(id) ?
+      this.state.selectedIds.filter(_id => _id !== id) :
+      this.state.selectedIds.concat(id)
+    );
   }
 
-  setSelectedIds = ({selectedIds}) => {
+  setSelectedIds = selectedIds => {
+    if (!Array.isArray(selectedIds)) {
+      throw new Error('selectedIds must be an array');
+    }
     this.setState({selectedIds}, () => {
       this.props.onSelectionChanged && this.props.onSelectionChanged(selectedIds.slice());
     });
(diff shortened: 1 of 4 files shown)


src/Table/BulkSelection/BulkSelection.js
```diff
@@ -1,8 +1,9 @@
 import React from 'react';
-import PropTypes from 'prop-types';
+import {string, number, arrayOf, oneOfType, func, any} from 'prop-types';
 import isEqual from 'lodash/isEqual';
+import without from 'lodash/without';
+import defaultTo from 'lodash/defaultTo';
```
argshook: it's just an or, i'd really prefer not to rely on lodash where we don't need

src/Table/BulkSelection/BulkSelection.js
```diff
@@ -33,87 +24,109 @@ export class BulkSelection extends React.Component {
...
-      isSelected: index => !!this.state.selections[index],
-      toggleItem: this.toggleItem,
-      toggleAll: this.toggleAll,
-      setSelection: this.setSelection,
-      toggleBulkSelection: this.toggleBulkSelection,
-      getBulkSelectionState: () => getNextBulkSelectionState(this.state.selections),
-      getNumSelected: () => getSelectionsCount(this.state.selections),
-      isAnySelected: () => getSelectionsCount(this.state.selections) > 0
-    };
+  setSelectedIds = ({selectedIds}) => {
```
argshook: unnecesarry destructuring, i'd just pass `selectedIds`, not an object

src/Table/BulkSelection/BulkSelection.js
```diff
@@ -33,87 +24,109 @@ export class BulkSelection extends React.Component {
...
   }
 
-  toggleItem = index => {
-    const selections = this.state.selections.slice();
-    selections[index] = !selections[index];
-    this.setSelection({selections});
+  toggleSelectionById = id => {
+    let newSelectedIds;
+    if (this.isSelected(id)) {
+      newSelectedIds = without(this.state.selectedIds, id);
```
argshook: just `filter` it

src/Table/BulkSelection/BulkSelection.js
```diff
@@ -33,87 +24,109 @@ export class BulkSelection extends React.Component {
...
+    } else {
+      newSelectedIds = [...this.state.selectedIds, id];
+    }
+    this.setSelectedIds({selectedIds: newSelectedIds});
```
argshook:
```
this.setSelectedIds(
  this.isSelectedId(id) ?
    this.state.selectedIds.filter(i => i !== id) :
    this.state.selectedIds.concat(id)
)
```

src/Table/Table.spec.js
```diff
@@ -53,45 +54,69 @@ describe('Table', () => {
...
-      const selections = [false, false];
-      const {driver, wrapper} = createEnzymeDriver(<Table {...defaultProps} selections={selections}/>);
-      selections[0] = true;
-      wrapper.setProps({selections});
+    it('should select rows according to selectedIds prop given numeric ids', () => {
+      const ID_1 = 1234, ID_2 = 1235;
+      const driver = createDriver(
+        <Table
+          {...defaultProps}
+          data={[{id: ID_1, a: 'value 1', b: 'value 2'}, {id: ID_2, a: 'value 3', b: 'value 4'}]}
```
argshook: can't `data` include objects with `selected` property and derive `selectedIds` from there? i don't think it's a good idea to split data and selectedIds, cause consumer would need to handle two things.

i might be missing something and it's hard to implement it?

src/Table/BulkSelection/BulkSelection.js
hepiyellow: ok

src/Table/BulkSelection/BulkSelection.js
hepiyellow: thanks. added a test for it.

src/Table/BulkSelection/BulkSelection.js
hepiyellow: ok. Although i think it is more readable to use `without`

src/Table/BulkSelection/BulkSelection.js
hepiyellow: done. thanks.

src/Table/Table.spec.js
hepiyellow: OMG: I wrote you 20 lines here and pressed "Cancel" by mistake )-:

PR #1867, opened 2018-06-21
<Table/> - [2] Remove BulkSelectionConsumer when showSelection is false

hepiyellow:
When there is no selection column, then we don't need the BulkSelection consumer.


src/Table/Table.js
```diff
@@ -16,7 +16,9 @@ const TableDefaultProps = {
 
 export const TableContext = createReactContext(TableDefaultProps);
 
-function createColumns({tableProps, bulkSelectionContext}) {
+function createColumns({tableProps,
```
lbelinsk: why the extra enter?

PR #1871, opened 2018-06-21
<Icons/> - use new set of icons:

Faradey27:
### What changed

 - add migration script
 - add proxy wix-ui-icons-common to wsr script
 - update package.json with new build command and new bin command
 - update wix-ui-icons-common to v2
 - minor component fixes

### Why it changed

- as part of the migration to new icons

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green


package.json
```diff
@@ -19,12 +22,13 @@
     "pretest:e2e": "./scripts/ensure-dist.sh",
     "test:e2e": "concurrently 'npm run test:e2e-only' 'node scripts/e2e-server.js' --names 'protractor,e2e-server' --success first --kill-others",
     "test:e2e-only": "NODE_ENV=production yoshi test --protractor",
-    "build": "npm run lint && svg2react-icon && yoshi build && build-storybook && npm run export-components",
+    "build": "npm run lint && npm run proxy-wix-ui-icons && svg2react-icon && yoshi build && build-storybook && npm run export-components",
```
shlomitc: as we don't expose the new icons yet, I think we can ditch this step for now

Faradey27: but without this line, we can not do any migration, because new-icons folder will not appear

we did not document new icons anywhere and we added all proxy and migration scripts, I think it is not a bad idea to leave this, in general, such line will appear in any case "pull request", so why not to test it before)

but if you think, that it is really important to remove it, I will, as it not so critical

@shlomitc

shlomitc: ok, I see it is necessary because your script will migrate also the internal components of WSR to use the `new-icons` folder.

You could import internally for WSR directly from `wix-ui-icons-common` without copying them, however, this will be an issue in the case where you import the entire `index` file.

package.json
```diff
@@ -19,12 +22,13 @@
...
     "pr-postbuild": "npx teamcity-surge-autorelease@^1.0.0 --dist=storybook-static",
     "storybook": "start-storybook -p 6006",
     "postpublish": "npx teamcity-surge-autorelease@^1.0.0 --dist=storybook-static",
     "export-components": "import-path --path src",
-    "lint": "yoshi lint"
+    "lint": "yoshi lint",
+    "proxy-wix-ui-icons": "node scripts/proxy-wix-ui-icons.js"
```
shlomitc: maybe instead of running this script it in the `build`, you can run it in `postinstall`, as these icons will not change between installation.

wdyt?

package.json
```diff
@@ -124,7 +129,7 @@
     "slate": "^0.20.3",
     "wix-animations": "^1.0.141",
     "wix-ui-backoffice": "1.0.260",
-    "wix-ui-icons-common": "1.0.39",
+    "wix-ui-icons-common": "2.0.0",
```
shlomitc: I think 2.0.2 is already released

scripts/proxy-wix-ui-icons.js
```diff
@@ -0,0 +1,92 @@
...
+  const content = `/*eslint-disable*/Object.defineProperty(exports, '__esModule', {value: true});\n${indexFileContent}function _interopRequireDefault(obj) {return obj && obj.__esModule ? obj : {default: obj};}`;
+  fs.writeFileSync(`${iconsDir}/index.js`, content);
+};
+
+const copySync = (src, dist) => {
+  shell(`mkdir -p ${dist}`);
+  shell(`cp -r ${src}/* ${dist}`);
+};
+
+const copyIconsToSrcForBackcompability = iconsDir => {
```
shlomitc: isn't this the new icons folder? why do you call it Backcompability?

package.json
Faradey27: @shlomitc we talked about this a month ago, and decided that better to proxy wsr and to use icons from wsr directly

scripts/proxy-wix-ui-icons.js
Faradey27: because old icons inside wsr/src/icons and new icons should be inside wsr/icons, but wsr/icons will not work with current storybook solution, so i need also to copy them to swr/src/icons

Faradey27: @shlomitc

Faradey27:
@shlomitc updated

scripts/migrate-to-icons-v2/transforms.js
```diff
@@ -0,0 +1,137 @@
...
+  transformWSRComponents({node, newIconName, oldIconName: iconName});
+};
+
+const updateImports = ({root, j, file, onError, onTick}) => {
+  const iconNames = [];
+  const imports = root.find(j.ImportDeclaration);
+  imports
+    .forEach(node => {
+      const iconName = getOldIconName(node);
+      const importedIconNames = (getListOfImportedIcons(node) || []);
```
Faradey27: namedImportIcons

scripts/migrate-to-icons-v2/transforms.js
```diff
@@ -0,0 +1,137 @@
...
+      ? value.replace(pathes.commonSrcPath, `/${ICONS_FOLDER}/${newIconName}`)
+      : `wix-style-react/${ICONS_FOLDER}/${newIconName}`;
+  } else if (value.endsWith(pathes.commonPath)) {
+    node.value.source.value = isWSRMigration
+      ? value.replace(pathes.commonPath, `/../${ICONS_FOLDER}/${newIconName}`)
+      : `wix-style-react/${ICONS_FOLDER}/${newIconName}`;
+  }
+};
+
+const logIconSpreadImportMigrationStatus = ({file, node, importedIconNames, onTick, onError}) => {
```
Faradey27: rename to 'not log'

scripts/migrate-to-icons-v2/transforms.js
```diff
@@ -0,0 +1,137 @@
...
+  const imports = root.find(j.ImportDeclaration);
+  imports
+    .forEach(node => {
+      const iconName = getOldIconName(node);
+      const importedIconNames = (getListOfImportedIcons(node) || []);
+      if (iconName) {
+        importedIconNames.push({value: iconName});
+      }
+      if (importedIconNames[0] && importedIconNames[0].type !== 'all') {
+        logIconSpreadImportMigrationStatus({file, node, importedIconNames, onTick, onError});
```
Faradey27: named import

scripts/migrate-to-icons-v2/transforms.js
```diff
@@ -0,0 +1,137 @@
...
+        onTick(logObj);
+      } else {
+        onError(logObj);
+      }
+    });
+};
+
+const transformSpreadImports = ({iconNames, importedIconNames, iconName, node}) => {
+  const newIconName = getNewIconName(iconName);
+  iconNames.push(...importedIconNames.map(icon => icon.value));
```
Faradey27: oldIconNames

PR #1877, opened 2018-06-22
<RichTextArea/> - fix Placeholder doesn’t update when props change

Faradey27:
### What changed

Placeholder doesn’t work

### Why it changed

https://github.com/wix-private/wsr-issues/issues/34

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green

### DIFF
--- a/src/RichTextArea/RichTextArea.js
+++ b/src/RichTextArea/RichTextArea.js
@@ -122,7 +122,9 @@ class RichTextArea extends WixComponent {
   }
 
   componentWillReceiveProps(props) {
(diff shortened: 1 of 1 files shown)


hepiyellow:
If I see a PR with "Files Changed 1" then I know I can automatically request changes and say: "missing test"

src/RichTextArea/RichTextArea.js
```diff
@@ -122,7 +122,9 @@ class RichTextArea extends WixComponent {
   }
 
   componentWillReceiveProps(props) {
```
hepiyellow: Would be nice to rename `props` to `nextProps`

PR #1878, opened 2018-06-24
`<DataTable/>` - (optinization) Add data.id as React key

hepiyellow:
Applying  data.id as row's React key, helps cases where 
the data is replaced. Before, the keys would stay the same, and now keys are different so React doesn't need to check the rest of the row structure.

This change is coming because with the new `<Table/>` component we add selection by id, so we thought it'd be better if the row keys are by id as well.


src/DataTable/DataTable.driver.js
```diff
@@ -37,6 +37,7 @@ const dataTableDriverFactory = ({element, wrapper, component}) => {
     getRowWithDataHook: dataHookName => element.querySelector(`[data-hook="${dataHookName}"]`),
     /** Returns an array representing the text content of the cells in a given row `index`.  */
     getRowText: index => values(getRows()[index].querySelectorAll('td')).map(td => td.textContent),
+    getRowKey: index => getRow(index).getAttribute('key'),
```
lbelinsk: document? or is it only for internal usages anyway?

src/DataTable/DataTable.spec.js
```diff
@@ -156,6 +157,25 @@ describe('Table', () => {
     expect(driver.getRowsWithDataHook(rowDataHook).length).toBe(defaultProps.data.length);
   });
 
+  describe('row keys', () => {
+    it('should assign data.id as row keys', () => {
+      const props = createDefaultProps();
+      props.data[0].id = '000';
+      props.data[1].id = '111';
+      const wrapper = mount(<DataTable {...props}/>);
+      const getRowKey = (wrapper, index) => wrapper.find('tbody tr[data-table-row="dataTableRow"]').at(index).key();
```
lbelinsk: You added a method in the driver, so maybe you just forgot to remove it there?

src/DataTable/DataTable.spec.js
```diff
@@ -156,6 +157,25 @@ describe('Table', () => {
...
+      expect(getRowKey(wrapper, 0)).toBe('000');
+      expect(getRowKey(wrapper, 1)).toBe('111');
+    });
+
+    it('should use rowIndex as keys', () => {
+      const wrapper = mount(<DataTable {...createDefaultProps()}/>);
+      const getRowKey = (wrapper, index) => wrapper.find('tbody tr[data-table-row="dataTableRow"]').at(index).key();
```
lbelinsk: Duplication of this function here

src/DataTable/DataTable.driver.js
hepiyellow: Oh, I should remove it. I tested this inline in the spec. I don't think consumers need this.

Issue #1664, opened 2018-04-10
Input component is missing a counter


Jefoss:
### Summary:
The InputArea component has the option to display a counter (current length / max length) 

### Priority:
- [ ] Blocker
- [ ] Urgent
- [ ] Normal
- [x] Minor

### Expected behavior:
The Input component should have a similar counter prop like the InputArea has

### Relevant Components:
Input, InputArea

milkyfruit:
Text Field (Input),
Text Area (incl. Rich Text Area),
Text Field With Selection

Has character counter feature by design.
![image](https://user-images.githubusercontent.com/1684587/38920291-986aa43a-42fb-11e8-863f-133a3985cf7d.png)

[Indexed Specs.](zpl.io/anRW6Ja)
[behaviour specs](zpl.io/2Egjola)

shlomitc:
Hey, we (Arijus) lately added the `FormField` component which is a replacement for the `TextField` composition.
It allows you to compose any type of component (not only input) and also gives you the ability to display the counter, as requested.

See how it works here:
https://wix-wix-style-react.surge.sh/?selectedKind=Components&selectedStory=FormField&full=0&addons=0&stories=1&panelRight=0

Choose `Input with char counter` from the `children*` list and type something

![image](https://user-images.githubusercontent.com/6093192/42664915-04ad5d4e-8646-11e8-8054-620bc57d5458.png)

Can you please try it out and tell us if it fit your needs? if so, we will declare this component as the right solution for compositions.

mozesoron:
@shlomitc it seems fine but I can't see the full example of how the counter is added. need to fix the story.

shlomitc:
Take a look in the top section in that page, you will find a code example for the counter

mozesoron:
@shlomitc design wise looks fine. what I mean is that I cant see the code in the story.

shlomitc:
![d4af6236-57cc-432f-8920-228545fcd0b5](https://user-images.githubusercontent.com/6093192/42733263-0e814180-8837-11e8-9692-313133c57b85.png)

shlomitc:
No further comments so I’m closing this one. If any issue is raised, please reopen

PR #1879, opened 2018-06-24
Input: add counter same as text area #185

mozesoron:
### What changed
added new counter same as text area
...

### Why it changed
#185
...

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green

### DIFF
--- a/src/Input/Input.driver.js
+++ b/src/Input/Input.driver.js
@@ -7,6 +7,7 @@ const inputDriverFactory = ({element, wrapper, component}) => {
   const input = element && element.querySelector('input');
   const clearButton = element && element.querySelector(`.${styles.clearButton}`);
   const suffixNode = element && element.querySelector(`.${styles.suffix}`);
+  const counterSelector = '[data-hook="input-counter"]';
 
   const driver = {
     trigger: (trigger, event) => ReactTestUtils.Simulate[trigger](input, event),
@@ -76,7 +77,9 @@ const inputDriverFactory = ({element, wrapper, component}) => {
     setProps: props => {
       const ClonedWithProps = React.cloneElement(component, Object.assign({}, component.props, props), ...(component.props.children || []));
       ReactDOM.render(<div ref={r => element = r}>{ClonedWithProps}</div>, wrapper);
-    }
+    },
+    getHasCounter: () => !!element.querySelectorAll(counterSelector).length,
+    getCounterValue: () => element.querySelector(counterSelector).textContent
   };
 
   return driver;
--- a/src/Input/Input.js
+++ b/src/Input/Input.js
@@ -17,7 +17,8 @@ class Input extends Component {
   static Group = Group;
 
   state = {
-    focus: false
+    focus: false,
+    counter: (this.props.value || this.props.defaultValue || '').length
   };
 
   componentDidMount() {
@@ -74,7 +75,8 @@ class Input extends Component {
       tooltipPlacement,
       onTooltipShow,
       autocomplete,
-      required
+      required,
+      hasCounter
     } = this.props;
 
     const onIconClicked = () => {
@@ -137,6 +139,7 @@ class Input extends Component {
     //https://github.com/wix/wix-style-react/issues/1691
     return (<div className={styles.inputWrapper}>
       {prefix && <div className={styles.prefixSuffixWrapper}><div className={styles.prefix}>{prefix}</div></div>}
+      {hasCounter && maxLength && <span className={styles.counter} data-hook="input-counter">{this.state.counter}/{maxLength}</span>}
 
       { inputElement }
       { visibleSuffixCount > 0 && <div className={styles.prefixSuffixWrapper}><InputSuffix
@@ -214,7 +217,7 @@ class Input extends Component {
     if (this.props.type === 'number' && !(/^[\d.,\-+]*$/.test(e.target.value))) {
       return;
     }
-
+    this.props.hasCounter && this.setState({counter: e.target.value.length});
     this.props.onChange && this.props.onChange(e);
   }
 
@@ -393,6 +396,8 @@ Input.propTypes = {
   tooltipPlacement: PropTypes.string,
   type: PropTypes.string,
   unit: PropTypes.string,
+  /** When true a letters counter will appear */
+  hasCounter: PropTypes.bool,
 
   /** Inputs value */
   value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
--- a/src/Input/Input.scss
+++ b/src/Input/Input.scss
@@ -15,6 +15,16 @@ $material-gap: 8px;
   @include FontRoman();
 }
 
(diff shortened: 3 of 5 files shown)


src/Input/Input.driver.js
```diff
@@ -76,7 +77,9 @@ const inputDriverFactory = ({element, wrapper, component}) => {
     setProps: props => {
       const ClonedWithProps = React.cloneElement(component, Object.assign({}, component.props, props), ...(component.props.children || []));
       ReactDOM.render(<div ref={r => element = r}>{ClonedWithProps}</div>, wrapper);
-    }
+    },
+    getHasCounter: () => !!element.querySelectorAll(counterSelector).length,
```
lbelinsk: Can you please just rename it to `hasCounter`. It is a little weird like that :)

src/Input/Input.spec.js
```diff
@@ -602,6 +602,18 @@ describe('Input', () => {
...
     });
   });
+
+  describe('counter', () => {
+    it('should show correct value when hasCounter and maxLength present', () => {
+      const driver = createDriver(<Input hasCounter maxLength={30} value={'abc'}/>);
+      expect(driver.getCounterValue()).toEqual('3/30');
+    });
+
+    it('should not show counter when hasCounter is not present', () => {
```
lbelinsk: I would also test when only maxLength is not present, and hasCounter is present (division by zero that you handled in the code :))

src/Input/Input.js
```diff
@@ -393,6 +396,8 @@ Input.propTypes = {
   tooltipPlacement: PropTypes.string,
   type: PropTypes.string,
   unit: PropTypes.string,
+  /** When true a letters counter will appear */
+  hasCounter: PropTypes.bool,
```
lbelinsk: I think a more react prop fitting name can be nicer here. I just can't think of one. If you find one, great, if not, never mind :)

src/Input/Input.spec.js
```diff
@@ -602,6 +602,18 @@ describe('Input', () => {
...
+      const driver = createDriver(<Input/>);
+      expect(driver.getHasCounter()).toBeFalsy();
+    });
+  });
```
lbelinsk: I would also test the case where we start with defaultValue, to make sure the counter is correct

src/Input/Input.js
```diff
@@ -137,6 +139,7 @@ class Input extends Component {
     //https://github.com/wix/wix-style-react/issues/1691
     return (<div className={styles.inputWrapper}>
       {prefix && <div className={styles.prefixSuffixWrapper}><div className={styles.prefix}>{prefix}</div></div>}
+      {hasCounter && maxLength && <span className={styles.counter} data-hook="input-counter">{this.state.counter}/{maxLength}</span>}
```
lbelinsk: why do you need to keep the counter on the state?
{this.props.value.length}/{maxLength}

Is it due to performance reasons, not to calculate the length on each render?
I think it is a tiny operation, we can remove from the state. wdyt?

lbelinsk:
@mozesoron Thanks for the pr.
I left some comments. Is there a reason why you didn't use the `prefix` prop for that?
You tagged some pr of breadcrumbs, but I didn't understand the context... What is this pr for? Did someone approve the design?

@milkyfruit Is this approved? This pr is adding a word counter field to the input as a prefix.
image

src/Input/Input.driver.js
mozesoron: this is the same name as in the textarea - I think it is better to use the same names for the same func. WDYT?

src/Input/Input.js
mozesoron: same as in textarea

src/Input/Input.js
mozesoron: so same as other comments - implemented same as in textarea. also the counter should be on the component state. it is internal and related only to the component

src/Input/Input.spec.js
mozesoron: I'v extended the test case

src/Input/Input.driver.js
lbelinsk: I guess so.

src/Input/Input.js
lbelinsk: It is related to the component, sure, The user won't need to take care of it. but it is only data that you have on the prop, no need to keep on the state

src/Input/Input.scss
```diff
@@ -15,6 +15,16 @@ $material-gap: 8px;
...
 }
 
+.counter{
+  position: absolute;
+  top: -30px;
+  right: 0;
+  font-size: 14px;
+  line-height: 1.3;
+  @include Placeholder($GR60);
+  color: #7a92a5;
```
lbelinsk: Please use D40

src/Input/Input.spec.js
lbelinsk: but you need to add some expectation on the value as well...

milkyfruit:
@mozesoron @lbelinsk this is not followed by the guidelines. The char counter is defined to be outside the input field:
image

Link to specs: https://zpl.io/2Egjola

lbelinsk:
@mozesoron It means that this should be added to TextField, not Input component. 
https://wix-wix-style-react.surge.sh/?selectedKind=3.%20Inputs&selectedStory=3.1%20Text%20Field&full=0&addons=0&stories=1&panelRight=0

I am closing this pr just because it is not approved by the design.
I can leave the branch if it helps

If you would just copy the same implementation we have in InputArea to TextField I will not ask for changes or additional stuff. I will do them if needed, so please just add the way you see right

shlomitc:
@argshook - I think the counter functionality should sit in the input, because the state of the characters length sits there and not in the form field component.

Having it in the field will create a complex API for the user for a simple use case. Also. Think about it like an error message. It could have been internally inside the input prefix and not separate.

argshook:
@shlomitc no, disagree.

input supports prefix & suffix, they accept nodes and so user of `Input`
can decide what to display there, might as well be custom counter.

making a special case for a thing that does not exist in design system
is asking for trouble in the long run.

Field component API will be this:

```js
<Field
  maxLength={100}
  value={yourState.value}
  onChange={yourHandler}
  >
  <Input/>
</Field>
```

it's not complex at all.
I'll give it for review today.

if counter functionality sits in input, then it also has to sit in InputArea, then RichTextArea then possibly more.

unless this is urgent for @mozesoron , i don't see why rush with it, let's follow design system

lbelinsk:
@argshook assigning you as a reviewer for this one to help you remember to close / add this fix in your generalized solution

shlomitc:
@mozesoron - please see my comment [here](https://github.com/wix/wix-style-react/issues/1664#issuecomment-404683231)
if it fit your needs, let's close this PR

shlomitc:
I see that you're good with the new component, so closing this PR. thanks!

PR #1882, opened 2018-06-24
<Table/> - [3] Add setSelectedIds instance method

hepiyellow:
expose the `setSelectedIds` method, to be used in Uncontrolled mode, in case something outside of the table needs to update the selected ids.


src/Table/Table.spec.js
```diff
@@ -102,6 +102,17 @@ describe('Table', () => {
     });
   });
 
+  describe('setSelectedIds', () => {
+    it('should set select rows when setSelectedIds is called', () => {
```
lbelinsk: 'should set select' --> 'should select'

PR #1884, opened 2018-06-25
`<Table>`- Add Card to Example. Add Header styles. Add README

hepiyellow:


src/Table/index.js
```diff
@@ -1 +1,3 @@
 export {default} from './Table';
+/*eslint no-duplicate-imports: 0*/
```
lbelinsk: why?

stories/Table/Table.story.js
```diff
@@ -70,7 +73,13 @@ export default {
   examples: (
     <div>
       <h1>Examples</h1>
-      <TableExample/>
+      <div className={s.examples}>
+        <div className={s.example}>
+          <CodeExample title="WithBulkActions" code={TableExampleRaw}>
```
lbelinsk: this is the title that we will see in the story, better not to have it as a variable name but as a readable title: `With Bulk Actions`

src/Table/index.js
hepiyellow: Because of eslint of course (-:
It is not possible to do:
`export *,{default} from './Table';`

Do you know of a way to export the default and also export all constants?

argshook: or maybe just have everything right here in this file and don't bother with useless proxy file overhead? :)

argshook:
regarding `exampleProps` you can check https://github.com/wix/wix-style-react/blob/master/stories/Search/index.story.js or https://github.com/wix/wix-ui-backoffice/blob/master/stories/LabelWithOptions.story.tsx stories

src/Table/README.md
```diff
@@ -0,0 +1,69 @@
...
+| render | function | - | true | A function to render column cells. The function will be called with each row's data and should return a jsx element. Signature: `render(rowData, rowNum)` |
+| width | string | - | - | The width to apply to the column. No value means column will try to contain its children, if possible.  |
+| important | bool | false | - | Whether font color should be stronger, more dominant |
+| sortable | bool | false | - | Enables sorting by column |
+| sortDescending | bool | - | - | Pass false - for ascending sort, true - for descending|
+| infoTooltipProps | object | - | - | Props object for [tooltip](https://wix-wix-style-react.surge.sh/?selectedKind=7.%20Tooltips&selectedStory=7.1.%20Tooltip&full=0&addons=0&stories=1&panelRight=0). Note: dataHook, moveBy and children will not be passed to tooltip. |
+
+### Children
+
+The `<Talbe/>` component acts as a Table context provider, and the compound components which consume it are:
```
argshook: typo

PR #1889, opened 2018-06-25
WIP Feat/FormField component

argshook:
### What changed

new `Field` component to eventually take place of `TextField` and friends

### Why it changed

as planned

---

- [X] Change is tested
- [X] Change is documented
- [ ] Build is green


hepiyellow:
Asterix has height 26:

screen shot 2018-06-26 at 17 35 10

src/Field/driver.js
```diff
@@ -0,0 +1,25 @@
+import {tooltipTestkitFactory} from '../../testkit';
```
hepiyellow: For consistency...
rename `driver.js` to `Field.driver.js`

src/Field/protractor.driver.js
```diff
@@ -0,0 +1,14 @@
+const find = component => selector => component.$(selector);
```
hepiyellow: For consistency:
rename `protractor.driver.js` to `Field.protractor.driver.js`

src/Field/index.spec.js
```diff
@@ -0,0 +1,151 @@
+/* global describe it expect */
```
hepiyellow: For consistency...
Rename `index.spec.js` to `Field.spec.js`

src/Field/index.js
```diff
@@ -0,0 +1,125 @@
...
+  info: PropTypes.node,
+
+  /** any value to be passed to children element. When string and maxLength is set, it is used to calculate remaining characters */
+  value: PropTypes.any,
+
+  /**
+   * custom value length for cases when `value` prop does not represent actual value length.
+   * For example, value may be `<p>Hello</p>` (12 chars) but is displayed as just `Hello` (5 chars)
+   */
+  valueLength: PropTypes.number,
```
shlomitc: see the comment for `value`

src/Field/index.js
```diff
@@ -0,0 +1,125 @@
...
+
+  /** optional handler that will be proxied to children.
+   * Mostly for convenience, as it can still be defined on children.
+   *
+   */
+  onChange: PropTypes.func,
```
shlomitc: see the comment for `value`

src/Field/index.js
```diff
@@ -0,0 +1,125 @@
...
+
+  /** number used to display remaining characters left. Should be used together with `value` or `valueLength`.
+   * It's `maxLength - valueLength || value.length`.
+   * Please note this does not validate data because there can be any type of children and value
+   */
+  maxLength: PropTypes.number,
```
shlomitc: see the comment for `value`

src/Field/driver.js
```diff
@@ -0,0 +1,25 @@
...
+  element.querySelector(`[data-hook*="${hook}"]`);
+
+const getInfoIcon = element =>
+  findByHook(element, 'field-infoicon');
+
+const fieldDriver = ({component, element}) => ({
+  exists: () => !!element,
+  element: () => element,
+  component: () => component,
+  getChildren: () => findByHook(element, 'field-children'),
```
shlomitc: see all comments in the protractor driver

src/Field/index.js
```diff
@@ -0,0 +1,125 @@
...
+        ...(value ? {value} : {}),
+        ...(onChange ? {onChange} : {})
+      }
+    );
+  }
+
+  return children;
+};
+
+const Field = ({children, label, required, info, dataHook, value, valueLength, onChange, maxLength}) =>
```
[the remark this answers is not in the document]
shlomitc: any reason why not to use it here?

src/Field/index.js
```diff
@@ -0,0 +1,125 @@
...
+  <div
+    className={styles.infoIcon}
+    data-hook={`field-infoicon${isInline ? '-inline' : ''}`}
+    >
+    <Tooltip
+      content={content}
+      theme="dark"
+      dataHook="field-infotooltip"
+      >
+      <div>
```
shlomitc: is this extra div necessary?

src/Field/index.js
```diff
@@ -0,0 +1,125 @@
+import React from 'react';
```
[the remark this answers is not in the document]
argshook: im strongly against this "pattern" it just adds unnecessary file with unnecessary proxy for no reason

src/Field/components/max-length-counter.js
```diff
@@ -0,0 +1,32 @@
...
+  return (
+    <div
+      className={styles.counter}
+      data-hook="field-counter"
+      >
+      <Text
+        size="small"
+        skin={count > 0 ? 'standard' : 'error'}
+        secondary
+        children={count}
```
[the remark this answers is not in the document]
argshook: because that's the same thing but without `>{children}</Field>` code

src/Field/Field.e2e.js
```diff
@@ -0,0 +1,35 @@
+import eyes from 'eyes.it';
+import {fieldTestkitFactory, getStoryUrl, waitForVisibilityOf} from '../../testkit/protractor';
+import autoExampleDriver from 'wix-storybook-utils/AutoExampleDriver';
+
+const storyUrl = getStoryUrl('12. Other', '12.6 Field');
+const driver = fieldTestkitFactory({dataHook: 'storybook-field'});
+
+describe('Field', () => {
```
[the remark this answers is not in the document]
argshook: why tho

src/Field/Field.e2e.js
```diff
@@ -0,0 +1,35 @@
...
+  beforeAll(() =>
+    browser.get(storyUrl));
+
+  eyes.it('should render with label', () => {
+    autoExampleDriver.setProps({label: 'hello'});
+    waitForVisibilityOf(driver.element(), 'Cannot find Field component')
```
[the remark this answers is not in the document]
argshook: why

src/Field/Field.e2e.js
```diff
@@ -0,0 +1,35 @@
...
+  });
+
+  eyes.it('should render info icon when specified and no label', () => {
+    autoExampleDriver.setProps({info: 'hello'});
+    waitForVisibilityOf(driver.element(), 'Cannot find Field component')
+      .then(() => {
+        expect(driver.getInfoIcon()).not.toBe(undefined);
+      });
+  });
+});
```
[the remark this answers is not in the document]
argshook: i don't want to duplicate unit and e2e, so either do unit or e2e tests. I figure a few snapshots is enough and all logic is covered in unit. If you insist, i can test logic in e2e but in my point of view that will be a nightmare to maintain.

src/Field/components/max-length-counter.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import Text from 'wix-style-react/Text';
+
+import styles from '../styles.scss';
```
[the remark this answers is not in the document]
argshook: why rename? folder name tells you what component files belong to

src/Field/components/max-length-counter.js
```diff
@@ -0,0 +1,32 @@
...
+
+const maxLengthCounter = ({maxLength, value, valueLength}) => {
+  const count = typeof value === 'string' ? maxLength - (valueLength || value.length) : maxLength;
+
+  return (
+    <div
```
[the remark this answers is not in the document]
argshook: this component is part of Field and is not meant to be used stand alone. I extracted function to dedicated file just to reduce noise in Field component implementation. I can extract `margin-left` but i don't think it matter where it's set for internal component.

src/Field/components/max-length-counter.js
```diff
@@ -0,0 +1,32 @@
...
+      className={styles.counter}
+      data-hook="field-counter"
+      >
+      <Text
```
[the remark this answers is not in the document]
argshook: nooot sure about it. We have text component tailored for exactly this usecase but want to avoid using it?

src/Field/components/max-length-counter.js
```diff
@@ -0,0 +1,32 @@
...
+        children={count}
+        />
+    </div>
+  );
+};
+
+maxLengthCounter.propTypes = {
+  maxLength: PropTypes.number,
+  valueLength: PropTypes.number,
+  value: PropTypes.any
```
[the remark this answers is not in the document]
argshook: if i do `PropTypes.string` and consumer passes something else, console will complain. and i don't know if `value` is always gonna be string cause Field is generic, children can be any node that handle any kind of value.

src/Field/driver.js
argshook: yeah about that, i know most components are `Component/Component.thing.js` but this feels awfully repetitive for me. I can do that if you reaaally insist but in general i don't see a point. Butter is buttery is what it becomes :)

src/Field/index.js
argshook: yes, Tooltip requires it

src/Field/index.js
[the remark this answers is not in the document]
argshook: i didn't find a reason to use it. i thought we tend to move away from unnecessary inheritance. The only useful thing in WixComponent is `onClickOutside` but i don't need it here.

src/Field/index.js
```diff
@@ -0,0 +1,125 @@
...
+    }
+  </div>;
+
+Field.displayName = 'Field';
+Field.propTypes = {
+  /** any kids to render, should be some form of input. Input, InputArea & RichTextArea work well */
+  children: PropTypes.node,
+
+  /** optional text labeling this form field */
+  label: PropTypes.node,
```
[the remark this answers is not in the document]
argshook: no, there's a design with no label, that's the `inline` thing

src/Field/index.js
```diff
@@ -0,0 +1,125 @@
...
+
+  /** whether to display an asterisk (*) or not */
+  required: PropTypes.bool,
+
+  /** display info icon with tooltip. Node from this prop is content of tooltip */
+  info: PropTypes.node,
+
+  /** any value to be passed to children element. When string and maxLength is set, it is used to calculate remaining characters */
+  value: PropTypes.any,
```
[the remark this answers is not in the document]
argshook: >  it makes it the internal component's value unpredictable.
how? i don't interfere, i pass i pass it through

> shouldn't interfere with the internal component logic nor create a new API for it.
bbbut i don't interfere with logic and i don't introduce new api

i think we need a call so i can show what it's all about. Your suggestion with children as function seems over engineered to me. what is the `setCurrentCharLength` for? why introduce `onCharLimitChange` api in input at all?

src/Field/index.spec.js
```diff
@@ -0,0 +1,151 @@
...
+      });
+
+      it('should display result of maxLength - value.length', () => {
+        const driver = createDriver(<Field label="hello" value="12345" maxLength={87987}/>);
+        expect(driver.getCounter().innerHTML).toMatch('87982');
+      });
+
+      it('should display with skin="error" when result < 0', () => {
+        const driver = createDriver(<Field label="hello" value="12345" maxLength={4}/>);
+        expect(driver.getCounter().innerHTML).toMatch('skin="error"'); // TODO: use Text testkit from wix-ui-backoffice
```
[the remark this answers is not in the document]
argshook: as noted in TODO i should use Text testkit from wix-ui-backoffice but didn't want to spend time on figuring that out

src/Field/styles.scss
```diff
@@ -0,0 +1,52 @@
+@import '../common.scss';
+
+.root {
+  display: flex;
+  flex-wrap: wrap;
```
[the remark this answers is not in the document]
argshook: i need it for inline at least, i can move it there

src/Field/styles.scss
```diff
@@ -0,0 +1,52 @@
...
+
+.suffixesInline {
+  display: flex;
+
+  .asterisk {
+    margin-top: 3px;
+  }
+
+  .infoIcon {
+    margin-top: 5px;
```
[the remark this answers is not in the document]
argshook: not sure, i was pushing pixels and centering didn't work ok for me

src/Field/index.spec.js
```diff
@@ -0,0 +1,151 @@
...
+        expect(driver.getInfoIcon()).not.toEqual(null);
+      });
+
+      it('should display value of `info` in tooltip', () => {
+        const driver = createDriver(<Field info="hello from tooltip"/>);
+        const tooltip = driver.getInfoTooltip();
+
+        tooltip.mouseEnter();
+
+        return resolveIn(500).then(() => {
```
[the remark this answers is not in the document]
argshook: what if i told you `wix-eventually` is `setTimeout` buried under dependencies? :)
but sure, i'll change to it

src/FormField/index.js
```diff
@@ -0,0 +1,110 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+import Text from 'wix-style-react/Text';
```
shlomitc: see previous comments regarding using a native `label` and apply the right typography.
also the `htmlFor` prop

src/FormField/index.js
```diff
@@ -0,0 +1,110 @@
...
+      data-hook="formfield-asterisk"
+      className={styles.asterisk}
+      children="*"
+      />;
+
+  renderChildren() {
+    const {children} = this.props;
+    if (typeof children === 'function') {
+      return children({
+        setLengthLeft: lengthLeft => {
```
shlomitc: might want to define this function once on the class instead of re-creating on each render

src/FormField/index.js
```diff
@@ -0,0 +1,110 @@
...
+class FormField extends React.Component {
+  static displayName = 'FormField';
+  static propTypes = {
+    /**
+     * any kids to render, should be some form of input. Input, InputArea & RichTextArea work well
+     *
+     * `children` can be React node or a function
+     *
+     * when function, it receives object with:
+     * * `setLengthLeft` - function accepts a number and will display it on top right of `FormField` component
```
shlomitc: let's talk about this api name.
it is supposed to display the character limit so `setCharactersLeft` sounds to me like more explicit.

src/FormField/index.js
```diff
@@ -0,0 +1,110 @@
...
+    children: PropTypes.oneOfType([PropTypes.node, PropTypes.func]).isRequired,
+
+    /** optional text labeling this form field */
+    label: PropTypes.node,
+
+    /** whether to display an asterisk (*) or not */
+    required: PropTypes.bool,
+
+    /** display info icon with tooltip. Node from this prop is content of tooltip */
+    info: PropTypes.node,
```
shlomitc: `infoContent`?

src/FormField/driver.js
```diff
@@ -0,0 +1,42 @@
...
+    return parseInt(count, 10);
+  }
+
+  return null;
+};
+
+const formFieldDriver = ({component, element}) => ({
+  exists: () => !!element,
+  element: () => element,
+  component: () => component,
```
shlomitc: component is not something we want to give our end users, so let's not expose in the driver

src/FormField/driver.js
```diff
@@ -0,0 +1,42 @@
...
+  getChildren: () => findByHook(element, 'formfield-children'),
+  getLabel: () => findByHook(element, 'formfield-label'),
+  isRequired: () => !!findByHook(element, 'formfield-asterisk'),
+  isInline: () => !!findByHook(element, 'formfield-inline-suffixes'),
```
shlomitc: not relevant for the user

src/FormField/FormField.e2e.js
```diff
@@ -0,0 +1,44 @@
...
+import autoExampleDriver from 'wix-storybook-utils/AutoExampleDriver';
+
+const storyUrl = getStoryUrl('12. Other', '12.6 FormField');
+const driver = formFieldTestkitFactory({dataHook: 'storybook-formfield'});
+
+describe('FormField', () => {
+  beforeAll(() =>
+    browser.get(storyUrl));
+
+  eyes.it('should render with label', () =>
```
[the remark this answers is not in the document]
argshook: why?

src/FormField/driver.js
argshook: so how should i test it myself then?

src/FormField/components/MaxLengthCounter.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import Text from 'wix-style-react/Text';
```
shlomitc: same here, don't use `Text` component, only typography

src/FormField/FormField.scss
```diff
@@ -0,0 +1,52 @@
...
+  align-items: center;
+  flex: 1 0 auto;
+  margin-bottom: 6px;
+}
+
+.children {
+  flex: 1 0 100%;
+}
+
+.childrenInline {
```
shlomitc: I would give it a semantic meaning, like `childrenWithoutLabel`

src/FormField/FormField.e2e.js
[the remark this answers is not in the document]
shlomitc: suffixesInline

src/FormField/driver.js
argshook: i want this test, how should i do it without `component`? https://github.com/wix/wix-style-react/pull/1889/files#diff-ce3ed13c1d51d395722495fd0f9cecbdR21

shlomitc: shouldn't be part of the driver. just mount the component and make sure it is there

src/FormField/FormField.e2e.js
```diff
@@ -0,0 +1,29 @@
...
+const driver = formFieldTestkitFactory({dataHook: 'storybook-formfield'});
+
+describe('FormField', () => {
+  beforeAll(() =>
+    browser.get(storyUrl));
+
+  eyes.it('should render with label', async () => {
+    await autoExampleDriver.setProps({label: 'hello'});
+    await waitForVisibilityOf(driver.element(), 'Cannot find FormField component');
+    expect(driver.getLabel().getText()).toMatch('hello');
```
shlomitc: missing `await`
`expect(await driver.getLabel().getText()).toMatch('hello');`

src/FormField/index.js
argshook: ok, i'll use label but how do you imagine `htmlFor` to work?

src/FormField/components/MaxLengthCounter.js
argshook: ok, i'll do it but it's weird as hell

src/FormField/index.js
shlomitc: this will probably require to pass in the same `id` you will you in your input.
so another prop is needed for the `FromField` component

src/FormField/components/MaxLengthCounter.js
shlomitc: why?

src/FormField/FormField.js
```diff
@@ -0,0 +1,110 @@
...
+      <div
+        data-hook={dataHook}
+        className={styles.root}
+        >
+        { label &&
+        <div
+          className={styles.label}
+          data-hook="formfield-label"
+          >
+          <Text appearance="T1" children={label}/>
```
hepiyellow: I think we should use a native `<label>` that wraps the `<Field/>`'s children, so that we conform to HTML behavior, and support ScreenReaders etc...

You might want to use the existing `<Label/>` component.

src/FormField/components/MaxLengthCounter.js
argshook: because of using already precooked component that can handle such usecase exactly, i need to do it manually. and doing manually adds up quickly.

but i'll do it, ignore me

src/FormField/FormField.scss
```diff
@@ -0,0 +1,52 @@
...
+
+.children {
+  flex: 1 0 100%;
+}
+
+.childrenWithoutLabel {
+  flex: 1 0 auto;
+}
+
+// TODO: get SVG from designers
```
shlomitc: do we need an svg here?

src/FormField/FormField.protractor.driver.js
```diff
@@ -0,0 +1,12 @@
+const fieldDriverFactory = component => {
+  const byHook = hook => component.$(`[data-hook*=${hook}]`);
+
+  return {
+    element: () => component,
+    getLabel: () => byHook('field-label'),
+    getAsterisk: () => byHook('field-asterisk'),
```
shlomitc: create a better abstraction

src/FormField/FormField.scss
argshook: there was comment in zeplin about it but i think we don't need an svg, a simple * is good enough. It's same what is used right now in TextField.

src/FormField/FormField.js
```diff
@@ -0,0 +1,127 @@
...
+
+  static defaultProps = {
+    required: false
+  }
+
+  state = {
+    lengthLeft: undefined
+  }
+
+  renderAsterisk = () =>
```
shlomitc: wouldn't it make it as an API in the AutoDocs?
Also, since it's static, worth taking it out of the class.

same for the `renderCharactersLeft`

src/FormField/README.md
```diff
@@ -0,0 +1,33 @@
...
+> NOTE: when `children` is function (a.k.a. render prop), it receives `setLengthLeft` which
+> can be called with `number`
+
+```js
+<FormField
+  label="Tweet"
+  maxLength={140}
+  >
+  {({setLengthLeft}) =>
+    <Input onChange={event => setLengthLeft(100 - event.target.value)}/>
```
shlomitc: `event.target.value.length`?
should probably have an e2e test for that

stories/FormField.story.js
```diff
@@ -0,0 +1,72 @@
...
+  {label: 'Input with char counter',
+    // eslint-disable-next-line react/prop-types
+    value: ({setCharactersLeft}) =>
+      <Input
+        placeholder={placeholder}
+        onChange={e => setCharactersLeft(100 - e.target.value.length)}
+        id={ID}
+        />
+  },
+  {label: 'InputArea',
```
shlomitc: all the examples from here are not relevant for the `FormField`.
it should only show how the component behaves, the content is less important.

we need other sections in the storybook to have these examples.
For example, the `TextArea` story will have examples of the `FormField` with `InputArea` according to the spec
image

src/FormField/README.md
argshook: e2e? there's unit test for that

stories/FormField.story.js
argshook: disagree, they show exactly how to use FormField for exactly components it was tailored for + other potential candidates.
just today i saw more usecases than Input & InputArea and FormField already can handle it.

other sections in storybook can be updated in other PR

shlomitc: you can have dozens of examples with every component that exist in WSR, but are they really needed? I guess not. This is just more code to maintain, which doesn't have a lot of value.

It should have the bare minimum way to understand how the API works, it's not intended to cover all use cases.

The real depo pages cover the usages we intend to have

TextField => FormField + Input
TextArea => FormField + InputArea
Range => FormField RangeInput
etc...

src/FormField/README.md
shlomitc: Fact is that this piece of code is not tested anywhere (and not working).
This can be an example (instead of just a regular code snippet).

argshook: that's not a fact at all, it's tested here https://github.com/wix/wix-style-react/blob/72c38e2fce6794dbd7b19e102dfa87396afb45e6/src/FormField/FormField.spec.js#L93-L134
there are two examples here https://github.com/wix/wix-style-react/blob/72c38e2fce6794dbd7b19e102dfa87396afb45e6/stories/FormField.story.js#L18-L40 and they both work

the thing i missed is `setLengthLeft` should be `setCharactersLeft` in readme, i'll update

shlomitc:
when counter is 0, the counter is broken
image

stories/FormField.story.js
```diff
@@ -0,0 +1,73 @@
...
+    value: <DatePicker value={new Date()}/>
+  },
+
+  {label: 'Dropdown',
+    value: <Dropdown placeholder="Select dominant hand" options={[{id: 0, value: 'Left'}, {id: 1, value: 'Right'}, {id: 2, value: 'Ambidextrous'}]}/>
+  }
+];
+
+export default {
+  category: '12. Other',
```
shlomitc: I think it should sit in a different area in the storybook, something like `APIs` or the formerly known `Componentes` category
the `12. Other` is used in the design system sketch

src/FormField/README.md
```diff
@@ -0,0 +1,30 @@
...
+</FormField>
+```
+
+With length count:
+
+> NOTE: when `children` is function (a.k.a. render prop), it receives `setCharactersLeft` which
+> can be called with `number`
+
+```js
+<FormField label="Tweet">
```
shlomitc: well, I think this is a little bit redundant as it doesn't have any actual effect, it's only text.
if it was in the `examples` section as a code example + actual rendering, then it would make much more sense in my opinion

stories/FormField.story.js
argshook: so `APIs` or `Componentes`?

PR #1894, opened 2018-06-26
Tooltip fix(react 16)

Faradey27:
### What changed

- fixed tooltip for react 16

### Why it changed

- to support react 16

---

- [x] Change is tested
- [x] Change is documented
- [ ] Build is green

### DIFF
--- a/src/Tooltip/Tooltip.js
+++ b/src/Tooltip/Tooltip.js
@@ -158,47 +158,8 @@ class Tooltip extends WixComponent {
     this.props.onClickOutside && this.props.onClickOutside(e);
   }
 
-
   componentDidUpdate() {
-    if (this._mountNode && this.state.visible) {
-      const arrowPlacement = {top: 'bottom', left: 'right', right: 'left', bottom: 'top'};
-      const position = this.props.relative ? 'relative' : 'absolute';
-      const tooltip = (
-        <TooltipContent
-          onMouseEnter={() => this._onTooltipContentEnter()}
-          onMouseLeave={() => this._onTooltipContentLeave()}
-          ref={ref => {
-            if (this.props.relative) {
-              this.tooltipContent = ref.tooltip;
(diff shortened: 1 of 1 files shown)


Faradey27:
@argshook ping

argshook:
@Faradey27 i went through the usecases that i saw crashing browser and can no longer reproduce
didn't test with react 16 though. otherwise looks ok to me

hepiyellow:
Who asked for this?
This doesn't look like a small trivial fix, so I think we should not make such fixes. WSR is not officially supporting React 16 as far as I know, so it doesn't worth the effort.
We need to test it with React 16 if we want to support it.
@shlomitc ?

shlomitc:
@hepiyellow - this is a request that came from several users and is low in our priority.
As this was a small fix done by Andrii, it was ok from my side to add it while declaring this is not officially supported.

The previous fix caused issues so this attempts to solve them.
So this is kind of a quick win. if it still causes us issue - it will go back to the backlog and won't be fixed until we rewrite the Tooltip logic.

hepiyellow:
Just wanted to mention that Oleg from adi came to me the other day, asking for a fix in Tooltip in WSR with React 16, and I finally recommended him do use `wix-ui-adi`, which he found satisfying.
Just wanted to make sure that we offer to use wix-ui-core (and derived libraries) when it comes to React 16.

Faradey27:
@shlomitc so what the correct approach here? should we close pull? or merge(build is green)

shlomitc:
merge

shlomitc:
@argshook - did you test it on a React16 project?

argshook:
@shlomitc no, i was more interested in not breaking v15. it doesn't so im okay with it :)

ipanasenko:
Doesn't work with React16. No errors, no tooltip. It just doesn't show, that's all

PR #1896, opened 2018-06-26
Dropdown add min and max width props, deprecate width prop

Faradey27:
### What changed

- added minWidth props
- added maxWidth props
- deprecated width prop

### Why it changed

- https://github.com/wix-private/wsr-issues/issues/112

---

- [x] Change is tested
- [x] Change is documented
- [ ] Build is green


hepiyellow:
@shlomitc 
1) Why do we need this?
The consumer can always wrap the Input with a `<span>`, and the `<Input>` will adapt it's width.
```
<span style={{minWidth, maxWidth}}>
  <Input/>
</span>
```
2) If we want to make it convenient to the consumer, we could open 2 props:
 - `className` property which we will append to the  `<input>`'s className.
 - `style` property that will be merged to the style of the `<input>`

Adding a prop for each css attribute clutters the component API, driver API, tests,...

I think we should be shifting into an approach a "Recommended design system" rather than "Closed design system".

argshook:
> I think we should be shifting into an approach a "Recommended design system" rather than "Closed design system".

that's a dream of mine since 2016 or something

Faradey27:
@hepiyellow @argshook  just my small comment:
As we are doing tool which should be same over all clients in same design with same transtions, etc. it is better to stay as strict as possible and provide customization just by themes.

Just we will allow more custom things, people will start to do super tricks and hack when they just lazy to read docs or when they UX guys just decided to not follow guidlines

Faradey27:
@shlomitc can you merge or close?

shlomitc:
I'll split it into two parts.

**Part 1**
So this is what I think about the different customization approaches:

1. Adding `span` with some `max-width/min-width` prop can be an option when the components stretch to 100%. if it doesn't, then it might not be effective. The downside of it is that it adds another `span` which is, well, another `span` in the DOM.
2. Opening `className` might be a good option in the simple cases but more problematic for compositions. in the `Dropdown` for example, will it go to the `input` or to the `dropdown` menu? All of these questions are solved when coming to some theming mechanism like `JSS`, `stylable` or any other tools, but as we don't use it in these components, it is not trivial.
3. `style` prop is not a good approach, I'd rather not recommend using it if not really needed (DOM size calculations for example).
4. Adding prop is usually very declerative but requires adding a prop and maintaining it which sucks.

so the answer should be given according to the use case. Regarding `Dropdown`, I have some doubts because of my second part

**Part 2**
I see a conflict regading the `Dropdown` size as is has some specific characteristics regarding minimum and maximum length and different when in a grid.
image

1. It needs to have a minimum 192px width and be able grow.
2. It should be possible to limit the max size of it (to allow it to grow but not completely)
3. On the other hand, it should be stretching to 100% so it will fit to in our `Grid`, which doesn't go well with the min/max width requirements.
So there is a conflict here. maybe @milkyfruit can explain a bit better.

milkyfruit:
@Faradey27 @shlomitc 

Regular dropdown is designed for submit forms:
image
Forms are constructed using the grid, so dropdown's width is always 100%.

Rounded dropdown is designed as a table filter:
image
To my knowledge, the table header doesn't have a grid, so the dropdown's width need to be limited. Video example (Wix Stores): https://drive.google.com/open?id=1qPKslkdvmC7WBEWR7rApq0KD37fQ03lH

For this reason the min-max width specs are added.

shlomitc:
OK, so it looks like a different task and a new feature: support expendable input/dropdown according to content (which will override the stretch behavior).

So I'm closing this PR, will open a new issue with the requirements

PR #1898, opened 2018-06-26
`<DataTable/>` - Add missing documentation comments on prop types + make `columns` required

hepiyellow:
I haven't yet switch DataTable's story to use AutoDocs, but I want to use these prop types in the AutoDocs of `<Table>`


lbelinsk:
It won't be breaking even if a user does not provide columns, it will just trigger console errors

PR #1902, opened 2018-06-28
<Page/> - add prop to disable headerContainer padding-bottom

hepiyellow:
This is need for the `<Table>` in a `<Page>`.

### DIFF
--- a/src/Page/Page.js
+++ b/src/Page/Page.js
@@ -28,7 +28,7 @@ class Page extends WixComponent {
     this._handleResize = this._handleResize.bind(this);
 
     this.state = {
-      headerHeight: 0,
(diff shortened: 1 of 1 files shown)


src/Page/Page.js
```diff
@@ -264,7 +267,8 @@ Page.propTypes = {
     ) {
       return new Error(`Page: Invalid Prop children, unknown child ${children[key].type}`);
     }
-  }).isRequired
+  }).isRequired,
+  withoutHeaderContainerBottomPadding: PropTypes.bool
```
shlomitc: please give it a more semantic name

hepiyellow:
replaced by #1933

PR #1907, opened 2018-07-01
dont publish scripts

lbelinsk:
### What changed

...

### Why it changed

...

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


shlomitc:
@lbelinsk - any reason why not to do it? we actually have there a script to migrate users to new icons

Issue #1881, opened 2018-06-24
When clicking unit suffix in <Input />, an error is thrown


ChenRoth:
### Summary:

When passing a `unit` prop to `<Input />` and clicking on the DOM of that unit, an error is thrown:
```
react-dom.min.js:14 Uncaught TypeError: t.props.onInputClicked is not a function
    at onIconClicked (external "React":1)
    at Object.r (react-dom.min.js:14)
    at a (react-dom.min.js:12)
    at Object.s [as executeDispatchesInOrder] (react-dom.min.js:12)
    at f (react-dom.min.js:12)
    at m (react-dom.min.js:12)
    at Array.forEach (<anonymous>)
    at r (react-dom.min.js:15)
    at Object.processEventQueue (react-dom.min.js:12)
    at r (react-dom.min.js:14)
```

### Priority:
- [ ] Blocker
- [ ] Urgent
- [ ] Normal
- [X] Minor

### Steps to reproduce:
 1. https://wix-wix-style-react.surge.sh/?selectedKind=3.%20Inputs&selectedStory=3.1%20%2B%20Input&full=0&addons=0&stories=1&panelRight=0
 2. Click on the `$`

### Expected behavior:

No errors

### Relevant Components:

Input

lbelinsk:
Done

PR #1908, opened 2018-07-01
Fix when clicking unit suffix in <Input />, an error is thrown (#1881)

falinsky:
### What changed
Fixed invoking click handler while clicking on input unit.

### Why it changed
To stop throwing error if no `onInputClicked` props provided to `Input` component
---

- [*] Change is tested
- [ ] Change is documented
- [*] Build is green


lbelinsk:
Thanks!
It looks like a good fix for this issue.
I think it will be better to generalize this. For example, how `magnifyingGlass` is different than `unit`?
I guess it will cause the same issue.

falinsky:
@lbelinsk good catch. I've added few tests for `magnifyingGlass` and `menuArrow` suffixes due to they have behavior similar to `unit` suffix.

PR #1911, opened 2018-07-01
Tiny loader

justenor:


lbelinsk:
build fails on linting errors

src/Loader/Loader.js
```diff
@@ -52,6 +57,7 @@ export default class Loader extends WixComponent {
     const sizeInPx = sizesInPx[size];
     const lightArcAngle = arcsAngles[size].light;
     const darkArcAngle = arcsAngles[size].dark;
+    const supportsTooltip = size !== 'tiny'
```
lbelinsk: We are hiding here the text, so maybe a better name will be 

shouldShowText = size !== 'tiny'

lbelinsk:
close https://github.com/wix-private/wsr-issues/issues/7

ipanasenko:
Do you think `tiny` is okay name? We already have `x-large` and `x-small` in couple places

lbelinsk:
Yeah, now that you mention it, you might be right.
Not sure worth changing it though, as it is already released.
We can be quick about it and change it.
@shlomitc @wuwa wdyt?

ipanasenko:
Another solution is to deprecate it immediately and add an alias, just not to break anyone's project

PR #1919, opened 2018-07-02
Loader e2e tests

justenor:


src/Loader/Loader.e2e.js
```diff
@@ -13,15 +14,55 @@ describe('Loader', () => {
...
-        expect(loaderDriver.getText()).toBe('WUBBA LUBBA DUB DUB');
-      });
+    await waitForVisibilityOf(loaderDriver.element(), 'Cannot find <Loader/>');
+    expect(loaderDriver.isLarge()).toBe(true);
+    expect(loaderDriver.getColor()).toBe('white');
+    expect(loaderDriver.hasText()).toBe(true);
+    expect(loaderDriver.getText()).toBe('WUBBA LUBBA DUB DUB');
+  });
+
+  eyes.it('should render tiny loader', async () => {
```
lbelinsk: Instead of having multiple tests, I would do one test for `should render different Loader sizes` and inside the test use eyes.checkWindow to take screenshots of the different sizes

PR #1923, opened 2018-07-02
Create shadow palette

justenor:


lbelinsk:
Open questions for UX:
ColorPicker - Handle has a slightly different shadow. We are using now `s10`.

lbelinsk:
closes https://github.com/wix-private/wsr-issues/issues/5

PR #1953, opened 2018-07-05
<Loader/> - add error and success states

justenor:


src/Loader/Loader.e2e.js
```diff
@@ -73,4 +73,30 @@ describe('Loader', () => {
...
+
+    await waitForVisibilityOf(loaderDriver.element(), 'Cannot find <Loader/>');
+
+    eyes.checkWindow('loading state');
+
+    expect(loaderDriver.isLoading()).toBe(true);
+
+    autoExampleDriver.setProps({state: 'error'});
+
+    await waitForVisibilityOf(loaderDriver.element(), 'Cannot find <Loader/>');
```
justenor: redundant

src/Loader/Loader.e2e.js
```diff
@@ -73,4 +73,30 @@ describe('Loader', () => {
 
     expect(loaderDriver.getColor()).toBe('white');
   });
+
+  eyes.it('should render different loader states', async () => {
+    autoExampleDriver.setProps({state: 'loading'});
+
+    await waitForVisibilityOf(loaderDriver.element(), 'Cannot find <Loader/>');
+
+    eyes.checkWindow('loading state');
```
justenor: check if an async command. is so, add `await`

src/Loader/Loader.js
```diff
@@ -5,6 +5,10 @@ import WixComponent from '../BaseComponents/WixComponent';
 import Arc from './Arc';
 import css from './Loader.scss';
 import Text from '../Deprecated/Text';
+import FormFieldError from '../../new-icons/system/FormFieldError';
```
justenor: only a single `../` so it will take files from the src directory and not from the root

src/Loader/Loader.js
```diff
@@ -44,23 +64,29 @@ export default class Loader extends WixComponent {
     color: PropTypes.oneOf(['blue', 'white']),
 
     /** Node (usually text) to be shown below the loader */
-    text: PropTypes.node
+    text: PropTypes.node,
+
+    /** The state of the loader */
+    state: PropTypes.oneOf(['loading', 'success', 'error'])
```
justenor: think of a better name for this prop. maybe status

src/Loader/Loader.js
```diff
@@ -79,6 +105,10 @@ export default class Loader extends WixComponent {
             strokeWidth={strokeWidth}
             viewBoxSize={sizeInPx}
             />
+          <div className={css.stateIndicator}>
```
justenor: remove div if not in success or error state

src/Loader/Loader.js
```diff
@@ -44,23 +64,29 @@ export default class Loader extends WixComponent {
...
+    status: 'loading'
   };
 
   render() {
-    const {size, color, text} = this.props;
+    const {size, color, text, status} = this.props;
     const sizeInPx = sizesInPx[size];
-    const lightArcAngle = arcsAngles[size].light;
-    const darkArcAngle = arcsAngles[size].dark;
+    const lightArcAngle = status === 'loading' ? arcsAngles[size].light : 0;
```
shlomitc: maybe think of a semantic way to describe that only one arc is displayed

PR #1957, opened 2018-07-05
<Icons/> - add new 1.4 story to storybook

Faradey27:
### What changed

...

### Why it changed

...

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


stories/NewIcons/index.js
```diff
@@ -0,0 +1,32 @@
...
+import AllIconsRaw from '!raw-loader!./AllIcons';
+
+import CustomeIcon from './CustomIcon';
+import CustomeIconRaw from '!raw-loader!./CustomIcon';
+
+import RTLIcon from './RTLIcon';
+import RTLIconRaw from '!raw-loader!./RTLIcon';
+
+storiesOf('1. Foundation', module)
+  .add('1.4.1 New Icons', () => (
```
shlomitc: should be aligned to the indexing in the zeplin.
So I'd call it `1.4 Icons` and the old one `1.4 Icons - deprecated`

stories/NewIcons/RTLIcon.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import {Duplicate} from 'wix-style-react/new-icons';
```
shlomitc: `import {Duplicate} from 'wix-style-react/new-icons/Dupliate'`

stories/NewIcons/CustomIcon.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import {Duplicate} from 'wix-style-react/new-icons';
```
shlomitc: import {Duplicate} from 'wix-style-react/new-icons/Dupliate'

Faradey27:
@shlomitc updated

src/new-icons/README.md
```diff
@@ -0,0 +1,21 @@
+# Icons
+
+> SVG Icon base type.
+
+## Usage
+~~~js
+import Favorite from 'wix-style-react/new-icons/Favorite';  // General icon
```
argshook: @Faradey27 so i understand that consumer will need to import from `new-icons` path?

isn't this problematic? once consumers use this path and the `new-icons` will become regular `icon`, we'll have to support both paths. what will happen when there are newer than new icons?

stories/NewIcons/index.js
```diff
@@ -0,0 +1,32 @@
...
+import {storiesOf} from '@storybook/react';
+import Markdown from 'wix-storybook-utils/Markdown';
+import CodeExample from 'wix-storybook-utils/CodeExample';
+
+import Readme from '../../src/new-icons/README.md';
+
+import AllIcons from './AllIcons';
+import AllIconsRaw from '!raw-loader!./AllIcons';
+
+import CustomeIcon from './CustomIcon';
```
argshook: custome :)

src/new-icons/README.md
Faradey27: at the start, we thought that we will just replace old to new, but then understood that we can not do this because of back compatibility.

So we decided to create completely independent icons and force everyone to use them, in long future(maybe half a year) when everyone will migrate to new icons, we will just replace icons with new icons, so we will have
icons/
new-icons/
and they will be same, then we will run some migration script to rename new-icons to icons for our clients

stories/NewIcons/index.js
Faradey27: ahah) cool

src/new-icons/README.md
argshook: currently `Icons` path is taken but `icons` is not, so perhaps we can rely on case sensitive systems and have path just `icons` and not `new-icons`? :)

shlomitc: some systems might get confused by the lower/capital case, so I prefer not to take that chance.

shlomitc:
@milkyfruit - regarding the redundant icons, please open a separate issue.

regarding the sizes - it's hard to tell how users will use the icons. We're not limiting them as we allow to control the size prop and also allow passing a class name which gives them the freedom.

changing the icon to be either 24px or 18px instead of 3em is ok.

src/new-icons/README.md
argshook: yeah im just shooting blanks. "some systems" is basically windows :)

milkyfruit:
I'm ok with allowing flexibility of allowing em and px. Just suggested use examples in the snippets, that match real back-office use.

shlomitc:
@milkyfruit you can open in WSR for convenience

shlomitc:
@Faradey27 - so please change the example to the snippet

Faradey27:
@shlomitc updated

PR #1962, opened 2018-07-08
`<Card/>` - Add overflow: hidden

hepiyellow:
When putting Table in a Card (without Card.Content),
 then row's white background overflows and we don't see the Card's rounded corners.


argshook:
@hepiyellow hey, maybe you remember, is something like this needed? css property is wrong (`over-flow` and not `overflow`) so this doesn't really change anything, so maybe it's not needed at all?

PR #1975, opened 2018-07-10
fix link in Icons readme

ipanasenko:


argshook:
Thanks! there was one more quote at the end which i removed.

PR #1960, opened 2018-07-06
 <Icons/> - migrate <Label/>, <MultiSelectCheckbox/>, <SideMenu/>, <Tooltip/> stories

Faradey27:
### What changed

...

### Why it changed

...

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green
- [ ] Change of UX is approved by [wuwa](https://github.com/wuwa) or [milkyfruit](https://github.com/milkyfruit)

PR #1976, opened 2018-07-10
<Icons /> - mirgate Button story

Faradey27:
Merge after https://github.com/wix/wix-style-react/pull/1960

### DIFF
--- a/stories/Button/close.story.js
+++ b/stories/Button/close.story.js
@@ -2,7 +2,7 @@ import React from 'react';
 
 import Button from 'wix-style-react/Button';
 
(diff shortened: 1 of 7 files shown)


Faradey27:
@milkyfruit For this example we should use `<CloseButton/>` instead of the <Button children={Icon}> but currently we do not officially have `<CloseButton>`(we have it just for internal usage). So my proposal is to merge this example(as we will not have any regressions), and then after infra team will create public `<CloseButton>`, we will update an example.

@shlomitc @milkyfruit

shlomitc:
yes

PR #1986, opened 2018-07-12
Remove duplicate "Examples" title in auto-stories

hepiyellow:
after adding Examples title to AutoDocs, we can remove the manual Examples heading that we had in some stories.


argshook:
build fails due to eyes, i'm approving and merging, all is ok :)

PR #1993, opened 2018-07-16
Layout component

mariusju:
### What changed
New layouting components were added - Layout and Cell. They are a lightweight alternative to the bootstrap column layout.

### Why it changed
Long-term goal is to fully remove dependency of bootstrap.js if we manage to cover all the use cases with the new layouting components.

---

- [ ] Change is tested
- [+] Change is documented
- [+] Build is green


stories/Layout/samples/RestaurantsLunch.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Layout, Cell} from '../../../src/Layout';
+import styles from '../Example.scss';
+import Card from '../../../src/Card';
+
+export default () => (
+  <div data-hook="restaurants-menus" className={styles.exampleContainer}>
```
lbelinsk: I really think that it will be useful having an additional prop for doing something similar to `stretchViewsVertically` of the bootstrap grid.

In this new example, I would think that developers would want to have the left half and the right half in the same total height:

image

src/Layout/styles.scss
```diff
@@ -0,0 +1,19 @@
+.container {
```
lbelinsk: Bootstrap grid specifies a fixed maxWidth and minWidth. Is it something we want for the css grid as well?

src/Layout/index.js
```diff
@@ -0,0 +1,56 @@
...
+import PropTypes from 'prop-types';
+
+class Layout extends WixComponent {
+  static propTypes = {
+    children: PropTypes.node,
+    gap: PropTypes.string
+  };
+
+  render() {
+    const classes = classNames(styles.container);
```
lbelinsk: You don't really need the `classNames` here as you have only one class.

src/Layout/index.js
```diff
@@ -0,0 +1,56 @@
...
+    const inlineStyle = {};
+
+    if (this.props.gap) {
+      inlineStyle.gridGap = this.props.gap;
```
lbelinsk: Allowing to pass any value here to control the spacing is usually not what we want in wix-style-react as it should specify the uniform look for all backoffice applications, however, in this case, I agree that we should allow such flexibility.

src/Layout/index.js
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import styles from './styles.scss';
+import WixComponent from '../BaseComponents/WixComponent';
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+
+class Layout extends WixComponent {
```
lbelinsk: Can you please create separate files for `Layout.js` & `Cell.js` and then just to export them from this `index.js` file?

lbelinsk:
It looks really good! I think it could really help everyone...

src/Layout/styles.scss
mariusju: <Layout> occupies all the horizontal space available, so that can be accomplished by limiting width on the parent. Actually, every sample uses .example-container, which is constrained by the business-manager values for minWidth and maxWidth

src/Layout/index.js
mariusju: Yes, it should not be changed for the top-level layout. However, I added this prop when I found a nested grid used in the Restaurants - https://wix.invisionapp.com/share/VFMRX6GZ24C#/screens/302323234
This is a valid use case (approved by Domas from UX)

stories/Layout/samples/RestaurantsLunch.js
mariusju: The column occupies all the vertical space. The card is responsible to stretch itself. I added an additional example how this can be done..

src/Layout/Cell.js
```diff
@@ -0,0 +1,29 @@
...
+import styles from './styles.scss';
+import WixComponent from '../BaseComponents/WixComponent';
+import PropTypes from 'prop-types';
+
+export default class Cell extends WixComponent {
+
+  static propTypes = {
+    children: PropTypes.node,
+    span: PropTypes.number,
+    dataHook: PropTypes.string
```
lbelinsk: no need to add dataHook. by inheriting from WixComponent, you get it out of the box.
Also, what do you think about having default `span` value?
Not sure what value it should be.
Or, we can add `isRequired` maybe

src/Layout/Cell.js
```diff
@@ -0,0 +1,29 @@
...
+  render() {
+    const inlineStyle = {
+      gridColumn: `span ${this.props.span}`
+    };
+
+    return (
+      <div
+        style={inlineStyle}
+        className={styles.cell}
+        data-hook={this.props.dataHook}
```
lbelinsk: can remove this data-hook, WixComponent takes care of it for you

src/Layout/Layout.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import styles from './styles.scss';
+import WixComponent from '../BaseComponents/WixComponent';
+import PropTypes from 'prop-types';
+
+export default class Layout extends WixComponent {
+  static propTypes = {
+    children: PropTypes.node,
+    gap: PropTypes.string
```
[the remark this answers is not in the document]
mariusju: changed

src/Layout/Cell.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import styles from './styles.scss';
+import WixComponent from '../BaseComponents/WixComponent';
+import PropTypes from 'prop-types';
+
+export default class Cell extends WixComponent {
+
+  static propTypes = {
+    children: PropTypes.node,
+    span: PropTypes.number
```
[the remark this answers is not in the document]
mariusju: changed

src/Layout/styles.scss
```diff
@@ -0,0 +1,18 @@
+.layout {
+  display: grid;
+  grid-template-columns: repeat(12, 1fr);
+}
+
+// the following is a fallback for IE11,
+// which lacks support for css grid
+@media all and (-ms-high-contrast:none) {
+  .container {
```
shlomitc: shouldn't it be called `layout` to override the layout css?

src/Layout/README.md
```diff
@@ -0,0 +1,34 @@
+# Layout
+The purpose of the following layouting components is to provide a lightweight alternative to the currently used Bootstrap.js grid system.
```
shlomitc: Remove the part explaining why it’s a replacement for Bootstrap.js.

src/Layout/README.md
```diff
@@ -0,0 +1,34 @@
...
+- The layouting system exposes two components - `<Layout>` and `<Cell>`.
+- `<Layout>` is responsible for positioning the child `<Cell>` components.
+- `<Cell>` component defines how many columns out of 12 must this content span.
+- `<Cell>` must be immediate children of the `<Layout>`.
+- The behaviour when other elements than `<Cell>` are used as direct children for `<Layout>` is not specified.
+- To fit the whole horizontal space the span sum of all sibling `<Child>` components should be 12.
+- If the sum of sibling `<Child>` components exceeds 12, the exceeding `<Cell>` components are moved to the next row. It is fine as sometimes it is the desired behaviour. 
+- By nesting multiple `<Layout>` components it is possible to achieve width fractions which are not compliant with the Wix design. It is not advised and such cases should be discussed with the ux.
+
+### Layout
```
shlomitc: Components and API should be in a separate section - prefer an API tab.

stories/Layout/index.js
```diff
@@ -0,0 +1,61 @@
...
+
+import HolyGrail from './samples/HolyGrail';
+import HolyGrailRaw from '!raw-loader!./samples/HolyGrail';
+
+storiesOf('Components/Layout', module)
+  .add('Readme', () => (
+    <Markdown source={Readme}/>
+  ));
+
+storiesOf('Components/Layout/Examples', module)
```
shlomitc: examples should not be “restaurants/Menus”, but more feature oriented.

src/Layout/README.md
```diff
@@ -0,0 +1,34 @@
+# Layout
+The purpose of the following layouting components is to provide a lightweight alternative to the currently used Bootstrap.js grid system.
+
+## Wix Layout Design
+Wix design is based on a 12 column layout. Every content area must be horizontally divided into fractions n/12, where n is a whole number from 1 to 12.
+
+## Components
```
shlomitc: this main page should contain a simple example and explain how to import

src/Layout/Layout.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import styles from './styles.scss';
+
+const Layout = ({children, gap}) =>
```
shlomitc: `margin` prop is missing. it should be a minimum of 48px for the “Layout grid” and 30px for the “Content grid”.
  1. it should have a simple abstraction to define which of the grid type it is (not just a free number)
  2. This will replace the `Card.Content` component that adds pre-defined paddings. that way it can also be used in the MessageBox modal or other widgets.

src/Layout/README.md
```diff
@@ -0,0 +1,34 @@
...
+- The layouting system exposes two components - `<Layout>` and `<Cell>`.
+- `<Layout>` is responsible for positioning the child `<Cell>` components.
+- `<Cell>` component defines how many columns out of 12 must this content span.
+- `<Cell>` must be immediate children of the `<Layout>`.
+- The behaviour when other elements than `<Cell>` are used as direct children for `<Layout>` is not specified.
+- To fit the whole horizontal space the span sum of all sibling `<Child>` components should be 12.
+- If the sum of sibling `<Child>` components exceeds 12, the exceeding `<Cell>` components are moved to the next row. It is fine as sometimes it is the desired behaviour.
```
shlomitc: When not having a `Row` component, it becomes very hard to split your code. for example:

```jsx
<Layout>
  <Col span="4">...</Col>
  <Col span="4">...</Col>
  <Col span="4">...</Col>
  <Col span="4">...</Col>
  <Col span="4">...</Col>
  <Col span="4">...</Col>
</Layout>
```

I can't split it into two separate components that represent two rows, as this will not work:
```jsx
const FirstRow = () => (
  <span>
    <Cell span={4}/>
      <Cell span={4}>
        <Card>
          <Layout margin={"30px"}>
            <Cell span={12}>
              <FormField label="some label" infoContent="this is good">
                <Input/>
              </FormField>
            </Cell>
          </Layout>
        </Card>
      </Cell>
      <Cell span={4}/>
  </span>
);
```
The `span` will destory the structure and fail.

Maybe in React 16 when we use Fragments it works much nicer, but this is not the case here

argshook: maybe it's the name that trips you. `Layout` is pretty much what `Row`
is, but we can't call it `Row` cause that would be a big breaking
change. tell me if you have a better name

your example should be:

```js
const FirstRow = () => (
  <Layout>
    <Cell span={4}/>
      <Layout>
        <Cell span={4}>
          <Card>
            <Card.Content>
              <FormField label="some label" infoContent="this is good">
                <Input/>
              </FormField>
            </Card.Content>
          </Card>
        </Cell>

        <Cell span={4}/>
      </Layout>
  </Layout>
);
```

src/Layout/Layout.js
shlomitc: so two options are on the table:
1. Grid will not do any side margins, so it requires fixing the 30px gap to not apply when the Cell is on the side (only spacing).
2. Grid will do margins and then the `Layout` should have two modes

BTW Currently, the `Page` component allows you to restrict the width, but it does not apply and padding to the page content, which is different from how `Card` behaves (`Card.Content`)

src/Layout/README.md
shlomitc: Why call it `Row` will be a breaking change?

don't know if by intention - but you didn't use the grid inside the `Card.Content` in the example

argshook: > Why call it Row will be a breaking change?

yeah, you're right it wont be breaking. need to consider these:

1. implementation in this PR is `import {Layout, Cell} from 'wix-style-react/Layout';`
1. since `Layout` and `Cell` may be confusing, maybe it should be `import {Row, Col} from 'wix-style-react/Layout'`
1. however current "production" version is `import {Row, Col} from 'wix-style-react/Grid';`
until we eradicate this one, there would be two `Row` and `Col`, which might also be confusing

---

> don't know if by intention - but you didn't use the grid inside the Card.Content in the example

by intention, since `<Layout><Cell span={12}/></Layout>` is same as  ` ` (nothing)

if need more columns than one then use layout else there's no reason for it (it adds two `div`s that serve no purpose)

shlomitc: Whatever you think as long as it's clear for the users and in the docs

PR #1994, opened 2018-07-16
Reproduce docgen bug - TextLink - remove usless path segment

hepiyellow:
Making this PR, only to demonstrate the bug.

It seems like `react-docgen` is not following well composed props.
when I change:
`import TextLinkLayout from '../../BaseComponents/TextLinkLayout';`
to
`import TextLinkLayout from '../TextLinkLayout';`

Then it tried to lookup `TextLinkLayout` under: `wix-style-react/src/Backoffice/TextLinkLayout` instead of `wix-style-react/src/BaseComponents/TextLinkLayout`.

Migrating to yoshi 3, there is an eslint rule:
`import/no-useless-path-segments` which does this fix, and then we hit the bug.


argshook:
thanks! this should fix it https://github.com/wix/react-autodocs-utils/commit/bc4c8f99b3e6701705055592fc7db75bbd7a3e7b

will take a while until it's picked by wix-storybook-utils though

PR #1995, opened 2018-07-16
<MessageBox/> - Improve docs

yardnsm:
### What changed

* Removed the buttons to display the messages
* Use a footer in <MessageBoxFunctionalLayout /> example

### Why it changed

...

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


stories/MessageBox/ExampleStandard.js
```diff
@@ -5,77 +5,51 @@ import {
...
+      <h2>{`Example for <MessageBoxFunctionalLayout/>:`}</h2>
+      <MessageBoxFunctionalLayout
+        title="Look at me please!"
+        primaryButtonLabel="Got It"
+        confirmText="Confirm"
+        cancelText="Cancel"
+        theme="blue"
+        footerBottomChildren={(
+          <div>
+            <Text>Look at me, I am a footer. </Text>
```
shlomitc: It should be a different size according to spec.
size `small` if I remember correctly

stories/MessageBox/ExampleStandard.js
```diff
@@ -5,77 +5,51 @@ import {
...
+
+    <div style={layoutStyles}>
+      <h2>{`Example for <MessageBoxFunctionalLayout/>:`}</h2>
+      <MessageBoxFunctionalLayout
+        title="Look at me please!"
+        primaryButtonLabel="Got It"
+        confirmText="Confirm"
+        cancelText="Cancel"
+        theme="blue"
+        footerBottomChildren={(
```
shlomitc: Please add `box-sizing: border-box;` to the internal footer CSS? I think it's in `FooterLayout.scss`

stories/MessageBox/ExampleStandard.js
```diff
@@ -5,77 +5,51 @@ import {
...
 } from 'wix-style-react/MessageBox';
-import Button from 'wix-style-react/Button';
+import Text from 'wix-style-react/Text';
+import TextLink from 'wix-style-react/TextLink';
+
+const layoutStyles = {
+  margin: 30
+};
+
+export default () => (
```
shlomitc: It would be great to have an example for each one of the scenarios we suggest.
Take a look at this zeplin: https://zpl.io/bzyBXGa. look at `9.1 Alert` (which is the MessageBox).
So I think we need to create one for each, but you can start only with the simple one and the footer one.

yardnsm: Maybe we can add a playground there? So you can also play with theme and the text.

shlomitc: According to the Zeplin specs it should be pretty strict, so this is what should be in the code snippet.

Regarding the playground - we can create an AutoDocs page if it's simple, you can try integrating it.

PR #1998, opened 2018-07-16
WIP - Upgrade to Yoshi 3.x

hepiyellow:
### Rules we have in WSR that I chose to Remove:
  "comma-dangle": [
        "error",
        "never"
      ],
  "indent": 0,
  "indent-legacy": [
    2,
    2,
    {
      "SwitchCase": 1
    }
  ]

- [warning ] react/forbid-foreign-prop-types
- [error] import/no-unresolved

### Rule : `import/no-useless-path-segments`
The is (was) a bug in `react-autodocs-utils` which after fixing errors related to `import/no-useless-path-segments`, would break the storybook for `TextLink`.
For now this rule is disabled. After the patch for `react-autodocs-utils` will be released, we can enabled it.


shlomitc:
another PR is open, so closing #2024

PR #2000, opened 2018-07-16
WIP: Add <FullTextView/> component

Faradey27:
WIP till fix of small bug in wix-ui-backoffice


stories/Highlighter.story.js
```diff
@@ -2,7 +2,7 @@ import Highlighter from 'wix-style-react/Highlighter';
 
 export default {
   category: '12. Other',
-  storyName: '12.4 Highlighter',
+  storyName: 'Highlighter',
```
lbelinsk: sure no index is needed?

src/FullTextView/index.js
```diff
@@ -0,0 +1 @@
+export {default} from './FullTextView.js';
```
lbelinsk: If this is meant to be used outside by users, it should be wrapped with createHOC for the dataHook support

package.json
```diff
@@ -131,7 +131,7 @@
     "react-transition-group": "^2.2.1",
     "slate": "^0.20.3",
     "wix-animations": "^1.0.141",
-    "wix-ui-backoffice": "1.0.274",
+    "wix-ui-backoffice": "^1.0.278",
```
lbelinsk: don't use carret, we decided to have fixed versions

stories/Highlighter.story.js
Faradey27: according to https://app.zeplin.io/project/5864e02695b5754a69f56150/screen/5a1eee7ff4370f472fcaf9d9

src/FullTextView/index.js
Faradey27: but it already wrapped in wix-ui-backoffice

package.json
Faradey27: good catch

src/FullTextView/index.js
lbelinsk: ok

package.json
lbelinsk: ok, so only this

lbelinsk:
build fails, just see why and tou can merge

hepiyellow:
@wuwa @milkyfruit  Where is the spec for this component?

src/Highlighter/Highlighter.e2e.js
```diff
@@ -2,7 +2,7 @@ import eyes from 'eyes.it';
 import {highlighterTestkitFactory, getStoryUrl} from '../../testkit/protractor';
 
 describe('Highlighter', () => {
-  const storyUrl = getStoryUrl('12. Other', '12.4 Highlighter');
+  const storyUrl = getStoryUrl('12. Other', 'Highlighter');
```
hepiyellow: I like to extract a `storySettings.js` file for each component story. And consume it in both story and e2e test.

lbelinsk:
I agree with Erez about the name.
Actually, this is very similar to text component, only that it comes with a tooltip and does smarter dom calculations in order to know when to show the tooltip. We should think about this.
We want to kill the Text and rename it to this complex name? This is the Core text component

shlomitc:
@hepiyellow http://zpl.io/a7GjPPq at the bottom
we'll add a readme later, let's merge it so it won't block Lior and we can decide later how to expose it to the consumers

PR #2001, opened 2018-07-16
Heading

lbelinsk:


stories/Heading.story.js
```diff
@@ -1,27 +1,25 @@
...
-// TODO: autodocs doesn't see ellipsis and forceHideTitle props, because of this line in
-// wix-ui-backoffice/src/components/Heading/Heading.tsx:34
-// const StyledText = withStylable<CoreTextProps, Props>(
-// issue https://github.com/wix/react-autodocs-utils/issues/5
 const propify = value => ({label: value.toString(), value});
 const bools = [false, true].map(propify);
 
 export default {
   category: '1. Foundation',
   storyName: '1.3 Heading',
```
hepiyellow: `1.3` according to zeplin spec is "Shadows". Need to find another number for this. @wuwa

stories/Heading.story.js
```diff
@@ -1,27 +1,25 @@
 import Heading from 'wix-style-react/Heading';
+import {APPEARANCES} from '../src/Heading/Heading';
 
-// TODO: autodocs doesn't see ellipsis and forceHideTitle props, because of this line in
```
hepiyellow: Isn't it because you are using `omit`?

src/Heading/Heading.js
```diff
@@ -0,0 +1,53 @@
...
+    appearance: oneOf(Object.keys(APPEARANCES))
+  };
+
+  static defaultProps = defaultProps;
+
+  state = {tagName: this.props.appearance.toLowerCase()}
+
+  render() {
+    return (
+      <StyledText {...this.props} tagName={this.state.tagName}/>
```
hepiyellow: You are passing here `light` and `appearance` to `CoreText`. And `CoreText` doesn't know what to do with them.

src/Heading/Heading.js
```diff
@@ -0,0 +1,53 @@
...
+  H5: 'H5',
+  H6: 'H6'
+};
+
+const defaultProps = {
+  appearance: APPEARANCES.H1,
+  light: false
+};
+
+const StyledText = withStylable(
```
hepiyellow: No need to use `withStylable`, it only creates another class layer which is not needed here.
Simply in `<Heading>`'s render(), do something like this: 
```
const {light, appearance, ...coreTextProps} = this.props;
return <CoreText {...coreTextProps} {...style('root', {light, appearance}, this.props)} />

File: src/Heading/Heading.js [hunk 5]
```diff
@@ -0,0 +1,53 @@
...
+    /** is the text has dark or light skin */
+    light: bool,
+
+    /** typography of the heading */
+    appearance: oneOf(Object.keys(APPEARANCES))
+  };
+
+  static defaultProps = defaultProps;
+
+  state = {tagName: this.props.appearance.toLowerCase()}
```
hepiyellow: Why do you use `state` here?
If it is only to "cache" the lowerCase tagName so you wouldn't need to compute it every render. I  think it is better to compute it every render, and not have a state at all. In any way, if you do need caching, don't use state, you can have a class member.
Whenever you add state which initializes from props, you need to also implement `componentWillReceiveProps` and update the state.

src/Heading/Heading.e2e.js
```diff
@@ -0,0 +1,16 @@
+import eyes from 'eyes.it';
+import {getStoryUrl, waitForVisibilityOf} from 'wix-ui-test-utils/protractor';
+import {headingTestkitFactory} from '../../testkit/protractor';
+
+describe('Heading', () => {
+  const storyUrl = getStoryUrl('1. Foundation', '1.3 Heading');
+
+  beforeEach(() => browser.get(storyUrl));
+  eyes.it('should display correct content', () => {
```
hepiyellow: Missing snapshots fo ALL variations. (All appearance, and each one with light=false, and light = true)

stories/Heading.story.js
argshook: it used to be typescript `Omit`, now, since Heading transformed into javascript, it's lodash `omit`. Both are related to either `react-docgen-typescript` or `react-docgen` not supporting them.
relevant issue https://github.com/reactjs/react-docgen/issues/283

lbelinsk:
@hepiyellow the coreText is going to be replaced by the `FullTextView`, or something like that.
Will comment on each comment in the right place

lbelinsk:
This component implements the new design spec
image

src/Heading/Heading.js
lbelinsk: Refactored to not use withStylable, and cherry picking the props

src/Heading/Heading.js
lbelinsk: Yes, Done

stories/Heading.story.js
lbelinsk: When it will be updated in the zeplin, I will update as well. Makes sense to me to have it succeeding the 1.2 Text component

src/Heading/Heading.js
lbelinsk: Yes, It is for performance purposes. There are tradeoffs in any solution, will reconsider

src/Heading/Heading.e2e.js
lbelinsk: Will add

hepiyellow:
- @lbelinsk Where are the H5, H6 coming from ? they are not in the spec?
- How is the H1.1 and H2.1 implemented?

lbelinsk:
zpl.io/aMqKgY3

PR #2003, opened 2018-07-17
`<Dropdown/>`, `<Slider/>` - Fix colors

hepiyellow:


src/InputArea/InputArea.scss
```diff
@@ -30,7 +30,6 @@ $gap: 6px;
 }
 
 .errorIcon {
-    color: #f2564d;
```
hepiyellow: This color was redundant, since we add color white a few lines after

hepiyellow:
eyes should have failed, but it didn't.

PR #2005, opened 2018-07-17
<DropdownLayout/> - allow a divider option without an id

### LINKED ISSUE #1983: `<DropdownLayout/>` - divider option without id is not legal
In `DropdownLayout` there is a validity check for `options` which yields INVALID when options include a divider option without id `{value: "-"}`.

This check is done only at `componentWillReceiveProps`.

Need to:
- [ ] fix check to allow divider option without `id`
- [ ] run this check in propType custom validation.

yardnsm:
### What changed

* Allowing a divider option without an `id`
* Allow it also in the component propTypes
* ~~Dividers are rendered with a data hook that matches a different template: https://github.com/wix/wix-style-react/blob/6fa8072def153134c4c2610d64d0f86ace569d1f/src/DropdownLayout/DropdownLayout.js#L222~~
* Updated tests accordingly

### Why it changed

Fixes #1983.

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green

### DIFF
--- a/src/DropdownLayout/DropdownLayout.js
+++ b/src/DropdownLayout/DropdownLayout.js
@@ -219,7 +219,7 @@ class DropdownLayout extends WixComponent {
   renderOption({option, idx}) {
     const {value, id, disabled, title, overrideStyle, linkTo} = option;
     if (value === DIVIDER_OPTION_VALUE) {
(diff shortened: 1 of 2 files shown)


yardnsm:
**Update:** the divider's data-hook has been reverted to the original one, in order to prevent a breaking change.

PR #2009, opened 2018-07-18
Button loading state

Faradey27:
https://github.com/wix-private/wsr-issues/issues/193

### DIFF
--- a/src/Backoffice/Button/Button.driver.js
+++ b/src/Backoffice/Button/Button.driver.js
@@ -1,20 +1,26 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import ReactTestUtils from 'react-dom/test-utils';
+import loaderDriverFactory from '../../Loader/Loader.driver';
 
-export default ({element, wrapper, component}) => ({
-  exists: () => !!element,
-  click: () => ReactTestUtils.Simulate.click(element),
-  focus: () => ReactTestUtils.Simulate.focus(element),
-  blur: () => ReactTestUtils.Simulate.blur(element),
-  mouseEnter: () => ReactTestUtils.Simulate.mouseEnter(element),
-  mouseLeave: () => ReactTestUtils.Simulate.mouseLeave(element),
-  getButtonTextContent: () => element.textContent,
-  isButtonDisabled: () => element.getAttribute('disabled') === '',
-  isPrefixIconExists: () => element.innerHTML.indexOf('prefix') !== -1,
-  isSuffixIconExists: () => element.innerHTML.indexOf('suffix') !== -1,
-  setProps: props => {
-    const ClonedWithProps = React.cloneElement(component, Object.assign({}, component.props, props), ...(component.props.children || []));
-    ReactDOM.render(<div ref={r => element = r}>{ClonedWithProps}</div>, wrapper);
-  }
-});
+export default ({element, wrapper, component}) => {
+  const loaderDriver = loaderDriverFactory({element});
+
+  return {
+    exists: () => !!element,
+    click: () => ReactTestUtils.Simulate.click(element),
+    focus: () => ReactTestUtils.Simulate.focus(element),
+    blur: () => ReactTestUtils.Simulate.blur(element),
+    mouseEnter: () => ReactTestUtils.Simulate.mouseEnter(element),
+    mouseLeave: () => ReactTestUtils.Simulate.mouseLeave(element),
+    getButtonTextContent: () => element.textContent,
+    isButtonDisabled: () => element.getAttribute('disabled') === '',
+    isPrefixIconExists: () => element.innerHTML.indexOf('prefix') !== -1,
+    isSuffixIconExists: () => element.innerHTML.indexOf('suffix') !== -1,
+    loading: () => loaderDriver.exists(),
+    setProps: props => {
+      const ClonedWithProps = React.cloneElement(component, Object.assign({}, component.props, props), ...(component.props.children || []));
+      ReactDOM.render(<div ref={r => element = r}>{ClonedWithProps}</div>, wrapper);
+    }
+  };
+};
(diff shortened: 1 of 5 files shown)


src/Backoffice/Button/Button.js
```diff
@@ -41,32 +43,43 @@ class Button extends WixComponent {
...
-    this.addIcon(styles.prefix, this.props.prefixIcon, this.props.height);
+    this.addIcon(styles.prefix, this.props.prefixIcon, this.props.height, 'prefix');
 
   addSuffix = () =>
-    this.addIcon(styles.suffix, this.props.suffixIcon, this.props.height);
+    this.addIcon(styles.suffix, this.props.suffixIcon, this.props.height, 'suffix');
+
+  renderButtonContent() {
+    if (this.props.loading) {
+      return <Loader size="tiny"/>;
```
hepiyellow: ## size tiny
Shouldn't the Loader be bigger when the button size is "medium" or "large"? (According to spec it is all tiny, I would ask @milkyfruit  if this is a mistake)

## Button width is changing when in `loading=true`
In the storybook, when I toggle the `loading` to true, then the button width is changing (becomes smaller). I guess it is because we don't render the text (child). To solve that (if it is a requirement) we need render the text and suffixes, but make then `visible:hidden` .

src/Backoffice/Button/Button.js
```diff
@@ -41,32 +43,43 @@ class Button extends WixComponent {
...
   addPrefix = () =>
-    this.addIcon(styles.prefix, this.props.prefixIcon, this.props.height);
+    this.addIcon(styles.prefix, this.props.prefixIcon, this.props.height, 'prefix');
 
   addSuffix = () =>
-    this.addIcon(styles.suffix, this.props.suffixIcon, this.props.height);
+    this.addIcon(styles.suffix, this.props.suffixIcon, this.props.height, 'suffix');
+
+  renderButtonContent() {
+    if (this.props.loading) {
```
hepiyellow: ### Missing e2e test
Missing snapshot in loading state.

### Missing specs for Loader with different `themes`
@milkyfruit Is the loader supposed to be blue/white also when the button has other themes (`Distructive`, `Primium`, `Standard+Secondary`, etc...)

src/Backoffice/Button/Button.spec.js
```diff
@@ -84,6 +84,11 @@ describe('Button', () => {
     expect(driver.isPrefixIconExists()).toBeFalsy();
     expect(driver.isSuffixIconExists()).toBeTruthy();
   });
+
+  it('should show loader when loading', () => {
+    const driver = createDriver(<Button loading/>);
+    expect(driver.loading()).toBeTruthy();
```
hepiyellow: Currently the `driver.loading()` always returns `true` (Bug in driver).
All driver methods like this should have a test that they can actually return also `false`.

src/Backoffice/Button/Button.driver.js
```diff
@@ -1,20 +1,26 @@
...
-  isButtonDisabled: () => element.getAttribute('disabled') === '',
-  isPrefixIconExists: () => element.innerHTML.indexOf('prefix') !== -1,
-  isSuffixIconExists: () => element.innerHTML.indexOf('suffix') !== -1,
-  setProps: props => {
-    const ClonedWithProps = React.cloneElement(component, Object.assign({}, component.props, props), ...(component.props.children || []));
-    ReactDOM.render(<div ref={r => element = r}>{ClonedWithProps}</div>, wrapper);
-  }
-});
+export default ({element, wrapper, component}) => {
+  const loaderDriver = loaderDriverFactory({element});
```
hepiyellow: The `loaderDriverFactory` should not be passed the Button's root element, but rather need to add a `data-hook` to the `<Loader/>` and find the Loader element by the data-hook, and pass **it** to `loaderDriverFactory`.

src/Backoffice/Button/Button.driver.js
```diff
@@ -1,20 +1,26 @@
...
+    click: () => ReactTestUtils.Simulate.click(element),
+    focus: () => ReactTestUtils.Simulate.focus(element),
+    blur: () => ReactTestUtils.Simulate.blur(element),
+    mouseEnter: () => ReactTestUtils.Simulate.mouseEnter(element),
+    mouseLeave: () => ReactTestUtils.Simulate.mouseLeave(element),
+    getButtonTextContent: () => element.textContent,
+    isButtonDisabled: () => element.getAttribute('disabled') === '',
+    isPrefixIconExists: () => element.innerHTML.indexOf('prefix') !== -1,
+    isSuffixIconExists: () => element.innerHTML.indexOf('suffix') !== -1,
+    loading: () => loaderDriver.exists(),
```
hepiyellow: Maybe rename to `isLoaderExists`

src/Backoffice/Button/Button.js
```diff
@@ -41,32 +43,43 @@ class Button extends WixComponent {
...
       icon ?
-        <div className={className} data-hook={dataHook}>
+        <div className={className} key={key} data-hook={dataHook}>
           {React.cloneElement(icon, {size: iconSize})}
         </div> :
         null
     );
   }
 
   addPrefix = () =>
```
hepiyellow: Now that we have `renderButtonContent`, we can make inline this `addPrefix`, since it is used only once.

src/Backoffice/Button/Button.js
```diff
@@ -41,32 +43,43 @@ class Button extends WixComponent {
...
-    this.addIcon(styles.prefix, this.props.prefixIcon, this.props.height);
+    this.addIcon(styles.prefix, this.props.prefixIcon, this.props.height, 'prefix');
 
   addSuffix = () =>
```
hepiyellow: Now that we have `renderButtonContent`, we can make inline this `addSuffix`, since it is used only once.

src/Backoffice/Button/Button.js
milkyfruit: the loader always stays `tiny`.
@hepiyellow is right about `visible:hidden`, it should inherit the width from a state when it's with text.

src/Backoffice/Button/Button.js
milkyfruit: I think the loader looks ok with the same blue style in all the possible button themes:
image

PR #2012, opened 2018-07-18
fix wrong icons path

shlomitc:
fix broken icons path


Faradey27:
@shlomitc thank you, have no idea how i miss this again

PR #2019, opened 2018-07-22
`<Table/>` - Table/rounded corners

hepiyellow:
- Fix `<Card/> remove `over-flow` mistake, and don't use overflow at all.
- Apply Card "looks" inside `<TableToolbar/>` and `<TableContent/>`
- `<Table/>` Add `openEnded` prop - that makes it look like there are more rows (no lastRowDivider, and no bottom rounded corners)


hepiyellow:
This PR is freezed, because it introduces a mix of  both Sass and Stylable stylesheets for the same component (DataTable). This is potentially problematic and should be avoided.
**We should first convert DataTable's Sass to Stylable.**

shlomitc:
I'm closing it for now.
We might have some changes in the design so I think we should reconsider it when it again when it becomes relevant

PR #2020, opened 2018-07-22
Storybook-infra:  Add offline mode

hepiyellow:
Allows running storybook when there is no internet connection.
(For working on the train / plain)

Add a script:
`npm run storybook:offline`

- Takes React from node_modules
- Caches the `fontFace.css` file, so that next run if file exists then it is taken from local folder.


argshook:
@hepiyellow 
it's nice idea but i think it can be done in a way that is much more simple.

`start-storybook` supports `-s` (or `--static-dir`) flag for static
assets to be served. i'll PR how this can be done for `react` and
`react-dom` but for fonts it's more problematic.

`fontFace.css` file includes links to fonts but those fonts are hosted at http://static.parastorage.com/services/third-party/fonts/Helvetica/ and so downloading `fontFace.css` is rather useless.

argshook:
oh btw, we can simply do this as we already have all fonts under `assets` https://github.com/wix/wix-style-react/pull/2025/commits/81cf2c59cf5154f3dbb20643b9c91042ab7f1605

package.json
```diff
@@ -83,6 +84,7 @@
     "gh-pages-auto-release": "^1.1.2",
     "github-markdown-css": "^2.4.1",
     "highlight.js": "^9.8.0",
+    "http-server": "^0.11.1",
```
argshook: @hepiyellow 
i'm ok with all that except for new dependency. perhaps instead of `http-server` we can do `python -m http.server`? i'm pretty sure everyone has python these days

hepiyellow: @argshook 
```
$ python -m http.server 8080
/usr/local/opt/python@2/bin/python2.7: No module named http
```
I prefer something that works for everyone. hasle free.

argshook: this is `npm ls` of just `http-server`:

```
└─┬ http-server@0.11.1
  ├── colors@1.0.3
  ├── corser@2.0.1
  ├─┬ ecstatic@3.3.0
  │ ├── he@1.1.1
  │ ├── mime@1.6.0
  │ ├── minimist@1.2.0
  │ └── url-join@2.0.5
  ├─┬ http-proxy@1.17.0
  │ ├── eventemitter3@3.1.0
  │ ├─┬ follow-redirects@1.5.8
  │ │ └─┬ debug@3.1.0
  │ │   └── ms@2.0.0
  │ └── requires-port@1.0.0
  ├── opener@1.4.3
  ├─┬ optimist@0.6.1
  │ ├── minimist@0.0.10 extraneous
  │ └── wordwrap@0.0.3
  ├─┬ portfinder@1.0.17
  │ ├── async@1.5.2
  │ ├── debug@2.6.9 extraneous
  │ └─┬ mkdirp@0.5.1
  │   └── minimist@0.0.8 extraneous
  └─┬ union@0.4.6
    └── qs@2.3.3
```

it's 21 dependencies, far from hassle free.

everybody has python, too bad mac ships with version from 10 years ago. here's an `if`:

```sh
#!/usr/bin/env sh

VERSION=`python --version | awk '{print $2}' | awk -F '.' '{print $1}'`

if [ $VERSION -eq "3" ]; then
  python -m http.server 8080
else
  python -m SimpleHTTPServer 8080
fi
```

hepiyellow: cool. ok. I'll do it.

.gitignore
```diff
@@ -9,6 +9,7 @@ coverage
 .history
 storybook-static
 storybook-e2e
+.storybook-offline-config
```
shlomitc: why ignoring and copying it again and again instead of creating it once?

hepiyellow: Because I don't want to create 2 storybook configurations which need to be maintained.
This way we merge (on-the-fly) only the require changes (for offline work) into the current storybook configuration.

hepiyellow: @shlomitc Is this acceptable? can I merge this?

shlomitc:
I created this PR: https://github.com/wix/wix-style-react/pull/2321

Reasons to prefer my solution:
1. configurations are shared and not copied
2. using storybook to serve static files  - no extra dependencies nor assumptions on the OS)

shlomitc:
I'm closing this PR in favor of my PR.

PR #2023, opened 2018-07-23
`<Table/>` - Fix TableToolbar import path. Move Table Toolbar to own component folder TableToolbar.

hepiyellow:
Table examples included an import 
```
 import {...} from 'wix-style-react/Table/Toolbar';
```

Which was broken and didn't work for an actual consumer of WSR.

To fix it:
- I moved TableToolbar from a folder inside `Table`, to it's own folder (under `components`).
- also moved the it's story.

So now the import consumer should use is:
```
import {...} from 'wix-style-react/TableToolbar';
```

### Backward compatibility
In case people have already started using the TableToolbar, and imported it from `wix-style-react/dist/src/components/Table/Toolbar`,
then I added backward compatibility proxies, with deprecation logs.


src/Table/Toolbar/TableToolbar.js
```diff
@@ -1,38 +1,4 @@
...
-export const TableToolbar = Toolbar;
-
-// Aliases for convenience
-TableToolbar.ItemGroup = ItemGroup;
-TableToolbar.Item = Item;
-TableToolbar.Label = Label;
-TableToolbar.SelectedCount = SelectedCount;
-TableToolbar.Title = Title;
-TableToolbar.Divider = Divider;
+deprecationLog('Importing TableToolbar from wix-style-react/dist/src/components/Table/Toolbar/TableToolbar is deprecated. Please import from wix-style-react/TableToolbar');
```
shlomitc: just break them, no need for a deprecation log here.

PR #2025, opened 2018-07-23
chore(storybook): serve react & react-dom from local

argshook:
example for how we could make storybook work offline

with this, however, we're losing react & react-dom caching. To avoid that we'd need something more complicated and i'm not sure it's really that important

@hepiyellow


argshook:
closing for now

@yardnsm fyi, maybe it's gonna be useful for you if you'll work on offline storybook during infra day

PR #2031, opened 2018-07-25
<Table/> - Implement action column

yardnsm:
The action column is the last column in the table, which consists of (at least one) primary action, and can have multiple secondary actions. The primary action is visible only when hovering the row; the secondary action have the same behaviour by default but it can be changed (with `alwaysShowSecondaryActions`).

There are "visible" secondary actions, which are shown in the cell, and there are "hidden" secondary actions which goes into a `<PopoverMenu/>`. The number of visible secondary actions can be set using the `visibleSecondaryActions` prop.

The primary action is triggered when pressing the primary action button or the row itself, and a secondary action is triggered by pressing it.

### Problems and stuff to consider:

* Keyboard navigation: using the Tab key on a row should trigger the row hovering, thus showing the primary and secondary action.

* API: Currently, the behaviour of the action column is encapsulated in `<TableActionColumn/>`. The `<Table/>` component basically adds a new item to the `column` prop which renders the action cell. The properties of the primary and secondary action are very specific. A few option are:
	
	* Keep the API as is if we want the consumers to follow a "strictier" guidelines.
	* Let the users add the action column themselves, by providing the `<TableActionColumn/>` component. The `<Table/>` component won't receive any action-column-specific props. This will allow the users to perform, for example, conditional rendering of the action cell depending on the row content.
	* Letting the users implement their own action column, like we did with `<TableToolbar/>`. We can provide them components like: `<PrimaryActionItem/>`, `<SecondaryActionsGroup/>`, `<SecondaryActionItem/>`, etc.

	IMO option 2 is the best since we're giving the users a bit "freedom" to play with action column, but still making sure they'll follow the UX guidelines.

* Width calculation: currently, the column's width is set to an arbitrary value (`40%`). When the table is getting small, it doesn't look well:

	screen shot 2018-07-25 at 11 41 50
 
	A few options are:

	* Letting the users set the width themselves. This is currently working using the `actionColumnWidth` props.
	* Calculate the desired with of the action column and set a fixed width for it.
	* Try to put more secondary actions in the `<PopoverMenu/>` if there is no room (setting the value of `visibleSecondaryActions` according to the table's width).

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


argshook:
i said these thing during the meeting, but will summarize here

instead of taking care of many cases it would be much easier for both
maintainers and consumers to simply allow custom content in action area.
sort of inversion of control, if you will.

now this implementation talks about primary and secondary actions, their
click handling, data passing. Then we need to think about width and how
to squeeze items if they don't fit. What happens when consumers want
different actions for rows. What if one row is "disabled" and some
actions shouldn't show up or have different styling?

this imo doesn't scale well and could very well become maintenance
nigthmare.

---

and so instead i would go with different approach: improve table rows to
support `actions` area for any kind of content.

this way we have only one problem to take care: how to consistently
layout action elements so that consumer doesn't have to?

one possible solution: `if actions prop is array then wrap each with div[margin="whatever"] else render as is`

---

i understand there was quite some effort put into this but im trying to
to imagine ahead how it will be used, how strict it is and how tough new
requirements will be to handle

yardnsm:
@argshook I agree. As I said, I think its the best to allow the consumers to render the action column themselves. We can still provide the `<TableActionColumn/>` component, and it could be used as follows:

```js
import {Table} from 'wix-style-react/Table';
import ActionColumn from 'wix-style-react/Table/ActionColumn';

// ...

const MyTable = () => (
	<Table
		columns={[
			// ...
			{title: '', width: '40%', render: rowData => (
				<ActionColumn
					primaryAction={...}
					secondaryActions={[...]}
					visibleSecondaryAction={2}
					/>
			)}
		]}
		/>
);

// ...
```


> improve table rows to support actions area for any kind of content.

Wdym by that in practice? Smth like passing the content (node/array of nodes) as a prop to `<Table/>`?

> that's not good, what if i want to select and copy some text and instead primary action is triggered. 

Click on the row to trigger the primary action was a UX requirement :)

hepiyellow:
The suggested Usage for the Table actions:

```js
import {Table} from 'wix-style-react/Table';
import ActionColumn from 'wix-style-react/Table/ActionColumn';

// ...
const mainAction=(rowData)=> {console.log('clicked');
const MyTable = () => (
  <Table
    onRowClick={mainAction}
    columns={[
     // ...
     {  title: '', width: '40%',
        render: rowData, isHover => (
          <TableActions>
            {isHover &&  (
                 <Button onClick={mainAction}> //Main Action
                 <ButtonIcon onClick={secondaryAction}/>
                 <PopoverMenu>...</PopoverMenu>
             )}
          </TableActions>
        )}
	]}
/>
);

**Comment by yardnsm** [comment]
There are couple of problems with the `isHover` approach. First, on every mouse event (when the value of `isHover` changes), you'll have to re-render the cell, which is not best for performance.
Second, I've experienced an issue with mouse events, where you have the `<PopoverMenu/>` opened. When the user hovers the `<PopoverMenu/>`, the `onMouseLeaveRow` callback kicks in, and the row is considered not hovered.

My suggestion in to use css for the hovering thing, and have the `<PopoverMenu/>` with `appendToParent` so the hover event will persist even when the use hovers the `<PopoverMenu/>`:

```js
import {Table} from 'wix-style-react/Table';
import ActionColumn from 'wix-style-react/Table/ActionColumn';

// ...

const mainAction = (rowData) => {/* ... */};
const secondaryAction = (rowData) => {/* ... */};

const MyTable = () => (
	<Table
		onRowClick={mainAction}
		columns={[
			{
				title: '',
				width: '40%',
				render: rowData, () => (
					<ActionColumn showOnlyWhenHovering>
						{isHover &&  (
							<Button onClick={mainAction}>
							<ButtonIcon onClick={secondaryAction}/>
							<PopoverMenu>{/* ... */}</PopoverMenu>
						)}
					</ActionColumn>
				)}
			]}
		/>
);
```

yardnsm:
We've decided to choose the following as the final usage:

<details>
  <summary>Final usage</summary>

```jsx
import TableActionCell from 'wix-style-react/TableActionCell';

const primaryAction = () => { /* ... */ };

<Table
  onRowClick={primaryAction}
  columns={[
    {
        title: '',
        width: '40%',
        render: rowData => (
          <TableActionCell
            primaryAction={{
              name: 'Edit',
              theme: 'fullblue',
              trigger: () => primaryAction(rowData)
            }}
            secondaryAction={
              [
                name: 'Print',
                icon: <Print />
                trigger: () => { /* ... */ },
              ],
            },
            visibleActions={...}
            /* ... */
          />
        )
    }
  ]}
  />
```
</details>
<br />

I'll try so summarise the reasons for this kind of API:

This API is a bit strict, as it handles most of the logic and the design. This way we can make sure the consumers will follow the UX guidelines. However, this usage does not _completely_ take away the consumers' freedom, as they can simply render anything inside `render`. If there is a need for different actions for a specific row, the consumer can add a conditional rendering in `render` (although I think this kind of behaviour is forbidden according to the guidelines). 

The `<TableActionCell/>` also handles some specific logic, such as:

* Wrapping `onClick` events and calling `stopPropagation` for each action;
* Showing a placeholder icon when there is only a primary action;
* Adding `<Tooltip/>`s to visible secondary actions;
* Theming the buttons and tooltips so they match the guidelines.

We'll also want to implement *more* complex logic in the future, such as keyboard navigation and width calculation.

We can't expect consumers to handle this kind of logic. Even if we'll provide a snippet, it will be huge and ugly. What will happen when the guidelines will change?

Also, this API takes away the awareness of the hover state (the `isHover` thing). Does the consumers need this? They can implement this by themselves using mouse events.

src/TableActionCell/TableActionCell.js
```diff
@@ -0,0 +1,161 @@
...
+  /**
+   * An object containing the primary action properties: `name` is the action
+   * name (the text of the button), `theme` is the button theme (can be
+   * `whiteblue` or `fullblue`), `onActionTrigger` is the callback function for
+   * the action, whose signature is `onActionTrigger(rowData, rowNum)`.
+   */
+  primaryAction: PropTypes.shape({
+    name: PropTypes.string.isRequired,
+    theme: PropTypes.oneOf(['whiteblue', 'fullblue']),
+    onActionTrigger: PropTypes.func.isRequired
```
hepiyellow: Rename: I would prefer not having `trigger`. Either `onClick`, `onAction`, `onActionClick`
(Same for secondary actions)
I prefer `onClick`, since it is the closest to the Button api, and main action does become a button.

src/TableActionCell/TableActionCell.js
```diff
@@ -0,0 +1,161 @@
...
+  dataHook: PropTypes.string,
+
+  /**
+   * An object containing the primary action properties: `name` is the action
+   * name (the text of the button), `theme` is the button theme (can be
+   * `whiteblue` or `fullblue`), `onActionTrigger` is the callback function for
+   * the action, whose signature is `onActionTrigger(rowData, rowNum)`.
+   */
+  primaryAction: PropTypes.shape({
+    name: PropTypes.string.isRequired,
```
hepiyellow: ## Rename
Also here, I would not use `name` since it might imply that it has some semantic meaning.
I prefer `text`.
We used this terminology in WSR's SectionHelper, and in WUB FloatingHelperContent, when we get action props and create a Button from them.

src/TableActionCell/TableActionCell.js
```diff
@@ -0,0 +1,161 @@
...
+    theme: PropTypes.oneOf(['whiteblue', 'fullblue']),
```
hepiyellow: ### Rename
In the new WUB Button, we don't use the `theme` terminology anymore.
We use `skin`, and the values are `standard` (for blue) and `white`.
Let's try to stay close to the Button api.

src/TableActionCell/TableActionCell.js
```diff
@@ -0,0 +1,161 @@
...
+   * signature is `onActionTrigger(rowData, rowNum)`.
+   */
+  secondaryActions: PropTypes.arrayOf(PropTypes.shape({
+    name: PropTypes.string.isRequired,
+    icon: PropTypes.node.isRequired,
+    onActionTrigger: PropTypes.func.isRequired
+  })),
+
+   /** The number of secondary actions to show outside the PopoverMenu */
+  visibleSecondaryActions: PropTypes.number,
```
hepiyellow: ### Rename
`numOfVisibleSecondaryActions`

src/TableActionCell/TableActionCell.js
```diff
@@ -0,0 +1,161 @@
...
+
+   /** Whether to show the secondary action also when not hovering the row */
+  alwaysShowSecondaryActions: PropTypes.bool
```
hepiyellow: Just wanted to clarify that this affects both  the visible secondary actions and the secondary actions in the popover-menu.

src/TableActionCell/TableActionCell.e2e.js
```diff
@@ -0,0 +1,45 @@
...
+    const element = byDataHook(dataHook);
+
+    await waitForVisibilityOf(element, `Cannot find ${dataHook}`);
+    await scrollToElement(element);
+
+    await eyes.checkWindow(dataHook);
+    hoverElement(element);
+    await eyes.checkWindow(dataHook);
+
+    return tableActionCellTestkitFactory({dataHook});
```
hepiyellow: I think it is better to first create the testkit (instead of doing `const element = byDataHook(dataHook);`),
the do: `await waitForVisibilityOf(driver.element(), `Cannot find ${dataHook}`);`.
That way we also test that the testkit factory worked.

src/TableActionCell/TableActionCell.scss
```diff
@@ -0,0 +1,37 @@
+tr .actionsContainer {
+  display: flex;
+  align-items: center;
+  justify-content: flex-end;
+  height: 36px;
```
hepiyellow: The button and icons height is 30px.
Is there any reason to make this container 6px heigher?

src/TableActionCell/TableActionCell.e2e.js
```diff
@@ -0,0 +1,45 @@
...
+
+  const examples = {
+    'Blue primary action': 'story-primary-blue',
+    'White primary action': 'story-primary-white',
+    'Primary and secondary actions': 'story-primary-secondary',
+    'Primary and hidden secondary actions': 'story-primary-hidden-secondary',
+    'Always visible secondary actions': 'story-always-visible-secondary',
+    'Only secondary actions': 'story-only-secondary',
+    'Only visible secondary actions': 'story-only-visible-secondary'
+  };
```
hepiyellow: How about an rtl example snapshot?

src/TableActionCell/TableActionCell.scss
```diff
@@ -0,0 +1,37 @@
...
+  position: relative;
+}
+
+tr .actionsContainer button {
+  margin: 0 6px;
```
hepiyellow: Doesn't this affect also the PopoverMenu buttons?
Is that intentional? (by the spec)

src/TableActionCell/HoverSlot/HoverSlot.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import styles from '../TableActionCell.scss';
```
hepiyellow: Use Stylable

src/TableActionCell/TableActionCell.scss
```diff
@@ -0,0 +1,37 @@
...
+}
+
+tr .actionsContainer .hoverSlot.onHover,
+tr:hover .actionsContainer .hoverSlot.notOnHover {
+  visibility: hidden;
+}
+
+tr:hover .actionsContainer .hoverSlot.onHover,
+tr .actionsContainer .hoverSlot.notOnHover {
+  visibility: visible;
```
hepiyellow: `visibility: visible` is the default. I think you can delete this rule.
Unless you need it to override (by being more specify) some higher level rule that might set `visibility: hidden`, which I am not sure will work.

src/TableActionCell/TableActionCell.scss
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// Primary action placeholder
+tr .actionsContainer .placeholderIcon {
```
hepiyellow: no need for `tr ` and `.actionsContainer`.
In general, no need for `tr` unles when using `tr:hover`

src/TableActionCell/TableActionCell.js
```diff
@@ -0,0 +1,161 @@
...
+import Tooltip from '../Tooltip/Tooltip';
+import Button from '../Button';
+import PopoverMenu from '../PopoverMenu';
+import PopoverMenuItem from '../PopoverMenuItem';
+import ChevronRight from '../new-icons/ChevronRight';
+
+/* eslint-disable react/prop-types */
+function renderPrimaryAction({name, theme, onActionTrigger}) {
+  return (
+    <Button
```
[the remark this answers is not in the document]
shlomitc: No. Keep using the WSR button

src/TableActionCell/TableActionCell.scss
yardnsm: I know, but I wanted to keep to context clear so it'll be better to understand that the first selector is for "non-hovered" state while to next one is for the "hovered" state.

src/TableActionCell/TableActionCell.scss
yardnsm: Its actually `36px`, not `30px`.

src/TableActionCell/TableActionCell.scss
yardnsm: We are overriding it, take a look at the rules above this one.

src/TableActionCell/TableActionCell.js
```diff
@@ -0,0 +1,161 @@
...
+
+import styles from './TableActionCell.scss';
+import HoverSlot from './HoverSlot';
+import Tooltip from '../Tooltip/Tooltip';
+import Button from '../Button';
+import PopoverMenu from '../PopoverMenu';
+import PopoverMenuItem from '../PopoverMenuItem';
+import ChevronRight from '../new-icons/ChevronRight';
+
+/* eslint-disable react/prop-types */
```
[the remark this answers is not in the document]
yardnsm: Yep, it thinks that this is a SFC so it gets angry.

src/TableActionCell/TableActionCell.e2e.js
```diff
@@ -0,0 +1,45 @@
...
+  const storyUrl = createStoryUrl({kind: storySettings.kind, story: storySettings.storyName});
+
+  const verifyItem = async dataHook => {
+    const element = byDataHook(dataHook);
+
+    await waitForVisibilityOf(element, `Cannot find ${dataHook}`);
+    await scrollToElement(element);
+
+    await eyes.checkWindow(dataHook);
+    hoverElement(element);
```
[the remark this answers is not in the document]
yardnsm: The 'hoverElement` function is not `async`

src/TableActionCell/TableActionCell.js
yardnsm: We're no using the WUB button here. Should we make this change anyways? And convert the new values to the old one?

src/TableActionCell/HoverSlot/HoverSlot.js
yardnsm: I guess that Stylable is not an option...

screen shot 2018-09-02 at 16 08 41

src/TableActionCell/TableActionCell.scss
hepiyellow: What I meant is that the selector of the rule above and this rule are mutually exclusive.
(Meaning: `tr` and `tr:hover` can not apply at the same time)

src/Table/Table.js
```diff
@@ -55,7 +66,10 @@ export function getDataTableProps(tableProps) {
           'newDesign',
           'hideHeader',
         ),
-    newDesign: true
+    newDesign: true,
+    rowClass: classNames(tableProps.rowClass, {
+      [styles.rowHover]: !!tableProps.onRowClick
```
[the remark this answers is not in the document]
yardnsm: Only if there is a primary action

src/Table/Table.st.css
```diff
@@ -0,0 +1,53 @@
...
+:import {
+  -st-from: '../TableActionCell/TableActionCell.st.css';
+  -st-default: TableActionCell;
+}
+
+.root {
+  -st-states: clickable;
+}
+
+.emptyStateContainer {
```
hepiyellow: Need to verify with UX, if this empty state should include title bar or not.
anyways, why do we have the `border-top` here? (does it depend on the above question?)

src/TableActionCell/TableActionCell.spec.js
```diff
@@ -0,0 +1,216 @@
...
+};
+
+
+describe('Table Action Cell', () => {
+  const createDriver = createDriverFactory(tableActionCellDriverFactory);
+
+  beforeEach(() => {
+    // The action column uses <Tooltip/> and <PopoverMenu/> under the hood,
+    // which renders straight into document.body, this we need to clear it.
+    document.body.innerHTML = '';
```
hepiyellow: Are you sure?

src/TableActionCell/TableActionCell.spec.js
```diff
@@ -0,0 +1,216 @@
...
+  });
+
+  it('should display the action column for primary action', () => {
+    const driver = createDriver(<TableActionCell {...primaryActionProps()}/>);
+    expect(driver.exists()).toBeTruthy();
```
hepiyellow: not clear what are you testing here. or maybe the test title is wrong?
`driver.exists()` is always true

src/TableActionCell/TableActionCell.driver.js
```diff
@@ -0,0 +1,63 @@
...
+    }).init.menuItemDataHook('table-action-cell-popover-menu-item')
+      .init.parentElement(element);
+
+  return {
+    /** Get the element */
+    element: () => element,
+    /** Whether the element exists */
+    exists: () => !!element,
+    /** Get the primary action placeholder element */
+    getPrimaryActionPlaceholder,
```
hepiyellow: This should be a private.driver method. It is not interesting to any consumer.

src/TableActionCell/TableActionCell.driver.js
```diff
@@ -0,0 +1,63 @@
...
+    /** Get the driver of the primary action <Button/> from the action column */
+    getPrimaryActionButtonDriver,
+    /** Click the primary action button from the action column */
+    clickPrimaryActionButton: () => getPrimaryActionButtonDriver().click(),
```
hepiyellow: Consider removing all the `click` methods. 
You are exposing the drivers anyways.
These `click` methods make the driver a bit too verbose... (scary on first sight)

src/TableActionCell/TableActionCell.spec.js
```diff
@@ -0,0 +1,216 @@
...
+        />
+    );
+
+    driver.clickPrimaryActionButton(0);
+    expect(onPrimaryActionTrigger).toHaveBeenCalledTimes(1);
+  });
+
+  it('should display the action column for secondary actions', () => {
+    const driver = createDriver(<TableActionCell {...secondaryActionsProps()}/>);
+    expect(driver.exists()).toBeTruthy();
```
hepiyellow: `driver.exists()` is always true

src/TableActionCell/TableActionCell.spec.js
```diff
@@ -0,0 +1,216 @@
...
+  it('should display the primary action button', () => {
+    const onPrimaryActionTrigger = jest.fn();
+
+    const driver = createDriver(
+      <TableActionCell
+        {...primaryActionProps(onPrimaryActionTrigger)}
+        />
+    );
+
+    expect(driver.getPrimaryActionButtonDriver(0).exists()).toBeTruthy();
```
hepiyellow: no need to pass `0` to `getPrimaryActionButtonDriver`.

src/TableActionCell/TableActionCell.spec.js
```diff
@@ -0,0 +1,216 @@
...
+  it('should trigger the primary action on primary button click', () => {
+    const onPrimaryActionTrigger = jest.fn();
+
+    const driver = createDriver(
+      <TableActionCell
+        {...primaryActionProps(onPrimaryActionTrigger)}
+        />
+    );
+
+    driver.clickPrimaryActionButton(0);
```
hepiyellow: no need to pass 0

src/TableActionCell/TableActionCell.spec.js
```diff
@@ -0,0 +1,216 @@
...
+
+  it('should put visible secondary actions in the cell', async () => {
+    const driver = createDriver(
+      <TableActionCell
+        {...primaryActionProps()}
+        {...secondaryActionsProps()}
+        />
+    );
+
+    expect(driver.getVisibleActionsCount(0)).toEqual(2);
```
hepiyellow: no need to pass 0

src/TableActionCell/TableActionCell.spec.js
```diff
@@ -0,0 +1,216 @@
...
+
+    expect(driver.getVisibleActionButtonDriver(0).getButtonTextContent()).toEqual('Icon 0');
+    expect(driver.getVisibleActionButtonDriver(1).getButtonTextContent()).toEqual('Icon 1');
+
+    const tooltipDriver1 = driver.getVisibleActionTooltipDriver(0);
+    const tooltipDriver2 = driver.getVisibleActionTooltipDriver(1);
+
+    tooltipDriver1.mouseEnter();
+    await resolveIn(300);
```
hepiyellow: Don't use fixed timeouts in tests.
You can use `wix-eventually` to do a short interval polling on the expectation.
e.g.
```
import * as eventually from 'wix-eventually';
...
tooltipDriver1.mouseEnter();
await eventually(()=>expect(tooltipDriver1.getContent()).toEqual('Action 0'));
```

Fix this for all usage of `resolveIn`

src/TableActionCell/TableActionCell.spec.js
```diff
@@ -0,0 +1,216 @@
...
+  it('should allow to change the number of visible secondary actions', async () => {
+    const driver = createDriver(
+      <TableActionCell
+        {...primaryActionProps()}
+        {...secondaryActionsProps()}
+        numOfVisibleSecondaryActions={3}
+        />
+    );
+
+    expect(driver.getVisibleActionsCount(0)).toEqual(3);
```
hepiyellow: no need to pass 0

src/TableActionCell/TableActionCell.spec.js
```diff
@@ -0,0 +1,216 @@
...
+
+    driver.clickPopoverMenu(0);
```
hepiyellow: Is it necessary to click the PopoverMenu in order to call `driver.getHiddenActionsCount()`?
If so, then document it, if not then remove the click.

src/Table/Table.st.css
yardnsm: According to the Zeplin, this is the correct behaviour. Regarding the border, we need to set it since originally it comes from the table's TitleBar. Since we can't have it with EmptyState, we need to set this border.

src/TableActionCell/TableActionCell.spec.js
yardnsm: It was needed when we had the PopoverMenu appended to body, now we can remove it

src/TableActionCell/TableActionCell.spec.js
yardnsm: You're right, it was an old test when the action cell was coupled to `<Table/>`. Removed.

src/TableActionCell/TableActionCell.driver.js
yardnsm: Why? We want to verify it was rendered. I'll change it to `isPrimaryActionPlaceholderExists`.

src/TableActionCell/TableActionCell.spec.js
yardnsm: I noticed it heavily used in `<Tooltip/>`, but ok

src/TableActionCell/README.TESTKIT.md
```diff
@@ -6,11 +6,11 @@
 |--------|-----------|----------------|-------------|
 | `element` | - | `element` | Get the element |
 | `exists` | - | `bool` | Whether the element exists |
-| `getPrimaryActionPlaceholder` | - | `element` | Get the primary action placeholder element |
+| `primaryActionPlaceholderExists` | - | `element` | Whether the primary action placeholder exists |
```
hepiyellow: Fix return value type. should be `bool`

src/TableActionCell/TableActionCell.driver.js
hepiyellow: The placeholder is something that the consumer has no control over. So they don't need a method to verify if it exists or not.
It is only something we want to test in our (WSR) spec.

src/Tooltip/Tooltip.js
```diff
@@ -384,8 +384,9 @@ class Tooltip extends WixComponent {
     this._hideOrShow('focus');
   }
 
-  _onClick() {
+  _onClick(event) {
     this._hideOrShow('click');
+    event.stopPropagation();
```
ipanasenko: Guys, did you consider this line to be a breaking change?
We had BI tracking broken because of this.

Also, we had a case of Tooltip inside Tooltip (:)), and it also stopped working because of this

argshook: maybe it was done to prevent you from adding tooltip inside a tooltip? :)

jokes aside, we're aware of it and it brings issues for our own components even (https://github.com/wix/wix-style-react/commit/4d34669c7#r30580075)

think we'll revert this line

ipanasenko: Yeah, I know Tooltip nesting is not the best solution, we will definitely do something about it

argshook: @ipanasenko just fyi this was fixed in https://github.com/wix/wix-style-react/commit/bdc3bf35e069db26b209bc40b5fc69c9724e8d6d and will be available in 5.3.3 once released

src/Checkbox/Checkbox.js
```diff
@@ -80,6 +80,7 @@ class Checkbox extends WixComponent {
         onBlur={this.props.focusableOnBlur}
         {...focusableStates(this.props)}
         tabIndex={disabled ? null : 0}
+        onClick={e => e.stopPropagation()}
```
ipanasenko: One more `stopPropagation()`, that affected our app. Can this line be reverted?

yardnsm: Opened a fix for this: https://github.com/wix/wix-style-react/pull/2290

PR #2036, opened 2018-07-25
set notification to a fixed size

banuni:
### What changed
Set Notification size to a fixed 48px

### Why it changed
This makes the button CTA Notification the same size as the other notifications. this also protects against the bug that too-long CTA text might stretch the notification indefinitely

...

---

- [V ] Change is tested
- [ ] Change is documented
- [V ] Build is green

### DIFF
--- a/src/Notification/Notification.scss
+++ b/src/Notification/Notification.scss
@@ -11,6 +11,7 @@
   justify-content: center;
   overflow: hidden;
   padding: 12px;
+  height: 48px;
 }
 
 .label {


lbelinsk:
@wuwa is this what was decided?
Here is the relevant story example: https://wix-wix-style-react-pr-2036.surge.sh/
The height should always be 48px? If so, please update the zeplin as well.

src/Notification/Notification.scss
```diff
@@ -11,6 +11,7 @@
   justify-content: center;
   overflow: hidden;
   padding: 12px;
+  height: 48px;
```
argshook: there was a time when fixed height was removed intentionally to make notification size flexible. There can be cases with a lot of text or something and then fixed height fires back.

i think what we want is `min-height`

banuni: The thing is that size, currently, is not really flexible. (overflow: hidden)
the only thing that affects the toast size is an overflow in the CTA button\link. BUT, it also looks like a bug, since the button IS fixed height.

argshook: i'll talk with @milkyfruit tomorrow and we'll figure it out. don't really want to fix height, sooner or later we'll get a request for flexibility, which already happened once

perhaps it should be `min-height: 54px` so that it's same height regardless of button exists or not.

i suppose the bug you describe is when multiple notifications appear on top of each other?

shlomitc:
This was Ben's last response:
> they should not be placed on top of each other, but sequentially - ie when one is dismissed the next is only triggered then.

wuwa:
decided to do this fix for now, so they can be placed on top (same size for all - 48px height)

src/Notification/Notification.scss
milkyfruit: @banuni @lbelinsk discussed about this PR with @argshook, I support `min-heigth: 54px` as a "hot fix" and the stacking should be solved as a broader issue.

Without further research I see 3 different use-cases when multiple notifications appear at once:
- different themes (e.g. error, warning, standard all together)
- push the content + overlay types (e.g. push the content + timeout success msg)
- same types (e.g. 3 success messages each after another)

milkyfruit:
@wuwa ok, then will remove 12px paddings from the specs and set fixed height.

milkyfruit:
Updated the specs according to this PR: zpl.io/aRYodBN

shlomitc:
@banuni - I've sent you an invitation

banuni:
cool, so according to what's foing on on Zepplin, the current code is ok. yeah?

argshook:
yep

PR #2044, opened 2018-07-26
<Page>. Improve e2e test.

mykas:
Page e2e tests were improved and now it only takes half of the original time.
Made some changes to the protractor.driver.js. Removed private method and added new testkit API endpoint scrollUp(). Threshold value now is imported from Page.js.

Mainly Page.e2e.js for now tests scroll threshold feature and of course if renders at all.
Scroll threshold: 20
scroll_threshold


src/Page/Page.e2e.js
```diff
@@ -1,66 +1,58 @@
...
 import autoExampleDriver from 'wix-storybook-utils/AutoExampleDriver';
-import {PRIVATE} from './Page.protractor.driver';
 import {TESTS_PREFIX} from '../../stories/storyCategories';
 import {storybookConfig} from '../../stories/Page/storybookConfig';
 
 const {category, storyName} = storybookConfig;
 
-const SCROLL_TOP_THRESHOLD = 20;
-const SCROLL_TOP_MIN_STEP = SCROLL_TOP_THRESHOLD + 1;
+const SCROLL_TITLE_THRESHOLD = 20;
```
shlomitc: van this value be taken from the component in some way?

src/Page/Page.e2e.js
```diff
@@ -1,66 +1,58 @@
...
-      await driver.scrollDown();
-      await eventually(() => !driver.titleExists());
-    });
-
-    eyes.it('should display minimized with header background still visible', async () => {
-      const driver = await initTest(initTestConfig);
-      expect(await driver.titleExists()).toBeTruthy();
-      await driver[PRIVATE].setContentScrollOffset(SCROLL_TOP_MIN_STEP);
+      //Should display title while scroll is under scroll threshold
+      await expect(await driver.titleExists()).toBeTruthy();
```
shlomitc: There is too much logic going on in this one test.
I prefer having separate tests without refreshing/reloading the page

src/Page/Page.protractor.driver.js
```diff
@@ -1,25 +1,23 @@
...
+      `arguments[0].scrollTop = ${offset};`,
+      await element.getWebElement()
+    );
   };
 
   return {
     element: () => component,
     /** Scroll content down, enough to minimize the Page. Receives scroll offset as argument, defaults to a step which is enough to minimize the Page. */
-    scrollDown: () => setContentScrollOffset(SCROLL_CONTENT_BIG_STEP),
+    scrollDown: (step = 200) => setContentScrollOffset(step),
```
shlomitc: this 200 value doesn't mean a lot. previously it did make sense to call it a big step or somehow to indicate the page is minimized
you can create a new driver function to scroll until page header is minimized, that way you encapsulate the value inside.

src/AddItem/AddItem.e2e.js
```diff
@@ -8,9 +12,8 @@ describe('AddItem', () => {
     const driver = addItemTestkitFactory({dataHook: 'storybook-addItem'});
     browser.get(storyUrl);
 
-    waitForVisibilityOf(driver.element(), 'Cannot find AddItem')
-      .then(() => {
-        driver.click();
-      });
+    waitForVisibilityOf(driver.element(), 'Cannot find AddItem').then(() => {
+      driver.click();
```
shlomitc: it doesn't need to be between the curly braces `{ ... }`. you can simply do `.then(() => driver.click())`

src/Page/Page.e2e.js
```diff
@@ -1,66 +1,58 @@
...
+      //Should not display title while sroll is over the scroll threshold
+      await driver.scrollDown(SCROLL_TITLE_THRESHOLD + 1);
       await eventually(() => !driver.titleExists());
+      //Should display title again when scrolled back up
+      await driver.scrollDown(0);
```
shlomitc: `scrollDown(0)` doesn't make sense. I'd call it `scrollTo` or create `scrollToTop` function and encapsulate this value

src/Page/Page.e2e.js
mykas: I would too, but there are problems in running all of these with one `const driver = await initTest(initTestConfig);` . Some weird issues with async await. I tried to figure out together Arijus with no luck, but this is what improves the speed.

src/Page/Page.e2e.js
mykas: But then I would need to add new API endpoint for users in teskit? Can I do it?

src/Page/Page.e2e.js
mykas: Yep, this is defined in Page.js. Should I just import it from the Page.js or maybe create additional file just for such constants?

src/AddItem/AddItem.e2e.js
mykas: Oh I think is is my eslint change on save..I should exclude this from this pull request.

src/Page/Page.e2e.js
```diff
@@ -1,66 +1,61 @@
...
-    });
-
-    eyes.it('should display minimized with header background still visible', async () => {
-      const driver = await initTest(initTestConfig);
-      expect(await driver.titleExists()).toBeTruthy();
-      await driver[PRIVATE].setContentScrollOffset(SCROLL_TOP_MIN_STEP);
-      await eventually(() => !driver.titleExists());
+      //Should display title again when scrolled back up
+      await driver.scrollUp();
+      await eyes.checkWindow('Page title appears');
```
shlomitc: why is this check relevant?

shlomitc: also - are you sure the title actually disappeared? because there is an animation that might take a while.
previously there was `await eventually(() => !driver.titleExists());`

src/Page/Page.e2e.js
```diff
@@ -1,66 +1,61 @@
...
-    eyes.it('should display minimized (without Title)', async () => {
+    eyes.it('should hide title on scroll threshold', async () => {
       const driver = await initTest(initTestConfig);
-      expect(await driver.titleExists()).toBeTruthy();
+      //Should display title while scroll is under scroll threshold
+      await eyes.checkWindow('Page title shown');
+      await expect(await driver.titleExists()).toBeTruthy();
+      //Should not display title while sroll is over the scroll threshold
       await driver.scrollDown();
+      await eyes.checkWindow('Page title hidden');
```
shlomitc: this should be done after the `await eventually(() => !driver.titleExists());`

src/Page/Page.e2e.js
mykas: I mean I guess to check again if scroll back up works? I could of course assume that if you scroll down - then it works for any case.

src/Page/Page.e2e.js
mykas: Yes, this seems to be logical to wait until it gets resolved in eventually() and then to check the window.

src/Page/Page.e2e.js
shlomitc: ok, let's keep it like this

src/Page/Page.e2e.js
shlomitc: right, let's fix this one

mykas:
I will add README.TESKIT.md and I think it can be merged.

src/Page/Page.e2e.js
```diff
@@ -1,66 +1,62 @@
...
-    eyes.it('should display maximized (with Title)', async () => {
-      const driver = await initTest(initTestConfig);
-      expect(await driver.titleExists()).toBeTruthy();
-    });
-
-    eyes.it('should display minimized (without Title)', async () => {
+    eyes.it('should hide title on scroll threshold', async () => {
       const driver = await initTest(initTestConfig);
-      expect(await driver.titleExists()).toBeTruthy();
+      console.log(typeof driver.element());
```
shlomitc: remove console.log

src/Page/Page.e2e.js
```diff
@@ -1,66 +1,62 @@
...
+      //Should display title while scroll is under scroll threshold
```
shlomitc: these comments are redundant. the test should be clear from reading it - if not then this means the drivers or the tests are not clear

mykas:
I think this time is because of the error. I've seen my own eyes 1m:12s and my local runner shows significant increase.

shlomitc:
something went wrong with the rebase, tons of other commits were added, please fix the PR

src/Page/Page.e2e.js
```diff
@@ -1,66 +1,61 @@
...
     await browser.get(storyUrl);
     const driver = pageTestkitFactory({dataHook});
     await waitForVisibilityOf(driver.element(), 'Cannot find Page');
     await scrollToElement(driver.element());
-    props && await autoExampleDriver.setProps(props);
+    props && (await autoExampleDriver.setProps(props));
     return driver;
   };
 
   const runTestCases = initTestConfig => {
```
hepiyellow: I intentionaly had 2 snapshots of minimized state. One which is 1 pixel scrolled after minimization occurs. And another snapshot, after background image (or gradiant) is scrolled off of the screen. The 2ns snapshot is to make sure the oage looks visually “correct” also when bg image is out of screen. Although it seems redundant, it helps the review process, since it is a natural state. 
I think you’ve just removed that snapshot.

src/Page/Page.js
```diff
@@ -8,7 +8,7 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import {ResizeSensor} from 'css-element-queries';
 
-const SCROLL_TOP_THRESHOLD = 20;
+export const SCROLL_TOP_THRESHOLD = 20;
```
hepiyellow: I would not export this const from this file. Better create a constants.js file and export from there. The reason is that the protractor driver that imports this constant, is in a pure node environment , and the Page component is a browser / React environment. I wouldn’t want to mix the two. Might cause problems later. When we will migrate to Typescript this could bring build-time dependencies that will clash.

hepiyellow:
By the way: if you really want to reduce test time, did you consider parralelizing the tests? With multiple protractor browsers (if possible). 

Also i would check out the status of Mocha runner with eyes integration.

src/Page/Page.js
shlomitc: This makes sense, a small refactor can be done here

src/Page/Page.protractor.driver.js
```diff
@@ -1,25 +1,26 @@
...
+      `arguments[0].scrollTop = ${offset};`,
+      await element.getWebElement()
+    );
   };
 
   return {
     element: () => component,
     /** Scroll content down, enough to minimize the Page. Receives scroll offset as argument, defaults to a step which is enough to minimize the Page. */
-    scrollDown: () => setContentScrollOffset(SCROLL_CONTENT_BIG_STEP),
+    scrollDown: () => setContentScrollOffset(SCROLL_TOP_THRESHOLD + 1),
```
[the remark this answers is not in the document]
shlomitc: I’m ok with exposing it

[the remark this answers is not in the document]
mykas: I could set default argument for both methods and in this way it could be reusable with other arguments, but I still think this is much more human readable then setContentScrollOffset. What do you think @hepiyellow ?

src/Page/Page.js
mykas: I was actually thinking about using constants file. Thanks for the feedback!

mykas:
@hepiyellow Initially I have actually went with checking protractor configurations to increase the overall time, but I got lost in protractor configuration abstractions. It seems that npm-ci-scripts injects some configurations which we might want to look in the future. Parralelizing the tests only works with resourceful docker agents. In my previous job we have been running docker container for integration tests with 3 GB , 2 cores and if we increased chrome instances from 1 to 3 at least - chrome instances started to crash at random times.

src/Page/Page.e2e.js
mykas: hey, but there was not check against background image. It only checked the title at 1 pixel over the threshold. This test does the same but in the same eyes.it run.

mykas: Check this: https://eyes.applitools.com/app/test-results/00000251869181042602/00000251869180619209/steps/3/edit?accountId=mE99eP8CUn9UTSXSv98L3OSaJ0NZmAcY6Wndx6WSHODdY110

We have a new screenshot for this case to be approved.

PR #2047, opened 2018-07-27
Fix tooltip react 16 regression

Faradey27:
### What changed

Update order of tooltip position calculation funcs

### Why it changed

Because Tooltip was unworkable with React@16

---

- [x] Change is tested
- [x] Change is documented
- [ ] Build is green


src/Tooltip/Tooltip.js
```diff
@@ -308,14 +307,17 @@ class Tooltip extends WixComponent {
...
-          do {
-            const tooltipNode = ReactDOM.findDOMNode(this.tooltipContent);
-            if (tooltipNode) {
-              fw = this._getRect(tooltipNode).width;
-              this._updatePosition(this.tooltipContent);
-              sw = this._getRect(tooltipNode).width;
-            }
-          } while (!props.appendToParent && fw !== sw);
+          // we need to set tooltip position after render of tooltip into container, on next event loop
+          setTimeout(() => {
```
ipanasenko: did you notice that tooltips started blinking upon opening? For a moment they appear in top left corner, and then update their position to correct one

ipanasenko: I don't actually know for sure if this change caused it, but it's most likely candidate

PR #2054, opened 2018-07-30
ISSUE-1721: Fix focusable button in disabled state

Faradey27:
### What changed

Fast fix before migration to wix-ui-core solution

### Why it changed

...

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


Faradey27:
@shlomitc just added complex jsdom regression test

src/common/Focusable/FocusableHOC.spec.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import {mount} from 'enzyme';
+import Button from './../../Button';
```
shlomitc: just removed the first `./` which is not needed
`import Button from '../../Button';`

Faradey27: yea, I am usually left `./` because love the same syntax for `./Button` and `./../Button`

but in general not a big deal, will update

PR #2055, opened 2018-07-30
Fix autoSelect during editing text issue

ihork:
### What changed

Added focus for input before start editing

### Why it changed

Because there was an issue with autoSelect

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


src/Input/Input.puppeteer.driver.js
```diff
@@ -2,7 +2,10 @@ const inputDriverFactory = async (component, page) => {
   const input = await component.$('input');
   return {
     element: () => component,
-    enterText: text => input.type(text),
+    enterText: async text => {
```
shlomitc: Can you please change the syntax to use promises instead? The built file is not playing well with typescript if async-await is in use

PR #2057, opened 2018-07-31
chore(index.js): remove exports and add deprecation message

argshook:
since eslint fails build anyway, there's no point in keeping these
anymore


shlomitc:
We need this file to create the bundle for our sandbox: https://codesandbox.io/s/mymk2zn1mp.
I think it's ok to leave this file as it is. the eslint and tslint rules keep users from importing the entire bundle, but whoever still wants it is welcome to use the bundled version.

let's close this PR for now

argshook:
@shlomitc i don't know, if we have a chance i say let's remove it without overthinking. index.js has been bad for a long time. it doesn't export everything, it has aliases, it imports fonts.

sandbox idea is good in general but the way it works is contradicting to what we suggest.

shlomitc:
I'm closing for now as it's not a blocker in anyway, we'll discuss it later

PR #2063, opened 2018-08-01
chore(SideBar & LanguagePicker): remove documentation

argshook:
as components are deprecated for a long time.

components themselves should be removed in wix-style-react@5.0.0


src/LanguagePicker/README.md
```diff
@@ -0,0 +1,4 @@
+# LanguagePicker
+
+> **NOTE** This component is deprecated and will be removed in wix-style-react@5.0.0
```
shlomitc: Are you sure you can get the same functionality from the IconWithOptions?

also - worth adding a deprecation message in the console for this component
https://github.com/wix/wix-style-react/blob/master/docs/DEPRECATION_GUIDE.md

argshook: it's not exactly the same but most similar to what we have. the difference is basically styling.

by console log i guess you mean this kind of log? https://github.com/wix/wix-style-react/blob/master/src/utils/deprecationLog.js

shlomitc: Yes, but it's not that simple because the design for this component was handmade for a specific reason.
I think the best thing will be exporting it outside to a different package, so whoever is still using it will have it from there. it will not be maintained nor updated, but we can remove it from our lib

argshook: extracting to different package is totally not worth the effort. there are only two repos that use
LanguagePicker that i find, one of them is archived

wix-prvt/users and wix-a/account-recovery

@lbelinsk maybe i can PR users repo with a copy paste of language picker in there? it's been marked
as deprecated on wix-style-react for a very long time, and i want to get rid of "deprecated" section
in storybook

argshook: @shlomitc just fyi, `LanguagePicker` internally is `IconWithOptions` with dividers and nothing more. Protractor and unit drivers are proxies to IconWithOptions. It can easily be removed imo and the one or two projects that need it can migrate to IconWithOptions

shlomitc: Cool. As long as it’s clear and written, no problem to deprecate

argshook: @shlomitc please see https://github.com/wix/wix-style-react/pull/2094

PR #2072, opened 2018-08-02
<Checkbox>. Increasing e2e speed. 

mykas:
Adding new autoexample method remount() which unmounts the whole component from DOM. Increased speed locally from 18s to 5s ( > 3x). 

P.S. Tests in tc are failing because of eyes.it. Previously eyes.it was used with beforeEach() where it takes screenshot on rendered component.


shlomitc:
not required at the moment, closing to reduce noise

PR #2081, opened 2018-08-03
<ImageViewer/> - allow to specify tooltip content

astrauka:
### What changed

Tooltip content was hardcoded, now it can be customised.

### Why it changed

Need to support different locales.

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green


src/AddItem/AddItem.js
```diff
@@ -53,13 +54,18 @@ class AddItem extends WixComponent {
...
-  aspectRatio: PropTypes.oneOf(Object.keys(ratioClasses)),
+  /** The element's aspect ratio */
+  aspectRatio: PropTypes.oneOf(Object.keys(RATIO_CLASSES)),
   /** Element's height - overrides the asspect ratio */
   height: PropTypes.number,
-  /** Content of the tooltip */
+  /** Tooltip props, leave undefined for no tooltip */
+  tooltipProps: PropTypes.shape(Tooltip.propTypes),
+  /** Content of the tooltip
+   * @deprecated
```
shlomitc: I'm not sure we want to deprecate it. this is an easy way to use the component.
for more advanced use cases it makes sense to pass the entire tooltip props

so the behavior of not having a tooltip without it doesn't make sense to me

src/ImageViewer/ImageViewer.js
```diff
@@ -74,16 +80,48 @@ class ImageViewer extends WixComponent {
...
   errorMessage: PropTypes.string,
+  /** Error tooltip props */
+  tooltipErrorProps: PropTypes.object,
+  /** Error tooltip placement
+   * @deprecated
+   * @see tooltipErrorProps
+   */
   tooltipPlacement: PropTypes.string,
+  /** Add icon tooltip props */
+  tooltipAddProps: PropTypes.object,
```
shlomitc: following the previous comment - all of these tooltips should be simply the text and not the entire configuration.
configuration can be passed once and work for different use cases

astrauka: @shlomitc could you be more specific on what you are proposing?
I don't actually understand your point. Giving an example of what you suggest would be helpful.

shlomitc: I pursue simple APIs, meaning less nested objects and configuration
if a configuration is needed - pass it once as `tooltipProps`, as it will probably apply to kind of tooltips in this component.
Then, for each tooltip create a flat API.
`addImageInfo`
`removeImageInfo`
`updateImageInfo`

src/AddItem/AddItem.js
astrauka: > this is an easy way to use the component

As far as I imagine, a good library is the one that both:
* provides as much configurability as possible
* provides amazing defaults

Defaults are provided [here](https://github.com/wix/wix-style-react/blob/ac5aa0d740d045ab9b6e3fa610c279d15e78fcf3/src/AddItem/AddItem.js#L9).

Only allowing to pass `tooltipContent` breaks the first rule "as much configurability as possible".
If a user needs to pass custom tooltip configuration, then he cannot use the element. This is a VERY big problem of wix-style-react library.
When user sees `tooltipProps`, then he knows that he is provided with Tooltip component interface, which is already familiar to him and the same for whole library.

Could you elaborate on your guiding principles for writing a good library?

src/ImageViewer/ImageViewer.js
astrauka: 👍 will udpate

src/AddItem/AddItem.js
shlomitc: There is always a tradeoff. A simple API vs. as much as configurable.
No one solution of course.

I like the approach of making components composable when there are a lot of possible combinations. For example, a top bar of an application can hold many kinds of buttons/dropdowns/icons etc... In that case, it will be hard to create a flat API and the right solution is to allow a composition.

When you have a unit with some specific behavior like this ImageViewer, it doesn't make sense to allow composition of different buttons or tooltips. It is pretty much a standalone unit. In that case, simple titles and actions are more likely the complete configurations of internal parts.
The configuration of tooltips makes sense up to a certain point, so you don't want to configure each tooltip, it's enough to configure once and maintain simple API for configuring the content.

So my approach is making composable components with good and simple APIs.
When configurations are needed, try to extend the API slightly instead of creating a complex one.

astrauka:
@shlomitc updated

src/ImageViewer/ImageViewer.js
```diff
@@ -74,16 +82,39 @@ class ImageViewer extends WixComponent {
   }
 }
 
+ImageViewer.defaultProps = {
+  addImageInfo: 'Add Image',
```
shlomitc: can you add these new props in the `propTypes`

astrauka: @shlomitc added, good catch 👍

Issue #1244, opened 2017-12-10
Implementing a form using Container component inside MessageBoxFunctionalLayout component, generates a UI that its content width exceeds the dialog's width


mennyWix:
### Summary:
Implementing a form using Container component inside MessageBoxFunctionalLayout component, generates a UI that its content width exceeds the width's dialog
![withminwidth](https://user-images.githubusercontent.com/28685657/33804157-c16b1a4a-dda7-11e7-8fd6-46fed696255d.png)
![withoutminwidth](https://user-images.githubusercontent.com/28685657/33804158-c3728792-dda7-11e7-849c-bbbbf81b6202.png)

### Priority:
- [ ] Blocker
- [X] Urgent
- [ ] Normal
- [ ] Minor

### Steps to reproduce:
 1. Render MessageBoxFunctionalLayout with a Container as its children
 2.
 3.

### Expected behavior:
Expected to have a container width the same as MessageBoxFunctionalLayout, but the Container width exceeds the MessageBoxFunctionalLayout - see attached images

### Relevant Components:
MessageBoxFunctionalLayout, Container

### Additional notes:
This problem was found while using these components in codeEmbed application.
Current workaround is to unset min-width inside container

shlomitc:
@wuwa - what should be done in such case?
The `Grid` has a minimum width which is bigger than the `MessageBoxFunctionlLayout` size (600px)

wuwa:
grid inside the box should have width 100%

shlomitc:
The reason for that is that the `Container` component should not be used for the internal grid, but only for the entire page layout.

As `Card` usually contain forms, `<Card.Content/>` is responsible for the layout and padding.
Using it in this component doesn't make sense as it will add padding (and there is already padding in the `MessageBoxFunctionlLayout` component), so we need a similar component that knows how to the internal layout without padding.

We will soon handle all Grid issues and will focus on this one as well

shlomitc:
BTW - removing the padding from `MessageBoxFunctionlLayout` is possible by passing the `noBodyPadding` flag.

PR #2082, opened 2018-08-03
feat(Container): add `fluid` prop

argshook:
to disable min/max width that is set by default. This way grid can be
used inside cards, modals and anything else

closes https://github.com/wix/wix-style-react/issues/1244
closes https://github.com/wix-private/wsr-issues/issues/218


src/Grid/Grid.js
```diff
@@ -163,5 +167,5 @@ export {
   AutoAdjustedColumns,
   AutoAdjustedColumns as AutoAdjustedRow,
   Col,
-  Card
+  Card // TODO: next major version should remove this export. Card is available from wix-style-react/Card
```
shlomitc: Please add to v5 breaking changes issue and also add a deprecation warning here if possible

src/Grid/README.md
```diff
@@ -51,7 +51,12 @@ export default () =>
 
   Use as wrapper for main content. Only `<Row/>`s should be its children.
```
shlomitc: Might want to explain that it can now be a container in every context

src/Grid/README.md
```diff
@@ -51,7 +51,14 @@ export default () =>
...
 
-  It has `minWidth: 894px` and `maxWidth: 1254px`.
+  By default it has `minWidth: 894px` and `maxWidth: 1254px`. Add `fluid` prop to remove those widths.
+
+  `fluid` allows to use grid for any content, be it big (whole page layout) or small (form element layout, for example).
+
+  | propName | propType | defaultValue | isRequired | description                                           |
+  | ---      | ---      | ---          | ---        | ---                                                   |
+  | children | node     | -            | -          | Should only be `<Row/>`s although any node is allowed |
+  | fluid    | bool     | -            | -          | disable min/max width, use for smaller grids          |
```
shlomitc: should default value be false?

src/Grid/README.md
```diff
@@ -51,7 +51,14 @@ export default () =>
 
   Use as wrapper for main content. Only `<Row/>`s should be its children.
 
-  It has `minWidth: 894px` and `maxWidth: 1254px`.
+  By default it has `minWidth: 894px` and `maxWidth: 1254px`. Add `fluid` prop to remove those widths.
+
+  `fluid` allows to use grid for any content, be it big (whole page layout) or small (form element layout, for example).
```
shlomitc: so `fluid=true` means not a page layout?

src/Grid/README.md
argshook: i didn't explicitly set default value but i'll add to docs

src/Grid/README.md
argshook: `fluid` means explicit width, which means it can be used anywhere, in small layouts but also in big ones too. `fluid=true` doesn't mean you can't use it for page

name and behaviour is from bootstrap, should be familiar to most feds imo

PR #2091, opened 2018-08-06
<DateTable/> - remove the newDesign flag from the <DataTable/> compon…

Faradey27:
…ent and apply it as default

### What changed

...

### Why it changed

...

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


hepiyellow:
@Faradey27 @shlomitc 
Hi, I see that there are still references to the `newDesign` prop in DataTable.
Was this PR about deprecating the old look completely?
Because it seems that some old styles are still supported and some were removed.

https://github.com/wix/wix-style-react/blob/master/src/DataTable/DataTable.js

shlomitc:
The idea was to remove the old look completely.
Probably it fell in one of the merge conflicts.
What is the effect now? we have an inconsistent look and feel between different design?

PR #2096, opened 2018-08-07
Date picker - 2 months layout

tzabarc:


src/DatePicker/Calendar/Calendar.js
```diff
@@ -62,31 +63,33 @@ export default class Calendar extends WixComponent {
...
-
           onLeftArrowClick: () => this._setMonth(startOfMonth(addMonths(month, -1))),
-
           onRightArrowClick: () => this._setMonth(startOfMonth(addMonths(month, 1)))
         }}
         />
     );
 
+    let className = '';
+    twoMonths && (className += 'DayPicker--TwoMonths');
```
Faradey27: ```const className = twoMonths ? 'DayPicker--TwoMonths' : ''```

src/DatePicker/DatePicker.spec.js
```diff
@@ -704,4 +704,12 @@ describe('DatePicker', () => {
       expect(calendarDriver.isVisible()).toBe(false);
     });
   });
+
+  fdescribe('two months layout', () => {
```
Faradey27: remove `f`

src/DatePicker/DatePickerHead/index.js
```diff
@@ -66,7 +66,7 @@ const DatePickerHead = ({
       }
 
       <div
-        className={classnames(styles.arrow, styles.arrowRight)}
+        className={classnames(styles.arrow, styles.arrowRight, 'DayPicker-Nav--Right')}
```
Faradey27: can you left comment in file, why we are using global classNames here

stories/DatePicker.story.js
```diff
@@ -22,7 +22,8 @@ export default {
     shouldCloseOnSelect: true,
     showYearDropdown: false,
     showMonthDropdown: false,
-    locale: 'en'
+    locale: 'en',
+    twoMonths: true
```
Faradey27: should be false by default and better to add one more story where twoMonths will be true

src/DatePicker/Calendar/Calendar.js
```diff
@@ -62,31 +63,30 @@ export default class Calendar extends WixComponent {
...
-
       initialMonth: month,
       initialYear: month,
       selectedDays: parse(propsValue),
       month,
       year: month,
       firstDayOfWeek: 1,
       locale: typeof locale === 'string' ? locale : '',
-      fixedWeeks: true,
+      fixedWeeks: !twoMonths,
```
shlomitc: This might be problematic. Having multiple sizes would cause tooltip to flip when moving between dates. I think it should stay true always

src/DatePicker/DatePicker.driver.js
```diff
@@ -52,6 +53,7 @@ const datePickerDriverFactory = ({element, wrapper}) => {
     triggerKeyDown: params => ReactTestUtils.Simulate.keyDown(getFocusedDay(), params),
     isFocusedDayVisuallyUnfocused: () => getFocusedDay().classList.contains('DayPicker-Day--unfocused'),
     containsVisuallyUnfocusedDay: () => !!getVisuallyUnfocusedDay(),
+    usesTwoMonthsLayout: () => getMonthContainers().length === 2,
```
shlomitc: IsTwoMonthsLayout

PR #2099, opened 2018-08-08
<FormField/> - allow passing info tooltip props

astrauka:
### What changed

Added ability to define FormField info tooltip props.

### Why it changed

We want to have a fixed width info tooltip.
Now with various text lengths tooltips look very long.

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green


src/FormField/FormField.js
```diff
@@ -46,6 +46,9 @@ class FormField extends React.Component {
     /** display info icon with tooltip. Node from this prop is content of tooltip */
     infoContent: PropTypes.node,
 
+    /** info icon tooltip props */
+    tooltipProps: PropTypes.shape(omit(['content'], Tooltip.propTypes)),
```
argshook: why `omit`? if you want to control tooltip through `tooltipProps`, so be it, control it all

argshook: i'd probably call it `infoTooltipProps`, cause who knows how many other tooltips we'll have in future

argshook: i'd probably call it `infoTooltipProps`, cause who knows how many other tooltips we'll have in future

src/FormField/FormField.spec.js
```diff
@@ -15,150 +14,144 @@ import FormField from './';
...
 
   describe('`id` prop', () => {
+    const id = 'five';
+
     it('should be added to label as `htmlFor` prop', () => {
-      const wrapper = mount(<FormField id="five" label="hello"><div/></FormField>);
-      const label = wrapper.find(Label);
-      expect(label.prop('for')).toEqual('five');
+      const wrapper = mount(renderFormField({id, label}));
+      expect(wrapper.find(Label).prop('for')).toEqual('five');
```
argshook: if you extract `id`, then also `toEqual(id)` :)

src/FormField/components/InfoIcon.js
```diff
@@ -1,29 +1,30 @@
...
       </div>
     </Tooltip>
-  </div>;
+  </div>
+);
 
 InfoIcon.propTypes = {
-  content: PropTypes.any.isRequired
+  content: Tooltip.propTypes.content,
+  tooltipProps: PropTypes.shape(omit(['content'], Tooltip.propTypes))
```
argshook: same here about `omit`, i wouldn't omit content

src/FormField/FormField.js
```diff
@@ -81,6 +84,11 @@ class FormField extends React.Component {
     return children;
   }
 
+  renderInfoIcon = () => {
+    const {infoContent, tooltipProps} = this.props;
+    return infoContent && <InfoIcon content={infoContent} tooltipProps={tooltipProps}/>;
```
argshook: regarding `content` i would do like so

```js
return infoContent && <InfoIcon content={infoContent} tooltipProps={{content: infoContent, ...tooltipProps}}/>;
```

no omits, no lodash. i'd argue that omitting like that is more confusing as consumers who are familiar with tooltip props wouldn't expect such tricks

src/FormField/FormField.js
argshook: btw, it's `omit(object, paths)` not the other way around. I see you copy pasted this here as well as in https://github.com/wix/wix-style-react/pull/2101/files

src/FormField/FormField.js
astrauka: @argshook the idea was to not allow passing content in two ways.
Now one can pass content via `infoContent` and `tooltipProps.content`.
When I see some value can be passed in more than one way, I usually specify both to make sure it will work

argshook: i don't see a problem with that whatsoever.

most consumers will never need to supply `tooltipProps` and thus wont even know there're two ways. You know, cause you implemented it :)

those who will need `tooltipProps`, will give `placement` or `appendToParent` or `showDelay` or something. Don't you think it would come off as unexpected behaviour if they also give `content` but don't see it?

if we're willing to give full power, i say let's not do any trickery.

btw, small adjustment to `InfoIcon` and `content` prop can be removed:

```js
return infoContent && <InfoIcon tooltipProps={{content: infoContent, ...tooltipProps}}/>;
```

astrauka:
@argshook updated

PR #2101, opened 2018-08-08
<ImageViewer/> - Disable tooltip content passing via tooltipProps

astrauka:
### What changed

<ImageViewer/> - Disable tooltip content passing via tooltipProps

### Why it changed

It's better to have one way to pass content, easier to maintain, less confusion for users.

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green


src/ImageViewer/ImageViewer.js
```diff
@@ -43,7 +45,7 @@ class ImageViewer extends WixComponent {
     return (
       <div className={classes} style={{width, height}}>
         {!imageUrl &&
-        <AddItem data-hook="add-image" height={height} onClick={onAddImage} tooltipProps={{...tooltipProps, content: addImageInfo}}/>
+        <AddItem data-hook="add-image" height={height} onClick={onAddImage} tooltipProps={tooltipProps} tooltipContent={addImageInfo}/>
```
argshook: here is same argument i wrote in https://github.com/wix/wix-style-react/pull/2099#discussion_r208486169

astrauka: Have to omit content for tooltip props as those are used for 3 tooltips

Issue #2028, opened 2018-07-23
`<Page/>` does not work with `react-hot-loader`


hepiyellow:
### Priority:
- [x] Blocker
- [ ] Urgent
- [ ] Normal
- [ ] Minor

### Steps to reproduce:
 1. Generate a new Client project using yoshi-create
 2. Add a minimum `<Page/>` usage
 3. run `npm start`

You get a:
```
Warning: Failed prop type: Page: Invalid Prop children, must contain Page.Header
```
And you don't see the header and content. 
### Expected behavior:
Page should work
### Relevant Components:
There might be more components that are affected by this.

### Investigation notes:
This is happening because `<Page/>` is doing prop validation which uses the children's `type`.
According to this Issue of `react-hot-loader` , this doesn't work:
https://github.com/gaearon/react-hot-loader/issues/304

### Verifying it really is ReactHotLoader
I added this in a generated project's `client.js`:
```
import ReactHotLoader from `react-hot-loader`;
ReactHotLoader.disableProxyCreation = true;
```
And the problem disappeared.

PR #2130, opened 2018-08-14
chore(Card.EmptyState): remove additional component and simply use composition

argshook:
with addition of `EmptyState` to `Card` as `Card.EmptyState` all consumers who use just `Card` would suddenly get `EmptyState` in their bundles too, even if they never use it.

moreover, `EmptyState` wants to have `section` theme by default and in that case `Card.EmptyState`
only purpose is to set different padding.

so, instead of adding additional component, weigh into consumers bundle and create additional API,
it's much more simple to just set different padding in `Card.Content` when children is `EmptyState`.

no new API, practically no new code, only composition


src/Card/Content/Content.js
```diff
@@ -3,15 +3,19 @@ import {node} from 'prop-types';
 import styles from './Content.scss';
 
 class Content extends Component {
-
   static propTypes = {
     children: node
   };
 
+  _getChildName = children =>
+    children.type && (children.type.name || children.type.displayName);
```
shlomitc: `displayName` will not work with `HMR`  and I'm not sure if `type.name` has the same problem.
https://github.com/wix/wix-style-react/issues/2028

I think we need a custom static property that says `children.type.wsrName`, and then no one will hook on it.

argshook: surely we don't organize our code based on whether hmr works or not?

we don't need custom anything, `EmptyState` has `displayName` set, so it's available.

i can also check by instance but that would require importing `EmptyState` inside `Card.Content` which sorta defeats the purpose of removing `Card.EmptyState`

shlomitc: We're going to solve the HMR anyway. For example Page component doesn't work in newly generated projects. So it's not we're organizing our code, we're handling an issue.

If possible, I prefer not adding more `displayName` if statements, because we'll just need to solve it later. @yardnsm is trying to fix it now for the Page, so sync together regarding a solution.

argshook: what do you suggest then? `Card.EmptyState` is good as it is or should i think of other way how to add that 12px padding?

shlomitc: ok, keep it

PR #2131, opened 2018-08-14
<EmptyState/> - add links to component examples in the story

yardnsm:
This PR adds some link in the `<EmptyState/>` story that refer to other components that utilize `<EmptyState/>` within them.

This PR also adds an example for an `<EmptyState/>` with a long action text.


package.json
```diff
@@ -105,6 +105,7 @@
     "logLevel": "verbose"
   },
   "dependencies": {
+    "@storybook/addon-links": "^3.4.10",
```
argshook: @yardnsm  we have `import {createStoryUrl} from 'wix-ui-test-utils/protractor';` which, i think, is essentially same as `linkTo`. Perhaps we can use ours instead of adding new dep?

argshook: ah, yes, you're right, that url has `iframe` hard coded https://github.com/wix/wix-ui/blob/28bfafb33c0f7a79844c546fb2ca059add78c98b/packages/wix-ui-test-utils/src/protractor/protractor-helpers.ts#L30

we can, of course, add a flag or something,  but maybe it's supposed to be for testing and so let's not mix responsibilities.

it's just that i always twitch when see new dependency added, but i'll be okay :)

PR #2136, opened 2018-08-15
<AddItem> component refactor and additional features.

mykas:
### What changed

- Supports multiple themes based on new UX design https://zpl.io/blProv5.
- Style for image viewer is now under theme `image`. 

- Size of the icon and padding between text and icon can vary in between four size layouts.
  1. Large.
      + Max `padding: 12px` between text and +
      + AddItemLarge.svg
  2. Medium.
      + Max `padding: 12px` between text and +
      + AddItemMedium.svg
  3. Small.
      + Max `padding: 6px` between text and +
      + AddItemSmall.svg
  4. Tiny
      + Max `padding: 6px` between text and +
      + Add icon with width and height: 26px.

- Component width and height size is based on parent container.

- Aspect ratio prop has been removed. Planned separate hoc component for controlling aspect ratio.

- Children in string form is used on Text component.

- Theme `image` has restrictions: 
   1. `disable` prop will not work by design.
   2. `children` prop by design will not be allowed.
   3. `size` prop is ignored. Fixed size.

## Example usage

```javascript
<AddItem 
         theme="plain | dashes | filled | image"
         alignItems="left | right | center"
         size="large | medium | small | tiny"
         onClick={func}
         disabled={true | false}
         tooltipProps={object}
         tooltipContent="tooltip message"
>
Add New Item
</AddItem>
```

### Why it changed

- UX request.

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green


src/AddNewItem/AddNewItem.driver.js
```diff
@@ -0,0 +1,17 @@
+import textDriverFactory from '../Text/Text.driver';
+
+const addNewItemDriverFactory = ({element, eventTrigger}) => {
+  const byHook = hook => element.querySelector(`[data-hook*="${hook}"]`);
+  return {
+    exists: () => !!element,
+    element: () => element,
+    getText: () => textDriverFactory({element: byHook('addnewitem-text')}).getText(),
+    textExists: () => textDriverFactory({element: byHook('addnewitem-text')}).exists(),
+    elipsisExists: () => !!byHook('text-with-ellipses'),
```
shlomitc: I don't see you use it, can be removed

src/AddNewItem/AddNewItem.driver.js
```diff
@@ -0,0 +1,17 @@
...
+    getChildren: () => !!byHook('addnewitem-children'),
```
shlomitc: I don't see a real reason why testing it.
You just check only if children exist, so you're not "getting" the children.

anyway - you can just place a data-hook on the children content and verify it's rendered without exposing this in the driver

src/AddNewItem/AddNewItem.driver.js
```diff
@@ -0,0 +1,17 @@
...
+    iconExists: () => !!byHook('addnewitem-icon'),
```
shlomitc: redundant, check it in the E2E test

stories/AddNewItem/BreakpointsExample.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+
+
+import {Container, Col, Row} from '../../src/Grid';
```
shlomitc: as this example is translated to markdown, import from `wix-style-react/Grid`. we have an alias that make it work

stories/AddNewItem/BreakpointsExample.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+
```
shlomitc: redundant lines

stories/AddNewItem/BreakpointsExample.js
```diff
@@ -0,0 +1,48 @@
...
+      <Row stretchViewsVertically>
+        <Col span={3}>
+          <div style={{height: '240px'}}>
+            <AddNewItem dataHook="addItem-large">
+            Add New Item
+            </AddNewItem>
+          </div>
+        </Col>
+        <Col span={3}>
+          <div style={{height: '200px'}}>
```
shlomitc: so this is something we that we should discuss.
This means that this will be the API of the end user coming to implement this component.
it's not the nicest one in my opinion.

What do we think about having a `Proportion` like component that controls also the size?
Like this one https://pinterest.github.io/gestalt/#/Letterbox

stories/AddNewItem/VerticalExample.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
```
shlomitc: redundant lines

stories/AddNewItem/VerticalExample.js
```diff
@@ -0,0 +1,55 @@
...
+
+import {Col, Row} from '../../src/Grid';
```
shlomitc: fix imports as before

stories/AddNewItem/WithoutActionExample.js
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
```
shlomitc: same

src/AddNewItem/AddNewItem.scss
```diff
@@ -0,0 +1,123 @@
+@import '../common.scss';
+
+.root {
+  display: flex;
+  height: 100%;
+  width: 100%;
+  align-items: center;
+  justify-content: center;
+  box-sizing: border-box;
+  padding-top: 12px;
```
shlomitc: not necessary but you can write `padding: 12px 30px`

src/AddNewItem/AddNewItem.scss
```diff
@@ -0,0 +1,123 @@
...
+  flex-direction: column;
+  align-items: center;
+  justify-content:center;
+  color:inherit;
+
+
+}
+
+.plain {
+  background-color: #ffffff;
```
shlomitc: should be a color from the colors palette

src/AddNewItem/AddNewItem.scss
```diff
@@ -0,0 +1,123 @@
...
+  }
+  
+  &:active {
+    background-color: $B30;
+  }
+}
+
+.dashes {
+  box-sizing: border-box;
+  background-color: rgba(255, 255, 255, 0.48);
```
shlomitc: should be a color from the colors palette

src/AddNewItem/AddNewItem.scss
```diff
@@ -0,0 +1,123 @@
...
+}
+
+.center {
+  align-items: center;
+  justify-content: center;
+  text-align: center;
+}
+
+.disabled {
+  background-color: rgba(255, 255, 255, 0.48);
```
shlomitc: should be a color from the colors palette

src/AddNewItem/AddNewItem.js
```diff
@@ -0,0 +1,77 @@
...
+import styles from './AddNewItem.scss';
+import {getHeightBreakpoint} from './_helpers/breakpointCalculator';
+
+const textWitElipsis = children => (
+  <div data-hook="text-with-ellipses">
+    <Text weight="thin" size="small" style={{color: '#3899ec'}} dataHook="addnewitem-text" elipsis>{children}</Text>
+  </div>
+);
+
+class AddNewItem extends Component {
```
shlomitc: extend `WixComponent` and then you don't need to handle the `data-hook` below

src/AddNewItem/AddNewItem.js
```diff
@@ -0,0 +1,77 @@
...
+      return null;
+    }
+    if (typeof children === 'string') {
+      return textWitElipsis(children);
+    }
+    return <div data-hook="addnewitem-children">{children}</div>;
+  }
+
+  render() {
+    const {dataHook, onClick, disabled, theme, alignItems} = this.props;
```
shlomitc: when extending `WixComponent`, it takes care for the `dataHook`

src/AddNewItem/AddNewItem.js
```diff
@@ -0,0 +1,77 @@
...
+    const style = styles;
+    const breakpoint = getHeightBreakpoint(this.state.height);
+    return (
+      <div ref={x => this.height = x} className={classnames(style.root, styles[breakpoint], style[theme], style[alignItems], disabled ? style.disabled : '')} onClick={disabled ? null : onClick} data-hook={dataHook}>
```
shlomitc: please extract the className to a readable variable.
regarding the `disabled`css, you can just do this:
`className( { [style.disabled]: disabled } )`

src/AddNewItem/AddNewItem.e2e.js
```diff
@@ -0,0 +1,91 @@
...
+import {createStoryUrl, waitForVisibilityOf, scrollToElement} from 'wix-ui-test-utils/protractor';
+
+import {addNewItemTestkitFactory} from '../../testkit/protractor';
+import {storySettings} from '../../stories/AddNewItem/storySettings';
+
+describe('AddNewItem', () => {
+
+  const storyUrl = createStoryUrl({kind: storySettings.kind, story: storySettings.storyName, withExamples: true});
+
+  const createDriverFactory = async driver => {
```
shlomitc: as this pattern is repeated in many tests, extract it to a common function in `test/utils/storybook-helpers.js`

src/AddNewItem/AddNewItem.e2e.js
```diff
@@ -0,0 +1,91 @@
+import eyes from 'eyes.it';
+import autoExampleDriver from 'wix-storybook-utils/AutoExampleDriver';
+import {createStoryUrl, waitForVisibilityOf, scrollToElement} from 'wix-ui-test-utils/protractor';
```
shlomitc: `createStoryUrl` should be imported from `../../test/utils/storybook-helpers`

src/AddNewItem/AddNewItem.js
```diff
@@ -0,0 +1,77 @@
...
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+import Add from '../new-icons/Add';
+import Text from 'wix-style-react/Text';
+import styles from './AddNewItem.scss';
+import {getHeightBreakpoint} from './_helpers/breakpointCalculator';
+
+const textWitElipsis = children => (
+  <div data-hook="text-with-ellipses">
```
shlomitc: why is this extra div necessary?

src/AddNewItem/AddNewItem.js
```diff
@@ -0,0 +1,77 @@
...
+
+  renderText = () => {
+    const {children} = this.props;
+    if (!children) {
+      return null;
+    }
+    if (typeof children === 'string') {
+      return textWitElipsis(children);
+    }
+    return <div data-hook="addnewitem-children">{children}</div>;
```
shlomitc: why is this div needed?

src/AddNewItem/AddNewItem.js
```diff
@@ -0,0 +1,77 @@
...
+      this.setState({height: offsetHeight});
+    }
+  }
+
+  renderText = () => {
+    const {children} = this.props;
+    if (!children) {
+      return null;
+    }
+    if (typeof children === 'string') {
```
shlomitc: are you sure this check is needed? can't React just render children as a string?

src/AddNewItem/AddNewItem.js
```diff
@@ -0,0 +1,77 @@
...
+import classnames from 'classnames';
+
+import Add from '../new-icons/Add';
+import Text from 'wix-style-react/Text';
+import styles from './AddNewItem.scss';
+import {getHeightBreakpoint} from './_helpers/breakpointCalculator';
+
+const textWitElipsis = children => (
+  <div data-hook="text-with-ellipses">
+    <Text weight="thin" size="small" style={{color: '#3899ec'}} dataHook="addnewitem-text" elipsis>{children}</Text>
```
shlomitc: why is color need to be passed as a style prop? this looks wrong

src/AddNewItem/AddNewItem.js
shlomitc: I wouldn't call it `height`.
It's a reference to the dom element, even calling it rootRef will be better.
make sure you understand how `ref`s work

src/AddNewItem/AddNewItem.scss
```diff
@@ -0,0 +1,123 @@
...
+  color:$D55;
+  border: solid 1px #cbd3dc;
```
shlomitc: should be a color from the colors palette

src/AddNewItem/_helpers/breakpointCalculator.js
```diff
@@ -0,0 +1,18 @@
+const BREAKPOINTS = {
+  large: {min: 212, max: 2000},
+  medium: {min: 162, max: 212},
+  small: {min: 120, max: 162},
+  tiny: {min: 42, max: 120}
+};
+
+
+export const getHeightBreakpoint = height => {
+  const results = Object.keys(BREAKPOINTS)
```
shlomitc: the first `Object.keys` can be done outside of this function as it happens once and stays the same

src/AddNewItem/_helpers/breakpointCalculator.js
```diff
@@ -0,0 +1,18 @@
...
+    .filter(breakpoint => BREAKPOINTS[breakpoint].min <= height && BREAKPOINTS[breakpoint].max > height);
```
shlomitc: you can just use `find` instead of `filter`. it will return the first item that satisfies the condition or `undefined`.

src/AddNewItem/_helpers/breakpointCalculator.js
```diff
@@ -0,0 +1,18 @@
...
+  if (!results[0]) {
+    height > BREAKPOINTS.large.max ? results[0] = 'large' : null;
```
shlomitc: what happens when null returns? is this a valid use case?
I'd just set a default fallback in case it is not satisfied.

also, mutating `results[0]` is a bad practice. always prefer creating a new object instead

src/AddItem/AddItem.spec.js
```diff
@@ -1,102 +1,118 @@
...
+    it('should not render children as string when theme is `image`', () => {
+      const driver = createDriver(
+        renderAddItem({children: text, theme: 'image'})
+      );
+      expect(driver.textExists()).toEqual(false);
     });
 
+    it('should not render children as component when theme is `image`', () => {
+      const wrapper = mount(renderAddItem({children: child, theme: 'image'}));
+      expect(wrapper.find(`[data-hook*="child"]`).exists()).toEqual(false);
```
argshook: can this `find` be part of driver?

src/AddItem/components/ActionText.js
```diff
@@ -0,0 +1,43 @@
...
+  };
+  render() {
+    const {children, disabled, size} = this.props;
+    return (
+      <div style={{...SPACING[size], ...ALIGN_TEXT}}>
+        <Text
+          weight="thin"
+          size="small"
+          style={{
+            color: disabled ? '#CBD3DC' : '#3899ec'
```
argshook: better use color from single source, either in css with `@import '../../../colors.scss'` or in js by `import color from '../../../colors.scss'`

src/AddItem/components/ActionText.js
```diff
@@ -0,0 +1,43 @@
...
+    disabled: PropTypes.bool,
+    size: PropTypes.string
+  };
+  render() {
+    const {children, disabled, size} = this.props;
+    return (
+      <div style={{...SPACING[size], ...ALIGN_TEXT}}>
+        <Text
+          weight="thin"
+          size="small"
```
argshook: i think thin and small is that one combination which is restricted, please check with ux

src/colors.scss
```diff
@@ -67,6 +67,7 @@ $S10 : rgba(22, 45, 61, 0.12);
 $S20 : rgba(22, 45, 61, 0.48);
 $S30 : rgba(22, 45, 61, 0.06);
 $S40 : rgba(22, 45, 61, 0.18);
+$S50: rgba(255, 255, 255, 0.48);
```
argshook: new color approved by ux?

src/AddItem/AddItem.js
```diff
@@ -15,65 +36,123 @@ const DEFAULT_TOOLTIP_PROPS = {
...
-                {renderInnerAddItem()}
-              </Tooltip> :
-              renderInnerAddItem()
-          }
-        </div>
+      <div
+        className={root}
+        data-hook={dataHook}
+        onClick={disabled ? null : onClick}
+        onFocus={this.props.focusableOnFocus}
```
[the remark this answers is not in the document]
mykas: eslint yells...😔

src/AddItem/AddItem.spec.js
mykas: It can, but should we? I mean users can wrap their react component and create their driver however they like. Really not sure if we should provide children hook.

src/ImageViewer/ImageViewer.js
```diff
@@ -36,49 +36,70 @@ class ImageViewer extends WixComponent {
...
-              </Tooltip>
+        {!imageUrl && (
+          <AddItem
+            onClick={onAddImage}
+            theme="image"
+            dataHook="add-image"
+            tooltipProps={{...tooltipProps, content: addImageInfo}}
+            />
+        )}
+        {!!imageUrl && (
```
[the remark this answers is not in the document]
mykas: Automatic prettier lint formatted the code. No changes were made to the code.

src/colors.scss
mykas: Yes

src/AddItem/AddItem.scss
```diff
@@ -1,57 +1,109 @@
-@import '../common';
-@import '../mixins/layout';
+@import "../common.scss";
```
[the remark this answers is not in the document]
mykas: common is used for focusable too.

PR #2142, opened 2018-08-19
POC - complete app structure

shlomitc:
I think it's super important that we allow our consumers to have their complete page layout with minimal effort.

Tell me what you think about it

### DIFF
--- a/src/Structure/index.js
+++ b/src/Structure/index.js
@@ -1,0 +1,32 @@
+import React from 'react';
+import css from './App.scss';
+
+export const Main = ({children}) => (
+  <div className={css.main}>
+    {children}
+  </div>
+);
+
+export const Header = ({children}) => (
+  <header className={css.header}>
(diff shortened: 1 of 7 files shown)


src/Structure/index.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
```
argshook: this shouldn't live under `src` cause `import-path` will make it importable which may be mistaken for a real component.

unless that's by intention?

shlomitc:
By intention, to make it easy to create a structure.
Or do you think it should be just guidelines?

src/Structure/index.js
shlomitc: By intention, to make it easy to create a structure.
Or do you think it should be just guidelines?

shlomitc:
closing it until will get some time to finish it correctly

PR #2146, opened 2018-08-21
support focus(options) inside input with options

alonbt:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus

it will allow to pass a param called 'preventScroll'

### DIFF
--- a/src/InputWithOptions/InputWithOptions.js
+++ b/src/InputWithOptions/InputWithOptions.js
@@ -251,8 +251,8 @@ class InputWithOptions extends WixComponent {
     }
   }
 
-  focus() {
-    this.input.focus();
+  focus(options = {}) {
+    this.input.focus(options);
   }
 
   blur() {


alonbt:
Adding 'preventScroll: true' as a param will disable the scroll, thus allowing the developer more options to scroll when the time is right and not when the element focuses. it allows to easily animate the scroll in a custom way

shlomitc:
can you please add a test for it?

alonbt:
@shlomitc not sure how to test this. it seems that `input.focus()` wasn't tested so I don't have the easy spot to add a param and check it's passing it on. and I don't know how to get the reference of the actual input and listen to its 'focus' method with the new params. Do you have any ideas ?

https://github.com/wix/wix-style-react/pull/2146/files#diff-deaeb5c7e253bfa4f497fa8ace0b3607R181

hepiyellow:
I added the test. Using enzyme, I mock the `input.focus()` and see that the `options` is passed to it.
Otherwise, we need here a browser test (since jsdom does not support focus options).

src/Input/Input.driver.js
```diff
@@ -13,9 +17,13 @@ const inputDriverFactory = ({element, wrapper, component}) => {
 
   const driver = {
     trigger: (trigger, event) => ReactTestUtils.Simulate[trigger](input, event),
-    focus: () => {
-      input.focus();
-      ReactTestUtils.Simulate.focus(input);
+    focus: options => {
+      if (isReactWrapper(wrapper)) {
+        wrapper.instance().focus(options);
```
shlomitc: please don't use wrapper,  it's enzyme related only

hepiyellow: You mean not to use it in the driver?
and use it in the spec only?

shlomitc: Consumers of the driver don’t have the wrapper object as this is only an enzyme wrapper.
You can do it in the spec file instead without exposing it in the driver

PR #2151, opened 2018-08-24
wip(<SortableList/>) : Sortable, Draggable. POC Component

Wo1v3r:
WIP Drag and Drop components.
- General <Draggable/> behavior component under `DragAndDrop`
- <Sortable/> component without predefined styles for rendering a list of draggable and moveable items
- Example Component: <SortableList/> using <Sortable/>

We added a story under `Components` for example.

Major Gaps:
------
- Testing this in isolation is basically testing that `react-dnd` works which is redundant. 
IMO focus should be on providing a simple enzyme testkit.
-  Drop horizontally too via props e.g; `<Draggable vertically horizontally/>`
-  Move items between different Sortable lists


Wo1v3r:
@shlomitc 
Something small I noticed
Maybe we should export (expose)

`DragDropContextProvider` so whoever will use 'Draggable' won't have to install react-dnd and its' friends as dependency

shlomitc:
@Wo1v3r - agree
Anyway, I see that there is an issue placing multiple `DragDropContextProvider` in the same page.
This might be very tricky in hosted applications.

For example, this will blow:
```js
<div>
  <DragDropContextProvider backend={backend}>
    <SortableList/>
  </DragDropContextProvider>
  <DragDropContextProvider backend={backend}>
    <SortableList/>
  </DragDropContextProvider>
</div>
```

We might need to have a special provider that check if it already exists in the context and will add it only if it doesn't exist

Wo1v3r:
@shlomitc 
I asked around and was made aware they solved it somehow but still didn't get to have a look
https://github.com/palantir/react-mosaic

gnight :)

src/SortableList/SortableList.js
```diff
@@ -0,0 +1,67 @@
...
+import PropTypes from 'prop-types';
+
+/**
+ * Attaches Drag and Drop behavior to a list of items
+ */
+export default class SortableList extends WixComponent {
+  constructor(props) {
+    super(props);
+    this.state = {
+      items: props.items || []
```
Faradey27: defaultProps?

src/SortableList/SortableList.js
```diff
@@ -0,0 +1,67 @@
...
+  render() {
+    const {items} = this.state;
+    const {render, withHandle} = this.props;
+
+    return (
+      <div>
+        {items.map(({id, ...props}, i) => (
+          <div key={i}>
+            <Draggable
+              key={id}
```
Faradey27: no meed of key here

src/SortableList/SortableList.js
```diff
@@ -0,0 +1,67 @@
...
+  };
+
+  render() {
+    const {items} = this.state;
+    const {render, withHandle} = this.props;
+
+    return (
+      <div>
+        {items.map(({id, ...props}, i) => (
+          <div key={i}>
```
Faradey27: better to use id instead of index

src/DragAndDrop/Draggable/Draggable.spec.js
```diff
@@ -0,0 +1,47 @@
...
+import PropTypes from 'prop-types';
+import TestBackend from 'react-dnd-test-backend';
+import {DraggableSource} from './Draggable';
+import {DragDropContextProvider} from 'react-dnd';
+
+import {draggableTestkitFactory as enzymeDraggableTestkitFactory} from '../../../testkit/enzyme';
+import {mount} from 'enzyme';
+
+describe('Draggable', () => {
+  xit('is draggable', async () => {
```
Faradey27: .

src/SortableList/SortableList.js
shlomitc: @Wo1v3r - I remember you said there was an issue here, do you remember?

src/SortableList/SortableList.js
shlomitc: agree

src/DragAndDrop/Draggable/Draggable.spec.js
shlomitc: testing in JSDOM might be problematic I guess.
@Wo1v3r - can you please share the experience of writing this test?

PR #2159, opened 2018-08-27
<DropdownLayout/> - change the divider's data-hook

### LINKED ISSUE #2034: <DropdownLayout> - change divider's data-hook
Its a good idea to change the divider's data-hook to use a different template so we can distinguish it from a "regular" option item.

This has been done in https://github.com/wix/wix-style-react/pull/2005, but reverted since it could be a possible breaking change.

cc: @shlomitc

yardnsm:
Fixes #2034.

This could break existing tests, so it may be better to have this in the next major version.

### DIFF
--- a/src/DropdownLayout/DropdownLayout.js
+++ b/src/DropdownLayout/DropdownLayout.js
@@ -219,7 +219,7 @@ class DropdownLayout extends WixComponent {
   renderOption({option, idx}) {
     const {value, id, disabled, title, overrideStyle, linkTo} = option;
     if (value === DIVIDER_OPTION_VALUE) {
-      return this.renderDivider(idx, `dropdown-item-${id || idx}`);
+      return this.renderDivider(idx, `dropdown-divider-${id || idx}`);
     }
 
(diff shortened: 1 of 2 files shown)


hepiyellow:
@shlomitc 
Yeah it can theoretically break tests.
But :
1) Breaking a test is not a big deal, it doesn't affect produciton and is easy to fix.
2) Searching GitHUb, I see very few ocurrences of `dropdownLayoutDriver.optionByHook`, non of which look like they are looking for a divider.

I think it is safe to merge this.

hepiyellow:
Looking a little bit more into it.
From searching GitHub, no one is using `driver.optionByHook` with the explicit `dropdown-item-...`.
And internally in the driver, we only use the `dropdown-item` hook value in the `optionById()` method, which semantically suggests that it is by id and not by position. So it is not affected by the change.

I think we can merge it.

hepiyellow:
@yardnsm 
Please double check my comment. If you agree, then you can merge.

PR #2160, opened 2018-08-27
differ Component from PureComponent to fix warnings

rghorbani:
### What changed

Created new WixComponentPure to distinguish between some components like DataTable, Table, DrillView and others. WixComponentPure extends React.PureComponent and WixComponent extends React.Component which allows using `shouldComponentUpdate`.

### Why it changed

To fix warnings regarding 'shouldComponentUpdate should not be used when extending React.PureComponent'. 

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green


Faradey27:
Hi, several small comments and questions :)

Usage of PureComponent require all props to be immutable(each update -> new ref) and all state var also, to be immutable. 
Are you sure that all components that was updated use such approach?
Also, are you sure that clients who is using this components, also doing same?

And in general, I do not like extend component from component in wsr and hope in future we will move out from such approach.

https://reactjs.org/docs/composition-vs-inheritance.html

@rghorbani

rghorbani:
@Faradey27 I got your point but I did't actually change the the architecture but in 3 component that use `shouldComponentUpdate `. This way we avoid React warnings and `shouldComponentUpdate` works (`shouldComponentUpdate`does not in `PureComponent`).

src/BaseComponents/WixComponentPure.js
```diff
@@ -0,0 +1,87 @@
...
+class WixComponentPure extends React.PureComponent {
+
+  constructor(params) {
+    super(params);
+    this._addDataHook = this._addDataHook.bind(this);
+    this._supportOnClickOutside = this._supportOnClickOutside.bind(this);
+    this._onMouseEventsHandler = this._onMouseEventsHandler.bind(this);
+  }
+
+  checkIfEventOnElements(e, elem) {
```
hepiyellow: This can be a static function, and we can take it from `WixComponent`.
- Please try to reuse stuff from WixComponent, and not duplicate them.

hepiyellow:
I also don't like that all WSR component are PureComponents, although some of them do not conform to the requirements of a pure component.

You can always add this to workaround the PureComponent optimization:
```
  shouldComponentUpdate() {
     return true
   }
```

Can you specify which React warning you get and under what circumstances ?

rghorbani:
@Faradey27 @hepiyellow 
Actually I didn't change much, It just seems like it.
I renamed `WixComponent` to `WixComponentPure` and changed all the references.
Then I created the new component named `WixComponent` that extends React.Component.

I use latest version of react (16.4.2). I think the optimization you mentioned was removed in react 16.
I get the following warning.

screen shot 2018-08-28 at 9 37 58 pm

hepiyellow:
Wix-style-react does not officially support react 16.

shlomitc:
I'll share my thoughts

Things `WixComponent` does:
- makes sure every component is a pure component
  - useful for all besides very few components that wants to be not-pure.
- adds `dataHook`
  - useful for all
- attaches `onClickOutside` hook
  - useful for dropdown components and Tooltip.
- exposes `setStyles` for typography and styles
  - no one uses it
- exposes `componentElements` to get the actual dom node.
  - only `Tooltip` uses it

Another benfit: It's also a single point to control all components if needed in the future

We had a suggestion to replace it with some HOC but I don't like HOCs for every component because you need to hoist a lot of static data and methods. It's also complicated for `autodocs` to parse these files.

Removing it means more work and place to mistakes in general.

So there are few options:
1. Create `WixComponentNotPure` with less code duplication and just change the relevant components.
2. Suggest good solutions how to add the `dataHook` and `onClickOutside` with minimal hussle.

Other ideas are also welcome

rghorbani:
@Faradey27 @hepiyellow 
I think @shlomitc option 1 is better for now. I can change the names accordingly so it removes the confusion.
What do you think?

shlomitc:
Since there are only three relevant components (`Table`, `DataTable` and `DrillView`), let's make them not inherit the `WixComponent`. It will make things more simple.

Each one of the mentioned components should handle the `dataHook` on its own (put it on the root).

Does this make sense?

rghorbani:
@shlomitc @hepiyellow @Faradey27 
Reverted previous changes.
Implemented `WixComponentNotPure`.
It a copy of `WixComponent` but uses React.Component.
I also deleted `componentElements` since it is not necessary.

shlomitc:
@rghorbani - so I prefer eventually not to have two similar components, but just make these three extend `React.Component` and handle the `dataHook` on their own.

something like:
```jsx
const {dataHook} = this.props
render(
  <div data-hook={dataHook}}>
     ...
  </div>
)
```

And adding it to the `proptypes`.

Thanks and sorry for the mess, but it was a good discussion :)

rghorbani:
@shlomitc What about this [structure](https://github.com/wix/react-native-ui-lib/blob/master/src/commons/baseComponent.js)?

src/SideMenu/DrillView/DrillView.js
```diff
@@ -244,17 +244,19 @@ class SideMenuDrill extends WixComponent {
...
-          <SlideAnimation direction={slideDirection} animateAppear={false} isVisible={showMenuA} {...menuAHandlers}>
-            {this.renderMenu(menuA)}
-          </SlideAnimation>
-          <SlideAnimation direction={slideDirection} animateAppear={false} isVisible={!showMenuA} {...menuBHandlers}>
-            {this.renderMenu(menuB)}
-          </SlideAnimation>
-        </div>
-        {this.props.stickyFooter}
-      </SideMenu>
+      <div data-hook={dataHook}>
```
shlomitc: It looks like there was a bug in first place. The root element (previously `SideMenu`) will always be set with a data-hook.
So you don't need an extra `div` here, just propagate the `dataHook` property to the `<SideMenu dataHook={this.props.dataHook}>`. It will override `drill-view`, but no one uses it.

src/Table/Table.js
```diff
@@ -103,22 +102,25 @@ export class Table extends WixComponent {
...
-            allIds={this.state.data.map((rowData, rowIndex) => defaultTo(rowData.id, rowIndex))}
-            onSelectionChanged={this.props.onSelectionChanged}
-            >
-            {this.renderChildren()}
-          </BulkSelection>
-        ) :
-        this.renderChildren()
-        }
-      </TableContext.Provider>
+      <div data-hook={dataHook}>
```
shlomitc: this is a bit more complex case, at the moment, instead of adding the `div`, just add the data-hook to the `div` in line 99.
no need to handle the case that `withWrapper` is false

PR #2162, opened 2018-08-28
Migrate to new Heading Typography (Add Text mixins)

hepiyellow:


src/TableToolbar/TableToolbar.js
```diff
@@ -17,9 +18,9 @@ Title.propTypes = {
 
 export const SelectedCount = props => {
   return (
-    <span className={typography.t2}>
+    <Text weight="normal" size="medium">
       {props.children}
-    </span>
```
lbelinsk: Default props

PR #2170, opened 2018-08-29
D&D area

Faradey27:
### What changed

...

### Why it changed

...

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


src/DragAndDrop/Draggable/components/Container.js
```diff
@@ -0,0 +1,119 @@
...
+  setChildRef = node => this.childNode = node;
+
+  render() {
+    if (!this.props.connectDropTarget) {
+      return null;
+    }
+    return this.props.connectDropTarget(
+      <div className={this.props.className} ref={this.setRootRef}>
+        <div className={this.props.contentClassName} ref={this.setChildRef}>
+          {this.props.children}
```
shlomitc: why can't `contentClassName` be applied to the children instead? because of the `margin` calculations?
if so - I would also pass in the `ref` function to the children

src/DragAndDrop/Draggable/components/DragLayer.js
```diff
@@ -36,9 +38,12 @@ const CustomDragLayer = ({
   currentOffset
 }) => {
   const shouldRenderLayer = isDragging && id === item.id && itemType === draggedType;
-  const styles = dragLayerStyle({initialOffset, currentOffset});
+  const previewStyles = dragLayerStyle({initialOffset, currentOffset});
+  if (!shouldRenderLayer) {
```
shlomitc: this can come before calculating the `previewStyles`

src/DragAndDrop/Draggable/Draggable.js
```diff
@@ -1,125 +1,46 @@
...
 
-DraggableSource.propTypes = {
-  isDragging: PropTypes.bool,
-  connectDragSource: PropTypes.func,
-  connectDragPreview: PropTypes.func,
-  render: PropTypes.func,
-  id: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
-  withHandle: PropTypes.bool
+Draggable.defaultPropTypes = {
+  containerId: 'DraggableList'
```
shlomitc: Why is a default prop needed? for the simple `SortableList?
I think this is tricky as there might be multiple lists with the same Id.
Why not just force everyone to have one?

src/DragAndDrop/Draggable/components/Container.js
```diff
@@ -0,0 +1,119 @@
...
+      we send original position of item and new one, also id of item and original item state(
+        it required for case, when we moving item from 1 container to another
+      )
+    */
+    props.onHover(dragIndex, hoverIndex, {
+      id: monitorItem.id,
+      item: monitorItem.originalItem
+    });
+    /** set new index for item */
+    monitor.getItem().index = hoverIndex;
```
shlomitc: should this be done after triggering the `onHover` or before?

src/DragAndDrop/Draggable/components/Container.js
```diff
@@ -0,0 +1,119 @@
...
+};
+
+const target = {
+  drop(props, monitor) {
+    return {
+      containerId: props.containerId,
+      index: monitor.getItem().index
+    };
+  },
+  hover(props, monitor, component) {
```
shlomitc: worth unit-testing this part if possible

src/DragAndDrop/Draggable/components/DraggableTarget.js
```diff
@@ -0,0 +1,97 @@
...
+    /**
+      after drop released we send containerId and index of dropped item to dropResult,
+      so endDrag inside of drag source can use this data
+    */
+    return {
+      containerId: props.containerId,
+      index: props.index
+    };
+  },
+  hover(props, monitor, component) {
```
shlomitc: worth testing this part as well.

src/SortableList/SortableList.js
```diff
@@ -1,57 +1,82 @@
...
-  };
+  handleMoveOut = id => {
+    this.setState({items: this.state.items.filter(it => it.id !== id)});
+  }
 
-  _onMove = ({id, from}) => {
-    const to = this.state.items.findIndex(({id: _}) => _ === id);
+  handleHover = (removedIndex, addedIndex, options = {}) => {
+    this.setState(prevState => {
+      const nextItems = copy(prevState.items);
```
shlomitc: as people will probably just copy-paste these lines, can this be a common function everyone can use?

stories/DragAndDrop/SortableList/MultiAreaList.scss
```diff
@@ -0,0 +1,39 @@
...
+
+.itemPlaceholder {
+  border: 1px solid transparent;
+  background-color: #daeffe;
+  color: transparent;
+  cursor: -webkit-grabbing;
+}
+
+.itemPreview {
+  background-color: #eaf7ff;
```
shlomitc: We are going to need a scss file with the common styles.
let's talk about how we should export these styles for users to consume.

1. export a `scss` mixin and use in CSS.
2. export a `js` file the imports the `scss` and exposes the classNames. Then it's just `import {previewClassName} from 'wix-style-react/dnd-styles`

More options?

src/SortableList/SortableList.js
Faradey27: you mean create util with swap function
const arr = [a,b,c]
const nextArr = swap(arr, 0, 2); // [c, b, a]
or copy function?

src/DragAndDrop/Draggable/components/Container.js
Faradey27: because children is an array
@shlomitc

Faradey27: you want to have them as function?

```{
 childre({ref}) // we will apply margin styles by class name from SortableList, or some other component
}```

File: src/DragAndDrop/Draggable/components/DragLayer.js (hunk 2 above)
Comment (Faradey27) [reply src/DragAndDrop/Draggable/components/DragLayer.js]: make sense

File: src/DragAndDrop/Draggable/Draggable.js (hunk 3 above)
Comment (Faradey27) [reply src/DragAndDrop/Draggable/Draggable.js]: ok, make sense

File: src/DragAndDrop/Draggable/DragUtils.js [hunk 9]
```diff
@@ -12,6 +12,6 @@ export const dragCoordinates = ({monitor, component}) => {
   // Get pixels to the top
   const hoverClientY = clientOffset.y - hoverBoundingRect.top;
 
-  return {hoverMiddleY, clientOffset, hoverClientY};
+  return {hoverMiddleY, clientOffset, hoverClientY, hoverBoundingRect};
```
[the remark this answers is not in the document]
Faradey27: yea, i require this during dev, but currently i m not using this

stories/DragAndDrop/SortableList/MultiAreaList.scss
Faradey27: i like js approach more, but in any case it is not so critical, because with custom renderer people require own styles and our styles will come just like demo

PR #2174, opened 2018-08-30
`<Page/>` - Add ellipsis to Header title and subtitle

hepiyellow:
Needed to add css rules to make the containers of the title limited in width, so the ellipsis would work (otherwise the title element simply gets wider and overflows).

Fixes https://github.com/wix-private/wsr-issues/issues/224


shlomitc:
What happens if someone passes a node instead of a pure string text? For example, a drop down or an input. How will it affect it?

shlomitc:
as long as it doesn't have any unpredictable side effect on the node, I'm good with handling only the text

hepiyellow:
I think it doesn't have any side-effects.
I made [this PR](https://github.com/wix/wix-style-react/pull/2182) that makes sure there are no side-effects.
But I want to wait for @Faradey27 to give his opinion about it.

hepiyellow:
After talking with Andrii, I confirmed that there is no harmful side-effect.
Although, consumers should be aware that the Page title is using `text-overflow: ellipsis`, so they can pass some node which is valid for ellipsis to work.
I'll add this to the docs.

src/PageHeader/PageHeader.e2e.js
```diff
@@ -0,0 +1,32 @@
...
+    await browser.get(testStoryUrl(testName));
+    const driver = pageHeaderTestkitFactory({dataHook});
+    await waitForVisibilityOf(driver.element(), 'Cannot find PageHeader');
+    return driver;
+  };
+
+
+  describe('Title only', async () => {
+    eyes.it('should disaply an ellipsed title and subtitle', async () => {
+      await initTest('1. Ellipsed Title and Subtitle', {
```
shlomitc: the object is not used

src/PageHeader/PageHeader.js
```diff
@@ -137,12 +137,12 @@ PageHeader.propTypes = {
   hasBackgroundImage: PropTypes.bool,
   /** Wix-Style-React Breadcrumbs component */
   breadcrumbs: PropTypes.node,
-  /** The main title text */
+  /** The main title text. This text is wrapped with `<Heading ellipsis/>` so the passed node (or result of function) should be a valid node that `text-overflow: ellipsis` can be applied to. */
```
shlomitc: the new comments are redundant in my opinion.

stories/PageHeader/PageHeaderTestStories.js
```diff
@@ -0,0 +1,41 @@
...
+      title="PageHeader title - very very long very very long very very long very very long very very long"
+      subtitle="PageHeader subtitle - very very long very very long very very long very very long very very long very very long very very long"
+      actionsBar={
+        <Button>Action</Button>
+      }
+      showBackButton
+      onBackClicked={() => null}
+      />
+  </PageHeaderContainer>
+    ));
```
shlomitc: lint and indent this file

PR #2178, opened 2018-08-30
WIP - DONT MERGE - Version 4 ts alpha

hepiyellow:
Issues:
- I wanted to upgrade first to yoshi 3, in order to use new tsline eslint yoshi preset.
- Upgrading to yoshi 3 has an issue with stylsheets not loaded in storybook.


shlomitc:
I don't want typescript to be added at the moment to our project.
It has a lot of overhead and we will still lack the `d.ts` files as we allow `js` files. so there is no value for the consumer of the library.

Thanks for the understanding

PR #2179, opened 2018-08-31
Allowed custom className to be passed to Sidemenu components

rghorbani:
### What changed

Added className prop to Sidemenu components

### Why it changed

allowing to customize sidemenu style through passing className

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green


rghorbani:
Any thoughts? @shlomitc @Faradey27 
It will allow independent developers like me to customize their design without doing any harm to original Wix design (As falsy values are just ignored in [classnames](https://github.com/JedWatson/classnames#usage) if className is not set).

PR #2182, opened 2018-09-02
`<Heading/>` & `<Text/>` - Apply (proxy) ellipsis only of children is a string

hepiyellow:
@Faradey27 
Does this sound right to you?

Motivation is only to avoid unexpected side-effects when passing a React Node (not a string),
assuming that ellipsis doesn't really work on it.
If there are cases where passing a node would still work, then maybe we shouldn't have this restriction (to have a `string` child).

### DIFF
--- a/src/Heading/ProxyHeading.js
+++ b/src/Heading/ProxyHeading.js
@@ -7,7 +7,7 @@ import ellipsedStyle from '../common/EllipsedTooltip/EllipsedTooltip.st.css';
 
 const EllipsedHeading = withEllipsedTooltip({showTooltip: true})(Heading);
 
-const ProxyHeading = ({ellipsis, ...props}) => ellipsis ?
+const ProxyHeading = ({ellipsis, ...props}) => ellipsis && typeof props.children === 'string' ?
   <EllipsedHeading {...ellipsedStyle('root', {}, props)} {...props}/> :
   <Heading {...props}/>;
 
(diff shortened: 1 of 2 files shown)


src/Heading/ProxyHeading.js
```diff
@@ -7,7 +7,7 @@ import ellipsedStyle from '../common/EllipsedTooltip/EllipsedTooltip.st.css';
 
 const EllipsedHeading = withEllipsedTooltip({showTooltip: true})(Heading);
 
-const ProxyHeading = ({ellipsis, ...props}) => ellipsis ?
+const ProxyHeading = ({ellipsis, ...props}) => ellipsis && typeof props.children === 'string' ?
```
shlomitc: Not sure this will solve all of our issues.
what about translated components? and components that contain another `Text` component inside (for bold values)?
I think that we need to get the actual text content of the node and render it in the tooltip.

Can you explain also why you need this change?

Faradey27: agree

Faradey27:
@lbelinsk

src/Heading/ProxyHeading.js
hepiyellow: It is not supposed to solve them.
If you say that when passing a non-string child, then the result is unexpected (ellipsis not working in 100% of the cases), then we should limit the scope of the ellipsis solution to work only on strings, until we figure it out.

It should be clear.
Is so, then I guess we should also state it in propTypes, the `children` must be a string.

Faradey27: A lot of people will use this component together with translation(react-intl with FormatedMessage component or some other wix solution) in such cases children can be a component and not a plain string, but that component will render plain string.

i think we should support such case

hepiyellow: I agree that we should support it.
I am asking, what is the current status? 
- is it supported?
- When it is not supported (no ellipsis), are there any side-effects of the `withEllpsedTooltip` that we should avoid?

PR #2185, opened 2018-09-03
`<SideMenuDrill/> - Add ellipsis, vertically center arrow

### LINKED ISSUE #2106: <SideMenuDrill /> layout breaks on long texts
### Summary:
`<SideMenuDrill />` Arrow position breaks when used with long texts:
screen_shot_2018-08-09_at_09 39 33

### Priority:
- [ ] Blocker
- [ ] Urgent
- [x] Normal
- [ ] Minor

### Blocker:
Design guidelines don't cover situations with long texts and how it should wrap.

hepiyellow:
Fixes #2106


hepiyellow:
@milkyfruit 
Hi Domas,
The SubMenu arrow looks like it is a little bit "up" (not centerred).
Could you try and play with it and see if you can center it?
I coudn't.

milkyfruit:
@hepiyellow vertical alignment fixes when icon's parent `<span />`height is set to 24px:

screen shot 2018-09-03 at 17 48 15

By default it was 27px height for some reason...

milkyfruit:
Another thing, the yellow indicator should also align middle on longer texts:
screen shot 2018-09-03 at 17 53 18

hepiyellow:
@milkyfruit yeah. I didn't understand where that `27px` is coming from.

hepiyellow:
@milkyfruit 
about fixing the badge (pimple), I also thought about it.
But fixing that is a breaking change.
So I think we can wait with that fix until this components gets a new UX spec (as @wuwa mentiond in #2106).

src/SideMenu/DrillView/DrillView.scss
```diff
@@ -26,3 +26,7 @@
   flex-direction: column;
   flex: 1;
 }
+
+:global(.rtl) {
+  direction: rtl;
```
shlomitc: should be set on a specific class and not as a global css rule.
```css
:global(.rtl) {
  .drillViewContainer {
    direction: rtl;
  }
}
```

src/SideMenu/core/navigation/styles.scss
```diff
@@ -155,9 +162,7 @@ $padding-vertical: 14px;
 }
 
 :global(.rtl) {
-  .linkArrow {
-    left: 25px;
-    right: auto;
+  .linkArrow svg {
```
shlomitc: instead of targeting the `svg` - just pass the classname to the icon

PR #2191, opened 2018-09-04
FocusableHOC e2e tests improvement + docs for applying focusable.

mykas:


src/common/Focusable/Readme.md
```diff
@@ -0,0 +1,97 @@
+# `withFocusable`
```
hepiyellow: Rename `Readme.md` to `README.md` (its our convension)

PR #2198, opened 2018-09-05
`<ButtonLayout/>` - update typography

hepiyellow:
Completes https://github.com/wix-private/wsr-issues/issues/232

- Add TypographyV5.scss - with Text mixins according to new Typography UX specs.
> The old `Typography.scss` file is deprecated but will stay until next major version.
- Update `<ButtonLayout/>`:
  - Replace all `t1_1`, etc..  style with new mixins.
  - Simplify rules, by extracting all `font-size`, `font-family` to one place.

NOTE: Before merging, please make sue that https://github.com/wix/wix-style-react/pull/2192 is merged.


hepiyellow:
@shlomitc 
This can be considered a breaking change, since Button width change (grow) due to the font weight changing from thin to normal. So it can break designs.

*Shell I add it to PRs for version 6.0?*

shlomitc:
It's not a breaking change. It's just a fix to the typographies as defined by the UX design system

hepiyellow:
@milkyfruit @wuwa 
Hi, 
could you make a quick sanity check for Button typography update.
It supports the old `theme` property (which is not yet deprecated).

### You can use the Playground:
https://wix-wix-style-react-pr-2198.surge.sh/?selectedKind=5.%20Buttons&selectedStory=5.1%20Standard&full=0&addons=0&stories=1&panelRight=0

### Or this "All themes" test page:
https://wix-wix-style-react-pr-2198.surge.sh/iframe.html?selectedKind=Tests%2F5.%20Buttons&selectedStory=5.0%20ButtonLayout

In this test page you can control the `height` and `hover` props using the URL,
by simply adding `&height=small` or `&hover=true`.
For example:
https://wix-wix-style-react-pr-2198.surge.sh/iframe.html?selectedKind=Tests%2F5.%20Buttons&selectedStory=5.0%20ButtonLayout&hover=true&height=x-small

PR #2203, opened 2018-09-05
<Tag/> - align to new definitions

shlomitc:
## What was done:
- move margins from Tag to TagsInput
- fix the remove button styles
- add eyes test
- use up-to-date typography
- fix small and large sizes, margins and padding

## Still missing:
- Tag focus
- Use the Text's Tooltip (can't be used until the `inline` style issue is resolved)


milkyfruit:
@shlomitc one small note: 5.7 Close button's `theme` `close-dark` should be updated as well.
Right now it becomes blue on hover, but it should always remain dark.
[Close button](https://wix-wix-style-react-pr-2203.surge.sh/?selectedKind=5.%20Buttons&selectedStory=5.7%20Close&full=0&addons=0&stories=1&panelRight=0)
[Specs](https://zpl.io/VQqdZ3R)

shlomitc:
@milkyfruit - done.
Could you also take a look? I'm not sure I managed to align the text and the icon correctly there

milkyfruit:
Thanks, @shlomitc the button color looks great now.

I noticed two extra details:

- When `size: small` is set, `padding-right` should be 3px, right now it's 6px
image
For `size:large` it should  remain 6px.

- The close button is shifted off to bottom when is set `size: large`. I noticed that the close button is wrapped with some <a> tag which has a weird `line-height: 26px`,while the button's height should be 24px. Removing it, fixes its vertical positioning.
image

shlomitc:
@milkyfruit - I fixed all the issues but as it was a while ago and I did some changes, I'll be happy for one last review if possible. thanks!

shlomitc:
@milkyfruit - fixed

PR #2205, opened 2018-09-05
<Calendar/> - remove shadow and border-radius from component

yardnsm:
This PR removes the `box-shadow` and `border-raduis` properties from the `<Calendar/>` component. The container component (`<DatePicker/>`, `<Tooltip/>`) is responsible for setting these.

This PR uses the feature from https://github.com/wix/wix-style-react/pull/2200.

Resolves https://github.com/wix-private/wsr-issues/issues/246.


src/DatePicker/DatePicker.scss
```diff
@@ -8,6 +8,11 @@
   z-index: 1;
 }
 
+.calendarRoot :global .DayPicker {
```
shlomitc: that doesn't seem correct.
are you sure that `DayPicker` should be customized?
Why can't it be just a property of `calendarRoot`?

stories/Calendar/ExampleTooltip.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import Calendar from '../../src/Calendar';
```
shlomitc: all these imports should be to `'wix-style-react/Calendar'`, as this is a snippers that users will use

src/DatePicker/DatePicker.scss
yardnsm: All of the styles in `<Calendar/>` applied to the `.DatePicker` class. If we want to have the border radius and the shadow correctly, we have to set these styles to `.DatePicker`.

yardnsm: It is. We can set these styles for `.calendarRoot`, but we'll need to also set `overflow: hidden`.

milkyfruit:
@yardnsm calendar in `<DatePicker/>` and `<Range />` looks great.
`<Calendar/>` component also looks correct, only this example isn't right:
screen shot 2018-09-10 at 16 37 04
When a calendar appears inside a popover, popover's 8px border radius isn't visible. Adding `overflow: hidden` fixes the issue.

Faradey27:
@milkyfruit fixed

src/Calendar/Calendar.js
```diff
@@ -136,7 +137,7 @@ export default class Calendar extends WixComponent {
 
   render() {
     return (
-      <div className={styles.calendar}>
+      <div className={`${styles.calendar} ${this.props.className}`}>
```
shlomitc: use `classnames` package here

src/Tooltip/Tooltip.protractor.driver.js
```diff
@@ -1,5 +1,7 @@
 const tooltipDriverFactory = component => ({
   clickButton: () => component.click(),
+  clickOn: datahook => component.$(`[data-hook="${datahook}"]`).click(),
```
shlomitc: I don't really like the freedom done here. if the consumer has the `data-hook` in this section, it means he knows the component he's clicking. For example, if it's a button, he'll use the Button testkit

src/Tooltip/Tooltip.protractor.driver.js
```diff
@@ -1,5 +1,7 @@
...
+  togglePopover: () => component.$('[data-hook="popover-button"]').click(),
```
shlomitc: why not just `click`?

src/Tooltip/Tooltip.e2e.js
```diff
@@ -37,6 +37,19 @@ describe('Tooltip', () => {
...
     });
+
+    eyes.it('should show tooltip in popover mode', async () => {
+      const EC = protractor.ExpectedConditions;
+      const driver = tooltipTestkitFactory({dataHook});
+      await driver.togglePopover();
+      eyes.checkWindow('Check tooltip with popover styles');
+      await waitForVisibilityOf(driver.getTooltipContentElement(tooltipDataHook), 'Cannot find Tooltip content');
+      await driver.clickOn('tooltip-anchor');
+      await browser.wait(EC.not(EC.presenceOf(driver.getTooltipContentElement(tooltipDataHook))));
```
shlomitc: can't you use `waitForVisibilityOf` and negate the result?

src/Tooltip/Tooltip.e2e.js
```diff
@@ -37,6 +37,19 @@ describe('Tooltip', () => {
...
         });
     });
+
+    eyes.it('should show tooltip in popover mode', async () => {
+      const EC = protractor.ExpectedConditions;
+      const driver = tooltipTestkitFactory({dataHook});
+      await driver.togglePopover();
+      eyes.checkWindow('Check tooltip with popover styles');
+      await waitForVisibilityOf(driver.getTooltipContentElement(tooltipDataHook), 'Cannot find Tooltip content');
+      await driver.clickOn('tooltip-anchor');
```
shlomitc: see comment in the driver

src/DatePicker/DatePicker.scss
```diff
@@ -4,8 +4,12 @@
   display: inline-block;
 }
 
-.calendarRoot {
+.date-picker-calendar {
   z-index: 1;
+  :global .DayPicker {
+    box-shadow: $Shadow30;
```
shlomitc: can this be shared across Tooltip and DatePicker? looks like a common css.

src/Tooltip/Tooltip.js
```diff
@@ -13,6 +13,13 @@ const renderSubtreeIntoContainer = ReactDOM.unstable_renderSubtreeIntoContainer;
 //maintain a 60fps rendering
 const createAThrottledOptimizedFunction = cb => () => window.requestAnimationFrame(throttle(cb, 16));
 
+const popoverConfig = {
```
shlomitc: `theme` should be `light` by default

stories/Calendar/ExampleTooltip.js
```diff
@@ -0,0 +1,24 @@
...
+  <Tooltip
+    popover
+    placement="bottom"
+    alignment="left"
+    showArrow={false}
+    hideTrigger="click"
+    shouldCloseOnClickOutside
+    maxWidth="300px"
+    content={(
+      <Calendar value={new Date()} onChange={() => {}}/>
```
shlomitc: image

is it possible to fix in this PR as well or should be a separate one?
https://github.com/wix/wix-style-react/issues/2213

src/Tooltip/Tooltip.js
```diff
@@ -13,6 +13,13 @@ const renderSubtreeIntoContainer = ReactDOM.unstable_renderSubtreeIntoContainer;
...
+  padding: 0,
+  overflow: 'hidden',
+  showTrigger: 'click',
+  hideTrigger: 'click'
+};
```
shlomitc: the arrow should be lower by default as it causes it to overlap,
image

src/Tooltip/Tooltip.js
Faradey27: not understood

shlomitc: tooltip sets the default `theme` prop to `dark` (see in the story).
for popover - it should be the `light` one

src/Tooltip/Tooltip.e2e.js
Faradey27: i do not like idea to try,catch waitForVisibilityOf + see an error message inside of the logs

stories/Calendar/ExampleTooltip.js
Faradey27: will fix

src/Tooltip/Tooltip.js
Faradey27: position is correct, the problem is in overflow(and zIndex), i ll hide arrow for popover

src/Tooltip/Tooltip.js
Faradey27: done

stories/Calendar/ExampleTooltip.js
Faradey27: fixed

Faradey27:
@shlomitc all comments are fixed or commented

common.scss
```diff
@@ -0,0 +1,45 @@
+@import './common/Focusable/FocusableMixins.scss';
+@import './colors.scss';
+
+@mixin popoverTheme {
```
shlomitc: I this it should sit in a separate area, maybe next to the Tooltip.
having it here will make it impossible to find

src/Tooltip/Tooltip.driver.js
```diff
@@ -86,6 +86,11 @@ const tooltipDriverFactory = ({element, wrapper}) => {
       const values = content.style._values;
       return values.padding;
     },
+    getOverflow: () => {
```
shlomitc: I don't think we should have this kind of function in the driver.
I prefer getting the element instead and verify it in the test.
This is not a function consumers will use in their projects

src/Tooltip/Tooltip.js
shlomitc: the arrow direction is wrong. it points inside the tooltip instead of outside

src/Tooltip/Tooltip.protractor.driver.js
```diff
@@ -1,5 +1,7 @@
 const tooltipDriverFactory = component => ({
   clickButton: () => component.click(),
+  clickOnTooltipAnchor: (datahook = 'tooltip-anchor') => component.$(`[data-hook="${datahook}"]`).click(),
```
shlomitc: this is not part of the component, you can not expose it in the driver

src/Tooltip/Tooltip.protractor.driver.js
```diff
@@ -1,5 +1,7 @@
...
+  click: () => component.$('[data-hook="popover-button"]').click(),
```
shlomitc: same here

src/Tooltip/Tooltip.spec.js
```diff
@@ -366,6 +366,39 @@ describe('Tooltip', () => {
...
+      return resolveIn(30).then(() => {
+        expect(driver.isShown()).toBeTruthy();
+        driver.click();
+        return resolveIn(30).then(() => {
+          expect(driver.isShown()).toBeFalsy();
+        });
+      });
+    });
+
+    it('should not have a padding and should have overflow: hidden', () => {
```
shlomitc: just call it - it should display the right theme.
This is checked in the eyes section but it's ok if it's here as well

src/Tooltip/TooltipContent.js
```diff
@@ -10,6 +10,9 @@ class TooltipContent extends Component {
     /** alignment of the tooltip's text  */
     textAlign: PropTypes.string,
 
+    /** The tooltip overflow */
+    overflow: PropTypes.string,
```
shlomitc: why is this necessary?

typography.scss
```diff
@@ -0,0 +1,140 @@
+/*
```
shlomitc: I think this file shouldn't appear here

src/Calendar/Calendar.scss
```diff
@@ -50,6 +48,7 @@ $hoverBoxShadow: 0 0 0 $boxShadowWidth $F00;
     width: 17px;
     padding: 18px 0 12px;
     color: $D10;
+    word-wrap: normal;
```
shlomitc: is this necessary here by default or only in the case the Tooltip is wrapping it?

Faradey27: for me, it should be default

src/Tooltip/TooltipContent.js
Faradey27: how would i pass overflow inside?

common.scss
Faradey27: it used inside of the Tooltip and DatePicker, for me it should not be inside of the Tooltip

src/Tooltip/Tooltip.driver.js
Faradey27: you want same logic just inside of the test?

common.scss
shlomitc: you're right it shouldn't sit as part of the tooltip scss file, it should be in common section.
however - this file is a bad mixture, it doesn't have any meaning. colors, focus, fonts. 
so I think we should have a new scss file inside the Tooltip folder that shares this common behavior.
it's then easy to understand where styles should come from and for what purpose

src/Tooltip/Tooltip.driver.js
shlomitc: yes

src/Tooltip/Tooltip.protractor.driver.js
shlomitc: need to trigger the tooltip. if it comes as part of the tooltip, awesome. if not, find the relevant triggering element (button?) and click it

src/Tooltip/TooltipContent.js
shlomitc: why do you need the overflow at all?

typography.scss
shlomitc: rebase shouldn't do that. looks to me like this commit pushed it inside
https://github.com/wix/wix-style-react/pull/2205/commits/edbba59eb97578d57f4f8cc087aa0f9da85ed026#diff-35d7f79fc5e005a5d077c11beac4ece9

src/Tooltip/Tooltip.e2e.js
shlomitc: ok, keep it

src/Calendar/Calendar.scss
shlomitc: This is the default behavior in css, but the fact that you need to re-decalre it is because the Tooltip sets it differently

I think this css rule should be set only for tooltips and not popovers
https://github.com/wix/wix-style-react/blob/master/src/Tooltip/TooltipContent.scss#L20

src/Tooltip/TooltipContent.js
Faradey27: we disscuss this already, i need it to show correct border, see Domas comments

typography.scss
Faradey27: this file not visible on my machine, i think problem in case-sensetive

PR #2206, opened 2018-09-05
WIP Chore/card collapsable header

argshook:
### What changed

...

### Why it changed

...

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green

### DIFF
--- a/src/Card/Card.driver.js
+++ b/src/Card/Card.driver.js
@@ -1,3 +1,4 @@
 export default ({element}) => ({
-  exists: () => !!element
+  exists: () => !!element,
+  getChildren: () => element.innerHTML
 });
--- a/src/Card/Card.driver.js
+++ b/src/Card/Card.driver.js
@@ -1,0 +1,3 @@
+export default ({element}) => ({
+  exists: () => !!element
+});
--- a/src/Card/Card.js
+++ b/src/Card/Card.js
@@ -35,8 +35,7 @@ class Card extends React.Component {
   };
 
   static defaultProps = {
-    stretchVertically: false,
-    isOpen: true
+    stretchVertically: false
   };
 
   static Header = Header;
@@ -49,31 +48,38 @@ class Card extends React.Component {
   constructor(props) {
     super(props);
 
-    const isOpen = ensureArray(this.props.isOpen);
+    const isOpen = typeof this.props.isOpen !== 'undefined' ?
+      ensureArray(this.props.isOpen) :
+      ensureArray(this.props.children).map(() => true);
 
     this.state = {
-      childrenState: {
-        ...ensureArray(this.props.children).reduce((acc, curr, index) => {
-          acc[index] = {
-            collapsed:
-              typeof isOpen[index] !== 'undefined' ? !isOpen[index] : false
-          };
-          return acc;
-        }, {})
-      }
+      open: isOpen.reduce((acc, curr, index) => {
+        acc[index] = curr;
+        return acc;
+      }, {})
     };
   }
 
+  componentDidUpdate(prevProps) {
+    if (this.props.isOpen !== prevProps.isOpen) {
+      const isOpen = ensureArray(this.props.isOpen);
+
+      this.setState({
+        open: isOpen.reduce((acc, curr, index) => {
+          acc[index] = curr;
+          return acc;
+        }, {})
+      });
+    }
+  }
+
   getChildrenInterface = key => ({
-    isOpen: !this.state.childrenState[key].collapsed,
+    isOpen: this.state.open[key],
     toggle: () =>
       this.setState({
-        childrenState: {
-          ...this.state.childrenState,
-          [key]: {
-            ...this.state.childrenState[key],
-            collapsed: !this.state.childrenState[key].collapsed
-          }
+        open: {
+          ...this.state.open,
+          [key]: !this.state.open[key]
         }
       }),
     CONTENT_SPLIT
@@ -113,7 +119,7 @@ class Card extends React.Component {
           <div data-hook="card-visible-children">{visibleChildren}</div>
 
           <Collapse
-            isOpened={!this.state.childrenState[key].collapsed}
+            isOpened={this.state.open[key]}
             children={collapsableChildren}
             />
         </div>
--- a/src/Card/Card.js
+++ b/src/Card/Card.js
@@ -1,6 +1,7 @@
 import React from 'react';
-import {bool, node, string} from 'prop-types';
-import classNames from 'classnames';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import Collapse from 'react-collapse';
 
 import Content from './Content';
 import Header from './Header';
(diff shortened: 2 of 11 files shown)


hepiyellow:
My suggestion is:



# `<Collapsable/>`

## UnControlled Example
```js
export const collap1 = () => (
  <Collapsable initiallyOpened>
    <Collapsable.Control>
      {({toggle, isOpen}) => (
        <ToggleSwitch onChange={toggle} checked={isOpen}/>
      ) }
    </Collapsable.Control>

    <Collapsable.Content>
      My content
    </Collapsable.Content>
  </Collapsable>
);
```

## Controlled Example
```js
class MyCollapsable { 
  state= {isOpened: true};

  render() {
    <Collapsable isOpen={this.state.isOpened}>
      <ToggleSwitch onChange={this.handleToggle} checked={this.state.isOpened}/>

      <Collapsable.Content>
        My content
      </Collapsable.Content>
    </Collapsable>
  }
};
```

## Collapsable Props
| propName          | propType | defaultValue | isRequired | description    |
| ---               | ---      | ---          | ---        | ---            |
| initiallyOpened   | bool     | true         | -          |  Sets initial opened state in Uncontrolled mode.|
| isOpened   | bool     | -          | -          |  If defined, then sets isOpened ( Controlled)|

## CollapsableContext
Collapsable is a provider of a React context , called CollapsableContext.
```js
interface CollapsableContext {
  isOpened: boolean;
  toggle: ()=>void;
}
```

## <Collapsable.Control/>
A consumer of the `CollapsableContext`.

## <Collapsable.Content/>
Wrapps children with `Collapse` (from `react-collapse`),
and passes the `isOpened` from the `CollapsableContext`.

### Control within Content
It is possible to put `<Collapsable.Content/>` within a `<Collapsable.Control/>`, and then you can also put a Close button inside the collapsable content.
<details>

  <summary>Example</summary>

```js
export const collap1 = () => (
  <Collapsable initiallyOpened>
    <Collapsable.Control>
      {({toggle, isOpen}) => (
        <ToggleSwitch onChange={toggle} checked={isOpen}/>
      ) }
      <Collapsable.Content>
        My content
        <Button onClick={toggle}>Close</Button>
      </Collapsable.Content>
    </Collapsable.Control>
  </Collapsable>
);
```

</details>

### Multiple Contents
It is also possible to have multiple `<Collapsable.Content>` under one `<Collapsable>`,
so that they all collaple/expand together.

<details>

  <summary>Multiple Contents Example (Controlled)</summary>

```js
class MyCollapsable { 
  state= {isOpened: true};

  render() {
    <Collapsable isOpen={this.state.isOpened}>
      <ToggleSwitch onChange={this.handleToggle} checked={this.state.isOpened}/>
      
      <Heading>Part 1</Heading>
      <Collapsable.Content>
        Part 1 content
      </Collapsable.Content>

      <Heading>Part 2</Heading>
      <Collapsable.Content>
        Part 2 content
      </Collapsable.Content>
    </Collapsable>
  }
};
```
</details>


# `<Card/>`

## Examples 

### Simplest Uncontrolled (single collapsed Card.Content)
```js
export const card1 = () => (
  <Card collapsable initiallyOpened>
    <Card.Header title="Collapsable Header"/>
    <Card.Content key="content">My collapsable content</Card.Content>
  </Card>
);
```

#### Explanation
- The idea is to embed `<Collapsable/>` inside `<Card/>` for a super-easy-to-use API.
- When `<Card/>` has a `collapsable=true` prop, then it wraps it's children with a  `<Collapsable/>`.
- `<Card.Header/>` is aware of `<Card/>`'s `collapsable` prop, if collapsable=true, then it will render our old <Card.CollapsedHeader/> (which we will make a `CollapsableContext` consumer)
- `<Card.Content/>` is also aware of `<Card/>`'s `collapsable` prop, if `collapsable=true`, then it will wrap it's children with `<Collapsable.Content/>


### MultiSectionCard
```js

const CollapsableToggle = () => (
  <Collapsable.Control>
    {({toggle, isOpened}) => (
      <ToggleSwitch onChange={toggle} checked={isOpened}/>
    )}
  </Collapsable.Control>
);

const MultiSectionCard = () => (
  <Card>
    <Card.Header title="Card Header"/>
    
    <Collapsable initiallyOpened>
      <CollapsableToggle/>
      <Collapsable.Content>
        <Card.Content key="content">My collapsable content 1</Card.Content>
      </Collapsable.Content>
    </Collapsable>

    <Collapsable initiallyOpened>
      <CollapsableToggle/>
      <Collapsable.Content>
        <Card.Content key="content">My collapsable content 2</Card.Content>
      </Collapsable.Content>
    </Collapsable>
  </Card>
);
```

argshook:
The idea of this PR and layout cleanup in general is to reduce the
amount of components we have, in this case get rid of
`Card.CollapsedHeader`. You suggest the opposite - add another component
`Collapsable` and still use `Card.CollapsedHeader>` which doesn't seem
like something we want now.

`<Collapsable/>` looks pretty much like `react-collapse`. That one is used
inside `Card.CollapsedHeader` (and in `Card` in this PR). We can expose
`<Collapsable/>` but it would be just proxy to `react-collapse` and how
useful is that?

the following api is something me and @shlomitc came up and is
implemented in this PR:

```js
<Card>
  {({ toggle, CONTENT_SPLIT, isOpen }) => [
    <Card.Header key="0" suffix={<Button>{isOpen ? 'Close' : 'Open'}</Button>}/>,

    CONTENT_SPLIT,

    <Card.Content key="1">
      Hello
    </Card.Content>
  ]}
</Card>
```

functionality wise it's great, because:
* it works controlled or not,
* can have multiple children (multiple collapsable contents) ((this is not required, i added for fun, cause possible without additional api))
* any kind of header or content
* any kind of content visiblity trigger

also to note:
* no new components
* no deep nesting

but for consumer it kinda sucks, because:
* must return array
* must add keys to each item
* weird `CONTENT_SPLIT`

i'd focus to solve just above free, with maybe the following:

```js
<Card
  header={({toggle, isOpen}) =>
    <Card.Header
      title="hello"
      suffix={<Button onClick={toggle}>{isOpen ? 'Close' : 'Open'}</Button>}/>
    />
  }
>
  <Card.Content>
    Hello
  </Card.Content>
</Card>
```

@hepiyellow  @shlomitc wdyt?

shlomitc:
Regarding the multi-level card collapsing (`<Card isOpen={[true, false]}>`).
This is not a request I know about nor a common use case, so let's not add this feature now.
If someone wants a multi-level collapsing, we should advise him to use `react-collapse` or our own `Collapse` component if we need a specific one.

I'm tempted to suggest having the collapse behavior as a code snippet and not as part of the component.
1. Easier to maintain
2. Gives more freedom (e.g. multi-level)
3. Today, testing the `CollapsibleHeader` requires adding a the `RequestAnimationFrame` polyfill, so if we do add the collapsing capability to the card, it will be a breaking change to everyone.

I know it's not what we discussed first, but it might be the simplest solution and it's just few lines of code. WDYT?

argshook:
@shlomitc

> I'm tempted to suggest having the collapse behavior as a code snippet and not as part of the component.

something like this?

```js
import Card from 'wix-style-react/Card';
import Collapse from 'wix-style-react/Collapse';

// ...

const { open } = this.state;

return (
  <Card>
    <Card.Header
      title="cheap socks"
      suffix={
        <Button
          onClick={this.toggleOpen}
          children={open ? 'Close Deal' : 'Open Deal'}
        />
    />

    <Collapse open={open}>
      <Card.Content>
        Buy one for the price of two
      </Card.Content>
    </Collapse>
  </Card>
);
```

this means new component `<Collapse>` (proxies `react-collapsable`) and no more `Card.CollapsableHeader`

thumbs up and i'll do this :)

mykas:
I think @argshook last suggestions looks clean and without additional weird `content split`. Just one more question: do we intend to have `collapse` or `accordion`? It seems that the only difference is that `accordion` will only let to show one expanded card at the time.

hepiyellow:
I am ok with arius’s last example. 
It’s a subset of what i suggested.  We can extend it later (although nameing changes would probably be needed)

I thought that the majority of collapsable card usage is uncontrolled. And this solution is controlled, and will force consumer to manage state. 
Which is not a big deal. 

We can have a snippet that implements an uncontrolled Card wrapper, so at least consumers could copy&paste.

shlomitc:
OK, so I think we all agree with the suggested solution to have a snippet and a utility instead of binding the behavior to the general Card component.

Now questions are:
1. Do we need our own `Collapse` component or should we just recommend `react-collapsible`?
2. `react-collapsible` or something else?

For Q.1 I'd say that if we have well defined UX definitions for the collapse behavior, then we should have one.

argshook:
closing in favor of #2251

PR #2208, opened 2018-09-06
docs: rename Tags to MultiSelect in storybook

ChenRoth:
### What changed
name in storybook

### Why it changed

so it would match the component's name

---

- [ ] Change is tested
- [X] Change is documented
- [ ] Build is green


shlomitc:
The component should be named `Tags`. It's an old code which we didn't want to change the name of it in order not to break API.
If the intention is to make the docs better - just add it: `Tags (MultiSelect)`

shlomitc:
Also, changing the page name requires changing it also in the e2e tests
https://github.com/wix/wix-style-react/blob/92920bddd6b2464906dfa25f93f04cfca51a71de/src/MultiSelect/MultiSelect.e2e.js#L8

shlomitc:
@ChenRoth - will you do the requested change?

PR #2217, opened 2018-09-06
allowed custom update and remove buttons

rghorbani:
### What changed

Allowed to show or hide update or remove buttons in ImageViewer

### Why it changed

Wanted an option to hide update button

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green


shlomitc:
@milkyfruit - what do you think about such a feature?
Currently, in the design system, both edit and update buttons are mandatory.
Should it be possible to remove some of them? If the answer is yes - what should happen when both are removed?

rghorbani:
@shlomitc 
I think the option should be given.
If both are removed the focusing (image scaling a bit on hover) of image should be disabled (It's not right now).

rghorbani:
@shlomitc Sorry, Didn't see the mention.

milkyfruit:
@rghorbani Why do you need to remove these buttons? If both are removed, a user can't remove/change an image anymore. That's not a good experience.
Could you share the mockups with expected flow?

rghorbani:
@milkyfruit In some cases like mine I only need to have the remove option available. So I though I could provide the option to have only one of the functions or none of them depending on the goal.
This PR keeps Wix's original design but provides more freedom.

shlomitc:
@milkyfruit - allowing to remove only the update button make sense?

shlomitc:
Technically I'm fine with allowing to remove only the update button and not the remove button. @milkyfruit - is that ok with you?

milkyfruit:
@shlomitc @rghorbani yes, we can go with allowing to disable (hide) `update` button, but leaving the `remove`.

rghorbani:
@shlomitc 
Did the changes requested.
Thanks for the help.

PR #2219, opened 2018-09-08
<BadgeSelect/> - create new component

Noam3kCH:
### What changed
Guild Week task 18
Added new component badge group. which is badge with dropdown to choose other badges.
per zeplin: https://app.zeplin.io/project/5864e02695b5754a69f56150/screen/5a1eee8100d7a41d2f6c7801

- [x] Change is tested
_component tests completed. e2e tests **not**._  
- [x] Change is documented
_added storybook page under 'Other'_
- [ ] Build is green
_it will not be green until merged wix-ui-backoffice PR: https://github.com/wix/wix-ui-backoffice/pull/216
which this change relies on._
_it's in zeplin so we assume it is._

**Caveats - and further work to be done**
- since the badge could potentially change width on every selection, the dropdown width changes accordingly. this could be a problem in various cases. I suggest adding a width param to this component, and as a prerequisite, to the wix-ui-backoffice badge component.
- e2e tests were not added due to time restrictions.
- A driver factory was made per the paradigm of other components, we're unsure about how this is exported into a testkit.
- to test how this looks without merging first, you could build the badgeSize branch on wix-ui-backoffice and copy over the badge component.  

image


hepiyellow:
Sounds to me like the name `<BadgeGroup/>` is not right.
It's more of a `<BadgeSelect/>`. 
It behaves like a native select.
We save the term group for stuff like Radio-Group, Thumbnail-Group, etc...

@wuwa @milkyfruit Do you agree?

wuwa:
agree.

yardnsm:
#### TODO:

- [ ] Make the build green
- [ ] Add e2e test
- [ ] Support the focus styles as described in the Zeplin
- [ ] Expose testkit drivers

stories/Builders/BadgeSelectItemBuilder/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+This builder is being used in [`<BadgeSelect/>`](https://wix-wix-style-react.surge.sh/?selectedKind=12.%20Other&selectedStory=12.2%20BadgeSelect&full=0&addons=0&stories=1&panelRight=0) under the hood.
+
+### `badgeSelectItemBuilder({ id: string | number, text: string, skin: string }): DropdownLayoutOption`
+
+| Name | Type | Required | Description |
+| ---- | ---- | -------- | ----------- |
+| `id`   | `string`, `number` | `true` | The id of the option, should be unique. |
+| `text` | `string` | `true` | The text to render within the badge. |
+| `skin` | `string` | `true` | The badge's skin. See the [`<Badge/>`](https://wix-wix-style-react.surge.sh/?selectedKind=12.%20Other&selectedStory=12.1%20Badge&full=0&addons=0&stories=1&panelRight=0) story for the full list of available skins. |
```
shlomitc: Please don't have links to production urls. prefer having a link below this label that uses the storybook `linkTo` component

stories/Builders/BadgeSelectItemBuilder/Example.js
```diff
@@ -0,0 +1,29 @@
...
+
+const style = {
+  display: 'inline-block',
+  padding: '0 5px',
+  width: '240px',
+  lineHeight: '22px'
+};
+
+const options = [
+  {id: 0, value: 'Option 1'},
```
shlomitc: these are not relevant

src/BadgeSelectItemBuilder/BadgeSelectItemBuilder.js
```diff
@@ -0,0 +1,32 @@
...
+        >
+        {text}
+      </Text>
+    </span>
+  </div>);
+};
+
+badgeOption.propTypes = {
+  id: PropTypes.string.isRequired,
+  text: PropTypes.string.isRequired,
```
shlomitc: this should be a node since translations sometime appear as components

stories/Builders/BadgeSelectItemBuilder/Example.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
```
shlomitc: @LittleTal @wuwa - please review this page as well. this is the beginning for the "Item Builder" pattern we've been talking about
https://wix-wix-style-react-pr-2219.surge.sh/?selectedKind=WIP%2FBuilders&selectedStory=BadgeSelectItemBuilder&full=0&addons=0&stories=1&panelRight=0

src/BadgeSelect/BadgeSelect.js
```diff
@@ -0,0 +1,146 @@
...
+import PropTypes from 'prop-types';
+import map from 'lodash/map';
+import {badgeSelectItemBuilder} from '../BadgeSelectItemBuilder';
+import styles from './BadgeSelect.scss';
+import noop from 'lodash/noop';
+import Badge from '../Badge/Badge';
+import {ChevronDown} from 'wix-ui-icons-common';
+import ReactDOM from 'react-dom';
+
+export default class BadgeSelect extends React.Component {
```
shlomitc: please add a static displayName

src/BadgeSelect/BadgeSelect.js
```diff
@@ -0,0 +1,146 @@
...
+    }
+  }
+
+  /**
+   * Determine if a certain key should open the DropdownLayout
+   *
+   * @param {KeyboardEvent.key} key - The key name
+   * @return {boolean} - Whether the key should cause the DropdownLayout to open
+   */
+  isOpenKey(key) {
```
shlomitc: please add a `_` prefix to all internal functions

src/BadgeSelectItemBuilder/BadgeSelectItemBuilder.js
```diff
@@ -0,0 +1,32 @@
...
+        tagName="span"
+        weight="normal"
+        >
+        {text}
+      </Text>
+    </span>
+  </div>);
+};
+
+badgeOption.propTypes = {
```
shlomitc: please add documentation on top of each of the functions

src/BadgeSelect/README.TESTKIT.md
```diff
@@ -0,0 +1,42 @@
+# `<BadgeSelect/>` component
+
+## Enzyme / ReactTestUtils TestKit API
+
+The following TestKit is exposing the driver for the [`<Badge/>`
+component](https://wix-wix-style-react.surge.sh/?selectedKind=12.%20Other&selectedStory=12.1%20Badge&full=0&addons=0&stories=1&panelRight=0),
```
shlomitc: also here, prefer not having links, it will be very hard to maintain it

src/BadgeSelect/BadgeSelect.driver.js
```diff
@@ -0,0 +1,18 @@
...
+  const dropdownLayout = element.querySelector('[data-hook=badgeSelect-dropdownLayout]');
+  const badgeDriver = badgeDriverFactory({element: badgeWrapper.childNodes[0], wrapper: badgeWrapper, eventTrigger});
+  const dropdownLayoutDriver = dropdownLayoutDriverFactory({element: dropdownLayout, wrapper});
+
+  const driver = {
+    /** Returns 'true' wether the element exists */
+    exists: () => !!element
+  };
+
+  return {driver, badgeDriver, dropdownLayoutDriver};
```
shlomitc: The driver should be a simple abstraction over common actions, meaning it should expose only the relevant interactions to the consumer, and not all drivers.
In my opinion, the only functions that should be here are `selectOption(id)` and maybe `getSelectedValue`(or id).
can stay internally

src/BadgeSelect/BadgeSelect.protractor.driver.js
```diff
@@ -0,0 +1,14 @@
+const badgeSelectDriverFactory = component => ({
```
shlomitc: same comments as for the enzyme test driver

src/BadgeSelect/BadgeSelect.js
```diff
@@ -0,0 +1,146 @@
...
+    onSelect: PropTypes.func,
+    size: PropTypes.string,
+    type: PropTypes.string,
+    uppercase: PropTypes.bool,
+    dataHook: PropTypes.string
+  };
+
+  static defaultProps = {
+    onSelect: noop,
+    dataHook: ''
```
shlomitc: are you sure it is needed?

src/BadgeSelect/BadgeSelect.js
```diff
@@ -0,0 +1,146 @@
+import React from 'react';
+import DropdownLayout from '../DropdownLayout';
+import PropTypes from 'prop-types';
+import map from 'lodash/map';
+import {badgeSelectItemBuilder} from '../BadgeSelectItemBuilder';
+import styles from './BadgeSelect.scss';
+import noop from 'lodash/noop';
+import Badge from '../Badge/Badge';
+import {ChevronDown} from 'wix-ui-icons-common';
```
shlomitc: this will import the entire file, pick only the relevant icon `wix-ui-icons-common/system/ChevronDown`

src/BadgeSelect/BadgeSelect.e2e.js
```diff
@@ -0,0 +1,55 @@
...
+      await createDriverFactory();
+    });
+  });
+
+  describe('dropdown layout', () => {
+    beforeEach(async () => {
+      await browser.get(storyUrl);
+    });
+
+    eyes.it('should open the dropdown on click', async () => {
```
shlomitc: this is more of an `openDropdown` function

stories/Builders/BadgeSelectItemBuilder/Example.js
yardnsm: I want to show how they can be used with regular dropdown options

src/BadgeSelect/BadgeSelect.driver.js
yardnsm: There are a lot of tests that checks some Badge props, for example:

https://github.com/wix/wix-style-react/blob/56851f40162b62ca3157017743cde93df27a6be8/src/BadgeSelect/BadgeSelect.spec.js#L33-L38

Should we remove them?

src/BadgeSelect/BadgeSelect.e2e.js
yardnsm: wdym?

stories/Builders/BadgeSelectItemBuilder/Example.js
shlomitc: even if technically possible, the idea is to have only a group of badges, so design-wize it is not something they should do

src/BadgeSelect/BadgeSelect.driver.js
shlomitc: let's stick with the private/public driver approach like here
https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/button-next/button-next.driver.private.ts.

src/BadgeSelect/BadgeSelect.e2e.js
shlomitc: sorry, I meant to mark the `createDriverFactory` function

src/BadgeSelect/BadgeSelect.e2e.js
```diff
@@ -0,0 +1,55 @@
+import eyes from 'eyes.it';
+import autoExampleDriver from 'wix-storybook-utils/AutoExampleDriver';
+import {createStoryUrl, waitForVisibilityOf, scrollToElement} from 'wix-ui-test-utils/protractor';
+
+import {badgeSelectTestkitFactory} from '../../testkit/protractor';
+import {storySettings} from '../../stories/BadgeSelect/storySettings';
+
+// eslint-disable-next-line
+fdescribe('BadgeSelect', () => {
```
shlomitc: also, `fdescribe`

src/BadgeSelect/BadgeSelect.e2e.js
yardnsm: So add a new `openDropdown` function to the driver? Maybe `clickBadge` is a better name?

yardnsm:
> Chevron svg icon should be in the original size of 24px (without width and height props).

@LittleTal The problem is that the Chevron icon has some kind of padding in its svg. Also, `<Badge/>` component "shrinks" the icon to be `12x12`, and with the padding of the Chevron icon is does look very tiny...
I've managed to make the icons in `<Badge/>` to their original size, and with this change the Chevron icon looks ok. The problem is that change could potentially break other designs...

@shlomitc wdyt?

shlomitc:
@yardnsm - regarding Chevron size. looks to me like the Badge should not have padding, but use min-height instead. what is potentially breaking?

src/BadgeSelect/BadgeSelect.driver.private.js
```diff
@@ -0,0 +1,22 @@
...
+import badgeSelectPublicDriverFactory from './BadgeSelect.driver';
+
+const badgeSelectDriverFactory = ({element, wrapper, eventTrigger}) => {
+  const badgeWrapper = element.querySelector('[data-hook=badgeSelect-badge-wrapper]');
+  const dropdownLayout = element.querySelector('[data-hook=badgeSelect-dropdownLayout]');
+  const badgeDriver = badgeDriverFactory({element: badgeWrapper.childNodes[0], wrapper: badgeWrapper, eventTrigger});
+  const dropdownLayoutDriver = dropdownLayoutDriverFactory({element: dropdownLayout, wrapper});
+
+  const driver = {
+    ...badgeSelectPublicDriverFactory({element, wrapper, eventTrigger})
```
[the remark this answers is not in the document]
yardnsm: why? don't we want to extends the public driver?

src/BadgeSelect/BadgeSelect.driver.private.js
```diff
@@ -0,0 +1,22 @@
...
+  };
+
+  return {
+    driver,
+    badgeDriver,
```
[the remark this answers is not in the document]
yardnsm: We can get rid of `dropdownLayoutDriver`, but there are some function in `badgeDriver` that I don't think it's necessary to explicitly expose.

yardnsm:
@shlomitc My bad - the icons does look a lot better with the change, but it could break existing design.

### Before the change:

screen shot 2018-10-11 at 13 51 18
screen shot 2018-10-11 at 13 51 23
screen shot 2018-10-11 at 13 51 28
screen shot 2018-10-11 at 13 51 34

### After the change:

screen shot 2018-10-11 at 13 50 17
screen shot 2018-10-11 at 13 50 50
screen shot 2018-10-11 at 13 50 55
screen shot 2018-10-11 at 13 51 01

I'll open the PR in `wix-ui-backoffice`.

src/BadgeSelect/BadgeSelect.scss
```diff
@@ -0,0 +1,22 @@
...
+}
+
+.dropdown {
+  position: absolute;
+  left: 50%;
+  transform: translateX(-50%);
+
+  /* make dropdown layout height responsive */
+  /* TODO: make this a little less "hacky" */
+  & > div > div {
```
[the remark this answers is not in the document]
yardnsm: We also need a way to target a single option. For the current implementation, passing a class names seems to reasonable solution. Maybe it'll be better to also pass a class name for the container?

For container class name we'll use the prop `className`.
For option class name we'll use the prop `optionClassName`.

We can still keep the `isInContainer`, as a way to "cleanup" the DropdownLayout default container styles, but is it needed?

hepiyellow:
I suggest not having these huge PRs. 
Simply merge feature subsets, and put it under WIP in the storybook. 
This is impossible to review.

src/BadgeSelect/BadgeSelect.spec.js
```diff
@@ -0,0 +1,130 @@
...
+  it('should change badge text after an option is selected', () => {
+    const {driver, badgeDriver} = createComponent();
+    const selectedIndex = 3;
+
+    driver.click();
+    driver.clickAtOption(selectedIndex);
+    expect(badgeDriver.text()).toBe(options[selectedIndex].text);
+  });
+
+  describe('testkit', () => {
```
shlomitc: move this to the testkit/index.spec.js file

src/BadgeSelect/BadgeSelect.spec.js
```diff
@@ -0,0 +1,130 @@
...
+    it('should exist', () => {
+      expect(isTestkitExists(
+        <BadgeSelect options={options} selectedId={initialOptionId.toString()}/>,
+        badgeSelectTestkitFactory,
+      )).toBe(true);
+    });
+  });
+
+  describe('enzyme testkit', () => {
```
shlomitc: move this to the testkit/enzyme.spec.js file

src/BadgeSelect/BadgeSelect.scss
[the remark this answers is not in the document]
yardnsm: The problem is that an option has it's own padding. The `<BadgeSelect/>`'s options should have a different padding, and we need a way to target an option for this.

PR #2218, opened 2018-09-06
Added RTL support to TableToolbar & custom className to Sidemenu.Header

rghorbani:
### What changed

- added missing className prop for Sidemenu.Header from #2179 
- Added RTL support to TableToolbar

### Why it changed

- Fixed missing prop
- Added another RTL support

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green
- [ ] Change of UX is approved by [wuwa](https://github.com/wuwa) or [milkyfruit](https://github.com/milkyfruit)

PR #2222, opened 2018-09-09
Added RTL support to TableToolbar

rghorbani:
### What changed

Added RTL support to TableToolbar (split from #2218)

### Why it changed

Another RTL support

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green


rghorbani:
@shlomitc Evidence:

screen shot 2018-09-10 at 8 02 36 am

stories/TableToolbar/MainExampleRTL.js
```diff
@@ -0,0 +1,63 @@
...
+    ];
+
+    const filterOptions = [
+      {id: 0, value: 'All'},
+      {id: 1, value: 'Red'},
+      {id: 2, value: 'Cyan'}
+    ];
+
+    return (
+      <div className="rtl" style={{direction: 'rtl'}}>
```
shlomitc: The RTL concept of `wix-style-react` says that an `rtl` global class name can be applied to the top of the application and all content will be aligned properly.
so the `direction: 'rtl'` should be part of the stylesheet and not part of the example.

rghorbani: I changed it to
<div className="rtl" dir="rtl">
  <MainExample/>
</div>
I needed to apply direction to see the actual result.

PR #2224, opened 2018-09-10
<Table/> - fix broken <TableToolbar/> docs and <TableActionCell/> RTL

rghorbani:
### What changed

- Fixed rtl bug in PopoverMenuItem related to TableActionCell
- Fixed deprecated import of Table/Toolbar -> TableToolbar
- RTL bug in Chevron (@shlomitc Help Wanted!)

### Why it changed

Improved RTL view and fixed bugs

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green


rghorbani:
@shlomitc Evidence:

screen shot 2018-09-10 at 8 02 36 am

src/TableActionCell/TableActionCell.st.css
```diff
@@ -15,6 +15,7 @@
 .hoverSlot.notOnHover {}
 .placeholderIcon {}
 
+/* TODO: This line does not work and even produces Stylable warning for unscoped native element `tr` */
```
hepiyellow: Stylable has no nesting feature (nesting is a Sass feature)

rghorbani: The new way doesn't work either. Changed only this part to Sass.

src/TableActionCell/TableActionCell.st.css
```diff
@@ -14,7 +14,3 @@
 .hoverSlot.onHover {}
 .hoverSlot.notOnHover {}
 .placeholderIcon {}
-
-:global(.rtl) tr .actionsContainer .placeholderIcon svg {
```
shlomitc: this should work in the `stylable` syntax instead of vreating a new `scss` stylesheet
```
:global(.rtl) .placeholderIcon svg {
  transform: scaleX(-1);
}
:global([dir="rtl"]) .placeholderIcon svg {
  transform: scaleX(-1);
}
```

I added the `[dir=rtl]` because the automated story is using it to emulate RTL.
Also - it looks like this will be the right way to do RTL, so we will probably will want to support both `className` and `dir` in the meanwhile until we move everyone to `dir`

src/PopoverMenuItem/PopoverMenuItem.scss
```diff
@@ -52,3 +52,16 @@
     margin-top: 4px;
   }
 }
+
+:global(.rtl) {
```
shlomitc: also worth adding here the global `.global([dir="rtl"])` if you want to see it displayed correctly in the preview window

src/TableActionCell/TableActionCell.st.css
rghorbani: Thanks for the help.

src/PopoverMenuItem/PopoverMenuItem.scss
rghorbani: I too actually prefer to have single style file also. I just added Sass because I couldn't get around the bug in Stylable.

PR #2230, opened 2018-09-12
<InfiniteScroll/> scrollElement as function, <Table/> infinite scroll example

Wo1v3r:
### What changed

- < Page / >Added `scrollableContentRef` optional prop to <Page/> to attach ref of scrollable content
- < Table / > Added an example to <Table/> stories of an infinite scrolled <Table/> with <Page/> inside

### Why it changed

- To support infinite scrolling in a <Table/> which is not the direct scrollable container

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


src/Page/Page.js
```diff
@@ -332,6 +337,9 @@ Page.propTypes = {
   gradientClassName: PropTypes.string,
   /** If false Gradient will not cover Page.Tail */
   gradientCoverTail: PropTypes.bool,
+  /** Is called with the Page's scrollable content ref **/
+  setScrollableContent: PropTypes.func,
```
shlomitc: call it `scrollableContentRef`

stories/Table/TableInfiniteScrollExample.js
```diff
@@ -0,0 +1,180 @@
...
+    const loadMoreData = () => {
+      this.setState({ count: this.state.count + 5 });
+      if (this.state.count > 20) {
+        this.setState({ hasMore: false });
+      }
+    };
+    setTimeout(loadMoreData, 1000);
+  }
+
+  scrollElement() {
```
shlomitc: rename to `getScrollElement`

stories/Table/TableInfiniteScrollExample.js
```diff
@@ -0,0 +1,180 @@
...
+    return this.refs['scrollableContent'] || window;
```
shlomitc: try to see it `window` can be removed or moved to the infinitescroll

src/DataTable/InfiniteScroll.js
```diff
@@ -11,7 +11,7 @@ export default class InfiniteScroll extends Component {
     threshold: PropTypes.number,
     useWindow: PropTypes.bool,
     isReverse: PropTypes.bool,
-    scrollElement: PropTypes.object,
+    scrollElement: PropTypes.oneOfType([PropTypes.object, PropTypes.func]),
```
shlomitc: as discussed - please see if changing the `scrollElement` can be done easily with a prop instead of passing a function

PR #2233, opened 2018-09-12
<MultiSelect/> - support reorderable tags

ChenRoth:
### What changed

Used the new `<SortableList/>` dnd to make `<MultiSelect/>` optionally reorderable

### Why it changed

It's a missing feature for `<MultiSelect/>` (3.8)

note the single test for the feature was put under ignore as there's no testkit for dnd yet.

---

- [ ] Change is tested
- [X] Change is documented
- [ ] Build is green

### DIFF
--- a/src/MultiSelect/InputWithTags.js
+++ b/src/MultiSelect/InputWithTags.js
@@ -7,6 +7,8 @@ import styles from './InputWithTags.scss';
 import omit from 'omit';
 import classNames from 'classnames';
 import isUndefined from 'lodash/isUndefined';
+import SortableList from '../SortableList/SortableList';
+import defaultDndStyles from '../../dnd-styles';
 
 class InputWithTags extends React.Component {
   constructor(props) {
@@ -14,6 +16,8 @@ class InputWithTags extends React.Component {
     this.focus = this.focus.bind(this);
     this.blur = this.blur.bind(this);
     this.select = this.select.bind(this);
+    this.handleHover = this.handleHover.bind(this);
+    this.renderReorderableTag = this.renderReorderableTag.bind(this);
 
     this.state = {inputValue: '', inputHasFocus: false, hasHover: false};
   }
@@ -52,6 +56,7 @@ class InputWithTags extends React.Component {
     const {
       tags,
       onRemoveTag,
+      onReorder,
       placeholder,
       error,
       errorMessage,
@@ -65,7 +70,7 @@ class InputWithTags extends React.Component {
     const isSelectMode = mode === 'select';
 
     const className = classNames({
-      [styles.tagsContainer]: true,
+      [styles.inputWithTagsContainer]: true,
       [styles.disabled]: disabled,
       [styles.error]: error,
       [styles.empty]: !tags.length,
@@ -99,18 +104,28 @@ class InputWithTags extends React.Component {
       rowMultiplier = 35;
     }
     const maxHeight = this.props.maxHeight || this.props.maxNumRows * rowMultiplier || 'initial';
-
     return (
       <div
         className={className}
         style={{maxHeight}}
         onClick={() => this.handleClick()}
-        onMouseOver={() => this.handleHover()}
-        onMouseOut={() => this.handleHover()}
+        onMouseOver={onReorder ? null : this.handleHover}
+        onMouseOut={onReorder ? null : this.handleHover}
         data-hook={this.props.dataHook}
         >
-        {tags.map(({label, ...rest}) => <Tag key={rest.id} disabled={disabled} onRemove={onRemoveTag} {...rest}>{label}</Tag>)}
-        <span className={classNames(styles.input, {[styles.emptyInput]: !tags.length})} data-hook="inner-input-with-tags">
+        {onReorder ?
+          <SortableList
+            contentClassName={styles.tagsContainer}
+            items={tags}
+            onDrop={onReorder}
+            renderItem={this.renderReorderableTag}
+            /> : tags.map(({label, ...rest}) =>
+              <Tag key={rest.id} disabled={disabled} onRemove={onRemoveTag} {...rest}>{label}</Tag>)
+        }
+        <span
+          className={classNames(styles.input, {[styles.emptyInput]: !tags.length})}
+          data-hook="inner-input-with-tags"
+          >
           <div className={styles.hiddenDiv} style={{fontSize}}>
             {this.state.inputValue}
           </div>
@@ -150,6 +165,30 @@ class InputWithTags extends React.Component {
     );
   }
 
+  renderReorderableTag({item: {id, label, ...itemProps}, previewStyles, isPlaceholder, isPreview, ...rest}) {
+    const {onRemoveTag, disabled} = this.props;
+    const classes = classNames({
+      [defaultDndStyles.itemPlaceholder]: isPlaceholder,
+      [styles.draggedTagPlaceholder]: isPlaceholder,
+      [defaultDndStyles.itemPreview]: isPreview,
+      [styles.draggedTag]: isPreview
+    });
+
+    return (
+      <div style={previewStyles}>
+        <Tag
+          id={id} disabled={disabled}
+          className={classes}
+          isPlaceholder={isPlaceholder}
+          onRemove={onRemoveTag} {...itemProps}
+          {...rest}
+          >
+          {label}
+        </Tag>
+      </div>
+    );
+  }
+
   focus() {
     this.input.focus();
   }
@@ -172,6 +211,7 @@ class InputWithTags extends React.Component {
 InputWithTags.propTypes = {
   onRemoveTag: PropTypes.func,
   tags: PropTypes.array,
+  onReorder: PropTypes.func,
   maxHeight: PropTypes.string,
   maxNumRows: PropTypes.number,
   onKeyDown: PropTypes.func,
@@ -190,7 +230,8 @@ InputWithTags.propTypes = {
 };
 
 InputWithTags.defaultProps = {
-  onRemoveTag: () => {},
+  onRemoveTag: () => {
+  },
   tags: [],
   placeholder: '',
   delimiters: []
--- a/src/MultiSelect/InputWithTags.js
+++ b/src/MultiSelect/InputWithTags.js
@@ -8,7 +8,7 @@ import omit from 'omit';
 import classNames from 'classnames';
 import isUndefined from 'lodash/isUndefined';
 import SortableList from '../SortableList/SortableList';
-import defaultDndStyles from '../../dnd-styles';
+import defaultDndStyles from '../dnd-styles';
 
 class InputWithTags extends React.Component {
   constructor(props) {
--- a/src/MultiSelect/InputWithTags.js
+++ b/src/MultiSelect/InputWithTags.js
@@ -179,7 +179,6 @@ class InputWithTags extends React.Component {
         <Tag
           id={id} disabled={disabled}
           className={classes}
-          isPlaceholder={isPlaceholder}
           onRemove={onRemoveTag} {...itemProps}
           {...rest}
           >
--- a/src/MultiSelect/InputWithTags.scss
+++ b/src/MultiSelect/InputWithTags.scss
@@ -44,7 +44,7 @@
   height: 0;
 }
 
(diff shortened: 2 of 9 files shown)


shlomitc:
didn't do a review for it yet, just played around with the storybook and it looks like the option is not moving correctly.
1. select two items.
2. try to move the first to the second
3. it is not changing the gray background to the right item

ChenRoth:
is that what it's supposed to do? is there a design for this?
all i see in zeplin is how the dragged item should behave

shlomitc:
The design should be the same as any D&D feature in our system (with the common styles).
So the colors look good to me (will need a UX review later on), but the functionality is not right,  it should flip the items when dragging

@milkyfruit - can you approve?

ChenRoth:
is there a link to a design that i can follow instead of doing this guesswork?

milkyfruit:
@ChenRoth This is the full description of how drag 'n' drop should work: https://zpl.io/bAEAWd6
@shlomitc is right, the styling is correct, but behaviour should follow this rule:

image

src/MultiSelect/InputWithTags.js
```diff
@@ -7,13 +7,16 @@ import styles from './InputWithTags.scss';
 import omit from 'omit';
 import classNames from 'classnames';
 import isUndefined from 'lodash/isUndefined';
+import SortableList from '../SortableList/SortableList';
+import defaultDndStyles from 'wix-style-react/dnd-styles';
```
Faradey27: you can not use such syntax inside of wix-style-react, we do not have such alias
instead you need to do like this:
```
import defaultDndStyles from '../../dnd-styles';
```

src/MultiSelect/InputWithTags.js
```diff
@@ -99,18 +103,28 @@ class InputWithTags extends React.Component {
...
     const maxHeight = this.props.maxHeight || this.props.maxNumRows * rowMultiplier || 'initial';
-
     return (
       <div
         className={className}
         style={{maxHeight}}
         onClick={() => this.handleClick()}
-        onMouseOver={() => this.handleHover()}
-        onMouseOut={() => this.handleHover()}
+        onMouseOver={onReorder ? null : () => this.handleHover()}
```
Faradey27: also, are you sure that, we do not need handleHover here? maybe we can do this in another way:
1) check that drag in is the progress(API is missed, but we can implement it)
2) if drag in progress - do not call handleHover(or do not do setState inside), otherwise, left everything as we have currently

and minor, but:
do we need arrow function here? maybe just `this.handleHover`

src/MultiSelect/InputWithTags.scss
```diff
@@ -44,7 +44,7 @@
   height: 0;
 }
 
-.tagsContainer {
```
Faradey27: why?

src/Tag/Tag.js
```diff
@@ -11,14 +11,16 @@ import Typography from '../Typography';
...
+    const {id, children, thumb, removable, onClick, onRemove, size, wrap, disabled, theme, maxWidth, draggable, className} = this.props;
 
-    const className = classNames({
+    const classes = classNames({
       [styles.tag]: true,
       [styles.large]: size === 'large',
       [styles.tagWrap]: wrap,
       [styles.disabled]: disabled,
-      [styles[`${theme}Theme`]]: true
+      [styles.draggable]: draggable,
```
Faradey27: not sure that Tag should know that it draggable, maybe let just pass to it some custom className with all required classes?

@shlomitc wdyt?

stories/MultiSelect/ExampleReorderable.js
```diff
@@ -0,0 +1,77 @@
...
+    const {tags} = this.state;
+    return (
+      <div>
+        <div className={styles.main}>
+          <MultiSelect
+            tags={this.state.tags}
+            onSelect={this.handleOnSelect}
+            onRemoveTag={this.handleOnRemoveTag}
+            onReorder={({addedIndex, removedIndex}) => {
+              const nextTags = copy(tags);
```
Faradey27: after some reasearch, i can say, that better to have here `nextTags = [...tags] or tags.slice()`

src/MultiSelect/InputWithTags.js
ChenRoth: i was following the current convention, i don't think this lambda is necessary either.
whenever i make changes in WSR, i make them as minimal as possible

src/MultiSelect/InputWithTags.scss
ChenRoth: because the DOM hierarchy has changed - tags are now contained within other divs

src/Tag/Tag.js
ChenRoth: removed `draggable`, shouldn't have been here

stories/MultiSelect/ExampleReorderable.js
ChenRoth: 👍

src/MultiSelect/InputWithTags.js
ChenRoth: 👍

src/MultiSelect/InputWithTags.js
```diff
@@ -99,18 +104,28 @@ class InputWithTags extends React.Component {
...
     const maxHeight = this.props.maxHeight || this.props.maxNumRows * rowMultiplier || 'initial';
-
     return (
       <div
         className={className}
         style={{maxHeight}}
         onClick={() => this.handleClick()}
-        onMouseOver={() => this.handleHover()}
-        onMouseOut={() => this.handleHover()}
+        onMouseOver={this.handleHover}
```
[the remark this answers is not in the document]
ChenRoth: no, changed it back to ternary condition.

Faradey27:
For me just 1 issue(with hover) left, which is blocked by this PR https://github.com/wix/wix-style-react/pull/2249

then we able to merge, @shlomitc any objections? new comments?

src/Tag/Tag.js
```diff
@@ -85,7 +86,16 @@ Tag.propTypes = {
   maxWidth: PropTypes.number,
 
   /** Whether to display ellipsis (...) for long content */
-  wrap: PropTypes.bool
+  wrap: PropTypes.bool,
+
+  /** Whether this tag is a placeholder for another  */
+  isPlaceholder: PropTypes.bool,
```
shlomitc: remove

src/Tag/Tag.js
```diff
@@ -85,7 +86,16 @@ Tag.propTypes = {
...
+
+  /** Whether this tag is being dragged  */
+  isPreview: PropTypes.bool,
```
shlomitc: remove

src/Tag/Tag.driver.js
```diff
@@ -10,6 +10,7 @@ const tagDriverFactory = ({element, wrapper, component}) => {
   const contentWithoutThumb = element.querySelector('span');
 
   return {
+    element: () => element,
```
shlomitc: where is this function used?

src/MultiSelect/InputWithTags.scss
```diff
@@ -86,3 +86,25 @@
 .hasHover {
   background: $B50;
 }
+
+.tagsContainer {
+  cursor: pointer;
+  display: flex;
```
shlomitc: @Faradey27 - is this the right way to make the list appear horizontally? if so - might want it in the common styles as well

src/MultiSelect/MultiSelect.spec.js
```diff
@@ -329,12 +344,43 @@ describe('MultiSelect', () => {
     expect(onSelect).toBeCalledWith([{id: options[0].id, label: options[0].value}]);
   });
 
+  // TODO: dnd testkit is missing - once it's available, this test has to be completed and run
+  xit('should allow reordering the tags', () => {
```
shlomitc: @Faradey27 - please update when we have the ability to test it

src/MultiSelect/MultiSelect.spec.js
```diff
@@ -329,12 +344,43 @@ describe('MultiSelect', () => {
...
+      <MultiSelect
+        draggable
+        options={options}
+        tags={tags}
+        onReorder={onReorder}
+        autoFocus
+        />
+    );
+    getTagDriverByTagId('Alabama').dragTo(getTagDriverByTagId('California3').element);
+    expect(onReorder).toBeCalled();
```
shlomitc: worth making sure it is called with the right parameters

src/MultiSelect/InputWithTags.js
```diff
@@ -150,6 +165,30 @@ class InputWithTags extends React.Component {
...
+    const {onRemoveTag, disabled} = this.props;
+    const classes = classNames({
+      [defaultDndStyles.itemPlaceholder]: isPlaceholder,
+      [styles.draggedTagPlaceholder]: isPlaceholder,
+      [defaultDndStyles.itemPreview]: isPreview,
+      [styles.draggedTag]: isPreview
+    });
+
+    return (
+      <div style={previewStyles}>
```
shlomitc: is this `div` necessary or it can be passed directly to the `Tag`?

src/MultiSelect/MultiSelect.spec.js
Faradey27: working on it

src/MultiSelect/InputWithTags.scss
Faradey27: yes it is right way, i was thinking about adding new prop to sortable list
`position: horizontal | vertical`

src/MultiSelect/InputWithTags.scss
```diff
@@ -86,3 +86,25 @@
...
+  flex-wrap: wrap;
+
+  .draggedTag {
+    background-color: transparentize($B30, 0.2);
+  }
+
+  .draggedTagPlaceholder {
+    background-color: $D60;
```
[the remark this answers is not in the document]
Faradey27: dnd-styles provide `background-color: $B40;` as default place holder color(according to general d&d guidelines)

but for tags, in zeplin you can find D60 color for placeholder

I think thats the reason why we have such override

@shlomitc @milkyfruit

src/MultiSelect/InputWithTags.scss
```diff
@@ -86,3 +86,25 @@
 .hasHover {
   background: $B50;
 }
+
+.tagsContainer {
+  cursor: pointer;
```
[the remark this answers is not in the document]
Faradey27: d&d styles provide cursor: -webkit-grabbing; for each draggable item

here, `cursor: pointer;` applied to whole container, so it should not be inside of d&d styles

src/MultiSelect/InputWithTags.scss
```diff
@@ -86,3 +86,25 @@
...
+
+  .draggedTagPlaceholder {
+    background-color: $D60;
+
+    &:hover {
+      background-color: $D60
+    }
+
+    * {
+      visibility: hidden;
```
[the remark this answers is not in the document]
ChenRoth: i want to maintain the space the content takes on screen without showing it.
this way, the placeholder is the same size as the actual content

src/MultiSelect/InputWithTags.js
ChenRoth: yes, dnd requires a `div`

ChenRoth:
@shlomitc pushed requested changes

src/MultiSelect/InputWithTags.js
```diff
@@ -150,6 +165,30 @@ class InputWithTags extends React.Component {
...
+        <Tag
+          id={id} disabled={disabled}
+          className={classes}
+          isPlaceholder={isPlaceholder}
```
Faradey27: do we need this?

ChenRoth: no, pushed

src/MultiSelect/InputWithTags.js
shlomitc: and what if we propagate the inline preview styles to the `Tag`? or it's still needed anyway

src/MultiSelect/InputWithTags.scss
shlomitc: what about just adding it as another `list` variation in the `dnd-styles`?

PR #2236, opened 2018-09-13
<Typography/> - convert  old Typography usage (from `.js` only)

hepiyellow:
- [x] EndorseContentLayout - refactor
- [x] FormField - refactor  (Add e2e test)
- [x] ThemedInput - use mixin
- [x] RadioButton - use `<Text/>` - updated disabled color
- [x] Loader - update `t5`->`h6`
- [x] ProgressBar - update `t5`->`h6` (Is this component deprecated?)
- [x] DataTable - [Refactor] Either add new Typography classes and use them (currently we have only mixins), or use `<Text/>` component, but that needs a performance check. Finally decided to use the mixins.
- [x] FieldWithSelection story
- [x] Tag

### DIFF
--- a/.storybook/stories.js
+++ b/.storybook/stories.js
@@ -114,7 +114,7 @@ import '../stories/Tag.story.js'; // 12.5 Tag
 import '../stories/Highlighter.story.js'; // Highlighter
 
 // Components API
-import '../stories/FormField.story.js';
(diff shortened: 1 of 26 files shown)


hepiyellow:
Replaced by #2268

PR #2242, opened 2018-09-16
<InputWithOptions/>- fix bug to allow re-selection of the same option

ChenRoth:
…option

### What changed
fix inability to re-select the same option after typing something
remove `selectedId` prop (it's an internal state)
### Why it changed
if you chose an option from the dropdown, typed something, and tried to re-select that option, it wouldn't let you.

also, the `selectedId` prop of `inputWithOption` relates to internal state and shouldn't be exposed to the user (there's no practical use case for selecting an option from the dropdown if the value in the input box is different)

---

- [ ] Change is tested
- [X] Change is documented
- [ ] Build is green


hepiyellow:
I suggest opening another PR with a simple fix as I proposed (calling onSelect when same option was selected).

In this PR, if you wish to follow it. Then it is needed to:
- remove selection from prop types
- Add `selectedId` to propTypes of `<Dropdown/>`, and fix it so that it does not pass `selectedId` to InputWithOptions

PR #2245, opened 2018-09-16
<InputWithTags/> - fix focus style

ChenRoth:
### What changed
removed background-color when component has focus

### Why it changed

so the component would be aligned with all other WSR input components.

---

- [ ] Change is tested
- [X] Change is documented
- [ ] Build is green


hepiyellow:
The fact that this does not break any eyes test, means that it is not tested.

ChenRoth:
either way, a test wouldn't fail now...

hepiyellow:
thanks.
BTW: this resolves https://github.com/wix-private/wsr-issues/issues/237

Faradey27:
@hepiyellow can we merge?

src/MultiSelect/MultiSelect.e2e.js
```diff
@@ -17,6 +17,7 @@ describe('MultiSelect', () => {
       .then(async () => {
         await browser.actions().sendKeys(protractor.Key.TAB).perform();
         await browser.actions().sendKeys(protractor.Key.TAB).perform();
+        await browser.actions().mouseMove(driver.element());
```
shlomitc: as this is now a different test - better have `eyes.checkWindow` before hovering, to make sure the previous focus styles are set.
do you think we should explain it also in the contribution guide?
https://github.com/wix/wix-style-react/blob/master/docs/contribution/TESTING.md#visual-testing

ChenRoth: @hepiyellow is this under your care now or do i need to do this?

PR #2251, opened 2018-09-18
<Collapse/> - add new component and revise <Card/> implementation

argshook:
replacing https://github.com/wix/wix-style-react/pull/2206

following discussion on #2206, the following changes were made

TL;DR;

1. new `<Collapse/>` component
1. deprecation log for `Card.CollapsedHeader`
1. update documentation accordingly

---

1. new `<Collapse/>` component

proxy `react-collapse` with practically no changes. The reason to proxy is so that we have ability
for more features and tweaks (think different animation, collapse direction, ability to change
underlying library etc)

2. deprecation log for `Card.CollapsedHeader`

* we prefer consumers to construct collapsed card themselves rather than rely on our opinionated
approach
* we also don't want everyone to get react-collapse in bundle even if they never need it (currently
just `import {Card} from 'wix-style-react/Card'` would also import react-collapse)

3. update documentation accordingly

* add `Collapse` component story page under `Components`
* remove `Card.CollapsedHeader` from docs
* update collapsed `Card` example


src/Card/CollapsedHeader/CollapsedHeader.js
```diff
@@ -30,6 +31,12 @@ class CollapsedHeader extends WixComponent {
     controlled: false
   };
 
+  componentDidMount() {
+    deprecationLog(
+      'Card.CollapsedHeader is deprecated, please use standalone <Collapse/> component instead.'
```
shlomitc: should use `<Collapse/>` + `<Card.Header/>`

stories/Collapse.story.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import Collapse from '../src/Collapse';
+
+import FormField from 'wix-style-react/FormField';
+import Input from 'wix-style-react/Input';
+
+const exampleChildren = [
+  {label: 'Simple text', value: 'Lorem perferendis sapiente quas facilis!'},
+  {
+    label: 'FormField with Input',
```
shlomitc: I don't think this example has any deep value of it.
worth adding a `controlled` example instead if possible.

argshook: i added this for fun just so we have more than one example.

regarding controlled example, is it really needed? not sure if it would add any value as `open` prop can already be played with

argshook:
@shlomitc could you review again? if ok, it's ready to merge

was thinking if testkit is needed for `<Collapse/>` but probably not. component renders or not its children so for consumer should be trivial to check whether children are rendered or not without a dedicated testkit. wdyt?

src/Card/CollapsedHeader/CollapsedHeader.driver.js
```diff
@@ -11,14 +11,18 @@ const CollapsedHeaderDriverFactory = ({element, wrapper, component}) => {
     subtitle: () => subtitle && subtitle.innerHTML,
     element: () => element,
     click: () => ReactTestUtils.Simulate.click(title),
-    findByDatahook: dataHook => element.querySelector(`[data-hook="${dataHook}"]`),
+    findByDatahook: dataHook =>
+      element.querySelector(`[data-hook="${dataHook}"]`),
     setProps: props => {
```
shlomitc: you can remove this

src/Card/CollapsedHeader/CollapsedHeader.driver.js
```diff
@@ -11,14 +11,18 @@ const CollapsedHeaderDriverFactory = ({element, wrapper, component}) => {
     subtitle: () => subtitle && subtitle.innerHTML,
     element: () => element,
     click: () => ReactTestUtils.Simulate.click(title),
-    findByDatahook: dataHook => element.querySelector(`[data-hook="${dataHook}"]`),
+    findByDatahook: dataHook =>
```
shlomitc: anyone uses it? looks to me redundant?

src/Card/CollapsedHeader/CollapsedHeader.js
```diff
@@ -19,7 +20,8 @@ class CollapsedHeader extends WixComponent {
     onCollapsedChange: func,
     buttonCollapseText: string,
     buttonExpandText: string,
-    controlled: bool
+    controlled: bool,
+    dataHook: string
```
shlomitc: don't need to add it if extending `WixComponent`

src/Card/CollapsedHeader/CollapsedHeader.js
```diff
@@ -86,36 +98,41 @@ class CollapsedHeader extends WixComponent {
...
-      toggleStyle
+      toggleStyle,
+      dataHook
     } = this.props;
 
     const {isCollapsed} = this.state;
 
     return (
-      <div>
+      <div data-hook={dataHook}>
```
shlomitc: don't need to add this as it extends `WixComponent`

src/Card/CollapsedHeader/CollapsedHeader.spec.js
```diff
@@ -1,11 +1,14 @@
 import React from 'react';
-import ReactTestUtils from 'react-dom/test-utils';
 import collapsedHeaderDriverFactory from './CollapsedHeader.driver';
 import {createDriverFactory} from 'wix-ui-test-utils/driver-factory';
 import CollapsedHeader from './CollapsedHeader';
 import {collapsedHeaderTestkitFactory} from '../../../testkit';
 import {collapsedHeaderTestkitFactory as enzymeCollapsedHeaderTestkitFactory} from '../../../testkit/enzyme';
 import {mount} from 'enzyme';
+import {
```
shlomitc: please move the sanity tests to `testkit/enzyme.spec.js` and `testkit/vanilla.spec.js`

src/Card/CollapsedHeader/CollapsedHeader.spec.js
```diff
@@ -53,7 +58,10 @@ describe('CollapsedHeader', () => {
   it('should call with collapse status', () => {
     const onCollapsedChange = jest.fn();
     const driver = createDriver(
-      <CollapsedHeader title="Header Title" onCollapsedChange={onCollapsedChange}>
+      <CollapsedHeader
```
shlomitc: We get warnings in the console because the components throws deprecation warnings. can you please fix it? maybe take a look how other components are using the deprecation messages

src/Card/README.md
```diff
@@ -4,7 +4,7 @@ A compound component for any content to be displayed in styled block.
 
 Most often used with `<Card.Header/>`, `<Card.Content/>` and `<Card.Divider/>`.
 
-Use [`<Row/>` and `<Col/>`](https://wix-wix-style-react.surge.sh/?selectedKind=Common&selectedStory=Grid&full=0&addons=0&stories=1&panelRight=0) to lay out content inside `<Card.Content/>`
+Use [`<Row/>` and `<Col/>`](https://wix-wix-style-react.surge.sh/?selectedKind=2.%20Layout&selectedStory=Grid&full=0&addons=0&stories=1&panelRight=0) to lay out content inside `<Card.Content/>`
```
shlomitc: The best thing will be adding these parts to be part of the story and not part of the readme file.
That way you can use the `Link` component of storybook instead of hard-coded links to production

src/Collapse/Collapse.spec.js
```diff
@@ -0,0 +1,27 @@
...
+      expect(wrapper.children().text()).toEqual(null);
+    });
+  });
+
+  describe('`dataHook` prop', () => {
+    it('should be passed to children', () => {
+      const wrapper = mount(
+        <Collapse dataHook="I'm hooked!" children={<div>text</div>}/>
+      );
+      expect(wrapper.children().prop('data-hook')).toEqual('I\'m hooked!');
```
shlomitc: I think it's problematic as it will override the data-hook of the inner component...
maybe not add it at all?

src/Card/CollapsedHeader/CollapsedHeader.driver.js
argshook: these changes are from prettier, i don't want to touch those files, because yes, maybe someone is using it, who knows. 
We're deprecating collapsedHeader anyway, not sure i need to fix driver of it

src/Card/README.md
argshook: while that's true for now these can be hardcoded imo.

the way to solve this i think is to make autodocs support mdx, which i
can gladly do in the upcoming days

src/Collapse/Collapse.spec.js
argshook: should be ok, added test to ensure it

stories/Collapse.story.js
argshook: also, Card storypage already has controlled example of `<Card/>` with `<Collapsed/>`

src/Card/README.md
shlomitc: might take a while until storybook will support it
https://github.com/storybooks/storybook/issues/4341

shlomitc:
I see a lot of console warnings in our tests
```
Style-React: [WARNING] Card.CollapsedHeader is deprecated. Instead please use <Card><Collapse><Card.Content/></Collapse></Card>
```

`CollapsedHeader.spec.js`, `StatsWidget.spec.js`, etc...

argshook:
@shlomitc i added `deprecationLog` to root of file, as is done in other components. Should i do something other than that?

shlomitc:
@argshook 
1. components that are using the deprecated components should be fixed, meaning stop using the deprecated components.
2. tests should either be skipped or at least override the `console.error` so it won't throw these warnings in the console

argshook:
@shlomitc

> 1. components that are using the deprecated components should be fixed, meaning stop using the deprecated components.

no component is using `Card.CollapsedHeader` in our codebase. The
problem is that `src/Card/Card.js` has `Card.CollapsedHeader = CollapsedHeader`.

this is because `CollapsedHeader` is available on `Card`, something we
want to remove but it is a breaking change.

thus, deprecationLog is fired even only `Card` is imported.

perhaps deprecationLog should happen in `componentDidMount` as it was before?

---

> 2. tests should either be skipped or at least override the console.error so it won't throw these warnings in the console

are you sure? what's the point of deprecation log if we ignore it
ourselves? i would let it annoy us and push us into fixing it. there are
many other deprecationLog messages from other components, not only
Collapsed.Header. should they all be ignored too?

shlomitc:
1. ok
2. so I suggest putting the deprecation only when component is actually used and not on import

PR #2252, opened 2018-09-19
Fix wrong position of drag item with connected handle

Faradey27:
### What changed

...

### Why it changed

...

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


Faradey27:
@shlomitc if it not clear, let do i call, i ll explain

PR #2254, opened 2018-09-20
<SortableList/> - disable drag for specific list items

Wo1v3r:
### What changed
Added disabled prop to draggable item
...

### Why it changed
Needed in order to have static elements in a DnD list for Price Quotes (Table header) or Trash like components
...

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green

### DIFF
--- a/src/DragAndDrop/Draggable/Draggable.js
+++ b/src/DragAndDrop/Draggable/Draggable.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import identity from 'lodash/identity';
 
 import WixComponent from '../../BaseComponents/WixComponent';
 import DraggableSource from './components/DraggableSource';
@@ -7,9 +8,19 @@ import DraggableTarget from './components/DraggableTarget';
 
 export class Draggable extends WixComponent {
   render() {
-    return (
+    const { item, id } = this.props;
+    return this.props.disabled ? (
+      this.props.renderItem({
+        ...this.props,
+        isPlaceHolder: false,
+        isPreview: false,
+        connectHandle: identity,
+        item,
+        id
+      })
+    ) : (
       <DraggableTarget {...this.props}>
-        <DraggableSource {...this.props}/>
+        <DraggableSource {...this.props} />
       </DraggableTarget>
     );
   }
@@ -39,7 +50,9 @@ Draggable.propTypes = {
   /** callback for drag start */
   onDragStart: PropTypes.func,
   /** callback for drag end */
-  onDragEnd: PropTypes.func
+  onDragEnd: PropTypes.func,
+  /** Disabled Drag and Drop Item */
+  disabled: PropTypes.bool
 };
 
 export default Draggable;
--- a/src/DragAndDrop/Draggable/Draggable.js
+++ b/src/DragAndDrop/Draggable/Draggable.js
@@ -8,7 +8,7 @@ import DraggableTarget from './components/DraggableTarget';
 
 export class Draggable extends WixComponent {
   render() {
-    const { item, id } = this.props;
+    const {item, id} = this.props;
     return this.props.disabled ? (
       this.props.renderItem({
         ...this.props,
@@ -20,7 +20,7 @@ export class Draggable extends WixComponent {
       })
     ) : (
       <DraggableTarget {...this.props}>
-        <DraggableSource {...this.props} />
+        <DraggableSource {...this.props}/>
       </DraggableTarget>
     );
   }
--- a/src/SortableList/SortableList.js
+++ b/src/SortableList/SortableList.js
@@ -111,6 +111,7 @@ export default class SortableList extends WixComponent {
               <Draggable
                 {...common}
                 key={`${item.id}-${index}-${this.props.containerId}`}
+                disabled={item.disabled}
                 id={item.id}
                 index={index}
                 item={item}
@@ -142,7 +143,7 @@ SortableList.propTypes = {
     inside of renderItem callback
   */
   dragPreview: PropTypes.bool,
-  /** list of items with {id: any} */
+  /** list of items with {id: any, disabled:bool} */
   items: PropTypes.array,
   /** callback for drag start */
   onDragStart: PropTypes.func,
--- a/stories/DragAndDrop/Introduction/IntroductionExample.js
+++ b/stories/DragAndDrop/Introduction/IntroductionExample.js
@@ -9,24 +9,31 @@ import styles from './IntroductionExample.scss';
 export default class IntroductionExample extends React.Component {
   constructor() {
     super();
-    this.state = {items: [
-      {
-        id: 'a',
-        text: 'Item 1'
-      },
-      {
-        id: 'b',
-        text: 'Item 2'
-      },
-      {
-        id: 'c',
-        text: 'Item 3'
-      },
-      {
-        id: 'd',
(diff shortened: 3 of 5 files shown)


Wo1v3r:
@Faradey27  Andrii pls have a look, maybe you want to move it to another file e.g `DraggableDisabled` seperately ?

src/DragAndDrop/Draggable/Draggable.js
```diff
@@ -7,9 +7,16 @@ import DraggableTarget from './components/DraggableTarget';
 
 export class Draggable extends WixComponent {
   render() {
-    return (
+    return this.props.disabled ? (
+      this.props.renderItem({
+        isPlaceHolder: false,
+        isPreview: false,
+        connectHandle: _ => _,
```
shlomitc: use `lodash.noop` here

stories/DragAndDrop/SortableList/SingleAreaList.js
```diff
@@ -26,6 +26,11 @@ export default class SingleAreaList extends React.Component {
       {
         id: 'd',
         text: 'Item 4'
+      },
+      {
+        id: 'e',
+        text: 'Disabled Item 5',
+        disabled: true
```
shlomitc: how will this affect the CSS of the non-dragged component?
for example - if there is a pointer cursor before drag - it should now not be set to the initial one.
I don't see this is propagated to the renderItem function, so I wonder how it affects it.

Wo1v3r: was thinking of adding another `isDisabled` but it is redundant as all of the above are passed via props therefore `disabled` would already be passed to the `renderItem` function, I can add relevant style change in the example. not sure how that will be clear in the API though (that all props pass through)

shlomitc: This looks to me like one of those values we should pass in a straightforward way like we pass the `id` - will also be easier to document it and understand the flow.

yes, style changes are needed. I think the current common styles lack some edge case, this one for example.

src/DragAndDrop/Draggable/Draggable.js
shlomitc: We use lodash.noop anyway in many other places so it’s not a new dependency. It also makes sure we do not create new functions accidentally and easier to explain required changes in PRs. So lodash.noop is exactly like the functions you wrote, I see no reason why not to use it

argshook: it is used here in Draggable and also in Tooltip, it's just two, not many :)

this doesn't matter much but i wouldn't use 3rd party helpers for things as trivial as this

Wo1v3r: Um ok originally I used identity there in connectHandle to return the handle unconnected, but I see it's noop now (e.g `() => undefined`).

I'll update in the PR with lodash.noop as it's already being used and will be easier to find and replace if @argshook  wishes (I agree it better be under utils than 3rd party [for one liners])

stories/DragAndDrop/SortableList/SingleAreaList.js
Wo1v3r: Cool np will update with example and with something similar to id;  <{...restProps} id={id} disabled={disabled} />

Wo1v3r:
@shlomitc  @Faradey27  @argshook  I updated the PR;
- used `lodash/identity` to also keep the handle in place etc, if user of Draggable will want otherwise, he can avoid rendering using the disabled prop, IMO the default should be to keep the component intact mostly
- passing `item` and `id` explicitly to `renderItem` to be more clear
- added styles to the example (colors are WSR but not sure they match styleguide)

src/DragAndDrop/Draggable/Draggable.js
```diff
@@ -1,15 +1,26 @@
...
 
 import WixComponent from '../../BaseComponents/WixComponent';
 import DraggableSource from './components/DraggableSource';
 import DraggableTarget from './components/DraggableTarget';
 
 export class Draggable extends WixComponent {
   render() {
-    return (
+    const { item, id } = this.props;
+    return this.props.disabled ? (
```
Faradey27: Draggable should not know internal api of DragSource, better to keep such logic inside of a DragSource

src/DragAndDrop/Draggable/Draggable.js
```diff
@@ -39,7 +50,9 @@ Draggable.propTypes = {
   /** callback for drag start */
   onDragStart: PropTypes.func,
   /** callback for drag end */
-  onDragEnd: PropTypes.func
+  onDragEnd: PropTypes.func,
+  /** Disabled Drag and Drop Item */
+  disabled: PropTypes.bool
```
Faradey27: maybe draggable? or fixed?

stories/DragAndDrop/SortableList/SingleAreaList.js
```diff
@@ -10,48 +10,61 @@ import styles from './SingleAreaList.scss';
...
+          id: 'e',
+          text: 'Disabled Item 5',
+          disabled: true
+        }
+      ]
+    };
   }
 
-  handleDrop = ({removedIndex, addedIndex}) => {
+  handleDrop = ({ removedIndex, addedIndex }) => {
```
Faradey27: wsr eslint will ask to remove spaces

Faradey27:
@Wo1v3r ping, let finish PR) the feature may be useful for a lot of people

Wo1v3r:
@Faradey27 ok pushing in a moment

shlomitc:
1. @Faradey27 - are you satisfied by this solution?
2. @Wo1v3r build is failing for lint errors and it is conflicting now with master, can you please fix it?

Wo1v3r:
@shlomitc  hopefully will get to it this thursday

Faradey27:
@shlomitc I think it enough to merge, but still I think <Draggable/> should not know what is renderItem

shlomitc:
I updated the branch (pushed-force), so notice that.
I also moved the example to the SortableList section.

Look however that the solution is a bit problematic.

for example, try to drag item 2 below item 3.
This will not trigger the drop area below it until it goes over item 4.

I think this is not the best user experience. @Faradey27 any other suggestions?

Faradey27:
> I updated the branch (pushed-force), so notice that.
> I also moved the example to the SortableList section.
> 
> Look however that the solution is a bit problematic.
> 
> for example, try to drag item 2 below item 3.
> This will not trigger the drop area below it until it goes over item 4.
> 
> I think this is not the best user experience. @Faradey27 any other suggestions?

sounds bad, so, I think when we say "disabled" we mean that we still want to drop on the item, but do not want to allow to drag it.

So, maybe let just prevent DragSource wrap? like inside of DragSource we will just call and return renderItem without using d&d HOCs(currenty in PR we are doing this inside of the Draggable). Also, maybe then let rename prop `disabled` to `draggable` or smth like this

Wo1v3r:
@Faradey27  I'm thinking of retracting this PR for now. the need came from a table use case, but with all the behaviour divs attached along the way, it makes it a tad problematic to use in a table.

I'm trying to implement on my own so we can discuss it (regarding the actual use case)

--

In general, I agree that it should be split into two different components (DragSource, DropTarget) and export as such (e.g no "Draggable" only)
then each of the HOCs will have its' own `disabled` prop with matching behaviour.
WDYT?

PR #2256, opened 2018-09-20
fix: fixed `<SideMenu/>` accepting dataHook prop

AtomicAlienZ:
### What changed

`<SideMenu/>` component now accepts `dataHook` prop and renders a div with it.

### Why it changed

Fixes the following scenario:
- Upgrade to wsr5
- See failing tests that test `DrillView` menu
- Dig in, discover that driver does not find the element
- Dig in deeper into `wix-ui-test-utils` package, see a selector by `data-hook` attribute
- Return to wsr sources, look into `<DrillView/>` render method
- Dig deeper into `<SideMenu/>` see hardcoded `data-hook` attribute

All exposed drivers for enzyme accept `dataHook` as props, so naturally it's expected to pass an actual data-hook to them.

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


shlomitc:
That's great, thanks a lot.
I think you can just remove the default `side-menu` dataHook. It doesn't make sense to have a default prop for this kind of attributes

After removing, you should add it to the right storybook example so the protractor test could locate it.
This is the test using it (no need to touch it unless you change the data-hook name):
https://github.com/wix/wix-style-react/blob/92920bddd6b2464906dfa25f93f04cfca51a71de/src/SideMenu/DrillView/DrillView.e2e.js#L9

And this is the place to add the data-hook:
https://github.com/wix/wix-style-react/blob/8dbe609baa68ddad485a1be76d30cbb2a2dfc805/stories/SideMenu/ExampleSideMenuDrill.js#L178

PR #2260, opened 2018-09-21
fix(Tooltip): remove event.stopPropagation

argshook:
there are cases when event bubbling is expected but does not happen
because of this line

https://github.com/wix/wix-style-react/commit/4d34669c7#r30580075
https://github.com/wix/wix-style-react/pull/2031#discussion_r219475350

### DIFF
--- a/src/Tooltip/Tooltip.js
+++ b/src/Tooltip/Tooltip.js
@@ -396,9 +396,8 @@ class Tooltip extends WixComponent {
     this._hideOrShow('focus');
   }
 
-  _onClick(event) {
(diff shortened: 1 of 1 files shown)


shlomitc:
I don’t know why it was added but I guess we can try to remove it - did you try to see how it affects the TableActionCell?

argshook:
i clicked around in storybook examples and didn't see any failures. it seems ok but i might not know some expected behavior or edge case

shlomitc:
If no test is failing - fee free to merge

PR #2261, opened 2018-09-23
<Table/> - fix state initialization warning

rghorbani:
### What changed

Fixed Table state initialization warning.

screen shot 2018-09-24 at 12 12 48 am

### Why it changed

Since React 16.5.0 initializing state with props directly generates a warning. This fix resolves this warning.

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green


rghorbani:
@Faradey27 I'm using React 16.5.0 and it does not produce an error for `componentWillReceiveProps` yet. It seems that for now the react team is only concern for the initialization of the component.

shlomitc:
@hepiyellow - why do we copy the props directly to the state instead of using them from props?

hepiyellow:
@shlomitc 
The props are assigned to the state since the whole `props` are use as the value of a React Context (React 16 context).
The context provider needs to be re-rendered whenever the `props` change.

>There is a comment about it [here](https://github.com/wix/wix-style-react/blob/268bcbde691e6d9542352c4b03787029d72d32f5/src/Table/Table.js#L96)

Seems though, that you are right, and there is no need for `state` we can use the `this.props` directly, since `shouldComponentUpdate` returns always `true`, if `props` change, then there is a re-render and the context provider's value is update. There is a test for it, and it passes when removing `state.

See https://github.com/wix/wix-style-react/pull/2296

PR #2264, opened 2018-09-25
WSR-ISSUES-237: <Tags/> - input box should not have hover color after being focused

Faradey27:
… 

### What changed

...

### Why it changed

...

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


Faradey27:
@lbelinsk good question, i think main problem that we have a lot of overrides(.hasFocus on parents of inputs)

@shlomitc wdyt?

lbelinsk:
I actually think we shouldn't have merged this because probably this should be implemented in the Input's stylesheet.

hepiyellow:
This was already fixed in #2245

PR #2265, opened 2018-09-25
<InputWithOptions/> - fix `onSelect` to be called also when re-selecting same option

ChenRoth:
### What changed

`onSelect` is triggered in <InputWithOptions/> even when option is supposedly already selected

### Why it changed

there's a bug where you can't reselect an option, even if you typed something in the input box

---

- [X] Change is tested
- [X] Change is documented
- [ ] Build is green


hepiyellow:
wait. This causes a double `onSelcet` call in `MultiSelectCheckbox`

ChenRoth:
what now then? we need this fix for a feature

hepiyellow:
Actually, I'll fix MultiSelectCheckbox in a separate PR

PR #2267, opened 2018-09-25
`<MultiSelectCheckbox/> - Fix  `onSelect` called twice

### LINKED ISSUE #2187: MultiSelectCheckbox's onSelect seems to be called twice
### Summary:

When creating this simple component, onSelect seems to be called twice for MultiSelectCheckbox's onSelect event:

```
<MultiSelectCheckbox
              onSelect={function(item) {
                console.log('WHATTT is this ', item);
                console.log('***********************');
              }}
              options={[
                {
                  id: 'Trending',
                  value: 'Trending',
                },
                {
                  id: 'New Features',
                  value: 'New Features',
                },
                {
                  id: 'Web Design & Wix Editor',
                  value: 'Web Design & Wix Editor',
                },
                {
                  id: 'Web Development & Wix Code',
                  value: 'Web Development & Wix Code',
                },
                {
                  id: 'Marketing & Business Tools',
                  value: 'Marketing & Business Tools',
                },
                {
                  id: 'Professional Services & Wix Products',
                  value: 'Professional Services & Wix Products',
                },
              ]}
            />
```

output when click on the first checkbox:

```
WHATTT is this  {id: "Trending", value: {…}}
***********************
WHATTT is this  Trending
***********************
```
### Priority:
- [ ] Blocker
- [X] Urgent
- [ ] Normal
- [ ] Minor

### Steps to reproduce:
 1. Create the component like above
 2. Check a checkbox :)

### Expected behavior:
Expected the "***********************" to be printed only once, because onSelect is supposed to be called once. It is printed twice - once from the "super".
### Relevant Components:
MultiSelectCheckbox
### Additional notes:
We believe it has to do with calling super._onSelect() on [this line](https://github.com/wix/wix-style-react/blob/master/src/MultiSelectCheckbox/MultiSelectCheckbox.js#L55)

hepiyellow:
Fixes #2187 
`onSelect` was called twice.
Once from `InputWithOptions`, and it was called with the option object as argument.
And once from `MultiSelectCheckbox`, and it was called only with `option.id`.

The fix remove the first call.


hepiyellow:
Need to add the original `option` object as second argument,
to conform to other such component we have (e.g Dropdown).

But need to somehow strip the added checkbox from the `option.value`.

lbelinsk:
@hepiyellow can you please sync with this pr, as it tries to solve the same issue
https://github.com/wix/wix-style-react/pull/2270

hepiyellow:
@lbelinsk 
It is not the same issue.
This PR solves `onSelect` from being called twice regardless if the same option was selected.

src/MultiSelectCheckbox/MultiSelectCheckbox.js
```diff
@@ -51,12 +51,23 @@ class MultiSelectCheckbox extends InputWithOptions {
     return styles.readonly;
   }
 
-  _onSelect(option) {
-    super._onSelect(option);
+  _onSelect(option, isSelectedOption) {
+    this.showOptions();
+
+    if (this.closeOnSelect()) {
```
Faradey27: || isSelectedOption

no need in 2 setState

src/MultiSelectCheckbox/MultiSelectCheckbox.js
```diff
@@ -51,12 +51,23 @@ class MultiSelectCheckbox extends InputWithOptions {
...
+      this.setState({showOptions: false});
+    }
+
+    if (isSelectedOption) {
+      this.setState({showOptions: false});
+    }
+
+    //The option object is not the original, since it was injected a checkbox
+    const originalOption = this.props.options.find(op => option.id === op.id);
     if (this.isSelected(option)) {
```
Faradey27: option or originalOption?

src/MultiSelectCheckbox/MultiSelectCheckbox.js
```diff
@@ -51,12 +51,23 @@ class MultiSelectCheckbox extends InputWithOptions {
...
-      this.props.onDeselect && this.props.onDeselect(option.id);
+      this.props.onDeselect && this.props.onDeselect(option.id, originalOption);
     } else {
-      this.props.onSelect && this.props.onSelect(option.id);
+      this.props.onSelect && this.props.onSelect(option.id, originalOption);
```
Faradey27: maybe originalOption.id for consistency?

src/MultiSelectCheckbox/MultiSelectCheckbox.js
hepiyellow: Thanks for that. But actually looking into it again, there is a bug here.
The old call to `super._onSelect(option)` did not pass the `isSelectedOption`, so this code was never run:
```js
if (isSelectedOption) {
      this.setState({showOptions: false});
}
```
So I shouldn't be introducing this behaviour now.
(-:

src/MultiSelectCheckbox/MultiSelectCheckbox.js
hepiyellow: It doesn't matter, both `option` and `originalOption` have the same id

src/MultiSelectCheckbox/MultiSelectCheckbox.js
hepiyellow: The whole idea here was to expose the whole option object and not only the id. (In other similar components like `<MultiSelect/>` which have an `onSelect` we pass the option object and not `option.id`)

PR #2270, opened 2018-09-26
ISSUE-2189: when trying to select the selected item, nothing happens

Faradey27:
### What changed

...

### Why it changed

...

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green

### DIFF
--- a/src/InputWithOptions/InputWithOptions.spec.js
+++ b/src/InputWithOptions/InputWithOptions.spec.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import ReactTestUtils from 'react-dom/test-utils';
 import {mount} from 'enzyme';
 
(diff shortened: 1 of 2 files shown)


hepiyellow:
I already fixed this.
In a different way.
Let me take a look at your solution.
https://github.com/wix/wix-style-react/pull/2265

hepiyellow:
@Faradey27 
You can merge your solution , it is better, since it doesn't change the API behavior.
In my solution I suggested that `onSelect` would be called also if the same option is selected.

hepiyellow:
@Faradey27 
BTW: Chen's solution was a bit towards your solution which "ties" the inputValue to the selectedOption. But he did it by eliminating the `selectedId` prop from `InputWithOptions`, and having the `DropdownLayout`'s `selectedId` begin controlled by `InputWithOption` according to the current `inputValue`.
I think that is a better solution , but it is a breaking change.

src/InputWithOptions/InputWithOptions.spec.js
```diff
@@ -430,6 +431,108 @@ const runInputWithOptionsTest = driverFactory => {
...
+        onSelect.mockClear();
+
+        inputDriver.focus();
+        inputDriver.enterText('First');
+
+        expect(dropdownLayoutDriver.isShown()).toBeTruthy();
+
+        driver.pressDownKey();
+        driver.pressTabKey();
+        expect(onSelect).toBeCalledWith(simpleOptions[0]);
```
hepiyellow: It's better to always also assert that the callback was called only once. (It seems we have a lot of double calls bug lately)

PR #2276, opened 2018-09-27
Docs: Add Dropdown Snippet story using FormField

hepiyellow:
According to https://github.com/wix-private/wsr-issues/issues/242

## What changed
- Moved Dropdown story under `Components` category.
- Add Dropdown snippet story at `4.1 Dropdown`
- Move `4.1 + DropdownComposite` story under new `Deprecated` category

## What next
- I guess we would want to make a `SnippetStoryTemplate` component to help us create snippet stories which look the same.
- Do we want to Remove `DropdownComposite`? (breaking change for next major)


.storybook/stories.js
```diff
@@ -128,6 +128,9 @@ import '../stories/TPA/TextLink';
 import '../stories/TPA/Input';
 import '../stories/TPA/Label';
 
+// Deprecated
+import '../stories/deprecated/DropdownComposite'; // 4.1 + DropdownComposite
```
argshook: duplicated path

stories/deprecated/DropdownComposite/index.js
```diff
@@ -3,10 +3,10 @@ import {storiesOf} from '@storybook/react';
 import InteractiveCodeExample from 'wix-storybook-utils/InteractiveCodeExample';
 import Markdown from 'wix-storybook-utils/Markdown';
 
-import Readme from '../../src/DropdownComposite/README.md';
+import Readme from '../../../src/DropdownComposite/README.md';
 import ExampleStandard from './ExampleStandard';
 
-storiesOf('4. Selection', module)
+storiesOf('Deprecated', module)
```
shlomitc: You can just remove this story instead of deprecating

Also - add a deprecation log for the `DropdownComposite` component asking to move to the `FormField` component instead

stories/DropdownSnippet/Example.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
```
shlomitc: I don't like the name `Snippet`.
What we're actually doing is a separation between how we index the library and which components we use.

So I suggest at first to create a `stories/components` folder, in which the auto example stories will be.

Then, the currently called snippet will just be the semantic name in the indexing system.

.storybook/stories.js
shlomitc: I think it's just a bug in the chrome extension we're using

shlomitc:
@wuwa - this is the first example for the separation between the components and the design system indexing.
Please suggest here how you think the `4.1 Dropdown` documentation should look like

stories/DropdownSnippet/Example.js
hepiyellow: We can do that.
Just bare in mind that for some components there is no difference between the UX indexing story and the component story. Like `<ToggleSwitch/>`.
So where should the `<ToggleSwitch/>` story be? should we have a UX story with almost nothing in it but a link to the component?

shlomitc: The `ToggleSwitch` for example has also a variation with a label (AKA FormField).

I think the auto-docs stories should sit in the components folder while the indexing should reflect the visuals of the Zeplin spec file

stories/deprecated/DropdownComposite/index.js
hepiyellow: done

stories/DropdownSnippet/Example.js
hepiyellow: ok

hepiyellow:
@wuwa @milkyfruit 
See this story:
https://wix-wix-style-react-pr-2276.surge.sh/?selectedKind=4.%20Selection&selectedStory=4.1%20Dropdown&full=0&addons=0&stories=1&panelRight=0

(Ignore the "(Snippet)" suffix)

hepiyellow:
@wuwa @shlomitc @milkyfruit 
Can I merge this?

hepiyellow:
@shlomitc @wuwa @milkyfruit @LittleTal 
## Hello UX team, where are you? 
>(Domas, you were on vacation, I know)

wuwa:
sorry, we were in a conference and moving offices also.
yes you can merge, but final indexing needs to be thought of - some hierarchy in the components section as well (i don't know if this change is something minor)

PR #2280, opened 2018-09-27
`<InputWithOptions/>` - Do not open options when focused

### LINKED ISSUE #2168: <InputWithOptions> cannot focus input without showing the options
### Priority:
- [ ] Blocker
- [ ] Urgent
- [ ] Normal
- [ x] Minor

### Steps to reproduce:
use inputWithOptions autoFocus prop or call focus() on ref 
focus triggers the display of the options list (which is arguably correct) but hideOptions() does not hide them so there is no way to focus without showing the options

hepiyellow:
Resolves #2168

### DIFF
--- a/src/InputWithOptions/InputWithOptions.js
+++ b/src/InputWithOptions/InputWithOptions.js
@@ -205,8 +205,12 @@ class InputWithOptions extends WixComponent {
   }
 
   _onInputClicked(event) {
-    if (this.state.showOptions && (Date.now() - this.state.lastOptionsShow > 2000)) {
-      this.hideOptions();
+    if (this.state.showOptions) {
+      if ((Date.now() - this.state.lastOptionsShow > 2000)) {
+        this.hideOptions();
+      }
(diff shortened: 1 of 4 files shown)


hepiyellow:
@shlomitc Sorry, this was supposed to be WIP

hepiyellow:
@shlomitc MultiSelect in read-only ('select') mode is not a special case.
I didn't yet implement to showOption on click.
Now that I've implemented it. It affects all inputs which are either read-only or not.
As @wuwa mentioned. Clicking should always open the options.
We removed the showOptions on focus only.

hepiyellow:
@shlomitc 
We had the concern that tests of consumers might fail now that focusing the InputWithOptions does NOT open the options. But looking at a sample of consumer code, I could not see cases where `focus` was used to open the options.
So I think we'll go ahead and release this under  `Changes`, and if a few tests will fail, they would use `driver.click()` to open the options.

PR #2286, opened 2018-10-01
<DatePicker/> Move z-index prop to the correct element in css

xmityaz:
### What changed

Minor styling fix for DatePicker component

### Why it changed

screen shot 2018-10-01 at 4 19 41 pm

`.calendarRoot` class exists in component JSX, but removed from styles.
Seems it got lost accidentally in this PR:
https://github.com/wix/wix-style-react/commit/8efe9fae798db43d3d32d8be4b9a18f53060536d#diff-9eea5f00ca05f3d2e814a03694c881e1L7

### DIFF
--- a/src/DatePicker/DatePicker.scss
+++ b/src/DatePicker/DatePicker.scss
@@ -4,8 +4,11 @@
   display: inline-block;
 }
 
-.date-picker-calendar {
+.calendar-root {
   z-index: 1;
+}
+
+.date-picker-calendar {
   :global .DayPicker {
     @include popoverTheme;
   }


Faradey27:
@shlomitc what do you think about zIndex: 1 as part of calendar.scss? to prevent such issue for all cases

shlomitc:
@Faradey27 - this is usually a popover issue (things getting on top of others), so I think it's a concern of the mechanism and not of the `Calendar` component

xmityaz:
@Faradey27 @shlomitc 
Guys, it quite simple change which fix critical bug for our project. And probably this issue affects other apps which are using datepicker.

Can we merge it, or fix it in other way?

PR #2293, opened 2018-10-03
<SideMenu/> - fix infinite loop when using React 16

banuni:
### What changed
Removed the `setState` inside `shouldComponentUpdate`
...

### Why it changed
Because it created an infinite loop (breaks React16 specifically)
...

---

- [v ] Change is tested
- [ ] Change is documented
- [ ] Build is green

### DIFF
--- a/src/SideMenu/DrillView/DrillView.js
+++ b/src/SideMenu/DrillView/DrillView.js
@@ -213,7 +213,7 @@ class SideMenuDrill extends React.Component {
     this.isAnimating = false;
     if (this.queuedUpdate) {
       this.queuedUpdate = false;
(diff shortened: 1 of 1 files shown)


hepiyellow:
FYI: You are welcome to open PRs as branches in wix-style-react.

hepiyellow:
Did you really reproduce an infinite loop?
oh, and could you add a test for this?

shlomitc:
@banuni - did you see this also in the React 15 usage or only in React 16?

banuni:
Only on React16, that's since the lifecycle is slightly different. I've
discussed this with Erez face to face on Wednesday.

PR #2294, opened 2018-10-03
<InputWithTags/> replace hasHover state with native css

ChenRoth:
### What changed

I've replaced the `hasHover` state with a native `:hover` css

### Why it changed

the `onMouseX` events conflict with the new d&d behaviour, thereby creating a disparity between non draggable and draggable versions of the component.

Also, the state seems redundant as native css can handle it, with one exception - the new code would make the component highlight _also when there are tags_. I think it's worth considering that this is the actual correct UX behaviour, as the highlight is meant to show the user that the component is actionable (we've seen how users are having a hard time understanding that without the highlight), so having tags doesn't mean the component is not actionable.

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


shlomitc:
The current state: no hover over the Tags container when in reorderable state. Regular tags has the hover effect only when no tags are selected.
Chen's initial suggestion: have the hover tag, regardless if tags are selected or not.

@milkyfruit - WDYT?

## Current behavior (while hovering)
image
image

image
image
image


## New behavior
image
image

image
image
image

milkyfruit:
I agree the new version would look more logical, however if we decide to head that direction, we need to update the <tag> color styles as well. Right now, <tag> is not designed to work together with input's blue hover effect.

shlomitc:
@ChenRoth - looks to me like the proper solution will be to support the native css with a condition of displaying tags or not (how the current version looks like, but support also the reorderable-no-tags)

shlomitc:
@milkyfruit - from what you wrote I understand that it is not approved. are you sure about your approval now?

ChenRoth:
any update?

milkyfruit:
@shlomitc @ChenRoth sorry guys, made an edit in the initial comment explaining why approved.

src/MultiSelect/MultiSelect.e2e.js
```diff
@@ -23,6 +23,15 @@ describe('MultiSelect', () => {
     // Should be focused and hovered
   });
 
+  eyes.it('should show hover style', async () => {
+    const driver = multiSelectTestkitFactory({dataHook: 'multi-select-standard'});
+    const element = driver.element();
+    await waitForVisibilityOf(element, 'Cannot find <MultiSelect/>');
+    await eyes.checkWindow('before hovered');
```
hepiyellow: This checkWindow is redundant. eyes take a snapshot right after `browser.get`

src/MultiSelect/MultiSelect.e2e.js
```diff
@@ -23,6 +23,15 @@ describe('MultiSelect', () => {
...
+    await mouseEnter(element);
+    await eyes.checkWindow('hovered');
```
hepiyellow: This checkWindow is also redundant. eyes take a snapshot at the end of every test.

ChenRoth:
now? :)

PR #2295, opened 2018-10-03
`<Text/>` (refactor) - remove usage of palette.st.css

hepiyellow:
We're about to remove `palette.st.css` from `wix-ui-backoffice`.
See https://github.com/wix/wix-ui-backoffice/pull/198

So we need to remove it's usage first in WSR.

### DIFF
--- a/src/Text/Text.st.css
+++ b/src/Text/Text.st.css
@@ -1,13 +1,13 @@
 :import {
-  -st-from: "wix-ui-backoffice/dist/src/palette.st.css";
-  -st-named: 
-    primaryText, secondaryText, primaryLightText, secondaryLightText,
-    success, danger, premium;
+  -st-from: "wix-ui-backoffice/dist/src/typography.st.css";
+  -st-named: fontLight, fontRoman, fontMedium, fontBold;
 }
 
 :import {
-  -st-from: "wix-ui-backoffice/dist/src/typography.st.css";
-  -st-named: fontLight, fontRoman, fontMedium, fontBold;
+  -st-from: "wix-ui-backoffice/dist/src/colors.st.css";
+  -st-named: 
+  D10, D20, D50,
+  G10, P10, R10, WHITE;
 }
 
 .root {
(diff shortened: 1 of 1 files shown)


src/Text/Text.st.css
```diff
@@ -1,13 +1,13 @@
...
-    primaryText, secondaryText, primaryLightText, secondaryLightText,
-    success, danger, premium;
+  -st-from: "wix-ui-backoffice/dist/src/typography.st.css";
+  -st-named: fontLight, fontRoman, fontMedium, fontBold;
 }
 
 :import {
-  -st-from: "wix-ui-backoffice/dist/src/typography.st.css";
-  -st-named: fontLight, fontRoman, fontMedium, fontBold;
+  -st-from: "wix-ui-backoffice/dist/src/colors.st.css";
```
shlomitc: please create `colors.st.css` in `wix-style-react` instead of using the `wix-ui-backoffice`

src/Text/Text.st.css
```diff
@@ -1,13 +1,13 @@
 :import {
-  -st-from: "wix-ui-backoffice/dist/src/palette.st.css";
-  -st-named: 
-    primaryText, secondaryText, primaryLightText, secondaryLightText,
-    success, danger, premium;
+  -st-from: "wix-ui-backoffice/dist/src/typography.st.css";
```
shlomitc: please use the `Text/Text.st.css` styles instead of the backoffice ones

src/Text/Text.st.css
shlomitc: we can merge, but I want to get rid of `wix-ui-backoffice` so sooner or later we have to remove it

src/Text/Text.st.css
hepiyellow: @shlomitc 
 Same thing here.
` fontFamily` variables do not exist in WSR,
we can duplicate them.
OR
We can create a `fonts.st.css` in WSR, and make it import from `backoffice`. Then when we deprecate WUB, we move the content from WUB to WSR.
That way at least we won't have to change many imports.

We can do the same for `colors.st.css`.

And, it is independent of this PR, so I'm merging.

PR #2302, opened 2018-10-08
Fix exposed `typography-helpers.scss`

hepiyellow:
## Why it changed

Initially, this PR comes to fix the exposed `typography-helpers.scss` which is at the root folder (after build).
It was broken since it had `@import './common.scss'` which had in it `@import './common/Focusable/FocusableMixins.scss';`. And there is no `common` folder under the root folder.

## What changed

Stuff done:
- Put public sass file in `sass` folder.
- Deprecate use of Sass files directly from `src`. Use `sass` folder instead.
- Keep `common.scss` under `src` (not copies to root on build)
- Extract `fonts.scss` out of `common.scss`
- rename `typography-helpers.scss` to `typography.scss` (there is no name clash with `src/Typography` folder when it is not under `src`), keep `src/typography-helpers.scss` for backwards compatibility.

Potentially breaking changes:
- `common.scss` is no longer on the root folder - but it was broken anyways.
- `mixins.scss` is no longer on the root folder.


hepiyellow:
After talking to @shlomitc , we decided to expose typography classes.
I think that in that case, there is no sense in exposing typography mixins.

I suggest:
- fixing `common.scss` by NOT having it at the root folder (broken import)
- remove typography-helpers
  - 1st step - add typography classes
 - 2nd step - refactor internal mixin usage to use classes
 - remove typography-mixins

@shlomitc what do you think?

hepiyellow:
Will do this after the typography classes PRs

Issue #2300, opened 2018-10-07
<DropdownLayout/> - accessibility issues


yardnsm:
# Bug Report

### Relevant Components

`<DropdownLayout/>`, also `<InputWithOptions/>`.

### Current Behavior

When pressing Enter (when the dropdown is focused), the dropdown layout won't open. Also, it **does** open when pressing Shift, which is not a desired behaviour.

### Expected Behavior

Should open the dropdown layout when pressing Enter.

### Steps to Reproduce

1. Go to: https://wix-wix-style-react.surge.sh/?selectedKind=4.%20Selection&selectedStory=4.1%20Dropdown&full=0&addons=0&stories=1&panelRight=0
2. Focus on the dropdown in the example
3. Press Enter / Shift

### Severity
- [ ] Critical
- [ ] Major
- [x] Low

PR #2310, opened 2018-10-09
<InputWithOptions/> - "refactor" onKeyDown event

yardnsm:
### What changed

This PR *partially* resolves https://github.com/wix/wix-style-react/issues/2300. It enables to open the DropdownLayout when the Enter key is pressed, and improves to readability of the a11y-related logic (mainly in `InputWithOptions#_onKeyDown`), by adding some comments that explain the behaviour.

However, it does not solve the issue where the DropdownLayout is being open when pressing Shift, Cmd, Alt/Option, and most of the keys that do not cause the input value to change. According to the a11y guidelines, the dropdown should open when the user start typing into the input.

@shlomitc maybe it'll be better to also fix the `Shift` in this PR before we merge this?

### Why it changed

https://github.com/wix/wix-style-react/issues/2300

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green

### DIFF
--- a/src/Dropdown/Dropdown.js
+++ b/src/Dropdown/Dropdown.js
@@ -68,17 +68,6 @@ class Dropdown extends InputWithOptions {
     this.setState({value: this.props.valueParser(option), selectedId: option.id});
     super._onSelect(option);
   }
-
-  _onFocus() {
-    if (this.props.disabled) {
-      return;
-    }
-    this._focused = true;
-    this.setState({isEditing: false});
-    if (this.props.onFocus) {
-      this.props.onFocus();
-    }
-  }
 }
 
 Dropdown.propTypes = InputWithOptions.propTypes;
--- a/src/InputWithOptions/InputWithOptions.js
+++ b/src/InputWithOptions/InputWithOptions.js
@@ -17,21 +17,24 @@ class InputWithOptions extends WixComponent {
 
   /**
    * An array of key codes that act as manual submit. Will be used within
-   * onKeyDown(event). The key codes are the values of {KeyboardEvent.key}.
+   * onKeyDown(event).
+   *
+   * @returns {KeyboardEvent.key[]}
    */
   getManualSubmitKeys() {
-    return [
-      'Tab',
-      ...(!this.isReadOnly ? ['Enter'] : [])
-    ];
+    return ['Tab', 'Enter'];
   }
 
   /**
-   * An array of key codes that should open the dropdown when in readOnly mode.
-   * Will be used within onKeyDown(event). The key codes are the values of {KeyboardEvent.key}.
+   * Returns a list of keys that should cause the DropdownLayout to be opened.
+   *
+   * @param {Object} options
+   * @returns {KeyboardEvent.key[]}
    */
-  getOpenKeys() {
-    return ['Enter', 'Spacebar', ' ', 'ArrowDown'];
+  getKeysForOpening() {
+    return this.isReadOnly ?
+      ['Enter', 'Spacebar', ' ', 'ArrowDown'] :
+      ['ArrowDown'];
   }
 
   constructor(props) {
@@ -258,32 +261,29 @@ class InputWithOptions extends WixComponent {
       this.setState({isEditing: true});
     }
 
-    // We'll always delegate if we're in readOnly mode
     const shouldDelegate = this.isReadOnly || !['Spacebar', ' '].includes(event.key);
+    const shouldOpenDropdown = this.getKeysForOpening().includes(event.key);
+
+    if (shouldOpenDropdown) {
+      this.showOptions();
+      event.preventDefault();
+    }
 
     if (shouldDelegate) {
 
       // Delegate event and get result
       const isHandledByDropdownLayout = this.dropdownLayout._onKeyDown(event);
 
-      if (!isHandledByDropdownLayout) {
-
-        if (this.readOnly) {
-
-          // For readOnly mode, we want to show the dropdown for specific keys
-          // only. Closing the dropdown should be handled by the DropdownLayout
-          // itself.
-          if (this.getOpenKeys().includes(event.key)) {
-            this.showOptions();
-            event.preventDefault();
-          }
-
-        // For editing mode, we want to *submit* only for specific keys.
-        // Otherwise, we'll show the dropdown and won't call preventDefault
-        // so the event will delegate to the Input.
-        } else if (this.getManualSubmitKeys().includes(event.key)) {
+      // For editing mode, we want to *submit* only for specific keys.
+      // Otherwise, we'll show the dropdown and won't call preventDefault
+      // so the event will delegate to the Input.
+      if (!isHandledByDropdownLayout && !this.isReadOnly) {
+        if (this.getManualSubmitKeys().includes(event.key)) {
           this._onManuallyInput(this.state.inputValue);
         } else {
+          // TODO: only show the option when the input value has changed.
+          // Currently, is also shows it when pressing keys like Shift, Alt,
+          // Cmd, etc.
           this.showOptions();
         }
       }
--- a/src/InputWithOptions/InputWithOptions.js
+++ b/src/InputWithOptions/InputWithOptions.js
@@ -14,11 +14,24 @@ class InputWithOptions extends WixComponent {
   dropdownClasses() {}
   dropdownAdditionalProps() {}
   inputAdditionalProps() {}
+
   /**
-   * An array of key codes (default ['Enter','Tab']) that act as manual submit. Will be used with the onKeyDown(event), tha key codes are values of event.key. When a manual submit key is pressed then onManuallyInput will be called.
+   * An array of key codes that act as manual submit. Will be used within
+   * onKeyDown(event). The key codes are the values of {KeyboardEvent.key}.
    */
   getManualSubmitKeys() {
-    return ['Enter', 'Tab'];
+    return [
+      'Tab',
+      ...(!this.isReadOnly ? ['Enter'] : [])
+    ];
+  }
+
+  /**
+   * An array of key codes that should open the dropdown when in readOnly mode.
+   * Will be used within onKeyDown(event). The key codes are the values of {KeyboardEvent.key}.
+   */
+  getOpenKeys() {
+    return ['Enter', 'Spacebar', ' ', 'ArrowDown'];
   }
 
   constructor(props) {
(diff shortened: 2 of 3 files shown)


src/InputWithOptions/InputWithOptions.js
```diff
@@ -245,13 +260,31 @@ class InputWithOptions extends WixComponent {
       this.setState({isEditing: true});
     }
 
-    const shouldDelegate = !['Spacebar', ' '].includes(event.key) || this.isReadOnly;
+    const shouldDelegate = this.isReadOnly || !['Spacebar', ' '].includes(event.key);
```
shlomitc: some code refactor
```
function shouldOpenDropdown(key, isReadOnly) {
  if(isReadOnly) {
    return Enter || DownArrow || Spacebar;
  } else {
    return DownArrow || pure text???
  }
}

//maybe improve this one
function shouldDelegate(key, isReadOnly) {
  if(isReadOnly) {
    return true
  } else {
    return !Spacebar
  }
}

if(shouldOpenDropdown) {
  openDropdown();
  preventDefault()
}

if (shouldDelegate()) {
  wasHandled = delegate();
  if(!(wasHandled || readOnly)) {
    return;
  }

  if(shouldDoManualSubmit()) {
    doManualInput()
  }
}
```

src/InputWithOptions/InputWithOptions.driver.js
```diff
@@ -22,6 +22,9 @@ const inputWithOptionsDriverFactory = ({element, wrapper, component}) => {
     pressSpaceKey: () => inputDriver.keyDown(' '),
     pressTabKey: () => inputDriver.keyDown('Tab'),
     pressEscKey: () => inputDriver.keyDown('Escape'),
+    pressShiftKey: () => inputDriver.keyDown('Shift'),
```
hepiyellow: I wouldn't add MORE verbose methods.
If anything, I would remove all these pressXxxxKey methods.
First of all it is missleading, since pressXxxxKey doesn't say on which element it triggers the key (input of dropdownlayout).

This driver exposes the inputDriver, which has a `keyDown` method that can and should be used.

I don't think this would be a breaking change, I don't thin consumers use it. you can search for it.
but I wouldn't bother, just break it (-:

src/InputWithOptions/InputWithOptions.js
```diff
@@ -14,11 +14,15 @@ class InputWithOptions extends WixComponent {
...
   /**
-   * An array of key codes (default ['Enter','Tab']) that act as manual submit. Will be used with the onKeyDown(event), tha key codes are values of event.key. When a manual submit key is pressed then onManuallyInput will be called.
+   * An array of key codes that act as manual submit. Will be used within
+   * onKeyDown(event).
+   *
+   * @returns {KeyboardEvent.key[]}
    */
   getManualSubmitKeys() {
-    return ['Enter', 'Tab'];
+    return ['Tab', 'Enter'];
```
hepiyellow: Any reason for changing the order? does the order have any meaning?

src/InputWithOptions/InputWithOptions.spec.js
```diff
@@ -146,22 +146,26 @@ const runInputWithOptionsTest = driverFactory => {
       });
     });
 
-    it('should show DropdownLayout on any key press', () => {
+    it('should show DropdownLayout only on specific keys', () => {
       const {driver, dropdownLayoutDriver} = createDriver(<InputWithOptions options={options}/>);
       driver.pressAnyKey();
```
hepiyellow: What is this anyKey? better make it explicit which key is pressed, and say in the test description:
`should NOT show DropdownLayout when a non WhiteListed key is pressed`

Or something similar...

src/InputWithOptions/InputWithOptions.spec.js
```diff
@@ -146,22 +146,26 @@ const runInputWithOptionsTest = driverFactory => {
...
 
     it('should show DropdownLayout on down key', () => {
       const {driver, dropdownLayoutDriver} = createDriver(<InputWithOptions options={options}/>);
-      driver.pressUpKey();
+      driver.pressDownKey();
       expect(dropdownLayoutDriver.isShown()).toBeTruthy();
     });
 
-    it('should show DropdownLayout on up key', () => {
+    it('should not DropdownLayout on modifier keys', () => {
```
hepiyellow: should not show (missing show)

src/InputWithOptions/InputWithOptions.js
yardnsm: This was a result of a few changes that I finally reverted manually. No reason 🙃

src/InputWithOptions/InputWithOptions.spec.js
yardnsm: I *think* that the `'Any'` key is a [non-standard key value](https://developer.mozilla.org/he/docs/Web/API/KeyboardEvent/key/Key_Values), but it is still triggering the `keydown` event.

For the sake of the tests, I think its good enough to keep it. However, this test is kind of problematic, as there are a lot of keys than can cause the dropdown to open (`Enter`, `Space`, `ArrowDown`, **and every key that changes the `<input/>` value**), so there isn't a *specific* "whitelist" of keys. I think we should combine this test with the `should not show DropdownLayout on modifier keys` case.

yardnsm:
@shlomitc @hepiyellow can we merge?

hepiyellow:
Please verify the required bahavior of arrow UP/DOWN opening the options.
The options can be opened in directionUp, in which case it is logical that up-arrow would open it (which you have now disabled).

I suggest having both down/up arrows open the options (independent of the options direction).
Please talk to @wuwa.

src/InputWithOptions/InputWithOptions.js
```diff
@@ -161,6 +165,41 @@ class InputWithOptions extends WixComponent {
...
+   *
+   * @param {KeyboardEvent.key}
+   * @returns {boolean}
+   */
+  shouldOpenDropdown(key) {
+    const openKeys = this.isReadOnly ?
+      ['Enter', 'Spacebar', ' ', 'ArrowDown'] :
+      ['ArrowDown'];
+
+    return openKeys.includes(key);
```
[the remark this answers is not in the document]
yardnsm: It is used in some more places in wsr, plus we got `babel-polyfill` so I think its ok.

src/InputWithOptions/InputWithOptions.driver.js
```diff
@@ -15,13 +15,7 @@ const inputWithOptionsDriverFactory = ({element, wrapper, component}) => {
...
-    pressDownKey: () => inputDriver.keyDown('ArrowDown'),
-    pressUpKey: () => inputDriver.keyDown('ArrowUp'),
-    pressAnyKey: () => inputDriver.keyDown('Any'),
-    pressEnterKey: () => inputDriver.keyDown('Enter'),
-    pressSpaceKey: () => inputDriver.keyDown(' '),
-    pressTabKey: () => inputDriver.keyDown('Tab'),
-    pressEscKey: () => inputDriver.keyDown('Escape'),
+    pressKey: key => inputDriver.keyDown(key),
     outsideClick: () => document.body.dispatchEvent(new Event('mouseup', {cancelable: true})),
     setProps: props => {
```
yardnsm: Remove it

src/InputWithOptions/InputWithOptions.driver.js
```diff
@@ -15,13 +15,7 @@ const inputWithOptionsDriverFactory = ({element, wrapper, component}) => {
     inputWrapper: () => inputWrapper,
     focus: () => inputDriver.focus(),
     blur: () => dropdownLayoutDriver.mouseClickOutside(),
-    pressDownKey: () => inputDriver.keyDown('ArrowDown'),
-    pressUpKey: () => inputDriver.keyDown('ArrowUp'),
-    pressAnyKey: () => inputDriver.keyDown('Any'),
-    pressEnterKey: () => inputDriver.keyDown('Enter'),
-    pressSpaceKey: () => inputDriver.keyDown(' '),
-    pressTabKey: () => inputDriver.keyDown('Tab'),
-    pressEscKey: () => inputDriver.keyDown('Escape'),
```
yardnsm: Add deprecation logs

yardnsm:
@hepiyellow @shlomitc Please take a look at the new changes. https://github.com/wix/wix-style-react/pull/2327 changed the scrolling behaviour by setting the focus on the hovered option, but did not returned the focus back to the `<Input/>`. The new change (https://github.com/wix/wix-style-react/commit/36374a45a733e51b9f748637f69cd03a6c854a4c) fixes that.

yardnsm:
@hepiyellow @shlomitc Please take a look at the new changes. https://github.com/wix/wix-style-react/pull/2327 changed the scrolling behaviour by setting the focus on the hovered option, but did not returned the focus back to the `<Input/>`. The new change (https://github.com/wix/wix-style-react/commit/36374a45a733e51b9f748637f69cd03a6c854a4c) fixes that.

yardnsm:
@hepiyellow @shlomitc Please take a look at the new changes. https://github.com/wix/wix-style-react/pull/2327 changed the scrolling behaviour by setting the focus on the hovered option, but did not returned the focus back to the `<Input/>`. The new change (https://github.com/wix/wix-style-react/commit/36374a45a733e51b9f748637f69cd03a6c854a4c) fixes that.

yardnsm:
@hepiyellow @shlomitc Please take a look at the new changes. https://github.com/wix/wix-style-react/pull/2327 changed the scrolling behaviour by setting the focus on the hovered option, but did not returned the focus back to the `<Input/>`. The new change (https://github.com/wix/wix-style-react/commit/36374a45a733e51b9f748637f69cd03a6c854a4c) fixes that.

yardnsm:
@hepiyellow @shlomitc Please take a look at the new changes. https://github.com/wix/wix-style-react/pull/2327 changed the scrolling behaviour by setting the focus on the hovered option, but did not returned the focus back to the `<Input/>`. The new change (https://github.com/wix/wix-style-react/commit/36374a45a733e51b9f748637f69cd03a6c854a4c) fixes that.

hepiyellow:
@yardnsm 
Did you try using `Element.scrollIntoView()`?
https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

src/InputWithOptions/InputWithOptions.spec.js
```diff
@@ -146,33 +146,35 @@ const runInputWithOptionsTest = driverFactory => {
...
-      driver.pressAnyKey();
-      expect(dropdownLayoutDriver.isShown()).toBeTruthy();
+      driver.pressKey('Any');
+      expect(dropdownLayoutDriver.isShown()).toBeFalsy();
     });
 
     it('should show DropdownLayout on down key', () => {
       const {driver, dropdownLayoutDriver} = createDriver(<InputWithOptions options={options}/>);
-      driver.pressUpKey();
+      driver.pressKey('ArrowDown');
```
yardnsm: Maybe change to smth else, e.g. `someKey` and add a comment that explains that this key is just triggering the onkeydown event

hepiyellow: You can extract a function `pressSomeKey` in the test (not the driver).
And add comment on the function.

src/Dropdown/Dropdown.js
```diff
@@ -68,17 +68,6 @@ class Dropdown extends InputWithOptions {
     this.setState({value: this.props.valueParser(option), selectedId: option.id});
     super._onSelect(option);
   }
-
-  _onFocus() {
```
hepiyellow: I usually add comments on my own PR, to help the reviewer understand stuff.
Like why this was removed (redundant)

src/InputWithOptions/InputWithOptions.spec.js
yardnsm: I think it's better this way

PR #2312, opened 2018-10-09
`<Text/>` - Update secondary+light color from D50 to D40

hepiyellow:
Update text according to UX spec


lbelinsk:
@milkyfruit worth updating this zeplin file, as we already decided what option to go with

PR #2317, opened 2018-10-11
`eyes.it` (test) - install eyes.it#refactor_with_configure_snapshots_2_MOCK

hepiyellow:
# This PR is only about testing changes in `eyes.it`

It installs `eyes.it` from a branch.

## Expected eyes results
- All should pass
- Except Button.e2e
   - [x] `should render disabled` - should have no snapshot at begining and end
   - [x] `should render prefix & sufix` - should have an END snapshot with different name ("*** THE END***")
   - [x] `render variations should display all themes with props={}` - should not have browser.get snapshot


shlomitc:
@hepiyellow - can we close this one?

hepiyellow:
Not need anymore

PR #2319, opened 2018-10-12
Added RTL support to stats

rghorbani:
### What changed

Added RTL support to Stats (fixed stat dividers)

Before:
screen shot 2018-10-12 at 3 35 51 pm

After:
screen shot 2018-10-12 at 3 33 12 pm

### Why it changed

...

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green


rghorbani:
@shlomitc Did the changes you requested.
Should I make another pull request extending `withRTL`  to other components too?

shlomitc:
I now see that the `className="rtl"` approach didn't change the direction.
It requires manual changes of the styles to support it, while `dir="rtl"` works out of the box

class name
image

dir
image

Which one of the strategies you use in your application?

rghorbani:
@shlomitc I mainly use className approach and I think as it is the original approach suggested by the docs and if we want to extend the `withRTL` to other components as well we should fix the problem.

rghorbani:
@shlomitc I personally tested it, but did not find any bug. As described in the docs I gave .rtl className a style `direction="rtl"` and everything is alright.

shlomitc:
@rghorbani - using `dir=rtl` work correctly.
Using the classname alone causes the bug. This is why I asked which approach you take.

try reproducing by until applying the classname, not the dir.

@lbelinsk - do you think we should stick to the `dir` only and ditch classnames?

rghorbani:
@shlomitc Actually It's because you have not defined .rtl className
`
.rtl {
  direction: rtl;
}
`

shlomitc:
I just added the class to the iframe content body.
did you need in your project to write this `.rtl { direction: rtl; }`?

rghorbani:
@shlomitc yes,
We always need to change the direction either by using `dir="rtl"` or defining direction in class.
Exception to all this is usually the `input` as it is in react-native as well.

PR #2320, opened 2018-10-12
<ButtonLayout/> - Added borderRadius prop

rghorbani:
### What changed

Added borderRadius prop to `ButtonLayout` to provide custom borderRadius
Also added test for the change.

### Why it changed

@milkyfruit 
Needed to add borderRadius to `Button` more easily (beside defining custom className) as I've seen in most libraries.

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green


shlomitc:
Hey. Thanks for the contribution.

`wix-style-react` is a component library that is designed to comply with the wix back-office design system. The base components design is strict and not supposed to be overridden. This assures the best quality of the product and alignment to the UX guidelines.

We do however allow passing `className` for use cases such as alignment, which gives you some customizability options. We do not encourage it, but also don't want to prevent you from doing it.

So I will unfortunately need to close this PR.

Thanks again

PR #2321, opened 2018-10-13
tech-debt: create an offline storybook configuration

shlomitc:
Another approach solving the offline storybook (following this PR: https://github.com/wix/wix-style-react/pull/2020)

Explanation:
1. I created two separate configurations that share a common config. The difference between the two is only the `manager-head.html` and `preview-head.html` files.
2. In the offline version, I mapped both `react` and `react-dom` folders and serve them from the same storybook port

Known issue:
1. I didn't handle the fonts loading in the offline as it is bundled into the CSS somehow already (could be one of the stories or the `index.js` file).


hepiyellow:
How does stirybook serve react in offline mode? Where are the served files licated?

Can we serve non-minified version?

I would also use non-minified for the officual storybook. Some react error logs dont show in the minified versions, and it makes it difficult to debug/investigate.

shlomitc:
1. I serve the node_modules folder, specifically the react and react-dom folders: https://github.com/wix/wix-style-react/blob/0f5645df449e7634206fa941cae13f2fac9c7e9e/package.json#L33
2. yes, it is possible to serve also the non-minified version (in the same folder), but let's not touch it in this pr

PR #2322, opened 2018-10-14
<Tag /> - add tiny & medium sizes

boriskor:
### What changed

Added tiny & medium sizes to `<Tag />`

### Why it changed

Closes https://github.com/wix-private/wsr-issues/issues/266

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green

### Notes
I followed the Zeplin, but there are some contradictions there, I wrote them there.
Anyway, 
```jsx
<Boris tShirtSize="x-large">
```
😇


src/Tag/Tag.e2e.js
```diff
@@ -12,20 +12,18 @@ describe('Tag', () => {
     await browser.get(url);
   });
 
-  afterEach(() => {
+  beforeEach(() => {
```
shlomitc: why `beforeEach`?

src/Tag/Tag.driver.js
```diff
@@ -31,7 +54,11 @@ const tagDriverFactory = ({element}) => {
     isWrapped: () => isClassExists(element, 'tagEllipsis') && isClassExists(getContentWithoutThumb(element), 'tagTextEllipsis'),
     isDisabled: () => isClassExists(element, 'disabled'),
     getLabel: () => element.textContent,
-    getTitle: () => element.title
+    getTitle: () => element.title,
+    isCloseButtonSmall: () => !isCloseButtonLarge(),
+    isCloseButtonLarge,
+    getTextSize: () => getTextDriver(element).getSize(),
```
shlomitc: this is too much of an internal logic and not something the consumer needs.
as the E2E test is taking screenshot of it, I think it's enough to assume it is taken care there.


same for the textWeight and the close button size

shlomitc:
@LittleTal - please review the new sizes

src/Tag/Tag.e2e.js
boriskor: Wanted it to be aligned with https://github.com/wix/wix-style-react/blob/master/src/AddItem/AddItem.e2e.js#L20 & https://github.com/wix/wix-style-react/blob/master/src/common/Focusable/FocusableTestsE2E.js#L73 , but I see that the docs use after each. I'll revert.

src/Tag/Tag.driver.js
boriskor: I see.

I want to point you to the fact that the text weight was wrong before this PR. It was `thin` for all sizes of the tag (the default of `<Text />`) instead of being `thin` for `tiny' & `normal` for the rest of the sizes, like in Zeplin. I fixed it TDD way. Now it's explicitly written in the tests and the people would think twice before breaking it. If you think that Eyes tests are enough, I'll remove the unit tests.

But from what you're saying, you have a problem with exposing it to the consumers, not with testing it. Maybe we need to separate it to private & public drivers? This way we still will be able to test it properly using the internal driver, and to consumers we will expose the public driver.

WDYT?

shlomitc: Thanks for fixing.
We started with the private driver approach so it's actually a good idea, let's do it!

See this for example:
https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/button-next/button-next.driver.private.ts

This is the private driver we're using in our tests, which composes the public driver.

So all the new functions you added in this PR and should go to the private. let's keep the old ones in the public one for backward compatibility.

Thanks

boriskor:
@LittleTal , good catch, thanks!

boriskor:
@shlomitc , all sized are confirmed by @milkyfruit . Ready for review.

shlomitc:
👑

Issue #2169, opened 2018-08-29
<InputWithOptions> using arrows to scroll through many options causes current element not be visible


liorur:
### Priority:
- [ ] Blocker
- [ ] Urgent
- [ ] Normal
- [x ] Minor

### Steps to reproduce:

 create InputWithOptions with a long list of options
scenario 1 - scroll down using down arrow - current element slowly drift outside the pane
scenario 2 - scroll using up arrow - last element is not marked as current

shlomitc:
both scenarios sound legit to me, but I might be missing something.
can you explain what is the wanted behavior?

liorur:
the behaviour i expect is for the current element (marked by bluish background) to always be in view when using up/down arrows to brows options

shlomitc:
@wuwa - what do you think?

wuwa:
long one has a bug. short (medium) works completely fine in my eyes.

yardnsm:
The problematic line is:

https://github.com/wix/wix-style-react/blob/2a459fb574873dbc77ba02cf391cbf8c77de81e9/src/DropdownLayout/DropdownLayout.js#L124

We're missing a pixel... We need to change `35` to `36`, as this is the current height of a single option.

This could cause problems, though. There are options that has a height different than `36px`, so we need to rethink the strategy to calculating the `scrollTop` for the dropdown.

For now, maybe the change to `36` is good enough?

yardnsm:
Resolved in https://github.com/wix/wix-style-react/pull/2327

PR #2327, opened 2018-10-14
<DropdownLayout/> - fix scroll behaviour

yardnsm:
Fix for https://github.com/wix/wix-style-react/issues/2169.

Please take a look at the changes. I'm using `getBoundingClientRect()` to retrieve an element's height, maybe there is a more efficient way to do this?


hepiyellow:
@wuwa @LittleTal @milkyfruit 
Is this the required UX behaviour? (That the selected option will always be in the center)
It seems that the solution in say, native `<select/>` with chrome, or MaterialUI <Select/>,
is that when a selected option is not fully visible, then the scroll is applied the **"minimum adjustment needed"** in order to make it fully visible.

Whats your decision?

yardnsm:
Also, in the native `<select/>`, when the focus is on the last item and the user presses `ArrowDown`, the focus does not go the first element (in wsr it *does*). Same thing goes when pressing `ArrowUp` when the first item is in focus.

hepiyellow:
## Using focus
just one concern. It means that the options are focusable (have `tabIndex`).AM I right?
So need to make sure that TAB does NOT focus on them, but rather got OUT of the DropdownLayout.

## Cyclic navigation
This is a UX decision.
There are a few decisions to make:

### Optional
Can the consumer disable ALL cyclic behaviour?

### Repeated-Key
When cyclic navigation is enabled, it is still important to disable it when repeated-key is pressed.
Meaning that if you press down arrow and keep it pressed, so you navigate FAST through the options, in which case you don't want to have cyclic navigation. It is very annoying.

### Scroll dependent
Another logic may say that if the options container has scroll, then cyclic navigation should be disabled.

@wuwa , @milkyfruit @LittleTal What do you think?

yardnsm:
Regarding to focus: Tab does not focus on the items, as the `onkeydown` event in `<DropdownLayout/>` already prevents it.

milkyfruit:
@hepiyellow @yardnsm 

**Focus / Tab-Index**
The focus should apply to whole container and the selected list item should be switched by using arrow keys. [More about a11y.](https://v9yfnh.axshare.com/#g=1&p=dropdown)

**Cyclic navigation**
I don't have a strong opinion about it. MacOS and Material platforms don't support this feature by default. We can make it as optional property, which could be enabled if needed.

**Scrolling & Repositioning**
When list is longer than a container, on arrow down the list should scroll until the next nearest option, as Erez mentioned:

> It seems that the solution in say, native <select/> with chrome, or MaterialUI , is that when a selected option is not fully visible, then the scroll is applied the "minimum adjustment needed" in order to make it fully visible.

hepiyellow:
ok . So no cyclic behavior yet.

hepiyellow:
# Sorry guys, I think we need to revert this PR !
We can't make each option focusable, since the focus must stay on the input box in cases like AutoComplete.
(so that pressing keyboard letters would type them into the input).

What happens now is that typing letters doesn't do anything.
Furthermore, it is not possible to navigate by keyboard back to the input (I am talking about Autocomplete).

See this [old storybook](https://wix-wix-style-react.surge.sh/?selectedKind=4.%20Selection&selectedStory=4.1%20%2B%20AutoComplete&full=0&addons=0&stories=1&panelRight=0) where it works properly .

The [new storybook](https://wix-wix-style-react.surge.sh/?selectedKind=4.%20Selection&selectedStory=4.1%20%2B%20AutoComplete&full=0&addons=0&stories=1&panelRight=0) is broken.

PR #2329, opened 2018-10-15
Revert to older version of yoshi

mykas:
Quick fix until we resolve ellipsis problem.


shlomitc:
This won’t help - consumers are using yoshi as well meaning they will get the wrong version.

We should either find a solution or revert the yoshi changes until we understand what happened

PR #2335, opened 2018-10-16
Calendar: added selectedDays prop, deprecated value prop

malsomnus:
### What changed

Added selectedDays prop in Calendar
Deprecated value prop in Calendar


src/Calendar/Calendar.js
```diff
@@ -171,9 +198,15 @@ Calendar.propTypes = {
...
   rtl: PropTypes.bool,
 
-  /** The selected date */
+  /** (Deprecated) The selected date */
   value: PropTypes.object,
 
+  /** The selected date range */
+  selectedDays: PropTypes.oneOfType([
+    PropTypes.object,
+    PropTypes.array
```
hepiyellow: You can be more specific and use `PropTypes.arrayOf(PropTypes.object)` (or something like that)

src/Calendar/Calendar.js
```diff
@@ -43,6 +51,21 @@ export default class Calendar extends WixComponent {
...
+    const {
+      value,
+      selectedDays
+    } = props;
+
+    if (!selectedDays) {
+      return value;
+    }
+    else {
+      const firstObject = (Array.isArray(selectedDays)) ? selectedDays[0] : selectedDays;
```
hepiyellow: Let's not support multiple ranges for now.
So `selectedDays` can be either a single Date object or a single Range object.

src/Calendar/Calendar.js
```diff
@@ -51,13 +74,16 @@ export default class Calendar extends WixComponent {
...
+      selectedDays,
       rtl,
       twoMonths
     } = this.props;
 
-    const month = this.state.month || propsValue || new Date();
+    const month = this.state.month || this._getMonth(this.props) || new Date();
     const localeUtils = localeUtilsFactory(locale);
-
+    const from = (Array.isArray(selectedDays)) ? map(selectedDays, 'from') : (selectedDays || {}).from;
```
hepiyellow: selectedDays can also be a single Date object (not a Range object), to in that case you want to take `selectedDays` and not `selectedDays.from`.

src/Calendar/Calendar.spec.js
```diff
@@ -0,0 +1,66 @@
...
+//import ReactTestUtils from 'react-dom/test-utils';
+import calendarDriverFactory from './Calendar.driver';
+import {createDriverFactory} from 'wix-ui-test-utils/driver-factory';
+//import {checkboxTestkitFactory} from '../../testkit';
+import Calendar from './Calendar';
+//import {mount} from 'enzyme';
+
+const originalConsoleWarn = global.console.warn; // eslint-disable-line no-console
+
+describe('Old prop deprecation', () => {
```
hepiyellow: Please keep all tests under one decribe block ('Calendar' block).
You can nest `describe blocks within each other.

src/Calendar/Calendar.spec.js
```diff
@@ -0,0 +1,66 @@
...
+  const createDriver = createDriverFactory(calendarDriverFactory);
+
+  beforeEach(() => {
+    global.console.warn = jest.fn();
```
hepiyellow: I don't think you need to use `global` here

src/Calendar/Calendar.js
```diff
@@ -27,22 +27,31 @@ export default class Calendar extends WixComponent {
...
   componentWillReceiveProps(nextProps) {
-    this.setState({month: nextProps.value});
+    this.setState({month: this._getMonth(nextProps)});
   }
 
   _setMonth = month => this.setState({month});
 
   _handleDayClick = (value, modifiers = {}) => {
-    this.props.onChange(value, modifiers);
+    this.props.onSelectedDaysChange(value, modifiers);
```
hepiyellow: We are still supporting `onChange`, you should call it also.

src/Calendar/Calendar.spec.js
```diff
@@ -0,0 +1,66 @@
...
+  it('should show deprecationLog when the onChange prop is used', () => {
+    const driver = createDriver(
+      <Calendar
+        onChange={() => {}}
+        />
+    );
+    expect(driver.exists()).toBeTruthy();
+    expect(global.console.warn.mock.calls.length).toEqual(1);
+  });
+
```
hepiyellow: Add test for using `value` and `onChange` see that they still work.

src/Calendar/Calendar.js
```diff
@@ -171,8 +197,20 @@ Calendar.propTypes = {
   /** RTL mode */
   rtl: PropTypes.bool,
 
-  /** The selected date */
-  value: PropTypes.object,
+  /** (Deprecated) The selected date */
+  value: (props, propName) => {
+    if (props[propName]) {
+      const msg = 'Calendar\'s onValue prop is deprecated, please use onSelectedDaysChange instead.';
```
hepiyellow: in msg: `onSelectedDays` not `onSelectredDaysChange`

src/Calendar/Calendar.js
```diff
@@ -171,8 +197,20 @@ Calendar.propTypes = {
...
+      const key = msg;
+      deprecationLog(msg, key);
+    }
+  },
+
+  /** The selected date range. Can be either a single JS Date object, or a single range object {from: Date, to: Date} */
+  selectedDays: PropTypes.oneOfType([
+    PropTypes.object,
+    PropTypes.array
```
hepiyellow: We said we are not supporting multiple days or multiple ranges for now. so no need for array.

src/Calendar/Calendar.spec.js
```diff
@@ -0,0 +1,66 @@
...
+    const value = new Date(2018, 10, 1);
+    const selectedDays = new Date(2018, 9, 1);
+
+    const driver = createDriver(
+      <Calendar
+        value={value}
+        selectedDays={selectedDays}
+        />
+    );
+    expect(driver.getMonthCaption()).toEqual('October');
```
hepiyellow: shouldn't this be September?

hepiyellow:
Closed in favor of #2403

src/Calendar/Calendar.js
malsomnus: Changed to just `PropTypes.object` since we're not supporting multiple ranges.

src/Calendar/Calendar.spec.js
malsomnus: Nope, the months are 0-based.

Issue #2338, opened 2018-10-16
<DropdownLayout/> - allow to override the styles for a single option


yardnsm:
As we starting to move on with the concept of [builders](https://wix-wix-style-react-pr-2219.surge.sh/?selectedKind=WIP%2FBuilders&selectedStory=BadgeSelectItemBuilder&full=0&addons=0&stories=1&panelRight=0), we need to have a way to change an option's padding (for example, in `<BadgeSelect/>`.

One solution is to have a toggle that will disable the spacing within an option. That way, the builder will be responsible for setting that spacing, and the styles set by `<DropdownLayout/>` will only handle hover / active states.

We can have this toggle as a prop of `<DropdownLayout/>`, but then it'll be problematic if we're about to mix different builders with "regular" options. As a result, I think the best solution is to have a property (in the option object) that will be set by the builder itself.

So, a builder function would return something like this:

```js
{ id: '0', value: <div style={{ padding: 20 }}>Some value</div>, disablePadding: true }
```

### Severity

- [ ] Critical
- [ ] Major
- [x] Low

shlomitc:
@wuwa - I'm adding you to this discussion.
We talked about those item builders in the past and now it is becoming real.

`DropdownLayout` component is the list of items and has default item styles (like the single-select)
`BadgeSelectItemBuilder`  requires different padding than the regular items.

It makes sense that the padding will be the same for all different the list items.

Can you please map the different scenarios and check why we don't have consistency?

hepiyellow:
@yardnsm @shlomitc 
Why not ad a "RegularBuilder" so you can mix Builders.
And remove the option styling from DropdownLayout completely.

To be backward compatible, we can have a new prop `optionBuilders` which if defined, overrides the `options` prop. And add a deprecationLog to the usage of `options`.

> An alternative to that, while we're at it, is to have the builders as children instead of a prop.

yardnsm:
Closing, there's the `overrideStyle` property you can use. There's event an example for this in the `<DropdownLayout/>` story.

PR #2357, opened 2018-10-21
<DropdownLayout/> - add `disableSpacing` property to an option

yardnsm:
### What changed

Added a property to an option object called `disableSpacing`. When `true`, the spacing-related styles (`padding`, `display`, `align-items`) for the option will be disabled.

This is useful for builders that need to set a different spacing for an option. The builder can set `disableSpacing` and add its own styles in the root div.

### Why it changed

Resolves https://github.com/wix/wix-style-react/issues/2338.

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green

### DIFF
--- a/src/BadgeSelectItemBuilder/BadgeSelectItem.st.css
+++ b/src/BadgeSelectItemBuilder/BadgeSelectItem.st.css
@@ -16,6 +16,10 @@
   -st-states: skin(string);
   white-space: nowrap;
   line-height: 24px;
+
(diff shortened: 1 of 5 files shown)


hepiyellow:
See discussion comment here: https://github.com/wix/wix-style-react/issues/2338

shlomitc:
@yardnsm - what is the status of this PR?

shlomitc:
let's revise this when becomes relevant

Issue #2359, opened 2018-10-22
<Table/> - column  align:center does not work with infoTooltip


liorur:
```
{
        title: 'title',
        render: () =><div>text</div ,
        align: 'center',
        infoTooltip: {
          content:'content'
        },
      },
```

### Relevant Components
<Table>

### Current Behavior
th is not centred, the tr below are fine

### Expected Behavior
th should be centred

### Severity
- [ ] Critical
- [ ] Major
- [ x] Low

PR #2333, opened 2018-10-15
<Table/> - support column alignments

yardnsm:
### What changed

Able to pass `align` in the that describes the table's columns, to control the text alignment for that column. For example:

```jsx
<Table
  dataHook="story-table-column-alignment-example"
  data={baseData}
  itemsPerPage={20}
  columns={[
      {title: 'Name', render: row => <span>{row.name}</span>, width: '30%', minWidth: '150px'},
      {title: 'SKU', render: row => <span>{row.SKU}</span>, width: '20%', minWidth: '100px', align: 'start'},
      {title: 'Price', render: row => <span>{row.price}</span>, width: '20%', minWidth: '100px', align: 'center'},
      {title: 'Inventory', render: row => <span>{row.inventory}</span>, width: '20%', minWidth: '100px', align: 'end'}
  ]}
  showSelection
  >
  <MainToolbar/>
  <Table.Content/>
</Table>
```

screen shot 2018-10-15 at 16 03 49

### Why it changed

https://github.com/wix-private/wsr-issues/issues/236

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green
- [ ] Change of UX is approved by [wuwa](https://github.com/wuwa) or [milkyfruit](https://github.com/milkyfruit)

PR #2360, opened 2018-10-22
<Table/> - Fix column alignment when using a tooltip

yardnsm:
### What changed

When using an `infoTooltip` with a column, the alignment does not apply for the header cell of the specific column (the rows are aligned fine). This PR fixes it.

Also, this PR fixes a typo in the `<Table/>`'s README (`infoTooltipProps` --> `infoTooltip`).

### Why it changed

Resolves https://github.com/wix/wix-style-react/issues/2359

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


src/Table/Table.js
```diff
@@ -154,7 +154,7 @@ Table.propTypes = {
     ]).isRequired,
     render: PropTypes.func.isRequired,
     sortable: PropTypes.bool,
-    infoTooltipProps: PropTypes.shape(Tooltip.propTypes),
+    infoTooltip: PropTypes.shape(Tooltip.propTypes),
```
hepiyellow: I suggest to keep it `infoTooltipProps` (it is what it is).
- fix DataTable to use `infoTooltipProps` instead of `infoTooltip`.
- Add `infoTooltip` to prop types DataTable with a deprecationLog

hepiyellow:
Also missing e2e test with column alignment.
Need to take a snapshot of `story-table-column-alignment-example`.

src/Table/Table.js
yardnsm: I'll revert these changes and open a new PR for that

yardnsm: https://github.com/wix/wix-style-react/pull/2367

yardnsm:
@wuwa @milkyfruit can you please review?

yardnsm:
The column alignment feature was actually in https://github.com/wix/wix-style-react/pull/2333 (which was merged without a UX review 😅).

Anyway, here's an example of a table with info tooltips and different alignments:

screen shot 2018-10-24 at 17 17 48

wuwa:
last one (right alignment - price column) looks slightly off. would be much better if there is a way to get a live preview link.

PR #2361, opened 2018-10-23
`<TextField/>` - deprecated component, and add snippet story using `<FormField/>`

hepiyellow:
The new story looks a bit lame.
I don't know which examples should we add.
And some UX design would probably help here.
@wuwa @milkyfruit @LittleTal - Take a look at [MaterialUI-TextFields docs page](https://material-ui.com/demos/text-fields/)

@shlomitc 
1. Does it make sense in cases like `<TextField/>`, to have ALL the examples in the TextField story,
and in the Input story, have only the auto-example playground?
1. The [Input story](https://wix-wix-style-react.surge.sh/?selectedKind=3.%20Inputs&selectedStory=3.1%20%2B%20Input&full=0&addons=0&stories=1&panelRight=0) is a mess. And it doesn't use AutoExample. Should I refactor it in this PR? Is it a priority?
1. Should we show and example of a `<FormField/>` in a `<form/>`?


milkyfruit:
@hepiyellow @shlomitc @wuwa

For situations like this, I think we should add more snippet examples demonstrating available and reasonable property combinations, like size, style, etc.

A good example in this POC story mockup (scroll to bottom to see):
https://wix.invisionapp.com/share/NKN85J759TD#/311364100_Page__Layout_1

shlomitc:
@milkyfruit - thanks for the invision file
I just updated the contribution guide regarding the documentation part.
The major understanding should be that we want to split the component APIs from the design system indexing.

See explanation here:
https://github.com/wix/wix-style-react/blob/master/docs/contribution/DOCUMENTING_COMPONENTS.md

So from the invision file, we should have a much more simplified version, which is more manual, like the examples you have at the bottom.
I think the relevant parts are the `size` and `suffixes`.
Everything that is related to the label, required and info, are part of the `FormField` component, so I'm not sure it should be in the index doc.

what do you think?

hepiyellow:
I agree with Shlomi about having only examples of <Input> in the UX index story.

- I don’t think we should have playground for composites. Only for Components.
- About examples layout. It’s nice to see a few example in a row, but it might be too small. Since the code part might be wider. We can consider having a few example variants in the same Code block.

Unrelated
Just wanted to reference Wix-Code documentation <https://www.wix.com/code/reference/wix-pay.html#startPayment>, for styling inspiration (or Layout)

hepiyellow:
@hepiyellow test

PR #2365, opened 2018-10-23
<DropdownLayout/> - fix scroll behaviour Pt.2

yardnsm:
### What changed

This is another fix for the scroll behaviour of `<DropdownLayout/>`. See https://github.com/wix/wix-style-react/pull/2327.

This is more like commit https://github.com/wix/wix-style-react/pull/2327/commits/9f56aad52b4b8179889cc190c06d4d3b124c17c3 on steroids.

### Why it changed

See:

* https://github.com/wix/wix-style-react/issues/2169
* https://github.com/wix/wix-style-react/pull/2327#issuecomment-431912097
* https://github.com/wix/wix-style-react/pull/2362

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


src/DropdownLayout/DropdownLayout.js
```diff
@@ -121,7 +121,32 @@ class DropdownLayout extends WixComponent {
     } while (!this.isSelectableOption(options[newHovered]));
 
     this.setState({hovered: newHovered});
-    this.options.scrollTop = (newHovered - 2) * 35;
+
+    const menuElement = this.options;
+    const hoveredElement = this.options.childNodes[newHovered];
+
+    const menuRect = menuElement.getBoundingClientRect();
```
hepiyellow: - Is the following logic consistent with scrollingIntoView on `componentDidMount` (when `selectedId` is provided)?
- It seems that in `componentWillReceiveProps` we don't scollIntoView, right?

src/DropdownLayout/DropdownLayout.js
```diff
@@ -121,7 +121,32 @@ class DropdownLayout extends WixComponent {
...
+    const hoveredRect = hoveredElement.getBoundingClientRect();
+
+    const overScroll = hoveredElement.offsetHeight / 3;
+
+    // We'll change the `scrollTop` only when needed. For example, when the
+    // hovered element is in the view, we won't change it.
+    if (hoveredRect.bottom + overScroll > menuRect.bottom) {
+
+      // The hovered element is *below* the view
+      menuElement.scrollTop = Math.min(
```
hepiyellow: This math is difficult to read and understand. A diagram would help?
Or we can sit on it together tomorrow.

hepiyellow:
@yardnsm 
Why didn't you use `scrollIntoView` or the smoothScroll polyfill?

src/DropdownLayout/DropdownLayout.js
yardnsm: 1. Yes.
1. We don't. `hoverNextStep` is being called only on the `keydown` event.
1. I'm not sure, but I think there isn't a significant one.

yardnsm:
@hepiyellow 

* Calling `scrollIntoView` with no arguments will scroll the whole page, not the parent element. We can solve this by calling to `scrollIntoView({block: 'nearest', inline: 'start'})`, but I'm not sure how well its supported across browsers. Each browser could have a different behaviour, and by using our own method we can make sure it'll work the same in all browsers.

* Not using the `smoothscroll` polyfill since I don't want to add more dependencies. Plus, this polyfill is mainly useful for a `smooth` scroll behaviour.

yardnsm:
@milkyfruit @wuwa Can you please take a look at this?

milkyfruit:
@yardnsm @hepiyellow the arrow key scroll works well now, the only concern is that after focused item is selected with 'enter' key down, it resets the focus highlight to first list item instead of leaving it to the selected one.

src/Dropdown/Dropdown.e2e.js
```diff
@@ -56,4 +56,39 @@ describe('Dropdown', () => {
...
+    });
+
+    beforeEach(async () => {
+      driver = dropdownTestkitFactory({dataHook: 'story-dropdown'});
+      await waitForVisibilityOf(driver.element(), 'Cannot find Dropdown');
+      await autoExampleDriver.setProps({options});
+    });
+
+    afterEach(async () => {
+      await autoExampleDriver.reset();
```
hepiyellow: I think arijus asked not to use `reset` any more, but use `remount` instead. Talk to him.

src/Dropdown/Dropdown.e2e.js
```diff
@@ -56,4 +56,39 @@ describe('Dropdown', () => {
...
+    });
+
+    eyes.it('should not change scroll when hovered option is in view', async () => {
```
hepiyellow: These tests are missing, either an expect on `scrollTop`.
Or an accurate description of how the scroll changes exactly,
something like: Item 9 should be visible and it's top should be aligned with the container's top.
Or if there is overScroll, 1/3 of item 8 should be visible and aligned with container top.

This of it as a spec for scroll behavior.

src/DropdownLayout/DropdownLayout.protractor.driver.js
```diff
@@ -1,8 +1,22 @@
...
+
+  const getDropdown = () => component.$(`[data-hook="dropdown-layout-options"]`);
+
+  return {
+    getDropdown,
+    getDropdownItem: index => component.$$(`[data-hook="dropdown-layout-options"] div`).get(index).getText(),
+    getDropdownItemsCount: () => component.$$(`[data-hook="dropdown-layout-options"] div`).getText().count(),
+    selectItemById: itemId => component.$(`[data-hook^="dropdown-item-${itemId}"]`).click(),
+
+    pressArrowDown: (times = 0) => repeat(times, () => browser.actions().sendKeys(protractor.Key.ARROW_DOWN).perform()),
```
hepiyellow: These should not be in a public driver

hepiyellow:
Also there is a failing test with DatePicker.
The new logic, when an initial selected option id is passed, it scrolls so that the option is aligned to top.
It used to show also a little of the previous option.

What does UX think about this? @wuwa @milkyfruit

yardnsm:
### Before the change

screen shot 2018-10-28 at 12 14 02

### After the change

screen shot 2018-10-28 at 12 13 51

wdyt? @wuwa @milkyfruit

wuwa:
i think it's better to show the previous option.

PR #2374, opened 2018-10-25
WIP - Migrate to Yoshi 3

hepiyellow:
# Issues

The issues which have a ~~strikethrough~~ are resolved or differed.

### ~~Deprecated `addStoriesToBaseGlob` yoshi feature~~

```
Warning: Invalid configuration object. Yoshi has been initialised using a configuration object that does not match the API schema.
 - configuration/features has an unknown property 'addStoriesToBaseGlob'. These properties are valid:
   object { externalizeRelativeLodash? }
   Field Description: Features of your project
```

I removed the settings.

### ~~import/no-unresolved~~
Getting this for stories using raw-loader:
```
/Users/erezm/Projects/wix-style-react/stories/DragAndDrop/SortableList/SingleAreaList/index.js
  8:31  error  Unable to resolve path to module '!raw-loader!./SingleAreaList'       import/no-unresolved
```
I disabled this `eslint` rule for now. 
Generaly, eslint can be configure to use webpack resolver so it should resolve raw-loader.
See [this](https://github.com/benmosher/eslint-plugin-import/issues/799)
But I suspect yoshi web-pack config does not have a raw-loader.
Maybe we configure it only for storybook.

This is one of the reasons I think the storybook should be a separate package in a WSR mono-repo. So they could have separate build / lint settings.

### ~~`react/forbid-foreign-prop-types`~~
I disabled this rule, we use it all the time.

### ~~144 Lint Warnings~~
Should we handle them now?

### ~~``  localhost pre-pended~~
- When rendering with `<Comp src="image.png"/>`, then expecting on `element.src` tha ctual value gets a prefix `http://localhost/image.png`. This is probably due to new `jsdom` version. But I didn't verify it. 
- I added the `http://localhost` prefix to all expected values.

### ~~Tooltip - can not accees property `tooptip` of null~~
See more info in separate comment below

### ~~TypeError: window.google.maps.places.AutocompleteService is not a constructor~~
- [x] Replace class mock functions (`()=> instance`) with real classes:
    ```
    class XXX {
       constructor() { return instance; }
    }
    ```
### ~~ImageViewer.spec - Tooltip ~~
Geeting this error:
```js
NotFoundError: The object can not be found here.

      450 |       const container = this._getContainer();
      451 |       if (container) {
    > 452 |         container.removeChild(this._mountNode);
          |                   ^
      453 |         container.removeEventListener('scroll', this._containerScrollHandler);
      454 |       }
      455 |       this._mountNode = null;
```

If I run only the `add image` test, is passes.
But if I run both `add image` and `update image` then the second fails.

It seems that reseting the `body` might not be working the same as before (maybe since jsdom 11.12.x) 
```js
beforeEach(() => {
      document.body.innerHTML = '';
    });
```
**Resolved by using dom-test0-cintainer (unmounting after each test)**


hepiyellow:
## Tooltip issue
The `ImageViewer.spec.js` test fails.
- Tooltip.js : 264: ref.tooltip  - Cannot read property 'tooltip' of null
- If I comment out in Tooltip.driver line 71 `mouseLeave()`, then it passes.
```js
ref={ref => {
            if (this.props.relative) {
              this.tooltipContent = ref.tooltip;
            } else {
              this.tooltipContent = ref;
            }
          }}
```
React calls the ref callback with null. It is natural.

hepiyellow:
Replaced by #2440

PR #1874, opened 2018-06-22
WIP - Proxy new TimePicker component from BackOffice.

alexgoncharwix:
Proxy TimePicker from BackOffice.
Add a story for TimePicker.

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green
- [ ] Change of UX is approved by [wuwa](https://github.com/wuwa) or [milkyfruit](https://github.com/milkyfruit)

PR #2376, opened 2018-10-27
Updated moment due to security issue

rghorbani:
### What changed

Updated moment dependency.

### Why it changed

Due to [Regular Expression Denial of Service](https://nodesecurity.io/advisories/532) issue.

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green


lbelinsk:
Thanks!
@shlomitc Since TimeInput is the only component that which uses moment, we should push forward the process of deprecating TimeInput in favor of the core TimeInputComponent with the bo theme.
That way, we will be able to remove moment dep entirely

argshook:
there was already effort from guildweek to do so, but it got rusty and not in priority, would be good to revive https://github.com/wix/wix-style-react/pull/1874

Issue #2375, opened 2018-10-26
<FormField/> - asterisk font is missing in the component


shlomitc:
# Bug Report

### Relevant Components

`<FormField/>`

### Current Behavior

The required asterisk is displayed differently than in specs.
![image](https://user-images.githubusercontent.com/6093192/47576386-02f32b00-d94d-11e8-8ec9-b973259d79c4.png)

### Expected Behavior

![image](https://user-images.githubusercontent.com/6093192/47576282-c1628000-d94c-11e8-91ec-ae703939847e.png)

### Steps to Reproduce
Just create a new project or remove the CSS fonts from the `body` in the storybook
![image](https://user-images.githubusercontent.com/6093192/47576436-27e79e00-d94d-11e8-8c2c-483d7ad7aceb.png)

### Severity

Critical / Major / Low

shlomitc:
done

PR #2377, opened 2018-10-27
remove the global fonts

shlomitc:
Following this issue https://github.com/wix/wix-style-react/issues/2375
will help identify what is wrong


hepiyellow:
I think you can merge this

shlomitc:
almost 50% of the tests fail, so no easy merge here.
looks like the autodocs components need to apply the fonts directly

PR #2381, opened 2018-10-29
<Dropdown/> - added live examples to story

yardnsm:
This PR adds a `<LiveCodeExample/>` to the `<Drodpown/>` story. Please wait for https://github.com/wix/wix-ui/pull/835 to be merged as an overflow issue were discovered when working on this PR.


stories/utils/Components/LiveCodeExample/index.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import LiveCodeExample from 'wix-storybook-utils/LiveCodeExample';
+
+import * as wsrScope from '../../../../src/index';
```
shlomitc: This might be problematic.
I'm not sure that the way components are exported from the index file is not the same as importing them from WSR.

We can with this solution and see what are the differences

shlomitc:
looks great to me besides the minor thing regarding importing index file.
However, I think we can merge it anyway. good job

stories/utils/Components/LiveCodeExample/index.js
yardnsm: You're right, there may be components that are not available in the index file (such as [FormField](https://github.com/wix/wix-style-react/pull/2381/files#diff-1fdf421c05c1140f6d71444ea2b27638R63)), but I couldn't find a better option.

PR #2396, opened 2018-10-31
[WIP] - migrate component-test Heading driver to Unidriver syntax

lbelinsk:
This pr is an example of how to migrate a component to use the unidriver pattern.
The thing is that this is a breaking change since every driver method is now async, and our consumers are using the methods today in a sync way.


lbelinsk:
Closing this pr because of even if the unidriver of a specific component is not changing the API, it is becoming async and therefore it must be considered as a breaking change. This means that our consumers will be required to do many changes in their tests.

One approach we can take is that only for new components we will add unidriver.

PR #2403, opened 2018-11-04
Calendar: added selectedDays and onChangeSelectedDays prop (value and…

malsomnus:
Calendar:
* Added `selectedDays` prop (`value` prop deprecated)
* Added `onSelectedDaysChange` prop (`onChange` prop deprecated)
* Added styling for selecting a date range
* Added a spec file

## Fixes to do together with Tal
- [ ] css extend instead of mixins

## Open Issues
- Should we support value as string ?

## TODO
- add componentWillReceive prop with nearest range edge logic


src/Calendar/Calendar.scss
```diff
@@ -99,11 +99,88 @@ $hoverBoxShadow: 0 0 0 $boxShadowWidth $F00;
...
 
-    &--outside,
     &--disabled {
       pointer-events: none;
       color: $D55;
     }
+
+    &.DayPicker-Day--selected:not(.DayPicker-Day--start):not(.DayPicker-Day--end):not(.DayPicker-Day--outside) {
+      background-color: $B50 !important;
+      border-radius: 0px !important;
```
argshook: nono sorry i can't allow this, i can't sleep well after seeing `important`s

are we certain there's no other way to achieve desired result? selectors could also be not as spooky

argshook:
sorry to be late to the party but could you tl;dr; me of why `value` and `onChange` must be deprecated? both props seem valid to me

src/Calendar/Calendar.scss
malsomnus: http://react-day-picker.js.org/examples/selected-range/
This is the reference I'm going by. Since this whole thing is pretty much a hack, just overriding the CSS of a third-party package, I'm not at all sure better options exist. This applies both to the `!important` and the horrifying selectors (first time I looked at those I didn't even realize this was legal CSS).

src/Calendar/Calendar.driver.js
```diff
@@ -44,6 +46,7 @@ const calendarDriverFactory = ({element, wrapper}) => {
     pressLeftArrow: () => ReactTestUtils.Simulate.keyDown(getFocusedDay(), {key: 'ArrowLeft', keyCode: 37}),
     pressRightArrow: () => ReactTestUtils.Simulate.keyDown(getFocusedDay(), {key: 'ArrowRight', keyCode: 39}),
     getSelectedDay: () => getSelectedDay().textContent,
+    getSelectedDays: () => map(getSelectedDays(), 'textContent'),
```
argshook: please use `Array.prototype.map` with `Array.from` instead of lodash dep.

```suggestion
    getSelectedDays: () => Array.from(getSelectedDays()).map(({textContent}) => textContent)
```

src/Calendar/Calendar.scss
malsomnus: Hmmm, I didn't know this was a feature.
It sounds like it might help a bit (I might have to write JS that is more complex than this CSS), but I'm also forced to use a bunch of pseudo-selectors so it might be impossible to replace all of it.
I'll try to remove the `!important` if possible.

argshook: please see if it's possible. the less crazy selectors we have the better :)

src/Calendar/Calendar.protractor.driver.js
```diff
@@ -14,12 +14,14 @@ const calendarDriverFactory = component => {
...
   const getCalendarWrapper = () => component.$('.DayPicker-wrapper');
   const getKeyboardSelectedDay = () => component.$('.DayPicker-Day:focus');
+  const getSelectedDays = () => component.$$('[role="gridcell"][aria-selected=true]');
 
   return {
     getElement: () => getCalendar(),
     exists: () => getCalendar().isPresent(),
     isYearInViewPort: year => browser.wait(EC.visibilityOf(getYear(year)), 5000),
     isVisible: () => getCalendar().isDisplayed(),
+    getSelectedDays: () => getSelectedDays().map(elem => elem.getText()),
```
hepiyellow: Add comment description, that this returns an array of day string number '1', '2',... in the currently displayed months

malsomnus: Errr... what private driver?

hepiyellow:
@argshook 
> sorry to be late to the party but could you tl;dr; me of why `value` and `onChange` must be deprecated? both props seem valid to me

Its a simple renaming:
`value` -> `selectedDays`
`onChange` -> `onSelectedDaysChange`

Calendar has selectedDays that can change, also month can change.

The will be new props:
`month`
`onMonthChanged`

I thought `value` and `onChange` are ambiguous. And it's better to be explicit.

src/Calendar/Calendar.scss
hepiyellow: I think this is going to be fixed, once we'll use the `dayRender` function, so we can render the days ourselves.

argshook:
@hepiyellow 

> Its a simple renaming:
> value -> selectedDays
> onChange -> onSelectedDaysChange

all user input components have `value` and `onChange`, it's a matter of consistency. I wouldn't change this. `<input type="radio"/>` has `onChange`, not `onRadioChanged`, so why do this for `Calendar`? As a consumer i wouldn't be happy about this change, it would require me to go and change my code for no real reason.

what will `onMonthChange` do? be called when user selects a month from dropdown? or be called when only month in value changes?

not sure about `month` though, is it going to be a setter? isn't `value` enough?

i'm probably "that guy" again, but as of now none of these changes seem valid to me. what do i miss?

src/Calendar/Calendar.driver.js
hepiyellow: I agree

hepiyellow: @argshook BW are there any lodash helpers that you DO think are worth using?
Can you make a list? Add it to the Contribution guide.

hepiyellow:
@argshook 
I don't see `<Calendar/>` as an classic "input", that needs to adhere to the 'value/onChange` paradigm.
I agree that this renaming is NOT compulsory, but I feel that it brings semantic clarity.

The `month/onMonthChange` is about the user navigating through the calendar months (without selecting days).

src/Calendar/Calendar.driver.js
argshook: i don't think we need to blacklist anything, what will we do in reviews then? :)

lodash does have useful things, like `zip`, `difference`, `intersection`, `debounce`, `throttle`, `uniq`, `flow` even `pick`, `omit`. it has many tools but most of the time i see `map`, `filter` or something like `isUndefined` being used.

argshook:
> @argshook
> I don't see `<Calendar/>` as an classic "input", that needs to adhere to the `value/onChange` paradigm.

hmm, why not? Consider a form with such fields:

`<Input/>`
`<FileInput/>`
`<TimePicker/>`
`<Calendar/>`
`<ToggleSwitch/>`
`<Checkbox/>`

everyone has `value` and `onChange`, except for `Calendar` which has
`selectedDays` and `onSelectedDaysChange`. why make it special?
Consistency is important.

> I agree that this renaming is NOT compulsory, but I feel that it brings semantic clarity which is worth while

where's semantic clarity in here? Is it not clear enough what would `onChange` do in a `Calendar` component?

> (+ It corresponds to the API of `react-day-picker` which we are using under the hood).

and what if tomorrow it will be something other than `react-day-picker`? will we change prop name again to adhere?

---

> The `month/onMonthChange` is about the user navigating through the calendar months (without selecting days).

so what does that mean? i assume `onMonthChange` will be called when
user changes month in dropdown. Now, how is that useful if `value` will
remain the same?

what will i give to `month` prop? a specific month? what if `value`
will have January but `month` will have February, what will be resolved?

---

i'm raising many questions here (and still have more). my point is that
if i need answers to many questions, then i'd say API design is flawed.
And for this reason `onChange` and `value` seem perfectly fine and not
worth changing.

hepiyellow:
@argshook 

- `<DatePicker/>` is in the input family, and forms would include that and not Calendar directly.
- What else can change. `month` for example.
- I mentioned react-day-picker as an example of someone else that thought that `selectedDays` is better than `value`.
- again. month is the currently displayed month. It could be that the selected day is not on the currently displayed month.

argshook:
Disagree, but i'll leave it here

src/Calendar/Calendar.scss
```diff
@@ -99,11 +119,83 @@ $hoverBoxShadow: 0 0 0 $boxShadowWidth $F00;
...
+        height: 100%;
+        top: 0;
+        right: 0;
+        text-align: center;
+        line-height: 36px;
+      }
+
+      .startBackground {
+        position: absolute;
+        background: linear-gradient(to left, $B50 0%,$B50 50%,#000000 50%,transparent 50%,transparent 100%); /* W3C */
```
hepiyellow: Where did you get this gradient from ?
I don't see it in the specs.

src/Calendar/Calendar.driver.js
```diff
@@ -28,6 +29,7 @@ const calendarDriverFactory = ({element, wrapper}) => {
     getCurrentMonthWithYear: () => getMonthAndYear() ? getMonthAndYear().map(element => element.textContent).join(' ') : '',
     getNthWeekDayName: (n = 0) => getNthWeekDayName(n) ? getNthWeekDayName(n).textContent : '',
     clickOnNthDay: (n = 0) => getNthDay(n) && ReactTestUtils.Simulate.click(getNthDay(n)),
+    clickOnNthDayOfTheMonth: (n = 0) => getNthDayOfTheMonth(n) && ReactTestUtils.Simulate.click(getNthDayOfTheMonth(n)),
```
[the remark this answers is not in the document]
malsomnus: I strongly object to redundant documentation. The function name in your example tells the reader clearly and unambiguously what the function does, and in the multi-month case I'm not sure we know what it should be doing yet either.

src/Calendar/Calendar.js
```diff
@@ -62,7 +63,7 @@ export default class Calendar extends WixComponent {
     } = props;
 
     const {from, to} = selectedDays || {};
-    return from || to || selectedDays || value;
+    return from || to || selectedDays || parse(value || new Date());
```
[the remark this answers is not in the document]
malsomnus: Good question about {}. Should we even allow it?

src/Calendar/Calendar.js
```diff
@@ -75,13 +122,28 @@ export default class Calendar extends WixComponent {
         />
     );
 
+    function renderDay(day, modifiers) {
+      const relevantModifiers = ['start', 'end', 'selected'];
+      for (const modifier of relevantModifiers) {
+        if (modifier in modifiers) {
+          return (
+            <div className={`${modifier}Background`}>
```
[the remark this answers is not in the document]
malsomnus: Not sure, I'll try it out.

src/Calendar/Calendar.scss
```diff
@@ -99,11 +119,83 @@ $hoverBoxShadow: 0 0 0 $boxShadowWidth $F00;
...
+        border-radius: 50% !important;
+        color: #FFFFFF;
+        z-index: 2;
+        position: absolute;
+        width: 100%;
+        height: 100%;
+        top: 0;
+        right: 0;
+        text-align: center;
+        line-height: 36px;
```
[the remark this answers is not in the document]
malsomnus: The text won't be vertically centered otherwise.

src/Calendar/Calendar.scss
malsomnus: This gradient is just the easy way to implement an element with half blue and half transparent background. Not an actual gradient.

src/Calendar/Calendar.scss
malsomnus: In a nutshell, at some point you have to compromise. I can either make the DOM heavier and hurt the users, or I can have some spooky selectors and `!important`s. I improved what I could, but am going to lean towards the latter.

src/Calendar/Calendar.js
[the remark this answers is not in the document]
malsomnus: This should be a component test, not e2e, right?

src/Calendar/Calendar.scss
```diff
@@ -99,11 +119,83 @@ $hoverBoxShadow: 0 0 0 $boxShadowWidth $F00;
...
+    }
+  
+    &.DayPicker-Day--selected.DayPicker-Day--start:not(.DayPicker-Day--outside),
+    &.DayPicker-Day--selected.DayPicker-Day--end:not(.DayPicker-Day--outside) {
+
+      .circle {
+        background-color: #3899ec !important;
+        border-radius: 50% !important;
+        color: #FFFFFF;
+        z-index: 2;
```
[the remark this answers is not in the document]
malsomnus: You are correct!

src/Calendar/Calendar.scss
hepiyellow: Oh

src/Calendar/Calendar.scss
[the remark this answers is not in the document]
hepiyellow: Well, you can either do vertical align (flex, and alignItems).
Or add `margin-top, margin-bottom = 9px` (like in the specs)
I think it's more elegant.
In fact, I'll add another comment on changing the font styling to use the new typography (which has line-height 18, and shouldn't be overwritten)

src/Calendar/Calendar.js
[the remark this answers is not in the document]
hepiyellow: component test indeed.
I think we should allow it.

src/Calendar/Calendar.driver.js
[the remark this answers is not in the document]
hepiyellow: - For consistency, we add documentation description for EVERY driver method.
- Multi month view is a very good question. I wouldn't ask you to fix old driver methods, but at least if you add a method, make it support multi-month.

One way I can think of supporting multi-month in an easy and clear way,
Is to have a function `clickDay(Date)`. it receives a date object.
If the date is visible (in any of the currently visible months) it will be clicked.
If not, nothing will happen.

hepiyellow:
@argshook We decided to revert, and leave it with `value/onChange` for the sake of NOT breaking consumers for no good enough reason. Thanks for being stubborn !

src/Calendar/Calendar.driver.js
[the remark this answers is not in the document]
malsomnus: I must be misunderstanding you, because as far as I can tell there is not a single driver method in all of WSR that is documented -.o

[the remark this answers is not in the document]
argshook: here's some context

in wix-style-react and wix-ui we try to automate our documentation. this
includes testkits. the approach that stood for a long time was to simply
document stuff manually. as in, open markdown file and write about each
method.

instead of manual work what we do now is use tools that read source code
and generate documentation pages (stuff you see in storybook) automatically.

in storybook you should see a table of all driver methods.
if there's a jsdoc type of comment (`/** comment with two stars in front */`) above
driver method, it will appear in aforementioned table.

now, this doesn't mean you need to write these comments for each and
every method. If there's anything to explain, please do so. but if
method name is informative enough, there's no need to duplicate content.

src/Calendar/Calendar.js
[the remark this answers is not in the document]
malsomnus: Something occurred to me while writing the tests... why exactly did the old code use `parse` if `props.value` was defined as an object and not a string?

src/Calendar/Calendar.scss
```diff
@@ -99,11 +119,83 @@ $hoverBoxShadow: 0 0 0 $boxShadowWidth $F00;
...
     }
+
+    &.DayPicker-Day--selected:not(.DayPicker-Day--start):not(.DayPicker-Day--end):not(.DayPicker-Day--outside) {
+      background-color: $B50 !important;
+      border-radius: 0px !important;
+      box-shadow: none;
+      color: $D10;
+    }
+    &.DayPicker-Day--selected:not(.DayPicker-Day--start):not(.DayPicker-Day--end):not(.DayPicker-Day--outside):last-of-type {
+      border-top-right-radius: 50% !important;
```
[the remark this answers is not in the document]
malsomnus: Tried, definitely doesn't work.

src/Calendar/Calendar.scss
```diff
@@ -99,11 +119,83 @@ $hoverBoxShadow: 0 0 0 $boxShadowWidth $F00;
...
+      border-radius: 50% !important;
+    }
+  
+    &.DayPicker-Day--selected.DayPicker-Day--start:not(.DayPicker-Day--outside),
+    &.DayPicker-Day--selected.DayPicker-Day--end:not(.DayPicker-Day--outside) {
+
+      .circle {
+        background-color: #3899ec !important;
+        border-radius: 50% !important;
+        color: #FFFFFF;
```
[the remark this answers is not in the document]
malsomnus: I used D80, actually. Is that alright?

src/Calendar/Calendar.scss
```diff
@@ -99,11 +119,83 @@ $hoverBoxShadow: 0 0 0 $boxShadowWidth $F00;
...
+    &.DayPicker-Day--firstOfMonth:not(.DayPicker-Day--start):not(.DayPicker-Day--end):not(.DayPicker-Day--outside) {
+      border-top-left-radius: 50% !important;
+      border-bottom-left-radius: 50% !important;
+    }
+    &.DayPicker-Day--lastOfMonth:not(.DayPicker-Day--start):not(.DayPicker-Day--end):not(.DayPicker-Day--outside) {
+      border-top-right-radius: 50% !important;
+      border-bottom-right-radius: 50% !important;
+    }
+
+    &.DayPicker-Day--singleDay:not(.DayPicker-Day--start):not(.DayPicker-Day--end):not(.DayPicker-Day--outside) {
```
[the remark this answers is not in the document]
malsomnus: Why not? Even if I didn't let a user select a range that starts and ends on the same day (which I do, because it seems like the most user-friendly way to go about it), there's nothing stopping a user from just providing that as a prop. I could handle this on the JS side instead of the CSS side, but I think that this is the easiest way to do it.

src/Calendar/Calendar.js
```diff
@@ -127,9 +126,7 @@ export default class Calendar extends WixComponent {
       for (const modifier of relevantModifiers) {
         if (modifier in modifiers) {
           return (
-            <div className={`${modifier}Background`}>
-              <div className={'circle'}>{day.getDate()}</div>
-            </div>
+            <div className={`circle ${t.text} ${t.sizeSmall} ${t.weightThin} ${modifiers.start || modifiers.end ? t.light : ''}`}>{day.getDate()}</div>
```
hepiyellow: Please use `classnames` package.
```js

<div className={classNames(
  styles.circle,
  t.text,
  t.sizeSmall,
  t.weightThin,
  t.light: modifiers.start || modifiers.end
)}>
  {day.getDate()}
</div>
```

src/Calendar/Calendar.js
```diff
@@ -220,17 +217,8 @@ Calendar.propTypes = {
...
-    if (props[propName]) {
-      const msg = 'Calendar\'s onChange prop is deprecated, please use onSelectedDaysChange instead.';
-      const key = msg;
-      deprecationLog(msg, key);
-    }
-  },
-
-  /** Callback function called whenever the user selects a day in the calendar */
-  onSelectedDaysChange: PropTypes.func,
+   /** Callback function called whenever the user selects a day in the calendar */
```
hepiyellow: We should add documentation on `onChange` that it can be a range.

src/Calendar/Calendar.js
```diff
@@ -244,17 +232,8 @@ Calendar.propTypes = {
...
-      const msg = 'Calendar\'s onValue prop is deprecated, please use selectedDays instead.';
-      const key = msg;
-      deprecationLog(msg, key);
-    }
-  },
-
-  /** The selected date range. Can be either a single JS Date object, or a single range object {from: Date, to: Date} */
-  selectedDays: PropTypes.object,
+  /** The selected date */
+  value: PropTypes.oneOfType([PropTypes.object, PropTypes.string]),
```
hepiyellow: Add
```js
 value : PropTypes.oneOfType([PropTypes.object, PropTypes.string, PropTypes.shape({
   from: PropTypes.oneOfType([PropTypes.object, PropTypes.string]),
   to: PropTypes.oneOfType([PropTypes.object, PropTypes.string])
})
]);
```

src/Calendar/Calendar.js
```diff
@@ -32,38 +32,40 @@ export default class Calendar extends WixComponent {
     };
   }
 
-  // TODO: Change to getDerivedStateFromProps with React ^16.0.0
```
hepiyellow: What if the consumer re-renders with different `value`?

src/Calendar/Calendar.scss
```diff
@@ -19,10 +20,7 @@ $hoverBoxShadow: 0 0 0 $boxShadowWidth $F00;
 
 .calendar :global .DayPicker {
   width: $calendarWidth;
-  //@include FontLight;
-  //font-size: 14px;
-  //line-height: 1.29;
-  color: $D40;
+  @include Text($size: small, $weight: thin);
```
hepiyellow: Actually, we'd rather use `@extend` than `@include`.
See [this](https://github.com/wix/wix-style-react/blob/master/stories/Typography/BasicClassesSassExample.scss)

It also appears in the [Storybook Typography](https://wix-wix-style-react.surge.sh/?selectedKind=Styling&selectedStory=1.2%20Typography%20Classes&full=0&addons=0&stories=1&panelRight=0)

malsomnus: I sat with someone for about an hour and a half on this particular thing, we read through SCSS references and looked for examples in the existing code, and together we completely failed to figure it out, so I'm not sure what exactly I should be doing here.

src/Calendar/Calendar.js
malsomnus: If a new `value` changes the displayed month, then selecting a range that extends into another month will by definition switch the display back to the first month, which is the behaviour I fixed here. If the consumer wants to re-render with a different `value` and wants to display a different month (for some reason that I can't imagine) then he'll just have to also provide the correct `month` prop...

src/Calendar/Calendar.js
malsomnus: Isn't that redundant when `PropTypes.object` is on the list? Also does this syntax capture the case of having only `from` or only `to`?

src/Calendar/Calendar.scss
```diff
@@ -93,17 +106,70 @@ $hoverBoxShadow: 0 0 0 $boxShadowWidth $F00;
 
     &--selected {
       background-color: $B10;
-      color: $D80;
+      @include Text($size: small, $weight: thin, $light: true);
       &:hover {
         background-color: $B20;
```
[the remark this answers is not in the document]
malsomnus: What exactly isn't working?

src/Calendar/Calendar.spec.js
```diff
@@ -0,0 +1,234 @@
...
+            selectionMode={'range'}
+            />
+        );
+
+        driver.clickOnNthDay(2);
+        expect(onChange.mock.calls.length).toEqual(1);
+        expect(onChange.mock.calls[0][0].from.getDate()).toEqual(1);
+        expect(onChange.mock.calls[0][0].to.getDate()).toEqual(3);
+      });
+      it(`should call onChange({from: $clickedDay, to: $from}) if the clicked day is earlier than the provided 'from'`, () => {
```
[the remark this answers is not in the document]
malsomnus: Actually I got the test wrong, it should have had `from` instead of `to`. 
It works both ways, of course.

src/Calendar/Calendar.js
```diff
@@ -104,13 +122,28 @@ export default class Calendar extends WixComponent {
         />
     );
 
+    function renderDay(day, modifiers) {
+      const relevantModifiers = ['start', 'end', 'selected'];
+      for (const modifier of relevantModifiers) {
+        if (modifier in modifiers) {
+          return (
+            <div className={`${modifier}Background`}>
```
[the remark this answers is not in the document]
malsomnus: I added `>` and made the class name more specific instead.

src/Calendar/Calendar.js
hepiyellow: Well, the use-case where a consumer may re-render with different value and month is exactly what the [`<CalendarPanel/>`](https://github.com/wix/wix-style-react/blob/master/src/CalendarPanel/README-SPEC.md) is all about. Presets.

And as you see in the spec for `<CalendarPanel/>` it uses a new `month` prop.

But since we also support having the `month` uncontrolled (for backward compatibility), we need a FULL solution for that. So we need to bring back the `componentWillReceiveProps`, and solve the range selection issue in a different way (like adding state, which sucks, but seems like the only way)

src/Calendar/Calendar.driver.js
```diff
@@ -26,6 +27,7 @@ const calendarDriverFactory = ({element, wrapper}) => {
...
```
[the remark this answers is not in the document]
hepiyellow: But it will be this week I hope, and I don't want to add a driver method which is not forward-compatible.

src/Calendar/Calendar.scss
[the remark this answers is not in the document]
hepiyellow: We went over it. If you have a pending commit that fixes it then push it.

src/Calendar/Calendar.js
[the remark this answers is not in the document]
hepiyellow: Renamed dayCircle and keep it global

src/Calendar/Calendar.scss
```diff
@@ -66,10 +66,30 @@ $hoverBoxShadow: 0 0 0 $boxShadowWidth $F00;
...
+
+      &:after {
+        position: absolute;
+        top: 0px;
+        bottom: 0px;
+        left: 0px;
+        right: 0px;
+        border-radius: 50%;
+        //border: 3px solid #aadbfc;
+        box-shadow: 0 0 0 3px #aabdfc;
```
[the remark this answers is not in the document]
hepiyellow: done

src/Calendar/Calendar.scss
```diff
@@ -99,11 +119,83 @@ $hoverBoxShadow: 0 0 0 $boxShadowWidth $F00;
...
     }
 
-    &--outside,
     &--disabled {
       pointer-events: none;
       color: $D55;
     }
+
+    &.DayPicker-Day--selected:not(.DayPicker-Day--start):not(.DayPicker-Day--end):not(.DayPicker-Day--outside) {
+      background-color: $B50 !important;
```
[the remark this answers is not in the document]
hepiyellow: It is needed

src/Calendar/Calendar.scss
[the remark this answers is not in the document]
hepiyellow: explained

src/Calendar/Calendar.scss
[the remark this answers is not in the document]
hepiyellow: Not possible to extend because of the global rule

src/Calendar/Calendar.scss
hepiyellow: EXTENDING IS PROBLEMATIC BECAUSE OF `:global` operator

src/Calendar/Calendar.js
```diff
@@ -55,8 +92,29 @@ export default class Calendar extends WixComponent {
       twoMonths
     } = this.props;
 
-    const month = this.state.month || propsValue || new Date();
+    const month = this.state.month || this._getMonth(this.props) || new Date();
     const localeUtils = localeUtilsFactory(locale);
+    const from = propsValue && propsValue.from && parse(propsValue.from);
+    const to = propsValue && propsValue.to && parse(propsValue.to);
+    const singleDay = !from && !to && propsValue;
```
hepiyellow: Should call `parse(propsValue)`

src/Calendar/Calendar.js
```diff
@@ -55,8 +92,29 @@ export default class Calendar extends WixComponent {
...
+      date.setDate(from.getDate() - 1);
+      selectedDaysProp = {after: parse(date)};
+    } else if (!from && to) {
+      const date = new Date(to);
+      date.setDate(to.getDate() + 1);
+      selectedDaysProp = {before: parse(date)};
+    } else if (from && to) {
+      selectedDaysProp = {from: parse(from), to: parse(to)};
+    } else {
+      selectedDaysProp = parse(propsValue);
```
hepiyellow: should use `selectedDaysProp = singleDay`;

src/Calendar/Calendar.js
```diff
@@ -27,22 +27,46 @@ export default class Calendar extends WixComponent {
...
+      onChange(value, modifiers);
+      shouldCloseOnSelect && this.props.onClose();
+    }
   };
 
+  _getMonth = props => {
+    const {value} = props;
+
+    const {from, to} = value || {};
+    if (!from && !to && !(value instanceof Date)) {
```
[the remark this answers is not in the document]
hepiyellow: We'll support it.

hepiyellow:
Replaced by #2533

PR #2406, opened 2018-11-04
enrich index file

lbelinsk:


argshook:
i thought we were going to remove this file entry file? especially that we have eslint rule not to import from this file

PR #2411, opened 2018-11-05
Upgrade <MessageBoxMarketerialLayout/> typography

mykas:
This PR forces `<MessageBoxMarketerialLayout/>` to use `<Heading>` and `<Text>` with upgraded typography. It also removes font face from button wrappers as buttons already comes with typography in mind.


mykas:
@milkyfruit

shlomitc:
@mykas @milkyfruit - I'm looking in the storybook and in the Zeplin after the fixes, and the logo alignment looks to me wrong:

Zeplin:
image

Storybook:
image

Did you review this fix as well?

shlomitc:
I don’t think so. This is the only definition we have which I know about

PR #2418, opened 2018-11-06
depracationLog() - issues with `key` values. 

mykas:
This PR fixes deprecationLog where it gets stuck if dev will not pass second parameter. In mentioned case key value is undefined and it gets saved in reportedKeys SET. This way all deprecationLogs are not logging anymore. 

I think we should save messages in SET instead of keys. What do you think @hepiyellow @shlomitc @argshook ?


src/utils/deprecationLog.js
```diff
@@ -1,27 +1,27 @@
 /* eslint-disable no-console */
-let deprecationLog = function () {};
+let deprecationLog = function () {
+  console.log('aha');
```
argshook: :)

mykas: :)

argshook:
my assumption is that this was done in order not to spam users with repetitive messages and print unique message only once. but implementation with `key` seems mistaken, as you point out, if it's not set, it will stop any other messages to be printed.

i agree with getting rid of `key`, not only it needs additional check for undefined but also requires user of `deprecationLog` to know about `key` and think of a name for it

mykas:
True, I will go through them.

mykas:
Looks like only you have used it in DataTable 😄 All others were not using it.

yardnsm:
There are also tests for the `key` arguments:

https://github.com/wix/wix-style-react/blob/80104c736e0871688e410a89ffb2b1459c70c0f9/src/utils/deprecationLog.spec.js#L18-L25

And in the Typography:

https://github.com/wix/wix-style-react/blob/80104c736e0871688e410a89ffb2b1459c70c0f9/src/Typography/index.js#L15-L18

mykas:
Thanks, fixed

hepiyellow:
@mykas @argshook @yardnsm 
ok guys. sorry for the mess.
My bug, I should have not put `undefined` key in the Set.

The idea of this `key` is some old pattern I remembered with generic logging.
Sometimes there is a log "type" , that in the message it has variables.
For example:
"This prop accepts 'red', 'blue', 'green', but you have passed 'hoochimama`"
Having the `hoochimama` in the log might be very useful.

So having a log "type" (key) can help filtering by it.

It might not be so relevant for deprecationLogs.

We'll add it if we need it in the future.

hepiyellow:
@mykas 
I thought that disabling multiple logging of a certain deprecationLog is the decision of the caller.
The idea was, that if `key` is NOT passed, then the log can repeat itself.

In your fix, you made it always filter out repeated messages.

## What do we want?
I guess that having only 1 deprecationLog is enough, we don't need to litter the console.

Type of deprecationLogs:
- deprecated component
- deprecated prop
- deprecated prop value
- deprecated css class (consumed though JS)

src/utils/deprecationLog.spec.js
```diff
@@ -12,16 +12,17 @@ describe('deprecationLog', () => {
 
   it('should log warning in dev mode', () => {
```
hepiyellow: Change test title to `should log the appropriate message content`

src/utils/deprecationLog.spec.js
```diff
@@ -12,16 +12,17 @@ describe('deprecationLog', () => {
...
     deprecationLog('Some message');
-    expect(global.console.warn).toBeCalledWith('Wix-Style-React: [WARNING] Some message');
+    expect(global.console.warn).toBeCalledWith(
+      'Wix-Style-React: [WARNING] Some message'
+    );
   });
 
   it('should log only once per class', () => {
```
hepiyellow: Change test description to `should log only once per message`

PR #2419, opened 2018-11-06
remove fonts from main index file

lbelinsk:
Should this be considered a breaking change?


argshook:
i think not. there's still eslint rule preventing from importing `index.js` and together fonts.

we can probably remove all the fonts in `assets` folder too

lbelinsk:
in `assets/` dir we have:
1. `fonts` dir
2. `helvetica.scss` with a lot of font-faces.
3. powered_by_google png

@argshook do you mean only to remove the `fonts` dir? I see that the png is being used only by GoogleAddressInput, so we should probably move it to sit near the component.

What about the Helvetica's file?

Issue #2369, opened 2018-10-24
<Table/> Sort arrows should have styles for when they are sorted by


Wo1v3r:
# Bug Report

### Relevant Components
<Table/>

### Current Behavior
Sort arrows are either rendered or not, with a callback for click and a state (up or down)

### Expected Behavior
- Sort arrows should have different style for when currently sorting by that column
- Sort arrows should display after clicking on the column header

Explained further on this ticket:
https://jira.wixpress.com/secure/RapidBoard.jspa?rapidView=699&projectKey=INV&view=planning&selectedIssue=INV-1738

### Severity
- [ ] Critical
- [ ] Major
- [ ] Low

yardnsm:
It's up to you to decide whether to render the sort arrow (and in what direction).

By setting the value of `sortDescending` for a specific column you can set the behaviour.

#### `sortDescending: false`

![screen shot 2018-11-01 at 17 40 30](https://user-images.githubusercontent.com/11786506/47862243-5b1ea700-ddfd-11e8-9490-b42fa4642e3a.png)

#### `sortDescending: true`

![screen shot 2018-11-01 at 17 40 35](https://user-images.githubusercontent.com/11786506/47862257-62de4b80-ddfd-11e8-9f40-90ec96f88919.png)

#### `sortDescending: undefined`

![screen shot 2018-11-01 at 17 40 42](https://user-images.githubusercontent.com/11786506/47862264-696cc300-ddfd-11e8-9573-eb5f49a333c0.png)

---

Yeah, the documentation for this is not great (is actually doesn't exist for passing `undefined`), so I think there's a room for improvement. Also, the name `sortDescending` is misleading.

I think we should deprecate the `sortDescending` property and instead have a `sortArrow` / `sortDirection` (or something similar) property that accepts one of the following strings: `'up'`, `'down'` or `'none'`.

@shlomitc @hepiyellow wdyt?

hepiyellow:
@shlomitc @wuwa 
this condradicts the ux requirement that there  should be only one column with a sort arrow at a time.

yardnsm:
### The idea

* We'll keep the `sortable` property for column and the `onSortClick` prop for the table
* We'll show the sorting icon when hovering the row, as specified in the specs (@wuwa am I right?)
* `sortDescending` will be deprecated, and instead:
	* `sortIconDirection` will determine the sorting icon direction. Acceptable values are `'desc'` (default) or `'asc'`.
[...]

PR #2433, opened 2018-11-07
[WIP] <DataTable/> - updates to the sorting behaviour

yardnsm:
This PR is a WIP - DO NOT MERGE YET. The API is not complete yet and is expected to be changed.

---

Man, I love tables.

As per https://github.com/wix/wix-style-react/issues/2369, we've decided to go with the following API:

We'll keep the `<DataTable/>`'s `onSortClick`. We'll deprecate the `sortDescending` property of a single column, and add the following:

| Prop | Type | Description |
| ---- | ---- | ----------- |
| `sortIconDirection` | `'asc' \| 'desc' \| 'none'` | Will set the direction of the sorting icon on the column's `<th>`. When `'none'`, not arrow will be shown. |
| `sortIconDirectionOnHover` | `'asc' \| 'desc' \| 'none' \| undefined` | Will set the direction of the sorting icon on the column's `<th>` **when the user hovers the row**. When `undefined`, the `sortIconDirection` will still be shown on hover. |

The `<DataTable/>` won't be responsible for any sorting - it's up to the consumer to implement this logic. The `<DataTable/>` will only display the arrows accordingly.

Some points to mention:

1. Normally, when the user clicks the row, the consumer will update the sorting state, thus the `sortIconDirectionOnHover` will be changed. Immediately after the click, the user is still hovering the row, and as a result, the arrow icon will be also changed.
As per the UX specs, after the user clicks the row, the last hovered row should be persisted. This involves having the last value of `sortIconDirectionOnHover` stored in the `<DataTable/>` internal state.
https://github.com/wix/wix-style-react/blob/c7beff4991f7becfc037ebad9bb2b85c7a5737cb/src/DataTable/DataTable.js#L309-L316
I'm not a fan of that. If someone got a better idea, please let me know.

2. Take a look at the story example for `<Table/>` regarding the sorting ([here](https://github.com/wix/wix-style-react/blob/c7beff4991/stories/Table/TableSortingExample.js)). It's pretty big. Are we really want that the consumer will have to manage all of that?
Of course, this is just one way to implement the sorting using the proposed API, and there may be a better (and a shorter) way to do that.

3. If we want to hide some of this logic from the consumers, we could have a different API. @lbelinsk suggested that the table would manage the sorting state by itself, and the consumers would simply pass a sorting function to it.
So, the consumer will pass a `compareFunction` to each column to make it "sortable", and using the `initialSortableColumn` (I'm bad in naming I know) prop, the consumer could determine the default column to be sorted by. Also, the consumers could still be aware to the current sorting state (and change it) using the table's `ref`.

	**Pros:**
	- This would dramatically decrease complexity in the consumer side.
	- API will become slimmer.
	- By defining a fixed behaviour, we could make the usage stricter to the UX guidelines.

	**Cons:**
	- The `<DataTable/>` would have to manage an internal state for the sorting behaviour.
	- The `<DataTable/>` would have to store a copy of the data.
	- What would happen when the data is changed, or more items will be added (for example, after a server calling)?
	- ~~Sorting by multiple columns will be difficult to manage.~~

Again, this is a question of "controlled" vs "un-controlled" (or "less-controlled"). Wdyt?

@shlomitc @hepiyellow @argshook

### DIFF
--- a/src/DataTable/DataTable.js
+++ b/src/DataTable/DataTable.js
@@ -308,7 +308,7 @@ class TableHeader extends Component {
 
     // When the cell is active, we want to preserve the *previous* state after
     // the click
-    const isCurrentCellActive = this.state.activeHeaderCell === colNum;
+    const isCurrentCellActive = sortIconDirectionOnHover && this.state.activeHeaderCell === colNum;
 
     const activeDirection = sortIconDirection;
     const hiddenDirection = isCurrentCellActive ?
@@ -413,7 +413,7 @@ class TableHeader extends Component {
       );
 
       // "Disable" the hover effect when using the legacy `sortDescending` property
-      sortIconDirectionOnHover = sortIconDirection;
+      sortIconDirectionOnHover = undefined;
 
       deprecationLog(
         'Property `sortDescending` of Table\'s `columns` prop is deprecated; use `sortIconDirection` instead.'
--- a/src/DataTable/DataTable.js
+++ b/src/DataTable/DataTable.js
@@ -268,8 +268,21 @@ class TableHeader extends Component {
     newDesign: PropTypes.bool
   };
 
+  state = {
+    // This state property indicates what header cell is currently active. A
+    // cell becomes "active" when it is clicked, and stops being active when
+    // the mouse leaves it.
+    activeHeaderCell: -1
+  }
+
   get style() {
     return styles;
+  }
+
+  setActiveHeaderCell = colNum => {
+    this.setState({
+      activeHeaderCell: colNum
+    });
   }
 
   renderSortingArrow = (column, colNum) => {
@@ -301,6 +314,10 @@ class TableHeader extends Component {
       return null;
     }
 
+    // When the cell is active, we want to preserve the *previous* state after
+    // the click
+    const isCurrentCellActive = this.state.activeHeaderCell === colNum;
+
     return (
       <span
         data-hook={`${colNum}_title`}
@@ -308,6 +325,9 @@ class TableHeader extends Component {
         >
         {ActiveSortingArrow && (
           <ActiveSortingArrow
+            style={{
+              display: isCurrentCellActive ? 'inherit' : undefined
+            }}
             height={12}
             className={this.style.activeSortingArrow}
             data-hook={`active_arrow_${sortingIconDirection}`}
@@ -315,6 +335,9 @@ class TableHeader extends Component {
         )}
         {HiddenSortingArrow && (
           <HiddenSortingArrow
+            style={{
+              display: isCurrentCellActive ? 'none' : undefined
+            }}
             height={12}
             className={this.style.hiddenSortingArrow}
             data-hook={`active_arrow_${sortingIconDirection}`}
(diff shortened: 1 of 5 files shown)


stories/Table/TableSortingExample.js
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  render() {
+    return (
+      <Card>
+        <Table
+          dataHook="story-table-column-alignment-example"
+          data={this.getSortedData()}
+          itemsPerPage={20}
+          onSortClick={(column, colNum) => this.handleSortClick(colNum)}
```
yardnsm: Maybe we can also put the `onSortClick` for each column? So we can create that function in `this.getSortingProperties` like so:

```js
  // This function gets `field`, not `colNum`.
  getSortingProperties(field) {
    // sortedColumn --> sortedField
    const {sortedField, sortingDirection} = this.state;
     const isCurentlySorted = sortedField === field;
     return {
      sortable: true,
      // ...
      onSortClick: () => this.handleSortClick(field)
    };
  }
```

And, in `this.handleSortClick` we could get rid of this ugly disaster:

```js
const fields = {
  0: 'name',
  3: 'price'
};
const fieldToSortBy = fields[sortedColumn];
```

yardnsm:
I've updated the example to be a bit more slimmer, also used lodash's `sortBy` function.

@hepiyellow @shlomitc @wuwa Can you please take a look?

hepiyellow:
@yardnsm 
I can't seem to scroll the Table in Page examples.
https://wix-wix-style-react-pr-2433.surge.sh/?selectedKind=10.%20Tables&selectedStory=10.1%20Table&full=0&addons=0&stories=1&panelRight=0

hepiyellow:
@yardnsm 
There is some issue with scrolling the Table in Page examples.
Sometimes it works, and sometimes it doesn't.

hepiyellow:
@yardnsm Need to update the README with new `sortIconDirection` column properties.

src/DataTable/DataTable.js
```diff
@@ -268,26 +268,85 @@ class TableHeader extends Component {
...
-      return (
-        <span
-          data-hook={`${colNum}_title`}
-          className={this.style.sortArrow}
-          >
-          <Arrow height={12} data-hook={sortDescending ? 'sort_arrow_dec' : 'sort_arrow_asc'}/>
-        </span>);
+
+    // Support old design
+    if (!this.props.newDesign) {
```
shlomitc: please remove the support for old design - it's already removed

yardnsm:
> There is some issue with scrolling the Table in Page examples.
> Sometimes it works, and sometimes it doesn't.

Not related to this PR.

Also, do you think we should document the changes also in the `<DataTable/>` story, or only in the `<Table/>` story?

hepiyellow:
@yardnsm Update documentation in DataTable and in Table

src/DataTable/DataTable.js
```diff
@@ -268,26 +268,85 @@ class TableHeader extends Component {
...
+    // When the cell is active, we want to preserve the *previous* state after
+    // the click
+    const isCurrentCellActive = sortIconDirectionOnHover && this.state.activeHeaderCell === colNum;
+
+    const activeDirection = sortIconDirection;
+    const hiddenDirection = isCurrentCellActive ?
+      this.state.lastSortDirectionOnHover :
+      sortIconDirectionOnHover || sortIconDirection;
+
+    const ActiveSortingArrow = this.getSortingArrow(activeDirection);
```
hepiyellow: No need for capitalized variable names

src/DataTable/DataTable.js
```diff
@@ -344,10 +403,36 @@ class TableHeader extends Component {
...
+      sortIconDirection = (
+        column.sortDescending === undefined ? undefined : (
+          column.sortDescending ? 'desc' : 'asc'
+        )
+      );
+
+      // "Disable" the hover effect when using the legacy `sortDescending` property
+      sortIconDirectionOnHover = undefined;
+
+      deprecationLog(
```
hepiyellow: I'd rather put the deprecation log in a custom PropTypes validator of `sortDescending`

yardnsm:
Closed in favour of https://github.com/wix/wix-style-react/pull/2485

PR #2434, opened 2018-11-07
<TextButton/>  - new component

mykas:
This is initial code for TextButton.

Proposed Component API:
```javascript
<TextButton
  onClick={callback}
  dataHook="data-hook"
  skin="standard | light | premium | dark"
  underlined="none | onHover | always"
  weight="thin | normal"
  size="small | medium"
  className="to override"
  disabled={true | false}
  prefixIcon={<Add />}
  suffixIcon={<Add />}
>
  Click me
</TextButton>
```

Proposed testkit API:

```js
   exists: async () => await base.exists(),
   click: async () => await base.click(),
   getText: async () => await base.text(),
   isDisabled: async () => !!(await base.attr('disabled'))
```
Should we include prefixIcon and suffixIcon check? 
 [erez] - no. the consumer can add `data-hook` on prefix/suffix icons, and check if they exists of not.


src/TextButton/TextButton.js
```diff
@@ -0,0 +1,48 @@
...
+      dataHook,
+      onClick,
+      suffixIcon,
+      prefixIcon,
+      children,
+      disabled,
+      className
+    } = this.props;
+    return (
+      <div className={backofficeTheme}>
```
mykas: How should we handle backofficeTheme wrapping in our components? We have issues with testing onclick when its not on root item. Any ideas @lbelinsk @shlomitc @hepiyellow ? Or should I just put it on a div?

src/TextButton/TextButton.driver.js
```diff
@@ -0,0 +1,11 @@
+import {baseUniDriverFactory} from 'wix-ui-test-utils/base-driver';
+
+export const textButtonDriverFactory = base => {
+  base = base.$('[data-hook="textButton-core"]');
+  return {
+    ...baseUniDriverFactory(base),
+    getTextContent: async () => await base.text(),
+    isDisabled: async () => !!(await base.attr('disabled')),
```
shlomitc: can a text be disabled?

src/TextButton/TextButton.js
```diff
@@ -0,0 +1,75 @@
...
+      undefined;
+    const underlined = allowedUnderlined.includes(this.props.underlined) ?
+      this.props.underlined :
+      undefined;
+    const weight = allowedWeight.includes(this.props.weight) ?
+      this.props.weight :
+      undefined;
+
+    return (
+      <div className={backofficeTheme}>
```
shlomitc: hmm... what happens if you put it as a sibling. will it work?

src/TextButton/TextButton.js
```diff
@@ -0,0 +1,75 @@
...
+    const {
+      onClick,
+      suffixIcon,
+      prefixIcon,
+      children,
+      disabled,
+      className
+    } = this.props;
+
+    const skin = allowedSkins.includes(this.props.skin) ?
```
shlomitc: isn't hits a prop validation done in `propTypes`? looks redundant to me.
just pass the prop directly to the `textButton`.

src/TextButton/TextButton.js
```diff
@@ -0,0 +1,75 @@
...
+      this.props.underlined :
+      undefined;
+    const weight = allowedWeight.includes(this.props.weight) ?
+      this.props.weight :
+      undefined;
+
+    return (
+      <div className={backofficeTheme}>
+        <ButtonNext
+          data-hook="textButton-core"
```
shlomitc: this should be the data-hook by the user. the extra div is really problematic

src/TextButton/TextButton.js
```diff
@@ -0,0 +1,75 @@
...
+          onClick={onClick}
+          className={className || textButton(skin, underlined, weight, size)}
```
shlomitc: I don't think `||` should be used here. it should combine classes with `classnames`

test/utils/dom-test-container.js
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
```
shlomitc: I don't think it should be copied from the core.
I prefer using existing solution we have in our current tests or reuse it from the core/test-utils

src/TextButton/TextButton.driver.js
```diff
@@ -0,0 +1,11 @@
...
+    getNative: async () => await base.getNative()
```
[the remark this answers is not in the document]
mykas: I thought its gonna give me raw react component buts actually bringing rendered output of component. I am bit confused on what kind of standard API should we exposed. I see from previous button in wsr we have things like `onMouseOver` etc.

src/TextButton/TextButton.js
mykas: it only complains in console.log, but does not throw error if prop is wrongly set. I just tested that. I am starting to like typescript just because it throws errors on compile.

src/TextButton/TextButton.js
mykas: sibling of ButtonNext?

src/TextButton/TextButton.js
mykas: combine? hm...so to override some ruleset, right? or add more?

test/utils/dom-test-container.js
mykas: We don't have unidriver solution in wix style react. Plus look at this implementation. It provide some good unmounting solutions which eliminates problems like "focus is not reseted on next test" etc.

src/TextButton/TextButton.js
shlomitc: Yes.
By the way, why is it still ButtonNext? Let’s rename it

src/TextButton/TextButton.js
mykas: sure sure, thats what I mean. Ok, good point

src/TextButton/TextButton.js
shlomitc: How is it different from any other component we have in WSR?

src/TextButton/TextButton.js
mykas: It is problematic. I am not sure how to solve this. Maybe we should actually get rid of wrapping components with a theme scope.

src/TextButton/TextButton.driver.js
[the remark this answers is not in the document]
shlomitc: I don’t see why it is needed. I think you can remove it for now

src/TextButton/TextButton.js
mykas: here the problem is that I pass this to `textButton(size)' and it resolves the stylesheet `premium` which was not suppose to be resolved by prop `size`

mykas: either we fix this in our props or we have to think about typings core. I would rather do this in core

src/TextButton/TextButton.js
mykas: I don't think this will work. The order of classes in css will change and it will not hit the correct css declaration

mykas: Do we have Button component in deprecated folder in core?

src/TextButton/TextButton.js
mykas: Anyway, this is a workaround now. I extend WixComponent which will pass dataHook to the parent div. In the driver I target this data-hook so that I could access values.

src/TextButton/TextButton.js
mykas: what if I pass `textButton({skin, size, underlined, weight})`. 

And then in core I would `export const textButton = ({skin,size,underlined,weight}) => getClassNames({skin,size,underlined,weight}, TextButton, 'textButton');`

mykas: Then in the `getClassNames` - `const getClassNames = (values, stylesheet, rootcls) => {
  const clsArray = object.values(values).map(cls => stylesheet[cls] || null);
  return classNames(stylesheet[rootcls], clsArray);`
};

mykas: Any downside to this? One problem is that developer will need to define accessible values in that file after creating his stylesheet. But maybe this is a good thing.

mykas: I've talked with Lior and I think we should leave this as is for now. This would mean though that if user will pass `size="premium"` it will resolve to premium class in the stylesheet.

src/TextButton/TextButton.js
shlomitc: yes

test/utils/dom-test-container.js
shlomitc: you can copy it for now

src/TextButton/TextButton.js
```diff
@@ -0,0 +1,65 @@
...
+import {string, func, element, bool, node, oneOf} from 'prop-types';
+import WixComponent from '../BaseComponents/WixComponent';
+
+class TextButton extends WixComponent {
+  static displayName = 'TextButton';
+
+  static propTypes = {
+    className: string,
+    skin: oneOf(['standard', 'light', 'premium', 'dark']),
+    underlined: oneOf(['none', 'onHover', 'underlined']),
```
shlomitc: should be called `underline`

stories/TextButton/index.story.js
```diff
@@ -0,0 +1,19 @@
+import TextButton from 'wix-style-react/TextButton';
+import {storySettings} from './storySettings';
+
+export default {
+  category: storySettings.kind,
+  storyName: storySettings.storyName,
+  component: TextButton,
+  componentPath: '../../src/TextButton',
+
+  componentProps: {
```
shlomitc: should have also the icons for the prefix/suffix

mykas: done

src/TextButton/TextButton.js
hepiyellow: I think we should have the css rule in ButtonNext defines so that you can add the theme on the component itself:
```css
BackofficeTheme .button,
BackofficeTheme.button, {
    .... rules....
}
```

So that we don't need the div wrapper.

lbelinsk: Check this rtl rule:
https://github.com/wix-private/wix-ui-poc/blob/master/src/core/themes/backoffice/button/button.st.css#L42

It follows the same logic @hepiyellow suggested.

If we want to follow this logic, I am suggesting it here:
https://github.com/wix/stylable/issues/411

src/TextButton/TextButton.js
```diff
@@ -0,0 +1,65 @@
...
+      className
+    } = this.props;
+
+    return (
+      <div className={backofficeTheme}>
+        <ButtonNext
+          data-hook="textButton-core"
+          onClick={onClick}
+          className={classNames(
+            className,
```
hepiyellow: @mykas @shlomitc 
Actually the `||` that was here before, rises a good question.
## The issue
What if the consumer passes a className which is one of ButtonNext's theme classes, like `sizeLarge`, while the default size is medium.
So the className will include both `sizeMedium sizeLarge` classes.
Which one will apply? 
I guess we want the consumer's class to apply, but that depends now on the order in which these classes are defined in the stylesheet.

## Solution 1
check passed class names, and throw error if one of them matches the ButtonNext classes.
hmm...... doesn't sound so good.

## Solution 2
Don't use the static `defaultProps`, and check it ourselves, and don't apply default if a class was passed.
Better.... but still....
> The `textButton(...)` function could also be passed the defaults, and the resolution would be inside.

## Solution 3
Ignore it.
Why would consumers want to do that, they can use the props.
Just need to maybe state this little warning in the documentation.

## Solution 4
Assuming we wouldn't add those "sugar" props that create classes.
Then we would still have to apply defaults somehow.
so doing it in the `textButton(...classes)` function is a good idea.

src/TextButton/TextButton.js
```diff
@@ -0,0 +1,65 @@
...
+      suffixIcon,
+      prefixIcon,
+      children,
+      disabled,
+      className
+    } = this.props;
+
+    return (
+      <div className={backofficeTheme}>
+        <ButtonNext
```
hepiyellow: ButtonNext supports also any native button attribute.
You should spread the `...rest` of props onto it.

src/TextButton/TextButton.js
mykas: Regarding the issue part. Why do you think consumer will try to use wix-ui-core stylesheets? we are not exposing those for now. Only the props will be exposed.

src/TextButton/TextButton.js
mykas: True

src/TextButton/TextButton.js
```diff
@@ -0,0 +1,65 @@
...
+    suffixIcon: element,
+    prefixIcon: element,
+    children: node
+  };
+
+  static defaultProps = {
+    skin: 'standard',
+    underline: 'none',
+    weight: 'thin',
+    size: 'premium'
```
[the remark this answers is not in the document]
mykas: Yes, I've demonstrated an example how this way it could resolve to premium anyway, because we have our stylesheet resolver is not the best solution

src/TextButton/TextButton.js
mykas: I agree with solution 4, but I am not sure we are ready to move our users from props to classes.

src/TextButton/TextButton.js
hepiyellow: @mykas So what did you decide here?
Please resolved conversations if needed.

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ Types of changes:
 - `<InputArea>` - add name prop to inputarea component [#2449](https://github.com/wix/wix-style-react/pull/2449)
 - `<GenericModalLayout/>` - add new component [#2439](https://github.com/wix/wix-style-react/pull/2439)
 
+- `<TextButton/>` - new component [#2434](https://github.com/wix/wix-style-react/pull/2434)
```
shlomitc: please don't add CHANGELOG changes in this PR. when merged, it will be placed in the wrong place

src/TextButton/README.TESTKIT.md
```diff
@@ -0,0 +1,86 @@
+# TextButton Testkits
```
shlomitc: @argshook - when is the auto-testkit panel going to be merged?

src/TextButton/TextButton.js
shlomitc: so instead, don't extend `WixComponent`, and explicitly add `dataHook` prop

stories/TextButton/icons-for-story.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
```
shlomitc: isn't this a generic thing and not just for TextButton?

src/TextButton/TextButton.js
```diff
@@ -0,0 +1,67 @@
...
+import {string, func, element, bool, node, oneOf} from 'prop-types';
+import WixComponent from '../BaseComponents/WixComponent';
+
+class TextButton extends WixComponent {
+  static displayName = 'TextButton';
+
+  static propTypes = {
+    className: string,
+    skin: oneOf(['standard', 'light', 'premium', 'dark']),
+    underline: oneOf(['none', 'onHover', 'underline']),
```
shlomitc: isn't last option should be called `always`?

mykas: true.it should.  And I should rename prop to `underlined`.

src/TextButton/TextButton.js
hepiyellow: The `<TextButton/>` dataHook should alway be on the root element.
- If you allow passing `...rest` onto `ButtonNext` then you shouldn't override it with your own data-hook. Maybe you can destructure `dataHook` from props, so that `...rest` doesn't have a `dataHook`.
- Another solution is simply NOT to add the `data-hook="textButton-core"`. You can always, in the driver, find the `<ButtonNext>` but the DOM structure, which you know. (`:first-child`).

stories/TextButton/icons-for-story.js
mykas: It is generic. Later I will move this to another folder and about every component to use it.

CHANGELOG.md
mykas: ok

src/TextButton/TextButton.js
mykas: we did not decide on this as we have your theming issue to resolve. Now we can't just make it work.

src/TextButton/TextButton.js
mykas: ok, thanks.

src/TextButton/TextButton.js
hepiyellow: @mykas @shlomitc 
Well about wether we have an issue here or not.
I think it is a good opportunity, to allow styling with className instead of props.
That way we can ask beta-volunteers to try it out. For that, we shouldn't block it.
And we'll be totally forward compatible.

And about Solution 4, I think defaults should be taken as low as possible in the flow.

src/TextButton/TextButton.js
```diff
@@ -0,0 +1,67 @@
...
+      className,
+      ...rest
+    } = this.props;
+
+    return (
+      <div className={backofficeTheme}>
+        <ButtonNext
+          {...rest}
+          data-hook="textButton-core"
+          onClick={onClick}
```
hepiyellow: Now that you spread `...rest`, you don't need this

src/TextButton/TextButton.js
```diff
@@ -0,0 +1,67 @@
...
+          className={classNames(
+            className,
+            textButton(skin, underline, weight, size)
+          )}
+          suffixIcon={suffixIcon}
```
hepiyellow: Now that you spread `...rest`, you don't need this `suffixIcon`, `prefixIcon`,`disabled`

src/TextButton/TextButton.js
mykas: oh yes, true.

src/TextButton/README.TESTKIT.md
argshook: it is merged since yesterday. testkit tab should automatically fill in with table of enzyme driver methods. comments can be added above each method, same as with proptypes

@mykas let me know if any of that doesn't work. you should no longer need markdown file

src/TextButton/TextButton.js
```diff
@@ -0,0 +1,55 @@
+import React, {Component} from 'react';
+import {ButtonNext} from 'wix-ui-core/button-next';
+import cx from 'classnames';
+import {textButton, backofficeTheme} from 'wix-ui-core/themes/backoffice';
+import {string, node, oneOf} from 'prop-types';
+
+class TextButton extends Component {
+  static displayName = 'TextButton';
+
+  static propTypes = {
```
shlomitc: disabled is missing here but appears in the story

stories/TextButton/examples/textbutton-affixes.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import CodeShowcase from 'wix-storybook-utils/CodeShowcase';
```
shlomitc: in general, prefer using the new code examples format as we have in the `4.1 Dropdown` section

shlomitc:
@milkyfruit - please do a UX review on this component

stories/TextButton/examples/textbutton-sizes.js
```diff
@@ -0,0 +1,42 @@
...
+
+import TextButton from 'wix-style-react/TextButton';
+import ChevronDown from 'wix-style-react/new-icons/ChevronDown';
+
+const example = `import React from "react";
+import TextButton from 'wix-style-react/TextButton';
+import ChevronDown from 'wix-style-react/new-icons/ChevronDown';
+
+export default () => (
+  <React.Fragment>
```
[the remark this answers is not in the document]
mykas: I can't use it like that otherwise we will have examples with certain color for background. We need to make examples as lean as possible from examples implementation. But maybe if I'll use the other example from Dropdown - then its gonna be ok. I'll check.

testkit/testkit-smoke.test.js
```diff
@@ -65,6 +65,7 @@ const FAILING_COMPONENTS = [
   'TableToolbar',
   'Tabs',
   'TextArea',
+  'TextButton',
```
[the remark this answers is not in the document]
mykas: TextButton is based on unidriver. I moved it here intentionally so at least our build will be green until we resolve this smoke test to work with unidriver.

src/TextButton/TextButton.js
mykas: disabled is defined in core button

stories/TextButton/examples/textbutton-affixes.js
mykas: ok

testkit/testkit-smoke.test.js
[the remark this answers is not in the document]
shlomitc: @argshook - can you assist here?

stories/TextButton/examples/textbutton-affixes.js
mykas: Leaving CodeShowcase until LiveCodeExample will support special color backgrounds for light based components and also shows components straight in line instead of a column.

src/TextButton/TextButton.js
mykas: Ah..now I see why I need to add it to proptypes. Storybook playground reads types here.

testkit/testkit-smoke.test.js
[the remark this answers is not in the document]
shlomitc: @mykas - so please create a sanity test for this file in `testkit/vanilla.spec.js`, `testkit/enzyme.spec.js`, `testkit/protractor.spec.js`/  `testkit/puppeteer.spec.js`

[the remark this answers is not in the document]
argshook: there is no `testkit/enzyme.spec.js` and `testkit/vanilla.spec.js` anymore.
we should either:
1. improve testkit-smoke.test.js to do sanity for protractor, pupeteer, unidriver
2. do sanity tests for TextButton inside TextButton.spec.js, until punto numero 1 is done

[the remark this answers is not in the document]
shlomitc: so I'm opting to number 2, but I don't want to mix the testkits inside the component spec file, it causes issues with `jest` watch mechanism. That was the whole point of extracting it out.

stories/TextButton/examples/textbutton-affixes.js
mykas: Solved the liveCodeExample. Rows in preview are not supported in  LiveCodeShowExample.

testkit/testkit-smoke.test.js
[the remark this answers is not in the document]
mykas: Arijus tweaked his code to support multiple asserts based on configuration. Now I have added unidriver asserters for enzyme and vanilla. Need to add test assert for puppeteer in wix-ui-test-utils and it should work.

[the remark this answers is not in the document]
shlomitc: Please update contribution guide after doing so

src/TextButton/TextButton.js
```diff
@@ -0,0 +1,58 @@
...
+    size: oneOf(['small', 'medium']),
+    suffixIcon: element,
+    prefixIcon: element,
+    disabled: bool,
+    children: node,
+    dataHook: string
+  };
+
+  static defaultProps = {
+    skin: 'standard',
```
hepiyellow: - The is no skin standard
- There are already default in the stylesheet it self. We don't need to set defaults for style props.
-  I can only imagine that you are doing this in order to satisfy AutoDocs to show a default value.
-  Can we solve it another way?
- a compromise will be to add comments on propTypes and write the default value there.

src/TextButton/TextButton.js
```diff
@@ -0,0 +1,58 @@
+import React, {Component} from 'react';
+import {ButtonNext} from 'wix-ui-core/button-next';
+import cx from 'classnames';
+import {textButton, backofficeTheme} from 'wix-ui-core/themes/backoffice';
+import {string, node, oneOf, element, bool} from 'prop-types';
+
+class TextButton extends Component {
+  static displayName = 'TextButton';
+
+  static propTypes = {
```
hepiyellow: Missing comments for documentation

src/TextButton/TextButton.js
```diff
@@ -0,0 +1,58 @@
...
+    className: string,
```
hepiyellow: Usually `className` prop is a class that is applied on the component's root element.
You are applying it to the nested ButtonNext, not the root element.
Either fix it.
Or make it very clear in the documentation (comments) of `className`.

src/TextButton/TextButton.driver.js
```diff
@@ -0,0 +1,10 @@
+import {baseUniDriverFactory} from 'wix-ui-test-utils/base-driver';
+
```
hepiyellow: Why aren't you importing and exporting the driver from `wix-ui-core`?

If you still have a reason to write a new one, then you should add tests for it.

src/TextButton/TextButton.js
```diff
@@ -0,0 +1,58 @@
...
+    skin: oneOf(['standard', 'light', 'premium', 'dark']),
```
hepiyellow: Since this order of skins makes it to the AutoDocs.
Please order the skins alphabetically (or in any rational order)

src/TextButton/TextButton.js
mykas: Yes, we can solve this later. This was only meant for AutoDocs.

src/TextButton/TextButton.js
mykas: Why should I apply it to the root when we have a wrapper that does nothing except for scope define? I think that classname is pretty great tool in this component to modify the default style of a button.

src/TextButton/TextButton.js
argshook: > Please order the skins alphabetically (or in any rational order)

:)

src/TextButton/TextButton.e2e.js
```diff
@@ -0,0 +1,52 @@
+import eyes from 'eyes.it';
+import autoExampleDriver from 'wix-storybook-utils/AutoExampleDriver';
+
+import {createStoryUrl} from '../../test/utils/storybook-helpers';
+import {storySettings} from '../../stories/TextButton/storySettings';
+
+describe('TextButton', () => {
```
hepiyellow: Please add sanity test for the textButtonDriver

mykas: Which sanity tests?

mykas: You mean puppeteer and protractor?

mykas: Later we will add automatic sanity tests for puppeteer and protractor but this will required more work to be done and it could be an infra task.

src/TextButton/TextButton.js
mykas: no

src/TextButton/TextButton.js
hepiyellow: @mykas The consumer needs to know where this class name goes.
For example if the consumer wants to put this TextButton as an item in a flex container,
and it wants to apply a class to the TextButton that will make the item `flex-grow: 1`.
The consumer might try to use the `className` prop (instead of adding a div wrapper for every TextButton), but it wouldn't work. Since the className doesn't go on the root element.

So again, either document it very clearly, or put it on the root element (and document it) which is the standard in WSR and wix-ui.

ipanasenko:
So, I can replace `<TextLink onClick />` with `<TextButtin onClick />`, right?

mykas:
Well TextButton actually wraps html tag -> button. If you are not using props like href, then sure you should use this one.

Issue #2435, opened 2018-11-07
<Button/> - x-small size with affixes has wrong margins


LittleTal:
In `<Button/>` component with suffix/prefix props- margin and padding are not aligned to guidelines. 
A fix is needed for all available height prop.

For example in `height: x-small`

Margin should be 0px instead of 10px.

Also, padding should be: 
**Prefix:**
    padding-right: 12px;
    padding-left: 6px;

**Suffix:**
    padding-right: 6px;
    padding-left: 12px;

Component should be aligned to [zeplin guidelines](https://zpl.io/aRq9ZKv):

![image](https://user-images.githubusercontent.com/23282301/48132245-96a4ef80-e29b-11e8-808d-fbc10cbcad19.png)

### Severity
Critical

mykas:
Hi, thanks for your input! I am working at this moment on introducing all the buttons aligned to specifications. It is already done in our core repo so it stay on.

shlomitc:
done

PR #2438, opened 2018-11-08
<Button/> - Fixing margins. Moving icon size control to css

mykas:
This pr fixes suffix and prefix icon margins and moves icon size control to css instead of js.

related to: https://github.com/wix/wix-style-react/issues/2435


mykas:
@LittleTal Hey, can you check?

PR #2440, opened 2018-11-08
Migrate to Yoshi 3

hepiyellow:
Replaces #2374 
Starts from latest master, cherry-pick relevant commits from #2374


hepiyellow:
Replaces by #2493

PR #2441, opened 2018-11-08
Add missed usage of InputArea data-hook

Faradey27:
### What changed

...

### Why it changed

...

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


shlomitc:
the data-hook is set automatically on the root node, no need to manually set it to the `textarea` component.

what exactly is being solved here?

PR #2442, opened 2018-11-08
BugFix: isOpen always returns true

zivl:
### What changed

`isOpen` function in modal driver

### Why it changed
Because of the way the modal is being rendered on the DOM `body`, which the wrapper is rendered empty when the modal is closed and not shown. `!!getContent()` always returned true since it looks for the wrapper DOM node which always exists.
Instead, there should be a deeper check up to see wether there is a content inside it or not
fix a bug where the `isOpen` always returns `true`

### DIFF
--- a/src/Modal/Modal.driver.js
+++ b/src/Modal/Modal.driver.js
@@ -11,7 +11,7 @@ const modalDriverFactory = ({element, wrapper, component}) => {
   return {
     exists: () => !!(getPortal()),
     element: () => element,
-    isOpen: () => !!(getContent()),
+    isOpen: () => !!getContent() && !!(getContent().textContent),
     isThemeExist: theme => !!getPortal().querySelector(`.${theme}`),
     getChildBySelector: selector => getPortal().querySelector(selector),
     isScrollable: () => !getPortal().classList.contains('portalNonScrollable'),


zivl:
@shlomitc - In the tests you're checking `getChildBySelector` to see wether rendering is happening for the modal content or not. but there is no tests which uses the `isOpen` function in the driver.
Sure, I'll add the tests, though, do you prefer to add new `test` blocks or shall I add another `expect` in the current tests mentioned above?

shlomitc:
You can add a new teat block. Thanks

shlomitc:
@zivl - any progress with this PR?

zivl:
@shlomitc yes I actually wrote the tests... just need to update the branch with origin and stuff...

shlomitc:
ping?

hepiyellow:
@shlomitc @zivl 
I don't think there is anything wrong with `isOpen`.
Shlomi, you mentioned to me that when the modal is being closed, it has a closeTimeout.
So maybe Ziv this is what you saw when you concluded that isOpen always returns true.

Please see this PR #2516 
It adds proper cleanup between tests without doing `document.body.innerHtml = ''`,
and it uses the `isOpen` in order to wait for the ReactModal to be removed from the body after each test (The  Modal is unmounted after each test).

It proves that `isOpen` works, it returns `false` when appropriate.

BTW: there was one test (`should wait closeTimeoutMS before removing the modal`) already that did assert:
`expect(driver.isOpen()).toBeFalsy();`
and passed.

zivl:
@shlomitc @hepiyellow  - i'll schedule a meeting so I can show you the problem (showed it to @lbelinsk before he left and he saw the issue)
but may be I'm wrong and you could help me fixing it

hepiyellow:
@meeting with Ziv this Monday

PR #2445, opened 2018-11-09
<IconButton/> - new component

mykas:
Initial code.

### DIFF
--- a/src/IconButton/IconButton.spec.js
+++ b/src/IconButton/IconButton.spec.js
@@ -1,0 +1,7 @@
+import IconButton from './IconButton';
+
+describe('IconButton', () => {
+  it('should have correct displayName', async () => {
+    expect(IconButton.displayName).toEqual('IconButton');
+  });
+});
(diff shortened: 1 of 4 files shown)


mykas:
@milkyfruit please review https://wix-wix-style-react-pr-2445.surge.sh/

src/IconButton/IconButton.spec.js
```diff
@@ -0,0 +1,7 @@
+import IconButton from './IconButton';
+
+describe('IconButton', () => {
+  it('should have correct displayName', async () => {
+    expect(IconButton.displayName).toEqual('IconButton');
+  });
```
argshook: please add to ensure `size="small"` adds appropriate size to children

mykas: done

ipanasenko:
Will this component have more sizes?

mykas:
For now it supports only two. But maybe we could support more. What do you think @milkyfruit?

milkyfruit:
Yup, as @mykas  mentioned this button [supports 2 sizes (36 and 30 px) and 2 themes](https://zpl.io/VQqdZ3m).
@ipanasenko could you share what use case you're trying to solve?

PR #2446, opened 2018-11-10
test(*): smoke test for all components vanilla & enzyme testkits

argshook:
### What changed

i noticed these files:  
https://github.com/wix/wix-style-react/blob/master/testkit/vanilla.spec.js  
and  
https://github.com/wix/wix-style-react/blob/master/testkit/enzyme.spec.js

and thought to myself, this can be automated.

this PR removes `vanilla.spec.js` & `enzyme.spec.js` files and adds
`testkit-smoke.test.js` which does exactly what prior two did but
automates it + has a list of components that currently don't work.

this scales effortlessly as well as let us know what components don't
have testkits or miss exports or something.


testkit/testkit-smoke.test.js
```diff
@@ -0,0 +1,121 @@
...
+import * as enzymeTestkitFactories from './enzyme';
+
+const cwd = path.resolve(__dirname, '..', 'src');
+const lowerFirst = a =>
+  a
+    .charAt(0)
+    .toLowerCase()
+    .concat(a.slice(1));
+
+const FAILING_COMPONENTS = [
```
argshook: some of these are not meant to have testkits, some are failing, some need required props (see next comment).
The goal should be to reduce this list to minimum

testkit/testkit-smoke.test.js
```diff
@@ -0,0 +1,121 @@
...
+  'Table',
+  'TableToolbar',
+  'Tabs',
+  'TextArea',
+  'TextField',
+  'Tooltip',
+  'VBox'
+];
+
+const setupProps = {
```
argshook: some components have required props, and they fail if used simply as `<Component/>`. thus, an object for those cases

hepiyellow:
@JoMarton @shlomitc @alisey 
I think this tool should be part of the `wix-ui` [autotools](https://github.com/wix-incubator/ui-autotools).

- The metadata infrastructure is already working, and it's exactly the place to define required props for testing.
- The autotools is extensible, so you can write a testkit-sanity test on top of it.
- As we learned from the Zelda Hackathon, there are always exceptions, so it's better to have manual registration API, and have the automation use that API. Meaning in our case, if we use the meta-data infrastructure, and that is our component registration API, then the import-all-component, could be simply a script that generates a default meta-data file for components which don't have it.

argshook:
@hepiyellow that's a good idea but perhaps until it manifests we can merge this into wsr, play with it, adjust as needed and then it will be more clear how should this tool look when extracted.

testkit/testkit-smoke.test.js
hepiyellow: Your GutHub comment should be an in-code comment

testkit/testkit-smoke.test.js
hepiyellow: Are these 4 the only components which have required props?
Shouldn't we add required props for all components, before we remove their manual sanity tests?

hepiyellow:
@argshook I'd rather not.
but if you insist...
then please add enough code comments, so that anyone can step in, understand what's going on, and maintain it.

argshook:
> I'd rather not.

why not? i don't want to merge whatever, if there's a good reason not to. in this case i think i do more good than harm and we have something to continue with

testkit/testkit-smoke.test.js
argshook: no, there're many more, there also are testkit sanity tests in component specs and not extracted here. As I mentioned somewhere, goal should be to reduce the list of `FAILING_COMPONENTS` to a minimum and eventually maybe don't have such list at all. Not it's here simply not to break our build.

If we're going to merge this, then we can continue to work on those components one by one. Could also be a well defined guildweek task.

argshook: > Shouldn't we add required props for all components, before we remove their manual sanity tests?

we can, sure, but that would mean one mega PR touching lots of component tests, possibly merge conflicts and whatnot. I'd prefer to start small and go gradually, you know, פרה פרה :)

Issue #2432, opened 2018-11-07
BadgeSelect shouldn't be a controlled component 🍄


aryelu:
# Bug Report

### Relevant Components
BadgeSelect

### Current Behavior
Select an option from the dropdown will change the `selectId`

### Expected Behavior
When selecting only invoke `onSelect` and let `onSelect` determine the outcome

### Steps to Reproduce
https://codesandbox.io/s/7w49xklnnx

### Severity
Critical

yardnsm:
The `<BadgeSelect/>` is actually *partially* controlled, like the `<Dropdown/>` component. It tries to be both controlled and uncontrolled in the same time.

You can see an example here: https://codesandbox.io/s/xr76q6r9lq

When you provide a new `selectedId`, it'll update it's internal state to match it. But, when the `selectedId` won't change, it update the internal state according to the user selection.

We can, maybe, set the component to be uncontrolled when you don't pass `selectedId`, but then we'll get a different behavior than the Dropdown's. 

@shlomitc

aryelu:
to make it clear I'm looking to set the `selectedId` in the `onSelect` prop.
my use case is to make the user approve his selection before the state will change using a modal.

aryelu:
Hi, I’ve tried the controlled badgeSelect (btw ty!!) and it has a weird issue, the selectId is controlled as I selected but the dropdown shows the one I’ve clicked (edited)
![image](https://user-images.githubusercontent.com/11705492/48371234-ff3d0380-e6c3-11e8-8c60-487218c81ced.png)

shlomitc:
fixed in v15.4.0

PR #2452, opened 2018-11-11
<BadgeSelect/> - make component controlled, add missing tests and refactor

shlomitc:
1. set as controlled if `selectedId` prop is provided.
2. add a missing test for default behavior when `selectedId` isn't supplied

This change is backward compatible.
- if `selectedId` was not passed, it keeps behaving like the uncontrolled version
- if `selectedId` was passed, it must have changed the wrapping state for consistency

@aryelu - please verify that it answers your [issue](https://github.com/wix/wix-style-react/issues/2432)


src/BadgeSelect/BadgeSelect.spec.js
```diff
@@ -86,21 +92,54 @@ describe('BadgeSelect', () => {
...
   });
 
-  it('should change badge skin after an option is selected', () => {
-    const {driver, badgeDriver} = createComponent();
-    const selectedIndex = 3;
 
-    driver.click();
-    driver.clickAtOption(selectedIndex);
-    expect(badgeDriver.getSkin()).toBe(options[selectedIndex].skin);
+  it('should should pick the first option if no selectedId given', () => {
```
lbelinsk: double `should`

src/BadgeSelect/BadgeSelect.spec.js
```diff
@@ -86,21 +92,54 @@ describe('BadgeSelect', () => {
...
+    const uncontrolledState = {selectedId: undefined};
+    const {badgeDriver} = createComponent(uncontrolledState);
```
lbelinsk: createComponent receives `props`, so maybe rename the argument to uncontrolledProp?

src/BadgeSelect/BadgeSelect.spec.js
shlomitc: done

src/BadgeSelect/BadgeSelect.spec.js
shlomitc: done

PR #2460, opened 2018-11-13
Input options numeric props

zivl:
### What changed
InputWithOptions.js `dropdownOffsetLeft` and `dropdownWidth` props' types

### Why it changed
When using `InputWithOptions` and the above props, we can the following React warning:
```
`InputWithOptions` was passed a numeric string value for CSS property `marginLeft` (value: `230`) which will be treated as a unitless number in a future version of React.
`InputWithOptions` was passed a numeric string value for CSS property `width` (value: `350`) which will be treated as a unitless number in a future version of React.
```

Hence those values must be changed sooner or later.
---

- [X] Change is tested
- [X] Change is documented
- [X] Build is green(https://github.com/milkyfruit)


shlomitc:
@zivl hey, I think the current string requirement is that you are required to pass a string with `px`.
otherwise, this means the component will need to know if you want `px/em/%` etc...

so passing a number doesn't look like the right way to go

zivl:
@shlomitc oh... ok.. closing this

PR #2464, opened 2018-11-13
<Checkbox /> - Generate proper ID for checkbox

mykas:
This PR generates proper id for checkbox.

### DIFF
--- a/src/Checkbox/Checkbox.js
+++ b/src/Checkbox/Checkbox.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import {node, bool, func, oneOf, string} from 'prop-types';
-import uniqueId from 'lodash/uniqueId';
 import classNames from 'classnames';
 import CheckboxChecked from 'wix-ui-icons-common/system/CheckboxChecked';
 import CheckboxIndeterminate from 'wix-ui-icons-common/system/CheckboxIndeterminate';
--- a/src/Checkbox/Checkbox.js
+++ b/src/Checkbox/Checkbox.js
@@ -9,6 +9,8 @@ import styles from './Checkbox.scss';
 import WixComponent from '../BaseComponents/WixComponent';
 import {withFocusable, focusableStates} from '../common/Focusable';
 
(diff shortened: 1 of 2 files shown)


argshook:
this can be good enough for now but i'd consider this a temporary solution. we should get rid of `id`s and `httmlFor` attributes entirely simply by nesting `<input/>` inside `<label/>`.

could be a good infra task

argshook:
@shlomitc `lodash.uniqueId` simply increments a number (https://github.com/lodash/lodash/blob/master/uniqueId.js), it's not really unique.

given multiple bundles on the same page it is very likely that those ids can clash. It's possible to have `Checkbox-1` multiple times on the page.

now since we do `<Label htmlFor={idOfCheckbox}/>` what's possible to happen is clicking one checkbox in one part of page could accidentally switch value of another checkbox with the same id.

---

what this PR does is creates a real GUID for `id`. It is overkill, but fixes issue of clashing `id`s

since it's an overkill, it should be temporary and eventually solved by not using IDs at all.

shlomitc:
@argshook - you mean not using internal `id`s? if so, I'm fine with that, but it's a breaking change so maybe doesn't worth making this change at all? If the `id` prop is already exposed, why not ask the consumer to use it?

mykas:
How is this a breaking change?

argshook:
@shlomitc

> @argshook - you mean not using internal ids? if so, I'm fine with that, but it's a breaking change so maybe doesn't worth making this change at all? If the id prop is already exposed, why not ask the consumer to use it?

yeah, internally we don't need that ID if we nest `<input/>` inside
`<label/>`. as long as we don't remove `props.id` it's not a breaking
change.

user can set id through prop and may continue to do so. but i don't
think that we should ask our users to change ids and ensure they're
unique enough, when we can avoid the problem altogether.


imo this can be merged and let's have infra task to resolve this and
other components properly.

shlomitc:
Whatever you think, I don’t have a strong opinion here

PR #2466, opened 2018-11-13
<DropdownLayout/> - support controlled mode selection and keyboard mark fixes

shlomitc:
## Why?
1. The original use case was to make the `BadgeSelect` controlled, so it can be prompt before changing to the selected value (see the added example and [this PR](https://github.com/wix/wix-style-react/pull/2452)). This didn't work properly as `DropdownLayout` holds the state regardless if it's controlled or not.
2. The keyboard navigation and option marking was wrong - it always started from the start and not from the selected option

## What's in?
1. In this PR, I assume that if `selectedId` and `onSelect` props are passed to the component, then it is considered `controlled` and should be updated only when props change.
2. I also changed the basic example of `Dropdown` to demonstrate the uncontrolled version (so selection works)
3. I removed the b2b theme that god knows why it was ever there

## The marking (hover) fixes:
1. When opening a dropdown - a selected item will not be hovered (in IconWithOptions for example)
2. When navigating with the keyboard from a selected item, it will mark the next item (DropdownLayout


argshook:
there's eye diff which im not sure if should be accepted

PR #2467, opened 2018-11-14
skeleton

eyaleizenberg:


.storybook/common/stories.js
```diff
@@ -24,6 +24,7 @@ import '../../stories/Page/FullPageExample.js'; // 2.5 + Page Example
 import '../../stories/EmptyState/EmptyState.story'; // 2.6 EmptyState
 import '../../stories/StatsWidget'; // 2.7 StatsWidget
 import '../../stories/AddItem/AddItem.story'; // 2.8 AddItem
+import '../../stories/Skeleton/index.story.js'; // 2.9 Skeleton
```
shlomitc: shouldn't be part of 2.9, just move to the Components API section in this file

src/Skeleton/SKELETON-SPEC.md
```diff
@@ -0,0 +1,22 @@
+# Content Skeleton
+## What is it?
+Content Skeleton is a “placeholder” to be shown while fetching data from a server (or any other long async task).
+
+## Why do we need it?
+The Content Skeleton’s main goal is to improve perceived performance. Instead of just show a blank page or a spinner, we show content which somewhat resembles the content which will soon be displayed on the screen.
+
+## Where do we need it?
+A Content Skeleton is already in use in the Business Manager’s Home (dashboard), inbox, and we now need one for the Contact Panel.
```
shlomitc: these are business concerns, so better not to mention them here in the library

src/Skeleton/Skeleton.js
```diff
@@ -0,0 +1,98 @@
...
+            <div data-hook="placeholder-chunk" className={classnames(styles.chunk, styles[size])}/>
+            {this.renderConcealer()}
+          </div>
+        </div>
+      </div>
+    );
+  }
+}
+
+export default class Skeleton extends WixComponent {
```
shlomitc: add `DisplayName`

src/Skeleton/Skeleton.js
```diff
@@ -0,0 +1,98 @@
...
+    spacing: PropTypes.oneOf(['small', 'medium', 'large'])
+  }
+
+  static defaultProps = {
+    alignment: 'start',
+    spacing: 'medium'
+  }
+
+  render() {
+    const {content, alignment, spacing, dataHook} = this.props;
```
shlomitc: since you extend `WixComponent`, the `dataHook` is automatically applied, so you don't need to manually set it on the root div

src/Skeleton/Skeleton.scss
```diff
@@ -0,0 +1,193 @@
...
+  }
+}
+
+.animatedBackground {
+  animation-duration: 1s;
+  animation-fill-mode: forwards;
+  animation-iteration-count: infinite;
+  animation-name: placeHolderShimmer;
+  animation-timing-function: linear;
+  background: linear-gradient(to right, #dfe5eb 9%, #eef2ef 18%, #dfe5eb 33%);
```
shlomitc: we use the colors from the `colors.scss` palette file.
@litalwix - can you please assist in defining the right colors for this component?

src/Skeleton/Skeleton.scss
```diff
@@ -0,0 +1,193 @@
...
+        left: mid-corner-position($mediumPercentage);
+      }
+      &.large {
+        left: mid-corner-position($largePercentage);
+      }
+    }
+  }
+}
+
+@mixin rightToLeft {
```
shlomitc: remove

src/Skeleton/Skeleton.scss
```diff
@@ -0,0 +1,193 @@
...
+
+}
+
+:global(.rtl) {
```
shlomitc: please use the `withRTL` mixin from here:
https://github.com/wix/wix-style-react/blob/eadd8d3d3e954e411a8c62ba678f40f666fb98f1/src/common.scss#L5

src/assets/skeleton-corner.svg
```diff
@@ -0,0 +1,8 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="6" height="11" viewBox="0 0 6 11">
```
shlomitc: I think it's better to have it next to the component and not in the assets folder

stories/Skeleton/index.story.js
```diff
@@ -0,0 +1,51 @@
...
+  componentPath: '../../src/Skeleton',
+
+  componentProps: () => ({
+    content: exampleContent[0].value,
+    dataHook: 'storybook-skeleton'
+  }),
+
+  exampleProps: {
+    content: exampleContent
+  }
```
shlomitc: you described in the spec that it can come in different use cases. I think adding some `examples` for actual usage will be important.

For example, inside a card

src/Skeleton/Skeleton.spec.js
```diff
@@ -0,0 +1,83 @@
...
+      expect(driver.inMiddleAlignment(driver.lines()[0])).toBe(true);
+    });
+
+    it('should have double the amount of concealers, but same number of lines', () => {
+      expect(driver.chunks().length).toBe(content.length);
+      expect(driver.concealers().length).toBe(2 * content.length);
+    });
+  });
+
+  describe('testkits', () => {
```
shlomitc: we're starting to move these tests away from the component.
there is the `testkit/testkit-smoke.test.js` file that automatically does that.
so if the build passes, you don't need to do anything

src/Skeleton/Skeleton.scss
```diff
@@ -0,0 +1,193 @@
...
+}
+
+.concealerRow {
+  display: flex;
+  width: 100%;
+  position: relative;
+}
+
+.chunk {
+  height: 11px;
```
shlomitc: extract to a variable, it is repeated in several places

src/Skeleton/Skeleton.scss
```diff
@@ -0,0 +1,193 @@
...
+  &.small {
+    margin-bottom: 10px;
+  }
+  &.medium {
+    margin-bottom: 16px;
+  }
+  &.large {
+    margin-bottom: 22px;
+  }
+  &.first {
```
shlomitc: this means the first item will get both `margin-bottom`
image

not sure if it's that that bad, but need to take into consideration.

maybe worth using a pseudo selector (`nth-child`) instead of applying a new `first` class anyway, it's just an extra logic that can be achieved by css

src/Skeleton/README.TESTKIT.md
```diff
@@ -0,0 +1,89 @@
+# Skeleton Testkits
```
shlomitc: @argshook - can you please explain how this should be transformed into autodocs readable content?

src/Skeleton/Skeleton.driver.js
```diff
@@ -0,0 +1,19 @@
+export const skeletonDriverFactory = ({element}) => {
+
+  return {
+    exists: () => !!element,
+    lines: () => element.querySelectorAll('[data-hook="placeholder-line"]'),
```
shlomitc: Drivers should return only primitive values (boolean, number, string, void, etc...)
In that case, I would return `getNumberOfLines()` or something

src/Skeleton/Skeleton.driver.js
```diff
@@ -0,0 +1,19 @@
...
+    chunks: () => element.querySelectorAll('[data-hook="placeholder-chunk"]'),
```
shlomitc: same comment as above

src/Skeleton/Skeleton.driver.js
```diff
@@ -0,0 +1,19 @@
...
+    concealers: () => element.querySelectorAll('[data-hook="placeholder-concealer"]'),
+    isFirstLine: el => el.className.includes('first'),
+    smallSize: el => el.className.includes('small'),
```
shlomitc: we had a lot of discussions about these kinds of drivers.
A driver should be a simple abstraction over common actions or interaction with the component, or an assertion over some state.

For example - selecting an option from a drop-down is a complex action that can be encapsulated in the driver.

styles assertions are interesting because they are props, but testing visual props are usually ending up with a mess of tests that don't help a lot in refactoring nor confidence.

if possible - please create `Skeleton.private.driver`, and have all the style validations there. you can see the `Tag.driver.private` for example

src/Skeleton/Skeleton.js
```diff
@@ -0,0 +1,98 @@
...
+class PlaceholderLine extends React.PureComponent {
+  static propTypes = {
+    size: PropTypes.oneOf(['small', 'medium', 'large']),
+    isFirst: PropTypes.bool.isRequired,
+    spacing: PropTypes.oneOf(['small', 'medium', 'large']),
+    alignment: PropTypes.oneOf(['start', 'middle'])
+  }
+
+  renderConcealer() {
+    return <div data-hook="placeholder-concealer" className={classnames(styles.concealer, styles[this.props.size])}/>;
```
shlomitc: in reusable components, we pass `data-hook` as a prop, that way you can make them unique when you query for them, for example `placeholder-concealer-${idx}`

src/Skeleton/README.TESTKIT.md
argshook: comments above method names in `component.driver.js` will be shown in storybook. same as with proptypes or interfaces in components. https://github.com/wix/wix-style-react/pull/2468/files#diff-7993db66a102a35dd5d07c3d962b5401R12

PR #2468, opened 2018-11-14
chore(Skeleton): grand refactora

argshook:


argshook:
this is a "review" PR for #2467

src/Skeleton/Skeleton.scss
```diff
@@ -1,33 +1,27 @@
-$smallPercentage: 30%;
-$mediumPercentage: 45%;
-$largePercentage: 75%;
-$skeletonCornerWidth: 6px;
-
-@function mid-corner-position($percentage) {
```
argshook: @eyaleizenberg i did not understand the need of corners and concealers. `border-radius` seems to work just fine and concealers feel entirely unnecessary. do i assume right? maybe they're needed for something?

src/Skeleton/Skeleton.scss
```diff
@@ -1,33 +1,27 @@
...
+  position: relative;
+  display: flex;
+  align-items: center;
+  height: 11px;
+  overflow: hidden;
+  background: #fff;
 
-@keyframes placeHolderShimmer {
-  0% {
-    background-position: -468px 0;
```
argshook: this is expensive animation and can make the whole page feel sluggish, especially if there are multiple skeletons in one page.

instead of moving background, it's far cheaper to `translateX` the whole "shimmer" item. that way there're no repaints at all.

another thing is magic numbers here. why is it `-468`, what does that mean? what if skeleton width is `1000px`?

for these reasons i changed animation into very simple:
```css
@keyframes moveRight {
  to { transform: translate(100%); }
}
```

src/Skeleton/Skeleton.driver.js
```diff
@@ -1,19 +1,39 @@
...
   return {
     exists: () => !!element,
-    lines: () => element.querySelectorAll('[data-hook="placeholder-line"]'),
-    chunks: () => element.querySelectorAll('[data-hook="placeholder-chunk"]'),
-    concealers: () => element.querySelectorAll('[data-hook="placeholder-concealer"]'),
-    isFirstLine: el => el.className.includes('first'),
-    smallSize: el => el.className.includes('small'),
-    mediumSize: el => el.className.includes('medium'),
-    largeSize: el => el.className.includes('large'),
-    inSmallSpacing: el => el.className.includes('small'),
```
argshook: @eyaleizenberg i think Shlomi wrote comments about it in your PR

driver API is always a matter of discussion and can differ from person to person. The important things in my book are:

* this api is public, so meaningful names are a must
* user should not need to know about internals of component
* the less arguments for methods, the better

here methods `chunks` and `concealers` mean nothing to the end user.
also, for example, `smallSize` can work well only when passed correct argument. but user should not know about internals.

Admittedly im not good at designing driver APIs, but try to follow those rules of thumb and i think it turns out good enough. but if you feel im wrong, feel free to educate me, seriously :)

argshook:
@eyaleizenberg just fyi, `README.md` is gone, because the amount of content there easily fits as comment above component implementattion. That comment is shown in storypage.

`README.TESTKIT.md` is gone also, because autodocs reads `Skeleton.driver.js` and shows data in storypage from there. also, because of copypaste, `README.TESTKIT.md` had protractor testkit guide, but there is no protractor testkit at all :)

Issue #2459, opened 2018-11-13
Icon size for x-small button is too big


naamagr:
Prefix icon in x-small button is too big (it should be like the size of icons in small button).
![image](https://user-images.githubusercontent.com/29533829/48415422-06aced00-e756-11e8-85f4-0367b1564812.png)
  
This issue can be reproduced in storybook.

mykas:
Hi, not so long ago fix for button affixes has been released. Are you sure that those are too big?

naamagr:
You're right.
Sorry I was using an older version.
However, margins are still a bit too small:
![image](https://user-images.githubusercontent.com/29533829/48421813-e8e78400-e765-11e8-9c3a-4c92a28c3dc6.png)

mykas:
We have UX guidelines for this and its 6px

mykas:
<img width="935" alt="screenshot 2018-11-13 at 17 13 56" src="https://user-images.githubusercontent.com/8919190/48422536-87281980-e767-11e8-8dc0-1c6fc9207e6f.png">

shlomitc:
@mykas - the spec is taking into consideration that the `tiny` button should use smaller icons
![image](https://user-images.githubusercontent.com/6093192/48428344-7a112780-e773-11e8-93f0-c21d08ae3fb8.png)

So unless I miss something, it looks like the component is implemented correctly but the zero margin is problematic for icons which are not Add or Cheveron icons.

@litalwix - can you please assist here?

mykas:
I've talked with @milkyfruit  and he agreed that we should add 3px margin to this. Thanks for noticing this @naamagr and  @shlomitc.

milkyfruit:
@naamagr updated Button's specs based on @mykas comment:
https://zpl.io/aRq9ZKv

![image](https://user-images.githubusercontent.com/1684587/48482513-8d76ce00-e819-11e8-8b3a-d9f6d0c877c7.png)

mykas:
https://github.com/wix/wix-style-react/pull/2470

shlomitc:
fixed

PR #2470, opened 2018-11-14
<Button /> - Increase margin for prefix and suffix icons on size x-small

mykas:
This Pr increases margin value for suffix and prefix icon on size `x-small`. 

image

https://github.com/wix/wix-style-react/issues/2459


mykas:
@milkyfruit  Can you check? https://wix-wix-style-react-pr-2470.surge.sh/

PR #2477, opened 2018-11-16
docs(Notification): WIP example with wix-storybook-utils sections

argshook:
Please don't merge, this is WIP and is just for showcase.

This is part of our storybook design update effort.
Idea is to have more flexibility in wix-storybook-utils which allows controlling what parts should be displayed in docs.

depends on https://github.com/wix/wix-ui/pull/900

### DIFF
--- a/.storybook/common/stories.js
+++ b/.storybook/common/stories.js
@@ -81,6 +81,7 @@ import '../../stories/Tooltip/Composite/CompositeStory';
 // 8. Notification Bars
 // 8.1 Standard, 8.2 Error, 8.3 Success, 8.4 Warning, 8.5 Premium
 import '../../stories/Notification';
+import '../../stories/Notification/index.story.js';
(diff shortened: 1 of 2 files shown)


hepiyellow:
@argshook @shlomitc @yardnsm @mykas 

I suggest not inventing a new `sections` API,
and provide a `<MetadataProvider/>`
and React components that consume it like:
- `<StoryPage/>`
- `<StoryTabs/>`
  - `<Usage/>`, `<API/>`, `<Testkit/>`
- `<Playground/>`
  - `<Playground.Controls/>`
  - `<Playground.Preview/>`
  - `<Playground.Code/>`
- `<PropsTable/>`
etc...

This is part of a wider API proposal which you can find in [this issue](https://github.com/wix/wix-ui/issues/943).

argshook:
Lets talk about it again on monday. What you propose is pretty much exactly what i want to achieve, except its more verbose and tangled with react and storybook. I would argue also that your proposal has more "API points" to invent. But lets talk on monday

PR #2480, opened 2018-11-19
<CloseButton/> - component refactor and expose to users

mykas:
This PR recreates `<CloseButton/>` and exposes it to users. 

Proposed Component API:
```javascript
<CloseButton
  onClick={callback}
  dataHook="data-hook"
  skin="standard | standardFilled | light | lightFilled | dark "
  size="small | medium"
  className="to add additional classes"
  disabled={true | false}
/>
```


stories/CloseButton/examples/index.js
```diff
@@ -0,0 +1,53 @@
...
+
+const halfColumn = {
+  display: 'flex',
+  flexDirection: 'column',
+  padding: '10px',
+  width: '48%',
+  lineHeight: '1.6'
+};
+
+const Container = ({children}) => <div style={controlledWidth}>{children}</div>; //eslint-disable-line
```
yardnsm: I would replace it with smth like:

```jsx
const Container = props => <div style={controlledWidth} {...props}/>;
```

in order to get rid of the eslint comment.

mykas: true

.storybook/common/stories.js
```diff
@@ -64,10 +64,9 @@ import '../../stories/Button/standard.story.js'; // 5.1 Standard
 import '../../stories/Button/white.story.js'; // 5.2 White
 import '../../stories/Button/icon.story.js'; // 5.3 Icon
 import '../../stories/TextButton/index.story'; // 5.3 Text Button
-import '../../stories/Button/error.story.js'; // 5.4 Error
+import '../../stories/CloseButton/index.story'; // 5.3 Close Button
```
yardnsm: Shouldn't it be `5.4` rather than `5.3`?

mykas: Yes

src/CloseButton/CloseButton.js
```diff
@@ -1,44 +1,50 @@
...
-};
-
-CloseButton.defaultProps = {
-  arialLabel: 'close button',
-  theme: 'close-transparent',
-  size: 'large'
-};
-
-CloseButton.displayName = 'CloseButton';
+class CloseButton extends Component {
```
yardnsm: Why using a class? It doesn't have any internal state.

mykas: There is no significant performance differences between classes and functions even though parsing functions is easier. I'd prefer functions - I don't think we as a team have any rules for this. If we gonna decide to go with functions - sure I'll change that. Right now it is not so important.

argshook: it doesn't matter, but it's interesting to see differences.

below is transpilled code of same component done with class and then with function

class
```js
var App = function (_React$Component) {
  _inherits(App, _React$Component);

  function App() {
    _classCallCheck(this, App);

    return _possibleConstructorReturn(this, _React$Component.apply(this, arguments));
  }

  App.prototype.render = function render() {
    return react__WEBPACK_IMPORTED_MODULE_0___default.a.createElement(
      'div',
      null,
      'yeah boi'
    );
  };

  // @ts-ignore
  App.prototype.__reactstandin__regenerateByEval = function __reactstandin__regenerateByEval(key, code) {
    // @ts-ignore
    this[key] = eval(code);
  };

  return App;
}(react__WEBPACK_IMPORTED_MODULE_0___default.a.Component);

var _default = App;
```

function:
```js
var App = function App() {
  return react__WEBPACK_IMPORTED_MODULE_0___default.a.createElement(
    'div',
    null,
    'yeah boi'
  );
};

var _default = App;
```

first glance it would seem that class looses. but this is very isolated example and bigger picture might be different.

if we compare bundles of a reasonable size codebase with classes and then with functions, maybe we'd see clear winner. until we do that, i guess it's a matter of taste and changing class into function for the sake of changing isn't worth it

testkit/enzyme.js
```diff
@@ -1,96 +1,96 @@
 import {
   enzymeTestkitFactoryCreator,
-  enzymeUniTestkitFactoryCreator,
+  enzymeUniTestkitFactoryCreator
```
yardnsm: where did the trailing comma go? 😲

mykas: well we have some new lint rules in yoshi3 which deletes it? ... I am not sure if I have correct setup of prettier in my vscode, but I am pretty sure that it uses yoshi one because it has changed.

mykas: Can you check on your side if this happens too?

yardnsm: The default lint rules _should_ keep those. Running `npm run lint` works with no errors?

yardnsm: I check the build in CI, `eslint` throws warnings on these, not errors. Same thing locally. Weird...

mykas: Ok, I have figure out. It seems when I use prettier plugin directly - it will ignore some of the rules from the yoshi. Don't know why. switched to eslint autosave functionality and now it will have trailling commas.

PR #2481, opened 2018-11-20
<TextButton /> - remove backoffice namespace

mykas:
This PR removes BackofficeTheme wrapper. The reason for this is because `TextButton` is variation of core Button. TextButton can only be achieved through classnames.


src/TextButton/TextButton.driver.js
```diff
@@ -1,6 +1,5 @@
 import {buttonNextDriverFactory} from 'wix-ui-core/dist/src/components/button-next/button-next.driver';
 
 export const textButtonDriverFactory = base => {
-  base = base.$('[data-hook="textButton-core"]');
   return buttonNextDriverFactory(base);
```
argshook: export { buttonNextDriverFactory as textButtonDriverFactory } from 'wix-ui-core/dist/src/components/button-next/button-next.driver';

mykas: THanks

PR #2482, opened 2018-11-20
`<Avatar/>` - initial (In WIP category)

hepiyellow:
Although open issue and missing feature, I think we can merge this, as Avatar is no in the WIP category.

## Open issues & Missing features
See [Project](https://github.com/wix/wix-style-react/projects/16)

## UX review
@wuwa @milkyfruit @LittleTal 
Please review.
The story is [here](https://wix-wix-style-react-pr-2482.surge.sh/?selectedKind=WIP&selectedStory=12.6%20Avatar&full=0&addons=0&stories=1&panelRight=0)
You may also use the Test stories:
 - [Sizes](https://wix-wix-style-react-pr-2482.surge.sh/?selectedKind=Tests%2FWIP%2F12.6%20Avatar&selectedStory=1.%20Avatar%20sizes&full=0&addons=0&stories=1&panelRight=0)
- [Colors](https://wix-wix-style-react-pr-2482.surge.sh/?selectedKind=Tests%2FWIP%2F12.6%20Avatar&selectedStory=2.%20Avatar%20colors&full=0&addons=0&stories=1&panelRight=0)
- [Placeholder](https://wix-wix-style-react-pr-2482.surge.sh/?selectedKind=Tests%2FWIP%2F12.6%20Avatar&selectedStory=3.%20Avatar%20placeholder&full=0&addons=0&stories=1&panelRight=0)
### Known issues:
- No focus ring - currently we decided that Avatar is not focusable. We may discuss it.
### To Be Fixed
- With size 36 and 30 ,the text is 1 px high (only in playground)
- In Sizes test page, if I enlarge the storybook sidebar, then the Avatars gets squashed (width!=height)


stories/Avatar/avatar.story.js
```diff
@@ -0,0 +1,30 @@
...
+    placeholder: false
+  },
+  exampleProps: {
+    size: ['size90', 'size72', 'size60', 'size48', 'size36', 'size30', 'size24', 'size18'],
+    color: ['blue', 'green', 'grey', 'red', 'orange'],
+    imgProps: [
+      {label: 'With Image', value: {src: IMG_REAL_URL}},
+      {label: 'With Invalid Image URL', value: {src: IMG_INVALID_URL}}
+    ],
+    icon: Object.entries(icons).map(([name, icon]) => ({label: name, value: React.createElement(icon)}))
```
mykas: Do you want here to show a list of icons to choose from?

src/Avatar/Avatar.e2e.js
```diff
@@ -0,0 +1,44 @@
...
+import {avatarTestkitFactory} from '../../testkit/protractor';
+import {storySettings} from '../../stories/Avatar/storySettings';
+import {getTestStoryKind} from '../../stories/storiesHierarchy';
+
+describe('Avatar', () => {
+  describe('AutoStory Page', () => {
+    const avatarDriver = avatarTestkitFactory({dataHook: storySettings.dataHook});
+    const url = createStoryUrl({kind: storySettings.category, story: storySettings.storyName});
+
+    describe('driver sanity', () => {
```
mykas: We have automatic testkit tests now in https://github.com/wix/wix-style-react/blob/master/testkit/testkit-smoke.test.js. Make sure to add your component in Components list with required props and unidriver flag.

src/Avatar/Avatar.js
```diff
@@ -0,0 +1,70 @@
...
+  const {size, color, dataHook, icon, placeholder, ...rest} = props;
+
+  return (
+    <div className={classNames(backofficeTheme, props.className)}>
+      <CoreAvatar
+        {...rest}
+        data-hook={dataHook}
+        icon={placeholder ?
+          (
+            <svg
```
mykas: 🤔 this is temporally, right?

src/Avatar/Avatar.js
```diff
@@ -0,0 +1,70 @@
...
+  imgProps: PropTypes.object
+};
+
+Avatar.propTypes = {
+  ...CoreAvatarPropTypes,
+  size: PropTypes.oneOf(['size90', 'size72', 'size60', 'size48', 'size36', 'size30', 'size24', 'size18']),
+  /* Bakcground color when text (initials) content is displayed */
+  color: PropTypes.oneOf(['blue', 'green', 'grey', 'red', 'orange']),
+  /* When true, shows a placeholder icon. Overrides the `icon` property if provided. */
+  placeholder: PropTypes.bool,
```
argshook: can this be the default? i mean, if no `icon` provided, then show some placeholder. then no need for this prop

hepiyellow: Actually, you are right, this is the default according to the spec.
But then how would the consumer choose to render initials?
The core Avatar component spec and logic is to render these as if provided by priority:
image -> icon -> test

When text is initials generated from the name.

So if consumer renders:
```js
<Avatar name="John Doe"/>
```
We will get a placeholder.
if they even render :
```js
<Avatar text="JD"/>
```
We will render a placeholder, since `icon` is not provided.

src/Avatar/Avatar.js
hepiyellow: why ?

stories/Avatar/avatar.story.js
hepiyellow: yes

src/Avatar/Avatar.e2e.js
hepiyellow: Well, since we have a WSR wrapper on the core's component.
It could be that the driver won't work.
For example if we would have had the backoffice theme div wrapper inside the component. It could break the driver.
So having the generic sanity smoke test is not enough.
We need to test all the driver methods.

hepiyellow:
@wuwa @milkyfruit @LittleTal 
Could someone review the Avatar?
See notes in the PR description on top.

src/Avatar/Avatar.js
hepiyellow: Removed `placeholder` prop

hepiyellow:
@milkyfruit @wuwa @shlomitc 
## Issue with `tiny` text having an odd `line-height` (`15px`)

About the `size30` and `size36` having the text 1px to high.
It's kind of your fault (-:
You defined the `line-height` of `tiny` typography as `15px`.
And `15px` can NOT be vertically centred properly in an even px height.
So it depends on the browser's current Zoom factor, if it is shown as centred or not.

I recall this issue also with Button with size `tiny`.

I can override the tiny style and make it line-height `16px` to solve it.

But I think you should either fix the `15px` in the Typography definition of `tiny` size.
Or you specify an even `line-height` in any spec that uses `tiny` text, and is expected to be vertically centred.

PR #2483, opened 2018-11-21
<MessageBoxMarketerialLayout> - add prop to disable primary button

liorur:
### What changed
add prop to disable primary button in MessageBoxMarketerialLayout

- [v ] Change is tested
- [v ] Change is documented
- [v ] Build is green


wuwa:
yes, although would prefer to see use case.

liorur:
@wuwa  - the use case is not to allow continuing before accepting terms:
screen shot 2018-11-21 at 20 43 25

wuwa:
cool, approved

stories/MessageBox/AnnouncementExamples/DisabledAction.js
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable react/prop-types */
+import React from 'react';
+import {MessageBoxMarketerialLayout} from 'wix-style-react/MessageBox';
+
+export default () => (
```
yardnsm: Please add an eyes test for this example.

https://github.com/wix/wix-style-react/blob/master/src/MessageBox/MessageBox.e2e.js#L80

PR #2484, opened 2018-11-21
Port dom-test-container

mykas:


test/utils/vanilla-uni-helpers.js
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import {reactUniDriver} from 'unidriver';
+
+// At the moment our tests support both Jsdom and browser environment.
```
hepiyellow: Remove comments about in browser tests

test/utils/vanilla-uni-helpers.js
```diff
@@ -0,0 +1,77 @@
...
+    return this;
+  }
+
+  unmountAfterEachTest() {
+    beforeAll(() => this.create());
+    afterEach(() => this.unmount());
+    afterAll(() => this.destroy());
+    return this;
+  }
+
```
hepiyellow: What about the `createLegacyRenderer`?
We should have it, just call it `createDriver`

test/utils/vanilla-uni-helpers.js
```diff
@@ -0,0 +1,77 @@
...
+      this.unmount();
+      this.destroy();
+    });
+    return this;
+  }
+
+  // Adapter for react based uni driver
+  createUniRenderer(driverFactory) {
+    return jsx => {
+      ReactDOM.render(jsx, this.node);
```
hepiyellow: This render is async.
Where did you take this file, it is NOT what there is in the core.
Please Port again (-:

mykas: this is not ready yet. Come on.

mykas: Ok, now I remember what happened. 
Here how it looked before:
```
public renderSync<P>(jsx: React.ReactElement<P>) {
    return ReactDOM.render(jsx, this.node);
  }
```

why do we need this method to be exposed?

mykas: I have moved the `ReactDOM.render(jsx, this.node);` directly to the renderer.

mykas: Oh I see you using it in avatar directly without a driver?

test/utils/vanilla-uni-helpers.js
mykas: I have named it createRenderer

mykas: Because it also renders the dom

test/utils/vanilla-uni-helpers.js
mykas: Added renderSync with a promised return

test/utils/vanilla-uni-helpers.js
```diff
@@ -0,0 +1,91 @@
...
+  destroy() {
+    this.node.remove();
+    return this;
+  }
+
+  renderSync(jsx) {
+    return new Promise(resolve => resolve(ReactDOM.render(jsx, this.node)));
+  }
+
+  // This function's signature should be:
```
hepiyellow: Remove Typescript comments

test/utils/vanilla-uni-helpers.js
```diff
@@ -0,0 +1,91 @@
...
+    return this;
+  }
+
+  destroy() {
+    this.node.remove();
+    return this;
+  }
+
+  renderSync(jsx) {
+    return new Promise(resolve => resolve(ReactDOM.render(jsx, this.node)));
```
hepiyellow: Why do you wrap with a promise? This method is sync

test/utils/vanilla-uni-helpers.js
```diff
@@ -0,0 +1,91 @@
...
+      this.unmount();
+      this.destroy();
+    });
+    return this;
+  }
+
+  // Adapter for driver
+  createRenderer(driverFactory) {
+    return jsx => {
+      this.renderSync(jsx);
```
hepiyellow: renderSync returns a promise. either change it NOT to return a promise (like it is in wix-ui-core) or add `await` here.

test/utils/vanilla-uni-helpers.js
```diff
@@ -0,0 +1,91 @@
+import React from 'react';
```
hepiyellow: Why do you call this file `vanilla-uni-helpers`?
It export a class called `ReactDOMTestContainer`, that could be a good name for the file as well.

In principal, it does 2 things:
- renders to a test container (with unmount at end of test)
- creates test drivers from the rendered element.

I would be happy also if those 2 things were separate. Meaning that the `createRenderer` and `createUniRenderer` would not be part of the `ReactDOMTestContainer`.

I mean , anyways, the consumer does something like:
```js
import { ReactDOMTestContainer } from '../../test/dom-test-container';

const createDriver = new ReactDOMTestContainer()
    .unmountAfterEachTest()
    .createLegacyRenderer(ImageViewerDriverFactory);
```

So they can do:
```js
import { ReactDOMTestContainer , driverFactory} from '../../test/';

const createDriver = driverFactory(
  new ReactDOMTestContainer().unmountAfterEachTest(), 
  ImageViewerDriverFactory
);
```

And for unidriver:
So they can do:
```js
import { ReactDOMTestContainer , uniDriverFactory} from '../../test/';

const createDriver = uniDriverFactory(
  new ReactDOMTestContainer().unmountAfterEachTest(), 
  ImageViewerUniDriverFactory
);
```

test/utils/vanilla-uni-helpers.js
mykas: Anyway, I've made this method async instead of sync now.

test/utils/ReactDOMTestContainer.js
```diff
@@ -0,0 +1,87 @@
...
+    return this;
+  }
+
+  destroy() {
+    this.node.remove();
+    return this;
+  }
+
+  renderAsync(jsx) {
+    return new Promise(resolve => resolve(ReactDOM.render(jsx, this.node)));
```
argshook: here's the implementation of `ReactDOM.render` [here](https://github.com/facebook/react/blob/master/packages/react-dom/src/client/ReactDOM.js#L665-L668)

as you see it accepts `callback` as third argument. you can use that in order to make `renderAsync` resolve at correct time
```suggestion
    return new Promise(resolve => ReactDOM.render(jsx, this.node, resolve));
```

mykas: Oh, I see

test/utils/vanilla-uni-helpers.js
mykas: This is mainly the reason for lint rule `ReactDOM.render() currently returns a reference to the root ReactComponent instance. However, using this return value is legacy and should be avoided because future versions of React may render components asynchronously in some cases. `

mykas:
I have added some tests. I think we can merge this.

hepiyellow:
@mykas
Can you remove renderWithRef?
It is not inuse, it is not tested,
And i have no idea what it does.

mykas:
Sure

hepiyellow:
`ReactDomTestContainer` is already merged (via the Yoshi 3 PR).
This introduces a few minor API changes.

I suggest we SOON move to using [`react-testing-library`](https://github.com/kentcdodds/react-testing-library) which is the new very popular standard.
It is very similar, and I estimate little effort.
It is good time to do it, before we make our ReactDomTestContainer deeply rooted.

With that said,
we can still use ReactDOMTestContainer as a Wrapper over `react-testing-library`,
and keep it's API.
Making it enhance the basic testing API, with sugaring for our test driver creation,
and maybe custom cleanup.

PR #2487, opened 2018-11-21
Remove lodash costly operators

giladsegal:
### What changed

Remove expensive lodash operators

### Why it changed

reduce the bundle size :)

---

- [X] Change is tested
- [X] Build is green

Action items (TBD):
- [ ] There are is still one usage of 3rd party `omit` package which I couldn't remove because it relies over iterating over `PropTypes` (it weights something like 1kb - but this is the only usage)
- [x] CardWithEditableSelector story uses `lodash/cloneDeep` - **doesn't affect production bundle size**
- [ ] GoogleAddressInput test file uses `lodash/fp` - **doesn't affect production bundle size**


src/Table/BulkSelection/BulkSelection.js
```diff
@@ -95,6 +94,17 @@ export class BulkSelection extends React.Component {
     });
   }
 
+  areSelectedIdsEqual = (selectedIds1, selectedIds2) => {
+    if ((selectedIds1 === undefined && selectedIds2 === undefined) || (selectedIds1 === null && selectedIds2 === null)) {
```
shlomitc: can be simplified?

src/DropdownLayout/DropdownLayout.js
```diff
@@ -305,19 +302,21 @@ class DropdownLayout extends WixComponent {
...
+    if (nextProps.options.some(option => !this.isLegalOption(option))) {
+      throw new Error(`DropdownLayout: Invalid options provided: ${nextProps.options}`);
+    }
 
-      if (this.state.hovered !== NOT_HOVERED_INDEX) {
-        this.setState({
-          hovered: findIndex(nextProps.options, item => item.id === this.props.options[this.state.hovered].id)
-        });
-      }
+    if (this.state.hovered !== NOT_HOVERED_INDEX && this.props.options[this.state.hovered].id !== nextProps.options[this.state.hovered].id) {
```
shlomitc: just make sure there is a test for this somewhere

src/DropdownLayout/DropdownLayout.js
```diff
@@ -305,19 +302,21 @@ class DropdownLayout extends WixComponent {
...
+      this.setState({
+        hovered: this.findIndex(nextProps.options, item => item.id === this.props.options[this.state.hovered].id)
+      });
     }
   }
 
+  findIndex(arr, predicate) {
```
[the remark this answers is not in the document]
gilad-segal: This is the only case, I used a function here because I wasn't sure about the edge cases - whether the options can be undefined and I didn't want to break anything

package.json
```diff
@@ -192,6 +192,7 @@
       "react/no-find-dom-node": 0,
       "react/no-string-refs": 0,
       "react/jsx-boolean-value": 2,
+      "no-unused-vars": ["error", { "ignoreRestSiblings": true }],
```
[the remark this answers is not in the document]
shlomitc: Yes, it's a bit annoying, but there shouldn't be too many of these and it means you know exactly what you do.
all other cases (unused vars) should be reported.

src/Table/BulkSelection/BulkSelection.js
shlomitc: Whatever you think

PR #2493, opened 2018-11-22
Migrate to Yoshi 3 

hepiyellow:
## Helping conflict resolution

- This PR includes 2 commits
  - Migrate to Yoshi 3
  - Run `lint --fix` 

After this PR is merged (Planned Monday at noon),
You will have difficult conflicts with master.

### How to resolve conflicts easily
The your PR's branch,
- cherry-pick the 1st commit of this PR (Migrate to Yoshi 3)
  - resolve conflicts if exist
- Run `npm run lint -- --fix`

You should be able to merge now with not conflicts. (or at least easy ones).

## Updating this PR with master changes

- Reset this branch to the 1st commit
- Rebase ontop of master
- run `npm run lint -- --fix`
- commit and push

## Code review
If you wish to review the migration, then look only at the first commit.

## Open Issues
- disable removal of propTypes
- `dist/es` issue


stories/Badge/ExampleBadges.js
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable no-duplicate-imports */
+/* eslint-disable import/no-duplicates */
```
yardnsm: Why not remove this comment and instead use

```js
import Badge, {SIZE, SKIN, TYPE} from 'wix-style-react/Badge';
```

test/dom-test-container.js
```diff
@@ -0,0 +1,87 @@
+import * as React from 'react';
```
yardnsm: What about https://github.com/wix/wix-style-react/pull/2484?

src/common/Focusable/FocusableHOC.js
```diff
@@ -153,5 +151,26 @@ export const withFocusable = Component => {
...
+ * So if we simply do FocusableHOC.propTypes = Component.propTypes, it is being stripped away.
+ * 
+ * This later becomes a problem if another component defines:
+ * <code>
+ * Comp.propTypes = {
+ *   prop1: SomeFocusableComp.propTypes.prop1
+ * }
+ * </code>
+ */
+function assignPropTypesHack(targetClass, propTypes) {
```
yardnsm: I checked and this plugin does remove the `PropTypes` for each component during build time.

I think we should remove the `babel-plugin-transform-react-remove-prop-types`. Since WSR is a library, we would like to always include PropTypes for our components, so our consumers will get warnings when they're using a component in a wrong way.

This would also eliminate the need to this "hack".

hepiyellow:
@yardnsm can you enlighten me about the `dist/es` thing.
Isn't this something that yoshi 3 does?
Does the `dist/es` suppose to support tree shaking?

src/common/Focusable/FocusableHOC.js
hepiyellow: @yardnsm I thought that the `babel-plugin-transform-react-remove-prop-types` works only when building production. But now that mention it, when we release WSR, we build with production flag. So you might be right here.

@ronami Do you recommend NOT using `babel-plugin-transform-react-remove-prop-types` in libraries?

stories/Badge/ExampleBadges.js
hepiyellow: done

yardnsm:
@hepiyellow 

To support tree-shaking, we also provide a WSR version which is tranpiled normally **but** retains the ES modules syntax (`import`s, `export`s). This version is located in `dist/es` and its main purpose is to support tree shaking, as most build tools (such as Webpack) relies on the static structure of this syntax.

I'm not sure what is the current configuration that make Yoshi able to do that, maybe its better to talk with @ranyitz or @ronami.

argshook:
does tree-shaking have to be solved here? we don't officially support it anyway.  i would do it as a separate task

src/common/Focusable/FocusableHOC.js
hepiyellow: @yardnsm BTW, also if we disable propTypes removal, when the consumer builds with propTypes removal, the prod code will break if we don’t keep this hack.

hepiyellow:
## Status report

So, in order to disable removal of propTypes,
the suggested solution is to use:
```
"babel": {
    "presets": [
      "env",
      "react"
    ],
```

But we need to "copy" the yoshi configuration of the `env` preset (according to [this logic](https://github.com/wix/yoshi/blob/master/packages/babel-preset-yoshi/index.js)), without the `babel-plugin-transform-react-remove-prop-types` plugin.

Good luck?
@argshook @mykas @yardnsm  Any volunteers of creating this `env` configuration?

Issue #2497, opened 2018-11-26
GenericModalLayout


adavidof:
### Relevant Components
GenericModalLayout

### Current Behavior
Bug with name in storybook

### Expected Behavior
Correct name: `GenericModalLayout`

### Steps to Reproduce
https://wix-wix-style-react.surge.sh/?selectedKind=Components&selectedStory=GenericModalLayout&full=0&addons=0&stories=1&panelRight=0
![image](https://user-images.githubusercontent.com/10529679/49011915-7375ad80-f180-11e8-8b75-fc128e609757.png)

### Severity
Low

shlomitc:
Thanks. Can you please add a displayName to the component?

adavidof:
@shlomitc 
already added
https://github.com/wix/wix-style-react/blob/master/src/GenericModalLayout/GenericModalLayout.js#L52

shlomitc:
merged

PR #2498, opened 2018-11-26
fix: GenericModalLayout component name in storybook

adavidof:
### What changed
Fix correct name

### Why it changed
Bug

#2497


adavidof:
@shlomitc 
In the example as well

In the story I import another component because I need button to show layout in the modal:
https://github.com/wix/wix-style-react/blob/master/stories/GenericModalLayout/GenericModalLayout.story.js#L14
https://github.com/wix/wix-style-react/blob/master/stories/GenericModalLayout/ExampleGeneric.js#L10

PR #2505, opened 2018-11-27
`<Tags/>` - add missing e2e tests and RTL tests

hepiyellow:


src/Tag/Tag.e2e.js
```diff
@@ -17,14 +18,32 @@ describe('Tag', () => {
...
-      await waitForVisibilityOf(tagDriver.element(), 'Cannot find <Tag/>');
-      autoExampleDriver.setProps({ size });
-      await eyes.checkWindow(`${size} size`);
-      autoExampleDriver.setProps({ removable: false });
-      await eyes.checkWindow(`${size} size: without remove button`);
+  function sizesTests(autoExampleProps) {
+    function setProps(props) {
+      return autoExampleDriver.setProps({
+        ...props,
+        ...createAutoExampleProps(autoExampleProps),
```
yardnsm: I would suggest adding this kind functionality (toggle between RTL on and off) to the `autoExampleDriver` itself. This would also beneficial for other libraries that uses the driver, such as `wix-ui-core`, and would reduce boilerplate code (for example, that line comment) and weird hacks such as [this query string](https://github.com/wix/wix-style-react/blob/186907ef5008dcf565860aaf36233c948f8c8255/stories/AutoExampleWrapper.js#L16).

The API could be like the following:

```js
// Toggle RTL on
await autoExampleDriver.setRtl(true);

// Toggle RTL off
await autoExampleDriver.setRtl(false);

// Calling `reset` will also reset the RTL state to off
await autoExampleDriver.reset();
```

This means that when viewing the story `iframe`, we should wrap the component in some div and add the appropriate attributes to it when we toggle on RTL, just like in the Playground.

Adding the dark background thing maybe also good?

@argshook wdyt?

Issue #2412, opened 2018-11-05
<Button />  - add min-width to component


milkyfruit:
EDITED: Added min-width default values for each button size.

# Feature Request

### Description
[`<Button />`](https://wix-wix-style-react.surge.sh/?selectedKind=5.%20Buttons&selectedStory=5.1%20Standard&full=0&addons=0&stories=1&panelRight=0) component needs a `min-width` property.
It should have a default value, which varies based on button size:

`min-width` default values by button size:

Large – `102px`
Medium – `84px` 
Small – `72px`
Tiny – `60px`

The overrides are allowed.

### Severity
Critical / Major

Wix Restaurants requested for this prop in their product.

### Link to specs
https://zpl.io/aRq9ZKv

shlomitc:
@milkyfruit - is this the requirement for the default behavior of a button? or have it as an optional way of controlling the min-width?

milkyfruit:
@shlomitc `min-width` should always exist with some default value. Each button size has its own default value. It should be allowed to override it if needed.

Updated [the specs](https://zpl.io/aRq9ZKv) specifying default `min-width` default values:
![image](https://user-images.githubusercontent.com/1684587/48545726-44815100-e8cf-11e8-988c-4ca57055cf61.png)

mykas:
Done

PR #2507, opened 2018-11-27
<Button/> - refactor

mykas:
Proposed Component API:
```javascript
<Button
  onClick={callback}
  dataHook="data-hook"
  skin="standard | inverted | destructive | premium | dark | light | transparent"
  priority="primary | secondary"
  size="tiny | small | medium | large"
  fullWidth={true | false}
  suffixIcon={Icon}
  prefixIcon={Icon}
  className="to add additional classes"
  disabled={true | false}
>
Button
</Button>
```

New features compared to the previous one:
- All buttons sizes are defined with specified `min-width`. [issue#2412](https://github.com/wix/wix-style-react/issues/2412)
- `MatchParent` prop is now defined as `FullWidth`.  
- FocusBox css - should have high z-index. [issue#1935](https://github.com/wix/wix-style-react/issues/1935)
- Create loading state

Deprecation log for previous Button:
`
'Using "Button" with current api is deprecated. In order to upgrade to the new Button api just pass prop "upgrade" and follow "5.1 Button" changed api docs. For old api refer to "5.1 Button" under "Deprecated" section.'
`

This PR will include:
-  A switcher between old Button and new Button just by passing prop `upgrade` to the Button [x]
-  Migration guide. [x]
- Deprecation log for previous button. [x]


protractor.conf.js
```diff
@@ -2,6 +2,18 @@ module.exports.config = {
   specs: ['test/**/*.e2e.js', 'src/**/*.e2e.js'],
   baseUrl: `http://localhost:6006/`,
   jasmineNodeOpts: { defaultTimeoutInterval: 60000 },
+  capabilities: {
```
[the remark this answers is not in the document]
mykas: not related, just tried to fix timeout. Gonna remove this one.

src/Button/Button.js
```diff
@@ -0,0 +1,65 @@
...
+      skin,
+      priority,
+      size,
+      children,
+      className,
+      dataHook,
+      ...rest
+    } = this.props;
+
+    const classNames = cx(className, button(skin, priority, size));
```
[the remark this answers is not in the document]
mykas: hmm... theoretically yes, but I think className ordering works only in the stylesheet maybe? Because this would mean that `class="class1 class2"` would resolve to class2? But how does it know in which order stylesheets are compiled?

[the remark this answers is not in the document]
yardnsm: From [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity):

> When multiple declarations have equal specificity, the last declaration found in the CSS is applied to the element.

So the order doesn't matter :)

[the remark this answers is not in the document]
mykas: @yardnsm `the last declaration found in the css is applied`.  This means that it matters, but what we can't really now where does custom `className` appears in the `<head><style src="index.css"/></head>` tag.

src/Backoffice/Button/Button.js
```diff
@@ -5,7 +5,11 @@ import WixComponent from '../../BaseComponents/WixComponent';
 import ButtonLayout from '../../ButtonLayout/ButtonLayout';
 import { withFocusable, focusableStates } from '../../common/Focusable';
 import { pickAccessibilityProps } from '../../common/accessibility';
+import deprecationLog from '../../utils/deprecationLog';
 
+deprecationLog(
+  'Using "Button" with current api is deprecated. In order to upgrade to the new Button api just pass prop "upgrade" and follow "5.1 Button" changed api docs. For old api refer to "5.1 Button" under "Deprecated" section.',
```
shlomitc: making it shorted if possible. you can remove the old api explanation.
Write the migration in a markdown file and add a link to it

src/Button/Button.e2e.js
```diff
@@ -0,0 +1,24 @@
...
+
+  beforeAll(async () => {
+    await browser.get(storyUrl);
+  });
+
+  beforeEach(() => {
+    return autoExampleDriver.remount();
+  });
+
+  eyes.it('Make a screenshoft of all Button examples', () => {
```
shlomitc: use the `eyesInstance` and disable `enableSnapshotAtBrowserGet`. this can actually be a default configuration in my opinion
https://github.com/wix/wix-style-react/blob/d1a176172bca8ff0065dda20261fabea928b59e0/src/Avatar/Avatar.e2e.js#L32

src/Button/Button.js
```diff
@@ -0,0 +1,79 @@
...
+      size,
+      fullWidth,
+      children,
+      className,
+      dataHook,
+      upgrade,
+      ...rest
+    } = this.props;
+
+    const styles = fullWidth ? { width: '100%' } : {};
```
shlomitc: can't it be done in CSS? I prefer having another class for it

stories/Button/examples/ExamplesLoading.js
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable */
+<div>
+  <Button upgrade><Loader size="tiny" /></Button>
```
shlomitc: the Loader is misaligned
image

stories/Button/index.story.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import Button from 'wix-style-react/Button';
+import { storySettings } from './storySettings';
+import icons from '../utils/icons-for-story';
+
+import ButtonStory from './examples';
+
+export default {
```
shlomitc: Should have here a notification that the old Button API is deprecated and give a link to the old docs.

stories/Button/index.story.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import Button from 'wix-style-react/Button';
+import { storySettings } from './storySettings';
+import icons from '../utils/icons-for-story';
+
+import ButtonStory from './examples';
```
shlomitc: testkit docs are missing. should add them manually if Autodocs can't get it

src/Button/Button.e2e.js
mykas: Ok

stories/Button/index.story.js
mykas: done

stories/Button/examples/ExamplesLoading.js
mykas: So here we have an example where if we don't control buttons internal css for loading - we can't get proper alignment. I remember Lior told me that we should not bundle the loading state inside our button, but what was the reason for this? It could be easier for me just render conditionally and also apply correct content styling.

src/Button/Button.js
mykas: Sure, it can.

stories/Button/examples/ExamplesLoading.js
mykas: Here we have problem with `line-height` though because button adds line-height based on size.

stories/Button/index.story.js
mykas: Notification in storybook?

shlomitc: just some text bold enough to describe the deprecation

stories/Button/examples/ExamplesLoading.js
shlomitc: bundling the `Loader` component will increase the button size, which is not required for most use cases.

Maybe we should have a class that lets you align icons - but then the question is why icon alignment is different the text alignment

mykas: Solved it by adding display:flex and align-items:center on content div.

stories/Button/index.story.js
mykas: Done

src/Backoffice/Button/Button.driver.js
```diff
@@ -2,26 +2,34 @@ import React from 'react';
...
-    );
-    ReactDOM.render(
-      <div ref={r => (element = r)}>{ClonedWithProps}</div>,
-      wrapper,
-    );
-  },
-});
+import { buttonDriverFactory } from '../../Button/Button.driver';
+
+export default ({ element, wrapper, component, upgrade }) => {
```
shlomitc: in user code, the `upgrade` flag should be propagated through `teskit/enzyme`'s `buttonTestkitFactory`

mykas: Hey, what do you mean? I have only an idea to update the `testkitFactoryCreator` from `wix-ui-test-utils` to accept a third argument `upgrade` which could be passed to the driver. And there I could switch between two testkits.

shlomitc: Sounds wrong in my opinion.
Better to have something on the component that helps understand if the rendered component is the upgraded one or not.
For example. Some data attribute on the root element. Based on that - return the right driver

mykas: I know. Tried it. Couldn't manage to make it work.

mykas:
@milkyfruit can you check?

src/Backoffice/Button/Button.driver.js
mykas: Solved

mykas:
@milkyfruit fixed. Please check again when storybook will be updated.

src/Backoffice/Button/Button.js
```diff
@@ -5,7 +5,11 @@ import WixComponent from '../../BaseComponents/WixComponent';
 import ButtonLayout from '../../ButtonLayout/ButtonLayout';
 import { withFocusable, focusableStates } from '../../common/Focusable';
 import { pickAccessibilityProps } from '../../common/accessibility';
+import deprecationLog from '../../utils/deprecationLog';
 
+deprecationLog(
```
hepiyellow: You should put this deprecationLog in PropTypes validation, so that if consumer do pass `upgrade` prop, they won't see this message.

mykas: This is different case. The whole button from backoffice folder is gonna be deprecated. If user will pass prop upgrade - it will resolve to another component.

Issue #2511, opened 2018-11-28
`<Table/>` - `dataHook` not applied anywhere when Table in Page


hepiyellow:
## The issue
When using Table in Page, we add the `withWrapper=false` prop.
In the case there is no root DOM element for the table.
So there is no dataHook.

## Solution
This is problematic....
The only solution I see is to add let the user pass dataHook to both Table.Content and Table.Titlebar,
and create 2 separate drivers from them.
But in each driver only a subset of them methods would work. The Titlebar driver would be used for querying header stuff, and the Content driver for row stuff.

## Alternatively
We could develop a NEW solution to Table in Page using `react-sticky` (or similar concept),
thus not having to split the Table into 2 parts.

hepiyellow:
Resolved by 
https://github.com/wix/wix-style-react/pull/2515

PR #2515, opened 2018-11-28
`<Table/>` - testkit - Add ability to use Table driver when using Table in Page

hepiyellow:
Solved #2511 
By allowing to pass `dataHook` onto Table.Titlebar and Table.Content,
thus allowing to create TableDriver from each of them.

Not ideal developer experience, but there is not choice, 
unless we change the table implementation to allow fixed header without splitting the Table.

### Example

> Unit Testing Example - Table in Page
```javascript
import React from 'react';
import {tableTestkitFactory} from 'wix-style-react/dist/testkit';
/***************
 enzyme example
***************/
 const dataHook = 'myDataHook';
const wrapper = mount(
  <Table
    withWrapper={false}
    data={tableData}
    showSelection
  >
    <Page>
      <Page.Header title="My Table Title" />
      <Page.FixedContent>
        <Card>
          <Table.ToolbarContainer/>
          <Table.Titlebar dataHook="test-table-titlebar"/>
        </Card>
      </Page.FixedContent>
      <Page.Content>
        <Card>
          <Table.Content
            titleBarVisible={false}
            dataHook="test-table-content"
          />
        </Card>
      </Page.Content>
    </Page>
  </Table>
);
 const titlebarDriver = enzymeTableTestkitFactory({
  wrapper,
  dataHook: 'test-table-titlebar',
});
 const contentDriver = enzymeTableTestkitFactory({
  wrapper,
  dataHook: 'test-table-content',
});
 //Do tests
expect(titlebarDriver.getBulkSelectionCheckboxDriver().isChecked()).toBeTruthy();
expect(contentDriver.getRowsCount()).toBe(defaultProps.data.length);
expect(contentDriver.isRowSelected(0)).toBeTruthy();
```


src/Table/Table.spec.js
```diff
@@ -410,6 +411,51 @@ describe('Table', () => {
     });
   });
 
+  describe('withWrapper', () => {
+    const testContainer = new ReactDOMTestContainer().unmountAfterEachTest();
+
+    it('should have working test drivers when without wrapper', () => {
+      testContainer.render(TableDriverFactory);
```
shlomitc: is this neeeded?

hepiyellow: yes. It's like the smoke test sanity.
We want to make sure we support dataHook, and while at it. See that creating a driver where the from where the data hook is (which is not the Root element, so not trivial) yields a driver that actually works.

PR #2516, opened 2018-11-28
`<Modal/>` - component test - do proper cleanup of document.body between tests

hepiyellow:
Resolved https://github.com/wix-private/wsr-issues/issues/331

- Add a driver method `isUnmountSideEffectsDone`. This method can be used by consumers to verify that the Modal was removed from the body.

## Open issues
- [ ] how to name this function? `isUnmountSideEffectsDone` is a bit long
- [ ] I also export this `isUnmountSideEffectsDone` apart from the driver, since it does not require an element (since it is supposed to be called after the component has unmounted). In fact, the fact that the `isUnmountSideEffectsDone` is included in the driver is a bit weird, but could be convenient.

### Also in this PR
- Use ReactDomTestContainer - so we unmount the Modal after each test
- Move testkit sanity to smoke test
- Remove jasmine timeout control (we are running in jest)
- Remove use of sinon (use jest mocks)


src/Modal/Modal.spec.js
```diff
@@ -1,26 +1,34 @@
...
+} from './Modal.driver';
+
+import { ReactDOMTestContainer } from '../../test/dom-test-container';
+
+const MODAL_CLOSE_TIMEOUT = 10;
 
 describe('Modal', () => {
-  const createDriver = createDriverFactory(ModalFactory);
+  afterEach(async () => {
+    await eventually(() => expect(isModalUnmountDone()).toBeTruthy(), {
```
shlomitc: I think it's not recommended to make assertions in the `afterEach` hooks.

src/Modal/Modal.driver.js
```diff
@@ -2,16 +2,19 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import ReactTestUtils from 'react-dom/test-utils';
 
+const getContent = () => document.body.querySelector('.ReactModal__Content');
+const isOpen = () => !!getContent();
+export const isUnmountSideEffectsDone = () => !isOpen();
```
shlomitc: no need to expose it separately

src/Modal/Modal.driver.js
```diff
@@ -36,6 +39,7 @@ const modalDriverFactory = ({ element, wrapper, component }) => {
         wrapper,
       );
     },
+    isUnmountSideEffectsDone,
```
shlomitc: you can call it `waitForModalToClose` and also do the polling here

src/Modal/Modal.spec.js
hepiyellow: I can change it to throw an error. You need to throw an error in order for eventually to work.

shlomitc: I think it's enough to return a rejected promise

src/Modal/Modal.driver.js
hepiyellow: Let's discuss this.
As I mentioned, when this method is used, the component is supposed to be already unmounted. So there is no `element`, non of the functions of the driver (which use `element`) would work.

If the driver was a class, I would make this `isUnmountSideEffectsDone` a static method.

shlomitc: yeah, but it's not a class and we're not exposing the module to the consumer, only the driver, so let's stick with the existing conventions in that case

src/Modal/Modal.driver.js
hepiyellow: Well, if I mention modal-closed in the name of this function, I might as well just keep it `isOpen` and consumer can use eventually with 
```
eventually(()=>expect(driver.isOpen()).toBe(false))
```

But I wanted to make some consistent (across drivers) function name that abstracts away  the "thing" that we should wait for. (The "thing" that means that the component's side-effects have been cleaned-up)

test/dom-test-container.js
```diff
@@ -78,6 +78,7 @@ export class ReactDOMTestContainer {
       return driverFactory({
         element: this.componentNode,
         wrapper: this.node,
+        component: jsx,
```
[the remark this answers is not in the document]
hepiyellow: I agree that the drivers should not have a `setProps` method (we have about 10 drivers that have that method, and use `component`).

Specifically , Modal.spec uses the `setProps` method.
I can change it to set the props from the test and not from the driver.

Do you think we should deprecate all `setProps` methods?

src/Modal/Modal.spec.js
```diff
@@ -79,61 +87,53 @@ describe('Modal', () => {
...
     describe('timeout', () => {
-      let originalTimeout;
-
-      beforeEach(() => {
-        originalTimeout = jasmine.DEFAULT_TIMEOUT_INTERVAL;
-      });
-      afterEach(() => {
-        jasmine.DEFAULT_TIMEOUT_INTERVAL = originalTimeout;
-      });
+      function delay(t, v) {
```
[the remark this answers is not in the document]
hepiyellow: I will

src/Modal/Modal.spec.js
hepiyellow: Ok, I will.

src/Modal/Modal.driver.js
hepiyellow: But that means that the consumer need to hold on to the `driver` after unmounting.
- Usually unmounting is done in an `afterEach` block
- For an `afterEach` block to have access to the driver, the driver variable can not be a `const` in the `it` case itself, but now has to be a `let` variable on the `describe` block.
- It is not so "cool" to force consumers to work that way (holding a `driver` variable outside of `it` cases)

src/Modal/Modal.spec.js
```diff
@@ -79,61 +87,53 @@ describe('Modal', () => {
...
-      props.onRequestClose = sinon.spy();
+      props.onRequestClose = jest.fn();
       props.shouldCloseOnOverlayClick = true;
       props.closeTimeoutMS = 0;
 
       const driver = createDriver(<Modal {...props} />);
       driver.clickOnOverlay();
 
-      expect(props.onRequestClose.calledOnce).toBeTruthy();
+      expect(props.onRequestClose.mock.calls).toHaveLength(1);
```
yardnsm: You can use `expect(props.onRequestClose).toHaveCalled()` or `expect(props.onRequestClose).toHaveCalledTimes(1)`.

src/Modal/Modal.spec.js
```diff
@@ -79,61 +87,53 @@ describe('Modal', () => {
...
-      props.onRequestClose = sinon.spy();
+      props.onRequestClose = jest.fn();
       props.shouldDisplayCloseButton = true;
       props.closeTimeoutMS = 0;
 
       const driver = createDriver(<Modal {...props} />);
       driver.clickOnCloseButton();
 
-      expect(props.onRequestClose.calledOnce).toBeTruthy();
+      expect(props.onRequestClose.mock.calls).toHaveLength(1);
```
yardnsm: Same as previous comment.

testkit/testkit-smoke.test.js
```diff
@@ -138,6 +138,12 @@ const COMPONENTS = {
       title: 'test title',
     },
   },
+  Modal: {
+    props: {
+      isOpen: false,
+      contentLabel: 'modal_' + Math.random(),
```
yardnsm: Why not using a fixed value?

src/Modal/Modal.driver.js
shlomitc: probably, but need to verify it in tests

test/dom-test-container.js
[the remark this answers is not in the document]
hepiyellow: Ok. I am removing the setProps from the driver.
Let's see if users break or shout.
If all goes well, we'll remove it from all component drivers.

src/Modal/Modal.driver.js
hepiyellow: @shlomitc What do you mean "verify it in tests"?

src/Modal/Modal.spec.js
hepiyellow: thanks

PR #2519, opened 2018-11-29
Test/eyes e2e protractor

argshook:


.storybook/config-offline/preview-head.html
```diff
@@ -1,5 +1,3 @@
 <script src="react.min.js"></script>
```
shlomitc: Just mentioning  that this is problematic. we don't have any caching for the scripts, meaning if we change to react16, users of the storybook will keep getting the old version.

This cannot be merged in this way

argshook: yes im aware, before i tackle it i want to see CI green with e2e

hepiyellow:
This PR is impossible to review. 
Please next time write description about what was the issue, and how it was solved.

PR #2530, opened 2018-12-03
`<Dropdownlayout/>` - Fix update of hovered index

### LINKED ISSUE #2529: `<DropdoanLayout/>` - Error - `id` of undefined
## The bug
Reported by @liorur 

> this line throws id of undefined in cases where inout with options filter returned less countries then the current index

https://github.com/wix/wix-style-react/blob/master/src/DropdownLayout/DropdownLayout.js#L385

## Blame
It happened in #2487 (i think)

hepiyellow:
Fixes #2529

### DIFF
--- a/src/DropdownLayout/DropdownLayout.spec.js
+++ b/src/DropdownLayout/DropdownLayout.spec.js
@@ -528,12 +528,39 @@ describe('DropdownLayout', () => {
       });
       driver.pressDownKey();
       driver.pressDownKey();
+      driver.pressDownKey();
+      driver.pressDownKey();
 
-      expect(driver.isOptionHovered(1)).toBeTruthy();
+      expect(driver.isOptionHovered(3)).toBeTruthy();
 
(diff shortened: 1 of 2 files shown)


src/DropdownLayout/DropdownLayout.spec.js
```diff
@@ -528,12 +528,39 @@ describe('DropdownLayout', () => {
...
-      expect(driver.isOptionHovered(1)).toBeTruthy();
+      expect(driver.isOptionHovered(3)).toBeTruthy();
 
       wrapper.setProps({ options: options.slice(1) });
 
-      expect(driver.isOptionHovered(0)).toBeTruthy();
+      expect(driver.isOptionHovered(2)).toBeTruthy();
+    });
+
+    it('should reset the hovered option when options change and hovered option does not exist anymore', () => {
```
hepiyellow: I added this test which is unrelated to the bug

PR #2533, opened 2018-12-03
`<Calendar/>: Add range support

hepiyellow:
Replaces #2403 because of a bad rebase.

The first commit "PART 1" is a squashed commit of all the commits prior to rebasing on master.


src/Calendar/Calendar.js
```diff
@@ -27,20 +28,103 @@ export default class Calendar extends WixComponent {
...
+      }
+    } else {
+      onChange(value, modifiers);
+      shouldCloseOnSelect && this.props.onClose();
+    }
+  };
+
+  _getMonth = props => {
+    const { value } = props;
+    const optionalParse = x => (typeof x === 'string' ? parse(x) : x);
```
hepiyellow: create `parseValue()` function that peels of any string date value, including inside range.
and use it, everywhere needed.

src/Calendar/Calendar.js
```diff
@@ -27,20 +28,103 @@ export default class Calendar extends WixComponent {
...
   }
 
-  // TODO: Change to getDerivedStateFromProps with React ^16.0.0
+  static dateToMonth(date) {
+    return date ? date.getFullYear() * 100 + date.getMonth() : null;
+  }
+
   componentWillReceiveProps(nextProps) {
-    this.setState({ month: nextProps.value });
+    if (this.props.value != nextProps.value) {
```
hepiyellow: This comparison is redundant.

src/Calendar/Calendar.js
malsomnus: (Including in willReceiveProps)

src/Calendar/Calendar.js
malsomnus: (Remove outer condition, it's always false)

src/Calendar/Calendar.js
```diff
@@ -36,22 +36,22 @@ export default class Calendar extends WixComponent {
...
-        const to = Calendar.dateToMonth(nextProps.value.to);
-
-        if (from && currentMonth < from) {
-          this.setState({ month: nextProps.value.from });
-        } else if (to && currentMonth > to) {
-          this.setState({ month: nextProps.value.to });
-        }
+    const currentMonth = Calendar.dateToMonth(this.state.month);
+
+    if (nextValue instanceof Date) {
```
hepiyellow: See if we can extract a `isSingleDay` method, and use it here, and in the `render`.

src/Calendar/Calendar.js
```diff
@@ -110,39 +110,45 @@ export default class Calendar extends WixComponent {
     }
   };
 
-  _getMonth = props => {
-    const { value } = props;
-    const optionalParse = x => (typeof x === 'string' ? parse(x) : x);
+  static optionalParse = x => (typeof x === 'string' ? parse(x) : x);
```
hepiyellow: You can rename `x` to a meaningful `dateOrString`

malsomnus: Done.

src/Calendar/Calendar.js
```diff
@@ -36,22 +36,22 @@ export default class Calendar extends WixComponent {
   }
 
   componentWillReceiveProps(nextProps) {
-    if (this.props.value != nextProps.value) {
-      const currentMonth = Calendar.dateToMonth(this.state.month);
+    const nextValue = Calendar.parseValue(nextProps.value);
```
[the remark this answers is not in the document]
malsomnus: It's impossible for `prevMonth` to be undefined.
Beyond that, I'm afraid I don't understand what you're talking about. What logic? What default month? How is this code related to `_getMonth`?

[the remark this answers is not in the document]
hepiyellow: The "logic" in `componentWillReceiveProps` is the one that sets the "next" month state to the month which is closest to the previous month and includes the range `start | end`.

You can extract this logic to a, lets call it, `getNextMonth(value, prevMonth)`.
The call to `_getMonth` in the constructor, is actually a `getInitialMonth`, which is the simple case where there is no prevMonth.

So if you extract a function called `_getMonth(value, prevMonth)`, (You can also call it `getNextMonth`, or `updateMonthState(propValue, prevMonth)`)
you can use it both in `componentWillRecieveProps`,
and also in the constructor, by calling `_getMonth(props.value)`.

So you can then delete the `_getMonth()` method.

hepiyellow:
Ready to merge, waiting for build.

hepiyellow:
Eyes diff approved.
Waiting for another green build

PR #2534, opened 2018-12-03
`<Tooltip/>` - Component Tests/Docs : Add proper cleanup

hepiyellow:
`Tooltip.spec`

It seems that for Tooltip we don't need to wait for the tooltip content to close, since when unmounting the Tooltip it remove it without any timer.

- Add proper cleanup using `react-testing-library`
- Remove `document.body.innerHTML = ''`
- Remove use of enzyme
- Move testkit sanity to smoke-test
- Update `README.TESTKIT.md`


src/Tooltip/README.TESTKIT.md
```diff
@@ -44,63 +44,78 @@ beforeAll(() => {
 
 ## Usage Example
 
-> Unit testing example
+### Cleanup
+
+When Tooltip is unmounted, it immediately removes any opened Tooltip content (with no timers). So It is enough to simply unmount it. (No need to wait for anything)
```
shlomitc: Are you sure? I still think it has a `setTimeout` when unmounting

src/Tooltip/README.TESTKIT.md
```diff
@@ -44,63 +44,78 @@ beforeAll(() => {
...
+
+### Example Enzyme
+
+```javascript
+import React from 'react';
+import {tooltipTestkitFactory as enzymeTooltipTestkitFactory} from 'wix-style-react/dist/testkit/enzyme';
+import waitForCond from 'wait-for-cond';
```
shlomitc: prefer `wix-eventually`, this is what we suggest our users to use

src/Tooltip/README.TESTKIT.md
hepiyellow: Yes I am sure.
https://github.com/wix/wix-style-react/blob/9a63887f5a658c76fc66dfe62a565ad58b9a0ea1/src/Tooltip/Tooltip.js#L201

src/Tooltip/README.TESTKIT.md
hepiyellow: ok

src/Tooltip/README.TESTKIT.md
```diff
@@ -44,63 +45,134 @@ beforeAll(() => {
...
+        </Tooltip>
+      </div>
+    );
+
+    // create testkits
+    const tooltipTestkit = tooltipTestkitFactory({ wrapper: container, dataHook: 'my-tooltip' });
+
+    //open the tooltip
+    tooltipDriver.mouseEnter();
+    expect(modalDriver.isShown()).toBeTruthy();
```
splintor: Where do this `modalDriver` came from? And don't we need to wait for `isShown()` to return `true`?
Maybe:
```
await eventually(() => expect(testkit.isShown()).toBeTruthy());
```

shlomitc: it's an old PR which was already merged, maybe worth creating a new issue mentioning this PR, asking for a fix.
It would also be great to have a PR fixing it :)

splintor: https://github.com/wix/wix-style-react/pull/3128

PR #2536, opened 2018-12-04
WIP: carousel component

mor-gilad:
### What changed

WIP - carousel component

### Why it changed

Because of wix code presets and our guild week :)

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


src/Carousel/Carousel.js
```diff
@@ -0,0 +1,142 @@
...
+                />
+              );
+            })}
+        </div>
+      </div>
+    );
+  }
+}
+
+Carousel.propTypes = WixComponent.propTypes;
```
shlomitc: redundant (since it extends WixComponent). however, propTypes definition are needed

src/Carousel/Carousel.js
```diff
@@ -0,0 +1,142 @@
...
+Carousel.defaultProps = WixComponent.defaultProps;
```
shlomitc: redundant

src/Carousel/Carousel.js
```diff
@@ -0,0 +1,142 @@
...
+  isDotActive(index) {
+    const activeIndex = this.state.activeIndex;
+    const originalImageCount = this.props.images.length;
+    if (activeIndex > originalImageCount - 1) {
+      return index === activeIndex - originalImageCount;
+    }
+    return index === activeIndex;
+  }
+
+  render() {
```
shlomitc: please separate component to smaller rendering components (can still be in this file for now)

src/Carousel/Carousel.js
```diff
@@ -0,0 +1,142 @@
...
+  onImageLoad() {
+    this.setState(state => {
+      const loadedImageCount = state.loadedImageCount + 1;
+      return {
+        loadedImageCount,
+      };
+    });
+  }
+
+  isLoading() {
```
shlomitc: prefix internal methods with `_`

src/Carousel/Carousel.js
```diff
@@ -0,0 +1,142 @@
...
+
+const duplicateIfTwoImages = images =>
+  images.length === 2 ? images.concat(cloneDeep(images)) : images;
+class Carousel extends WixComponent {
+  constructor(props) {
+    super(props);
+    const images = this.props.images || [];
+    this.state = {
+      activeIndex: 0,
+      images: duplicateIfTwoImages(images),
```
shlomitc: this should actually be implemented also on `componentWillReceiveProps`.
Since I think this adds complexity, maybe it doesn't worth adding this behavior at all and just require the consumer of this component to handle this on his side

src/Carousel/Carousel.js
```diff
@@ -0,0 +1,142 @@
...
+    return this.state.activeIndex === 0
+      ? this.state.images.length - 1
+      : this.state.activeIndex - 1;
+  }
+
+  onImageLoad() {
+    this.setState(state => {
+      const loadedImageCount = state.loadedImageCount + 1;
+      return {
+        loadedImageCount,
```
shlomitc: this triggers a re-render, which will happen on each image load. Maybe it's better just to hold this counter as a property value and hold the `isLoading` value on the state instead

src/Carousel/Carousel.scss
```diff
@@ -0,0 +1,85 @@
...
+        transform: translateX(-100%);
+    }
+
+    &.next {
+        visibility: hidden;
+        transform: translateX(100%);
+    }
+}
+
+img {
```
shlomitc: this will affect every `img` tag in an app, use class instead

src/Carousel/Carousel.scss
```diff
@@ -0,0 +1,85 @@
...
+    height: 100%;
+    display: flex;
+    flex-direction: column;
+}
+
+.gallery {
+    display: flex;
+    width: 100%;
+    height: 100%;
+    margin-bottom: 34px;
```
shlomitc: we usually work with multiples of 6 - are you sure this is the right margin?
I would also have it as a variable since you use it in multiple places

src/Carousel/Carousel.spec.js
```diff
@@ -0,0 +1,140 @@
...
+      );
+      driver.clickNext();
+      driver.clickPrevious();
+      await eventually(async () => {
+        expect(driver.getCurrentImageIndex()).toBe(0);
+      });
+    });
+  });
+
+  describe('when `loopImages` is true', () => {
```
shlomitc: I would call it `infinite` instead

src/Carousel/README.TESTKIT.md
```diff
@@ -0,0 +1,43 @@
+# Carousel component
```
shlomitc: This file is supposed to be automatically created for you (just try to remove it).

Just add the methods description to the driver as comments (`/** the comment */`)

src/Carousel/README.md
```diff
@@ -0,0 +1,16 @@
+# Carousel component
```
shlomitc: not needed, API table is auto-generated from prop-types when comments are added

testkit/index.js
```diff
@@ -470,3 +470,9 @@ import { iconButtonDriverFactory } from '../src/IconButton/IconButton.driver';
 export const iconButtonTestkitFactory = uniTestkitFactoryCreator(
   iconButtonDriverFactory,
 );
+
+import { carouselDriverFactory } from '../src/Carousel/Carousel.driver';
```
shlomitc: should be added also to the `enzyme` one

scripts/npm-start.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
 
 if scripts/ensure-dist.sh; then
-  npm run storybook & yoshi start
+  npm run storybook
```
shlomitc: I actually think this is a good change. @argshook any objections?

argshook: sure, but could trip over newcomers who might expect tests to run. but i myself never do `npm start` and always do `npm run storybook`

mor-gilad: hahaha
pushed it by mistake. 
But I can agree :)

mor-gilad: We were newcomers and we didn't expect 'npm start' to run the testkit

mor-gilad: also, we had an webpack error which was hidden by the test runner so we couldn't read it

shlomitc: are you sure it was a webpack error or `jest` watch mode error? usually, it's the later one and `brew install watchman` fixes it

src/Carousel/Carousel.js
mor-gilad: ok

src/Carousel/Carousel.js
mor-gilad: ok

src/Carousel/Carousel.js
mor-gilad: ok. Although I'm not sure I get the reason for that. What is internal? Everything that isn't react lifecycle method? Why is it good?
Will change it if this is how you work. No problem. I just don't understand why is it good.

src/Carousel/Carousel.js
mor-gilad: This indeed adds complexity. But I feel this is something the component should support. A carousel is maybe not the best component to have when you need to show only 2 images (ux). But. it still feels like basic functionality and I don't feel comfortable with publishing a component that doesn't support it.

src/Carousel/Carousel.spec.js
mor-gilad: Will do

src/Carousel/Carousel.js
shlomitc: I’m ok with putting this aside. We trust our developers to do with the components the right things, so asking them to give the right images seems like a basic request

src/Carousel/Carousel.js
shlomitc: AutoDocs parses methods as public API, but we intend to use it only for internal use. So this convention helps AutoDocs to understand it should have it in the docs

src/Carousel/Carousel.js
shlomitc: the idea is is to fix a use-case of less than three images, so the naming should be more descriptive (and tested of course)

src/Carousel/Carousel.js
```diff
@@ -0,0 +1,220 @@
...
+      pause = true;
+      setTimeout(() => {
+        pause = false;
+      }, time);
+      callback(...args);
+    }
+  };
+}
+
+const duplicateIfTwoImages = images =>
```
shlomitc: as this functionality is not tested and will require some more work - let's drop it.

src/Carousel/Carousel.js
```diff
@@ -0,0 +1,220 @@
...
+  images.length === 2 ? images.concat(cloneDeep(images)) : images;
+
+class Carousel extends WixComponent {
+  constructor(props) {
+    super(props);
+    const images = this.props.images || [];
```
shlomitc: use default props for that

src/Carousel/Carousel.js
```diff
@@ -0,0 +1,220 @@
...
+      images: duplicateIfTwoImages(images),
+      loadedImageCount: 0,
+    };
+    this._slide = throttle(this._slide.bind(this), 2000);
+  }
+
+  componentDidMount() {
+    super.componentDidMount();
+    if (this.props.autoplay) {
+      this._autoplay();
```
shlomitc: 1. cleanup in `componentWillUnmount`
2. halt it's operation when prop changes from true->false (or the other way around)

src/Carousel/Carousel.js
```diff
@@ -0,0 +1,220 @@
...
+
+  _autoplay() {
+    const intervalToken = setInterval(
+      () => this._slide(this._getNextIndex()),
+      4000,
+    );
+    this._haltAutoplay = () => clearInterval(intervalToken);
+  }
+
+  _onMouseOver() {
```
shlomitc: rename to `_stopSlideshow`

src/Carousel/Carousel.js
```diff
@@ -0,0 +1,220 @@
...
+    this.props.autoplay && this._haltAutoplay();
+  }
+
+  _onMouseOut() {
```
shlomitc: rename to `_continueSlideshow`

src/Carousel/Carousel.js
```diff
@@ -0,0 +1,220 @@
...
+        loadedImageCount,
+      };
+    });
+  }
+
+  _isLoading() {
+    return this.state.loadedImageCount < this.state.images.length;
+  }
+
+  _isDotActive(index) {
```
shlomitc: I'm not sure I understand the logic, shouldn't it just be the `activeIndex`?

src/Carousel/Carousel.scss
```diff
@@ -0,0 +1,85 @@
...
+
+.dot {
+    width: 12px;
+    height: 12px;
+    background-color: #daeffe;
+    border-radius: 6px;
+    margin: 6px;
+    
+    &.active {
+        background-color: #3899ec;
```
shlomitc: colors should be taken from the `colors.scss` file

stories/Carousel/index.story.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import Carousel from 'wix-style-react/Carousel';
+import storySettings from './storySettings';
+
+const imagesExamples = [
+  {
+    value: [
```
shlomitc: drop the 2 images logic for now

stories/Carousel/storySettings.js
```diff
@@ -0,0 +1,4 @@
+module.exports.storyName = 'Carousel';
+module.exports.category = 'WIP/Carousel';
```
shlomitc: should be just `WIP`

stories/Carousel/storySettings.js
```diff
@@ -0,0 +1,4 @@
...
+module.exports.kind = 'WIP/Carousel';
```
shlomitc: duplicated, merge with category

src/Carousel/Carousel.js
```diff
@@ -0,0 +1,220 @@
...
+            priority="secondary"
+            onClick={() => this._next()}
+          >
+            <ChevronRightLarge />
+          </IconButton>
+        </div>
+      </div>
+    );
+
+    const pagination = (
```
shlomitc: refactor to a standalone component (can still be in the Carousel folder).

it should not be aware to the `images`, but only the the amount of the required dots

src/Carousel/Carousel.js
mor-gilad: done

hepiyellow:
Does this component have any spec?
Please reference zeplin spec, and please right a dev spec.

hepiyellow:
Also, this sounds like a classic wix-ui-core component. 
It has complicated behaviour, it would actually be nice to have it separated from styling.

shlomitc:
@hepiylow - it’s ok. I’m on it. Please talk with me about it on Sunday and I’ll explain

src/Carousel/Carousel.scss
mor-gilad: done

src/Carousel/Carousel.js
hepiyellow: I didn't understand where the name `_isDotActive` is coming from. what is `dot`?
Can you rename this?

src/Carousel/Carousel.js
```diff
@@ -0,0 +1,237 @@
...
+}
+
+//update images on imageUpdate
+Carousel.propTypes = {
+  /** Array of strings where each string is a src of an image (in \<img src="your_src" /\>) */
+  images: PropTypes.array.isRequired,
+  /** Images loop endlessly */
+  infinite: PropTypes.bool,
+  /** Auto-playing of images */
+  autoplay: PropTypes.bool,
```
hepiyellow: I didn't see any handling of `coponentWillReceiveProps` that, for example, stops the slideShow if `autoplay` becomes `false`.
If you want this prop to be controlled, please add handling in `componentWillReceiveProps`.
If it is to be uncontrolled then please rename it to `initialAutoplay`.
(So that if we want to add controlled `autoplay` in the future, we won't have to rename the old `autoplay` to `initialAutoplay`, which is a breaking change)

Personally, I think it should be controlled.

src/Carousel/Carousel.js
```diff
@@ -0,0 +1,237 @@
...
+      }, time);
+      callback(...args);
+    }
+  };
+}
+
+const duplicateIfTwoImages = images =>
+  images.length === 2 ? images.concat(cloneDeep(images)) : images;
+
+class Carousel extends WixComponent {
```
hepiyellow: `WixComponent` is ` React.PureComponent`.
The Carousel has an `images` prop which is an array. So this can not be a PureComponent.
Please extends `ReactComponent`, and add `dataHook` to props, and spread it on root element.

src/Carousel/Carousel.js
```diff
@@ -0,0 +1,237 @@
...
+
+    if (this.props.autoplay) {
+      this._autoplay();
+    }
+  }
+
+  _autoplay() {
+    const intervalToken = setInterval(
+      () => this._slide(this._getNextIndex()),
+      4000,
```
hepiyellow: of course this should be a prop, right?

src/Carousel/Carousel.js
```diff
@@ -0,0 +1,237 @@
...
+    }
+    return index === activeIndex;
+  }
+
+  render() {
+    const gallery = (
+      <div className={styles.gallery}>
+        <div className={styles.buttonContainer}>
+          <IconButton
+            dataHook="prev-button"
```
hepiyellow: Instead of `dataHook` please use an `aria-label="previous image"` which is missing here anyways.

src/Carousel/Carousel.scss
```diff
@@ -0,0 +1,87 @@
+@import '../../colors.scss';
```
hepiyellow: @shlomitc should we use Stylable?

src/Carousel/Carousel.scss
```diff
@@ -0,0 +1,87 @@
...
+    height: 100%;
+    margin: 0 24px;
+
+    &.loading {
+        display: none
+    }
+}
+
+.image {
+    transition: 2000ms cubic-bezier(0.77, 0, 0.175, 1);
```
hepiyellow: please be specific with the transition property.

src/Carousel/Carousel.scss
```diff
@@ -0,0 +1,87 @@
...
+    flex-direction: column;
+    justify-content: center;
+}
+
+.loader {
+    overflow: hidden;
+    position: relative;
+    width: 100%;
+    height: 100%;
+    background-color: $paletteB5;
```
hepiyellow: Don't use `palleteXX` colors.

src/Carousel/Carousel.js
hepiyellow: Please don't Unresolve conversations, until the Fix is pushed.
You can comment with you intentions, so we know we are on the "same-line".

src/Carousel/Carousel.js
```diff
@@ -0,0 +1,237 @@
...
+    if (activeIndex > originalImageCount - 1) {
+      return index === activeIndex - originalImageCount;
+    }
+    return index === activeIndex;
+  }
+
+  render() {
+    const gallery = (
+      <div className={styles.gallery}>
+        <div className={styles.buttonContainer}>
```
hepiyellow: I don't think you need this `<div/>` only to wrap the IconButton with a `className={styles.buttonContainer}`.
The `<IconButton/>` can accept a `className` prop.

src/Carousel/Carousel.js
```diff
@@ -0,0 +1,237 @@
...
+                  [styles.active]: this._isDotActive(currentIndex),
+                })}
+              />
+            );
+          })}
+      </div>
+    );
+
+    return (
+      <div className={styles.carousel} data-ready={!this._isLoading()}>
```
hepiyellow: If the spec tests can check for a `data-hook="loader"`,
then the e2e test can also do that. And we don't need the `data-ready`.

If you change the driver to be a UniDrier, then you even can use the same method `isLoading`.

src/Carousel/Carousel.js
hepiyellow: Why not use the `<Pagination/>` component from `wix-ui-core`.
It takes care of cases where the number of images (dots) is larger than it fits in the given width.

src/Carousel/Carousel.js
```diff
@@ -0,0 +1,237 @@
...
+        {pagination}
+      </div>
+    );
+  }
+}
+
+//update images on imageUpdate
+Carousel.propTypes = {
+  /** Array of strings where each string is a src of an image (in \<img src="your_src" /\>) */
+  images: PropTypes.array.isRequired,
```
hepiyellow: What if the component is re-rendered with different `images` (Like external filtering feature).
We need to support that in `componentWillReceiveProps`
OR if you don't want to support it *yet* ...
Then rename it to `initialImages`.

The consumer can always render the `<Carousel/>` with a new `key` value, to cause it to re-mount (and reset state).

stories/Carousel/storySettings.js
```diff
@@ -0,0 +1,3 @@
+module.exports.storyName = 'Carousel';
```
hepiyellow: you can use :
```js
export const storyName = 'Carousel';
```

src/Carousel/Carousel.js
mor-gilad: done

src/Carousel/Carousel.js
mor-gilad: No. Because of the clone for two images which we talked about.

src/Carousel/Carousel.js
```diff
@@ -0,0 +1,220 @@
...
+    this.setState(state => {
+      const loadedImageCount = state.loadedImageCount + 1;
+      return {
+        loadedImageCount,
+      };
+    });
+  }
+
+  _isLoading() {
+    return this.state.loadedImageCount < this.state.images.length;
```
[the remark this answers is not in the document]
mor-gilad: We do not follow the entire spec for the first phase. You are correct. having many images will not work out. This is why this is a work in progress :)

src/Carousel/Carousel.js
shlomitc: sounds like an edge case, I wouldn't want to touch this at the moment

src/Carousel/Carousel.js
shlomitc: handled in componentDidUpdate

src/Carousel/Carousel.scss
shlomitc: not here

src/Carousel/Carousel.spec.js
```diff
@@ -0,0 +1,187 @@
+import React from 'react';
+import carouselDriverFactory from './Carousel.driver.private';
+import Carousel from './Carousel';
+import { createDriverFactory } from 'wix-ui-test-utils/driver-factory';
+
+describe('Carousel', () => {
+  const createDriver = createDriverFactory(carouselDriverFactory);
+  jest.useFakeTimers();
```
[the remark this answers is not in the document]
shlomitc: not relevant for here

src/Carousel/Carousel.spec.js
```diff
@@ -0,0 +1,187 @@
+import React from 'react';
+import carouselDriverFactory from './Carousel.driver.private';
+import Carousel from './Carousel';
+import { createDriverFactory } from 'wix-ui-test-utils/driver-factory';
```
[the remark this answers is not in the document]
shlomitc: can be refactored later

src/Carousel/Carousel.spec.js
```diff
@@ -0,0 +1,187 @@
...
+      expect(driver.isLoading()).toBeTruthy();
+    });
+
+    it('should hide the loader when images are loaded', () => {
+      const { driver } = createDriver(
+        <Carousel images={[{ src: 'image1.jpg' }, { src: 'image2.jpg' }]} />,
+      );
+      driver.loadImages();
+      expect(driver.isLoading()).toBeFalsy();
+    });
```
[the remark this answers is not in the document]
shlomitc: currently yes, we can optimize it later

src/Carousel/Carousel.spec.js
```diff
@@ -0,0 +1,187 @@
...
+        <Carousel images={[{ src: 'image1.jpg' }, { src: 'image2.jpg' }]} />,
+      );
+      driver.clickNext();
+      jest.runOnlyPendingTimers();
+      driver.clickPrevious();
+      expect(driver.getCurrentImageIndex()).toBe(0);
+    });
+  });
+
+  describe('infinite functionallity', () => {
```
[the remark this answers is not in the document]
shlomitc: `infinite` is a common naming for this cyclic navigation, so I'm ok with that

shlomitc:
There are few more TODOs after merging, but we can start with this

TODO
Important:
- get a UX review
- 4/3 aspect ratio is still missing in the component

regular:
- migrate to unidriver
- refactor the start/stop slideshow (duplicated logic)
- image loading: call setState only when all images are loaded instead of on each component load
- remove `button-container` if not needed
- use data-hook=“loader” in the loading test
- change tests to work with react-testing library

PR #2541, opened 2018-12-06
`<MessageBoxMarketerialLayout/>` (Announcement) - Add `illustration` prop

hepiyellow:
According to [spec](https://app.zeplin.io/project/5864e02695b5754a69f56150/screen/5891b72cd20d89ec14d1d217)


hepiyellow:
## API suggestion
I talked with @milkyfruit about adding a new prop `illustration`, and that we can not have both `illustration` and (`imageComponent` or `imgUrl`).

Right now there are 2 props:
`imageComponent`
`imgUrl`

They both get different positioning inside the header.

According to @milkyfruit, I suggest that the illustration will be passed in the `imageComponent` prop,
and we will add a new prop called something like: `imageOverlap` (which can get true / false).

But does it mean that the `imageComponent` is always limited to `height=150px`?

hepiyellow:
@wuwa @milkyfruit So what about the API suggestion, does `imageOverlap` prop make sense?

Or is "Anouncement" such a different creature that it shouldn't use `<MessageBoxMarketerialLayout/>` and shouldn't have `imageComponent` and `imgUrl` props?
(Meaning it's a new `<Announcement/>` component.

wuwa:
i think this component is very outdated and should be killed. is there any urgent request for this?

shlomitc:
no urgency, just cleanup.
can wait for the modals re-structure

PR #2542, opened 2018-12-06
Update dataHook on componentDidUpdate

### LINKED ISSUE #2538: Unexpected datahook behaviour
# Bug Report

### Relevant Components
Any component that inherits from `WixComponent` (e.g Button)

### Current Behavior
given the following app (assume trivial implementation of state and `toggleFoo`):
```
<div>
        <div>
          <Button dataHook="action" onClick={this.toggleFoo}>
            <span>Action</span>
          </Button>
        </div>
        <div>
          {this.state.isFoo ? (
            <Button dataHook="foo">Foo</Button>
          ) : (
            <Button dataHook="bar">Bar</Button>
          )}
        </div>
</div>
```
first render:
```
<div>
...
  <div>
    <button data-hook="foo">Foo</button>
  </div>
</div>
```
second render (after `toggleFoo`):
```
<div>
...
  <div>
    <button data-hook="foo">Bar</button>
  </div>
</div>
```
data-hook was not updated on the button's html.

### Expected Behavior

data-hook should update on the button's html:

second render (after `toggleFoo`):
```
<div>
...
  <div>
    <button data-hook="bar">Bar</button>
  </div>
</div>
```

### Steps to Reproduce

https://codesandbox.io/s/ov91jxjjr6

### Severity
This is very confusing, especially in tests as code will work as expected (As long as you don't rely on data-hook for your implementation) but a test that will check that barButton is now visible by the presence of data hook will fail.

yakirn:
### What changed

`dataHook` will now be updated when changed.

Fix #2538.

### Why it changed

See #2538.

---

- [x] Change is tested
- [ ] Change is documented
- [X] Build is green

I didn't find a suitable place for both documentation and tests, so I didn't add any. I'll be happy to add if you have one.


yakirn:
@hepiyellow I chose `componentDidUpdate` over `componentWillReceiveProps` as the latter is marked unsafe starting react 16.3.

https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#gradual-migration-path

yakirn:
@hepiyellow done.

src/BaseComponents/WixComponent.spec.js
```diff
@@ -0,0 +1,37 @@
...
+  }
+}
+
+const fixtureDriverFactory = ({ element }) => {
+  return {
+    getDataHook: () => element.dataset['hook'],
+  };
+};
+
+describe('BaseComponent', () => {
```
hepiyellow: Change to `WixComponent`

hepiyellow:
I just realized that adding a `componentDidUpdate` to `<WixComponent/>` will work in most component, but some component override `componentDidUpdate`, so the dataHook would not update for them.

I suggest we add a sanity test for all components (we have one, its called `smoke-test`) that the dataHook can be updated.

PR #2543, opened 2018-12-06
[DO NOT MERGE] <Counter/> - template for the component generator

yardnsm:
**Do not merge this PR, its meant for discussion purposes only.**

In order for us to create a great component generator, we first need to define a common component template including our best-practices and conventions. This PR includes a simple `<Counter/>` component, that resembles a new component added to the library.

Eventually, the component generator will generate this exact component, with its test driver and tests, but of course its naming will be different.

Do you think this is a good template? Please suggest any changes.

@shlomitc @argshook @mykas @hepiyellow

### DIFF
--- a/.storybook/stories.js
+++ b/.storybook/stories.js
@@ -119,6 +119,7 @@ import '../stories/Layout';
 import '../stories/Text/index.story';
 import '../stories/Skeleton.story.js';
 import '../stories/Collapse.story';
+import '../stories/Counter/index.story';
 
 // Styling
 import '../stories/Typography/TypographyClassesStory';
--- a/src/Counter/Counter.driver.private.js
+++ b/src/Counter/Counter.driver.private.js
@@ -1,0 +1,15 @@
+import { counterDriverFactory as publicDriverFactory } from './Counter.driver';
+
+export const counterPrivateDriverFactory = base => {
+  return {
+    ...publicDriverFactory(base),
+
+    /** Get the increment button text */
+    getIncrementText: async () =>
+      base.$('[data-hook="counter-increment"]').text(),
+
+    /** Get the increment button text */
+    getDecrementText: async () =>
+      base.$('[data-hook="counter-decrement"]').text(),
+  };
+};
--- a/src/Counter/Counter.driver.private.js
+++ b/src/Counter/Counter.driver.private.js
@@ -4,12 +4,6 @@ export const counterPrivateDriverFactory = base => {
   return {
     ...publicDriverFactory(base),
 
-    /** Get the increment button text */
-    getIncrementText: async () =>
-      base.$('[data-hook="counter-increment"]').text(),
-
-    /** Get the increment button text */
-    getDecrementText: async () =>
-      base.$('[data-hook="counter-decrement"]').text(),
+    // Add here driver methods that considered "private"
   };
 };
--- a/src/Counter/Counter.e2e.js
+++ b/src/Counter/Counter.e2e.js
@@ -1,0 +1,43 @@
+import {
+  createStoryUrl,
+  waitForVisibilityOf,
+  scrollToElement,
+} from 'wix-ui-test-utils/protractor';
+
+import { eyesItInstance } from '../../test/utils/eyes-it';
+import { counterTestkitFactory } from '../../testkit/protractor';
+import { storySettings } from '../../stories/Counter/storySettings';
+
+const eyes = eyesItInstance({ enableSnapshotAtBrowserGet: false });
+
+describe('Counter', () => {
+  const storyUrl = createStoryUrl({
+    kind: storySettings.category,
+    story: storySettings.storyName,
+  });
+
+  const createDriver = async (dataHook = storySettings.dataHook) => {
+    const driver = counterTestkitFactory({ dataHook });
+
+    await waitForVisibilityOf(
+      await driver.element(),
+      `Cannot find <Counter /> component with dataHook of ${dataHook}`,
+    );
+
+    await scrollToElement(driver.element());
+
+    return driver;
+  };
+
+  beforeAll(async () => {
+    await browser.get(storyUrl);
+  });
+
+  eyes.it('should render', async () => {
+    await createDriver();
+  });
+
+  eyes.it('should render live example', async () => {
+    await createDriver('story-counter-live-example');
+  });
+});
--- a/src/Counter/Counter.js
+++ b/src/Counter/Counter.js
@@ -1,0 +1,93 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styles from './Counter.scss';
+
+import Heading from '../Heading';
+import Button from '../Button';
+
+class Counter extends React.PureComponent {
+  static displayName = 'Counter';
+
+  static propTypes = {
+    dataHook: PropTypes.string,
+
+    /** The initial count */
+    initialCount: PropTypes.number,
+    /** Callback funtion to be called when the count has been changed. Signature is `onCountUpdate(newCount)`. */
+    onCountUpdate: PropTypes.func,
+    /** Text for the increment button */
+    incrementText: PropTypes.string,
+    /** Text for the decrement button */
+    decrementText: PropTypes.string,
+
+    /** Theme of the buttons */
+    theme: Button.propTypes.theme,
+  };
+
+  static defaultProps = {
+    initialCount: 0,
+    onCountUpdate: () => {},
+    incrementText: '+',
+    decrementText: '-',
+  };
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      count: props.initialCount || 0,
+    };
+  }
+
+  handleIncrement = () => {
+    this.setState(
+      ({ count }) => ({
+        count: count + 1,
+      }),
+      () => {
+        this.props.onCountUpdate(this.state.count);
+      },
+    );
+  };
+
+  handleDecrement = () => {
+    this.setState(
+      ({ count }) => ({
+        count: count - 1,
+      }),
+      () => {
+        this.props.onCountUpdate(this.state.count);
+      },
+    );
+  };
+
+  render() {
+    const { count } = this.state;
+    const { dataHook, incrementText, decrementText, theme } = this.props;
+
+    return (
+      <div className={styles.root} data-hook={dataHook}>
+        <Heading dataHook="counter-count">{count}</Heading>
+
+        <div className={styles.buttons}>
+          <Button
+            onClick={this.handleDecrement}
+            theme={theme}
+            dataHook="counter-decrement"
+          >
+            {decrementText}
+          </Button>
+          <Button
+            onClick={this.handleIncrement}
+            theme={theme}
+            dataHook="counter-increment"
+          >
+            {incrementText}
+          </Button>
+        </div>
+      </div>
+    );
+  }
+}
+
+export default Counter;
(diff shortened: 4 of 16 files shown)


testkit/enzyme.js
```diff
@@ -500,3 +500,9 @@ import { closeButtonDriverFactory } from '../src/CloseButton/CloseButton.driver'
 export const closeButtonTestkitFactory = enzymeUniTestkitFactoryCreator(
   closeButtonDriverFactory,
 );
```
argshook: @yardnsm im working on script that makes these exports work automatically, so probably no need to have these (enzyme, index, protractor, puppeteer) in generator

.storybook/stories.js
```diff
@@ -119,6 +119,7 @@ import '../stories/Layout';
 import '../stories/Text/index.story';
 import '../stories/Skeleton.story.js';
 import '../stories/Collapse.story';
+import '../stories/Counter/index.story';
```
argshook: how do we treat the whole  kebab-case thing?

src/Counter/Counter.e2e.js
```diff
@@ -0,0 +1,43 @@
...
+  createStoryUrl,
+  waitForVisibilityOf,
+  scrollToElement,
+} from 'wix-ui-test-utils/protractor';
+
+import { eyesItInstance } from '../../test/utils/eyes-it';
+import { counterTestkitFactory } from '../../testkit/protractor';
+import { storySettings } from '../../stories/Counter/storySettings';
+
+const eyes = eyesItInstance({ enableSnapshotAtBrowserGet: false });
```
argshook: not too related, but perhaps `enableSnapshotAtBrowserGet` should be the default? meaning this line is not needed

.storybook/stories.js
yardnsm: So `../stories/Counter/Counter.story`? I noticed some components are using `index.story.js` and some use `ComponentName.story.js`, what's better?

src/Counter/Counter.e2e.js
yardnsm: I think this is a good idea. @hepiyellow wdyt?

testkit/enzyme.js
yardnsm: No problem. I'll leave these changes in this PR so the CI will pass for now.

src/Counter/Counter.js
```diff
@@ -0,0 +1,93 @@
...
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      count: props.initialCount || 0,
+    };
+  }
+
+  handleIncrement = () => {
```
yardnsm:
```suggestion
  _handleIncrement = () => {
```

src/Counter/Counter.js
```diff
@@ -0,0 +1,93 @@
...
+      ({ count }) => ({
+        count: count + 1,
+      }),
+      () => {
+        this.props.onCountUpdate(this.state.count);
+      },
+    );
+  };
+
+  handleDecrement = () => {
```
yardnsm:
```suggestion
  _handleDecrement = () => {
```

src/Counter/Counter.js
```diff
@@ -0,0 +1,93 @@
...
+  render() {
+    const { count } = this.state;
+    const { dataHook, incrementText, decrementText, theme } = this.props;
+
+    return (
+      <div className={styles.root} data-hook={dataHook}>
+        <Heading dataHook="counter-count">{count}</Heading>
+
+        <div className={styles.buttons}>
+          <Button
```
yardnsm: use `IconButton` instead

src/Counter/Counter.js
```diff
@@ -0,0 +1,93 @@
...
+    initialCount: PropTypes.number,
+    /** Callback funtion to be called when the count has been changed. Signature is `onCountUpdate(newCount)`. */
+    onCountUpdate: PropTypes.func,
+    /** Text for the increment button */
+    incrementText: PropTypes.string,
+    /** Text for the decrement button */
+    decrementText: PropTypes.string,
+
+    /** Theme of the buttons */
+    theme: Button.propTypes.theme,
```
yardnsm: remove

src/Counter/Counter.js
```diff
@@ -0,0 +1,93 @@
...
+import Button from '../Button';
+
+class Counter extends React.PureComponent {
+  static displayName = 'Counter';
+
+  static propTypes = {
+    dataHook: PropTypes.string,
+
+    /** The initial count */
+    initialCount: PropTypes.number,
```
yardnsm: remove in order to simplify

.storybook/stories.js
argshook: no, i mean `kebab-case` as stated here https://github.com/wix/wix-ui/blob/master/docs/contribution-guide.md#project-structure

personally, i'd say no, those guidelines are only for wix-ui that core3 came up with and wix-style-react doesn't depend on them. but since we're thinking of generator which should provide best example, which should probably be consistent, it's something to think about.

¯\_(ツ)_/¯

src/Counter/Counter.js
```diff
@@ -0,0 +1,93 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styles from './Counter.scss';
```
hepiyellow: @shlomitc Did we say that new components should be written with Stylable or Sass?

src/Counter/Counter.js
```diff
@@ -0,0 +1,93 @@
...
+  };
+
+  static defaultProps = {
+    initialCount: 0,
+    onCountUpdate: () => {},
+    incrementText: '+',
+    decrementText: '-',
+  };
+
+  constructor(props) {
```
hepiyellow: There is no need for a constructor, if you only need to set `state` from `props`.
just do :
```js
state = {
      count: this.props.initialCount || 0,
};
```

src/Counter/README.md
```diff
@@ -0,0 +1,3 @@
+# `<Counter/>`
+
+> A simple counter component. This is the component's description.
```
hepiyellow: To encourage best-practice, we can have here real description.

src/Counter/Counter.js
```diff
@@ -0,0 +1,93 @@
...
+class Counter extends React.PureComponent {
+  static displayName = 'Counter';
+
+  static propTypes = {
+    dataHook: PropTypes.string,
+
+    /** The initial count */
+    initialCount: PropTypes.number,
+    /** Callback funtion to be called when the count has been changed. Signature is `onCountUpdate(newCount)`. */
+    onCountUpdate: PropTypes.func,
```
hepiyellow: `onCountChange`

src/Counter/Counter.js
hepiyellow: yeah, we don't want to encourage using foreign PropTypes

hepiyellow:
## Generator => Stubs

Generally, I think the generator should generate stubs (skeleton) without full implementation.
certainly not an example.
So for example, a generated `Counter.js` would be:
```js
import React from 'react';
import styles from './Counter.scss';

class Counter extends React.PureComponent {
   render() {
     return <div>Counter</div>
   }
}

export default Counter;
```
## `<BestPractice/>`

Another idea about Best-practices and convention , At first I thought it should be in the CONTRIBUTING docs, but I thought it would be cool to have a REAL component, which is called `<BestPractice/>`,
with real tests, full file structure, etc..
And we will update it whenever we decide on a new best-practice.
It can also appear in the storybook, why not.

src/Counter/Counter.js
```diff
@@ -0,0 +1,93 @@
...
+          >
+            {incrementText}
+          </Button>
+        </div>
+      </div>
+    );
+  }
+}
+
+export default Counter;
```
hepiyellow: I thought we want to encourage exporting like this:
`export {Counter}`;

And have the default Export only from the `index.js` file.
(For backwards consistency)

src/Counter/Counter.js
```diff
@@ -0,0 +1,93 @@
...
+
+    return (
+      <div className={styles.root} data-hook={dataHook}>
+        <Heading dataHook="counter-count">{count}</Heading>
+
+        <div className={styles.buttons}>
+          <Button
+            onClick={this.handleDecrement}
+            theme={theme}
+            dataHook="counter-decrement"
```
hepiyellow: Inspired by the `react-testing-library` that encourages best-practice.

Using `data-hook` is not a best-practice, but a last-resort.

If you were not to use it, you would ask yourself, how would I find this Button in tests, right?
And then you realize that you are missing :
```js
  aria-label="decrement"
```

Which in tests you can use `react-testing-library`'s `getByLabelText('decrement')`.

src/Counter/Counter.driver.private.js
```diff
@@ -0,0 +1,15 @@
+import { counterDriverFactory as publicDriverFactory } from './Counter.driver';
+
+export const counterPrivateDriverFactory = base => {
+  return {
+    ...publicDriverFactory(base),
+
+    /** Get the increment button text */
+    getIncrementText: async () =>
```
hepiyellow: why is this in a private driver?

src/Counter/Counter.spec.js
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import { createUniDriverFactory } from 'wix-ui-test-utils/uni-driver-factory';
+
+import Counter from './Counter';
+import { counterPrivateDriverFactory } from './Counter.driver.private';
+
+describe('Counter', () => {
+  const createDriver = createUniDriverFactory(counterPrivateDriverFactory);
```
hepiyellow: DONT use `createUniDriverFactory` ! It offers no cleanup.
Use `react-testing-library`.
Import `createRendererWithDriver` from [here](https://github.com/wix/wix-style-react/blob/14759064d183601b8568c3abe71cbee49a4b24a9/test/utils/react/index.js#L17) .

Actually you need to add a new function there `createRendererWithUniDriver`.

And don't forget to import `cleanup`, and do:
```js
afterEach(()=>cleanup());
```

src/Counter/Counter.spec.js
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
```
hepiyellow: General concern.
Are we intending to switch to mocha-runner (in-browser)?
If so we just need to be minded of that, and for example not use jest specific stuff (like `useFakeTimers` I guess).

src/Counter/Counter.js
yardnsm: The idea is to eventually let the user choose between sass or stylable in the component generator. We'll keep things simple for now, so we'll start with sass.

src/Counter/README.md
yardnsm: This description won't be hard-coded as it will come from the generator.

src/Counter/Counter.js
yardnsm: @shlomitc wdyt?

.storybook/stories.js
yardnsm: We'll keep the PascalCase for now, maybe change it in the future.

src/Counter/Counter.spec.js
yardnsm: We've decided to stick with `createUniDriverFactory` as `createRendererWithDriver` only support the vanilla (non-uni) drivers.

src/Counter/Counter.spec.js
argshook: at least in Vilnius office most people are migrating to jest, not sure we need to invest much into mocha

shlomitc:
@yardnsm - can we close this one?

PR #2545, opened 2018-12-07
Testkit Smoke Test - add BeforeAllHook and make RichTextArea teskit check work

mykas:
This PR removes RichTextArea from ignored list of `testkit-smoke.test.js` . RichTextArea is depended on window.getSelecion() which is defined in external library `slate`. Now we can define `beforeAllHook` in `driverconfig` with a function that can handle these types of issues before the test happens.


testkit/testkit-smoke.test.js
```diff
@@ -231,12 +239,15 @@ Object.entries(AllComponents).forEach(([name, component]) => {
   if (driverConfig) {
     const drivers = driverConfig.drivers || ['vanilla', 'enzyme'];
     const props = driverConfig.props || {};
+    const beforeAllHook = driverConfig.beforeAllHook || null;
```
argshook: why `null`?

testkit/testkit-smoke.test.js
```diff
@@ -231,12 +239,15 @@ Object.entries(AllComponents).forEach(([name, component]) => {
...
 
     if (driverConfig.unidriver) {
-      DRIVER_ASSERTS.enzymeUni({ name, component, props: {} });
-      DRIVER_ASSERTS.vanillaUni({ name, component, props: {} });
+      DRIVER_ASSERTS.enzymeUni({ name, component, props: {}, beforeAllHook });
```
argshook: noticed that `props` here is set as empty object. is this ok?

testkit/testkit-smoke.test.js
```diff
@@ -171,9 +172,13 @@ const AllComponents = importAllComponents({
   ignore: FAILING_COMPONENTS,
 });
 
+const handleBeforeAllHook = task =>
+  beforeAll(async () => task && (await task()));
```
argshook: perhaps together  you could do `afterAll`? could be useful as teardown block for other  components

testkit/testkit-smoke.test.js
mykas: well we already handle this in variable so I guess we can remove this.

testkit/testkit-smoke.test.js
mykas: you mean to add another handler in the same handling mechanism?

testkit/testkit-smoke.test.js
mykas: why not? I could do `undefined` too.

mykas: hm I guess if driverConfig.beforeAllHook is undefined it will resolve beforeAllHook undefined too?

testkit/testkit-smoke.test.js
argshook: yep, if  we support `beforeAll`, why  shouldn't we support  `afterAll` too

PR #2546, opened 2018-12-07
<Dropdown/>: Controlled when selectedId and onSelect are passed

Wo1v3r:
### What changed

- Updated <Dropdown/> controlled example
- Changed <Dropdown/> not to change local state when is controlled
- Updated <Dropdown/> readme to include this information
- Added a relevant test for <Dropdown/>

### Why it changed

- It is a requirement in Price Quotes Web, to prompt user before changing the dropdown value

---

- [x] Change is tested
- [x] Change is documented
- [ ] Build is green

### DIFF
--- a/src/Dropdown/Dropdown.js
+++ b/src/Dropdown/Dropdown.js
@@ -67,11 +67,20 @@ class Dropdown extends InputWithOptions {
     return { readOnly: true, value: this.state.value };
   }
 
+  _isControlled() {
+    return (
+      typeof this.props.selectedId !== 'undefined' &&
+      typeof this.props.onSelect !== 'undefined'
+    );
+  }
+
   _onSelect(option) {
-    this.setState({
-      value: this.props.valueParser(option),
-      selectedId: option.id,
-    });
+    if (!this._isControlled()) {
+      this.setState({
+        value: this.props.valueParser(option),
+        selectedId: option.id,
+      });
+    }
     super._onSelect(option);
   }
 }
--- a/src/Dropdown/Dropdown.spec.js
+++ b/src/Dropdown/Dropdown.spec.js
@@ -32,6 +32,20 @@ describe('Dropdown', () => {
     expect(inputDriver.getValue()).toBe('Option 1');
   });
 
+  it('should not set item value when controlled', () => {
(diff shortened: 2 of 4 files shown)


Wo1v3r:
Marked last bullet as there is no UX change in WSR flow (example is naive with `confirm()`)

hepiyellow:
## Breaking Change

This is a breaking change.
Some consumers pass `selectedId` and mean for it to be the `initialSelectedId` (since that's how the component works)
For example this [`LabeledDropdown`](https://github.com/wix-private/cashier/blob/afc498b1a1ed9978312004283e6609274f7688a8/cashier-payments-dashboard/src/components/LabeledDropdown/LabeledDropdown.tsx) from `cashier`
and it's [usage](https://github.com/wix-private/cashier/blob/afc498b1a1ed9978312004283e6609274f7688a8/cashier-payments-dashboard/src/components/TransactionsFilters/TransactionsFilters.tsx#L43) which passes 
```js
selectedId: DefaultOptionNoFilter
```
They also pass an `onSelect` prop.

Your change of making it controlled, will break them.

## How to make a non-breaking change

The problem is with naming, the `selectedId` should have been named `initialSelectedId` in the first place, but it wasn't (-:
Unfortunately, we have this pattern in other components as well.

Now the best-practice solution would be to:
- Add a `controlled` boolean prop. When `true` the `selectedIds` would act as a controlled prop.
- Add an `initialSelectedId` prop.
- Add deprecationLog that prints only when `selectedId` is used and `controlled=false`. The log should instruct the consumer to use `initialSelectedId` instead.

Later on, we can have a major version, where we make `selectedId` controlled and deprecate the  `controlled` prop.

hepiyellow:
@Wo1v3r  said he might have time for this by the end of this week.

shlomitc:
ping

Wo1v3r:
@shlomitc  Hey, looks like we'll still need this, will fix the PR this week to what @hepiyellow  mentioned

hepiyellow:
@Wo1v3r Closing this PR.
You can re-open it when you have time to work on it.

Wo1v3r:
@hepiyellow Starting work on this right now :]

PR #2549, opened 2018-12-09
Docs -  add CONTROLLED_VS_UNCONTROLLED.md

hepiyellow:


docs/contribution/COMPONENT_GUIDELINES.md
```diff
@@ -12,11 +12,14 @@ A component will be named with a `CamelCase` convention.
...
-1. *dataHook*: Every component should accept a `dataHook` property which is applied as a `data-hook` attribute on the component's root element. This attribute is use as a selector in tests. A component can extend `WixComponent` in order to be a pure component and have the `dataHook` prop applied automatically.
+## API (Props)
+
+| Prop Rule | Description |
+|--------|-----------|
+| *dataHook*| Every component should accept a `dataHook` property which is applied as a `data-hook` attribute on the component's root element. This attribute is use as a selector in tests. A component can extend `WixComponent` in order to be a pure component and have the `dataHook` prop applied automatically.
+| *className* | The convension for allowing styling of the component's root element is by exposing a `className` property.
+| *initialPropName*| props which set an initial value to an identical uncontrolled state, should be named with an `initial` prefix. [Read More About Controlled & Uncontrolled Components](./CONTROLLED_VS_UNCONTROLLED.md)
```
shlomitc: I don't think each component should have it. this is not a standard thing.
it's happening only in a case of uncontrolled components, which should be only a few.
it's ok to keep the controlled/uncontrolled section, but I wouldn't recommend it by default

PR #2550, opened 2018-12-09
`<Tooltip/>` - Fix testkit `isShown()` to be specific to the Tooltip instance

### LINKED ISSUE #2486: tooltip testkit isShown will return true for all tooltips in the tree
# Bug Report
If i have more than one tooltip in a component, it is enough for one to be shown in order for all the others to be considered shown as well 

### Relevant Components
Tooltip

### Steps to Reproduce
consider that in my App there are two tooltips one with dataHook 'first' the other with 'second' and that both are shown and hidden on mouse enter and mouse leave
now in my test files 

const firstTooltipTestKit = tooltipTestKitFactory(wrapper, dataHook: 'first');
const secondTooltipTestKit .= tooltipTestKitFactory(wrapper, dataHook: 'second');

firstTooltipTestKit.mouseEnter();

firstTooltipTestKit.isShown() will return true as expected
secondTooltipTestKit.isShown() . will also return true or in fact any other testkit for any other Tooltip

### Severity

Major

ehabshs:
Resolved #2486 

### What changed in Component
- tooltip testkit `isShown` looks for an element with a uniq dataHook of the form `tooltip-content-${tooltipDataHook}-${uniq_id}. Where `uniq_id` is uniqe to every Tooltip instance.
- The  `${tooltipDataHook}` is only for convenience.

### Why it changed
Previously isShown depended on querying the whole document for the class .tooltip, this means once a tooltip is shown tests will always return true for isShown for any other tooltip.

#### showImmediately

Make showImmediately NOT use any timer (instead of `setTimeout()` with delay 0)

(Due to failing test which may have been for this reason, the test didn't wait)

#### Component Test cleanup
- Remove `setProps` from driver (@shlomitc said we don't need to mind about breaking consumers who used this method)
- Use smoke-test
- Remove use of enzyme
- Remove cleanup hack of `document.body.innerHTML = ''`


src/Tooltip/Tooltip.driver.js
```diff
@@ -28,7 +28,12 @@ const tooltipDriverFactory = ({ element, wrapper }) => {
   const getTooltipContent = () => bodyOrWrapper.querySelector('.tooltip');
   const mouseEnter = () => ReactTestUtils.Simulate.mouseEnter(element);
   const mouseLeave = () => ReactTestUtils.Simulate.mouseLeave(element);
-  const isShown = () => !!bodyOrWrapper.querySelector('.tooltip');
+  const isShown = () => {
+    const hook = element.getAttribute('data-hook');
```
hepiyellow: What if `data-hook` is not provided?
This could happen if someone creates the driver themselves and not through the `wix-style-react/dist/testkit` driver creators (which use data-hook).

For example, they can:
```js
import {render} from 'react-testing-library';
import {createTooltipDriver} from 'wix-style-react/dist/src/component/Tooltip/Tooltip.driver';
const {container} = render(<Tooltip />);
const driver = createTooltipDriver(container.firstChild);
```

I think in this case we need to generate a unique id.
We can use `uuid` library for that (as a devDependency)

We could ONLY use generated ids, but actually it is good to use the data-hook in cases it is provided, to make debugging easier.

src/Tooltip/Tooltip.js
```diff
@@ -256,6 +257,7 @@ class Tooltip extends WixComponent {
       const position = this.props.relative ? 'relative' : 'absolute';
       const tooltip = (
         <TooltipContent
+          dataHook={this.props.dataHook}
```
hepiyellow: Better add the `tooltip-content-` prefix here.
So that `<TooltipContent/>` will be consistent with other components and simply pass on the dataHook onto it's root element unchanged.

src/Tooltip/Tooltip.spec.js
```diff
@@ -60,7 +61,12 @@ describe('Tooltip', () => {
       </div>
     );
     const driver = createDriver(
-      <Tooltip showDelay={5} hideDelay={5} content={buttonContent}>
+      <Tooltip
+        showDelay={5}
+        hideDelay={5}
+        content={buttonContent}
+        dataHook="myDataHook"
```
hepiyellow: dataHook is not needed for this test

src/Tooltip/Tooltip.spec.js
```diff
@@ -495,14 +501,12 @@ describe('Tooltip', () => {
   describe('testkit', () => {
     const createTooltipTestkitDriver = props => {
       const div = document.createElement('div');
-      const dataHook = 'myDataHook';
+      const dataHook = (props && props.dataHook) || _props.dataHook;
       const tooltipProps = { ..._props, ...props };
```
hepiyellow: I think you want to add the dataHook to tooltipProps

src/Tooltip/Tooltip.driver.js
ehabshs: I'm not well familiar with how standalone driver creators work but two questions here:
1) if i generated a uuid then how can i retrieve it in the implementation - currently we retrieve the  datahook by using element.getAttribute(), how do we retrieve it then?
2) does not the above code work if we did
const {container} = render(<Tooltip  dataHook="someDataHook"/>);

hepiyellow: 1. You put the generated id on the root element as well. And do element.getAttribute(data-test-id)
2. I didnt understand the question

src/Tooltip/Tooltip.spec.js
hepiyellow: If we add generated test id, it wont fail

src/Tooltip/Tooltip.spec.js
hepiyellow: Well the local dataHook is not used, so remove it

hepiyellow:
@ehabshs Forgive me, but I have taken over this PR (-:
- I implemented the `testId` feature that doesn't required `dataHook` in order for `isShown` to work.
- I made a big cleanup in the spec test, and removed a lot of shit.

ehabshs:
@hepiyellow Np, it is good with me, just notice that other methods in the tooltip testkit like getTooltipContent, hasErrorTheme, hasDarkTheme, hasLightTheme,  hasAnimationClass, hasArrow work by querying global class names, and so they have the same problem like isShown

hepiyellow:
@ehabshs Thanks, I'll take care of that.

hepiyellow:
Ready to merge, waiting for build to pass.

PR #2552, opened 2018-12-09
Add a component generator

yardnsm:
This PR adds a new component generator to `wix-style-react`. The generated component resembles the template in https://github.com/wix/wix-style-react/pull/2543/.

Run the generator via:

```console
$ npm run generate
```

It'll prompt you for the new component name and description. If you provide a description, a `README.md` file will be generated for the component.

#### Generation flow

The following is the process of a new component generation:

1. The generator checks if the working directory is clean, and will proceed only if it is. You can use the `--force` flag to skip this step.
1. The generator prompts the user for:
	* The component's name
		* Must be in `PascalCase`
		* Must not be an existing component name
	* A description
1. The generator will creating the component based on pre-defined templates:
	* A `README.md` file will be generated only if a description was provided.
1. The generator will run 3 codemods:
	* Add the new component's story to `.storybook/stories.js`
	* Add the new component to the index file
	* Add testkit exports (in `testkit/{index,enzyme,protractor,puppeeter}.js`).
1. The generator will run `yoshi lint --fix` on the generated component file in order to fix potential lint issues.
1. Profit.

#### About the codemods

Currently, the codemods are very primitive; they simply add a new line at the of the file. The idea is to improve them in the future, for example:

* Add an `import` to `.storybook/stories.js` after a specific marker comments or so.
* Only add the `import` / `export` statement of they do not exist.

Although we could simply append a new lines, the codemods are using `jscodeshift` so we could improve them more easily.

#### Testing the generated component

While the generator itself is tested, we also need to test a generated component with the entire WSR library, to make sure it'll interface properly with the storybook, and its tests will pass. Think of it as an end-to-end test for the generator.

The way it works is by permanently store a generated test component in the library, whose name is (unsurprisingly) `<GeneratedTestComponent/>`. There are some differences between this component and a "regular" generated component:

- The test component is not `export`ed from the index file.
- Its story is located under `Tests`, and not under `Component` (which is the default location for generated components).
- It's generation does run any codemods for now, but we'll probably want to run them when they'll become more robust.
- Since the component it not added to the index file, we can't use `<LiveCodeExample/>` in the story (since it's context depends on the index file). Therefore, the test component's story has no `<LiveCodeExample/>`.

When making changes to the generator (more specifically, to the template files), make sure to:

1. Update the snapshot tests, by running:

	```console
	$ npm run test -- -u
	```

1. Re-generating the test component, by running:

	```console
	$ npm run generate-test-component
	```

	The test component will be also generating during the `build` script.

The test component will also be generated in the CI (as it generated during build). If you forgot to update the test component after making changes, you'll see a message in the CI:

screen shot 2018-12-10 at 17 37 40


stories/GeneratedTestComponent/index.story.js
```diff
@@ -0,0 +1,31 @@
...
+  },
+
+  exampleProps: {
+    // Put here presets of props, for more info:
+    // https://github.com/wix/wix-ui/blob/master/packages/wix-storybook-utils/docs/usage.md#using-list
+  },
+
+  examples: (
+    <div style={{ maxWidth: 627 }}>
+      <GeneratedTestComponent
```
shlomitc: @yardnsm - this should use the live code editor by default. it is pointless to show an example without a snippet

scripts/component-generator/templates/component/src/Component/Component.e2e.js
```diff
@@ -0,0 +1,43 @@
...
+  createStoryUrl,
+  waitForVisibilityOf,
+  scrollToElement,
+} from 'wix-ui-test-utils/protractor';
+
+import { eyesItInstance } from '../../test/utils/eyes-it';
+import { {%componentName%}TestkitFactory } from '../../testkit/protractor';
+import { storySettings } from '../../stories/{%ComponentName%}/storySettings';
+
+const eyes = eyesItInstance();
```
shlomitc: should be `const eyes = eyesItInstance({ enableSnapshotAtBrowserGet: false });` or change it to be the default behavior in `eyesItInstance`

scripts/component-generator/src/generate-component.js
```diff
@@ -0,0 +1,41 @@
+const chalk = require('chalk');
```
shlomitc: it's a missing dependency in package.json

scripts/component-generator/templates/component/src/Component/Component.e2e.js
yardnsm: https://github.com/wix/wix-style-react/pull/2568

PR #2557, opened 2018-12-10
<InputArea/> - allow to grow while typing

jonathanadler:
### What changed

Added autoGrow feature to InputArea component

### Why it changed

Feature request

- [ v ] Change is tested
- [ v ] Change is documented
- [ ? ] Build is green


src/InputArea/InputArea.protractor.driver.js
```diff
@@ -1,11 +1,19 @@
...
   hover: () => mouseEnter(component.$('textarea')),
   click: () => component.$('textarea').click(),
   isFocused: () => isFocused(component.$('textarea')),
   isHovered: () => component.$(':hover').isPresent(),
+  sendKeys: keys => component.$('textarea').sendKeys(keys),
+  getOffsetHeight: () => component.$('textarea').getAttribute('offsetHeight'),
+  getScrollHeight: () => component.$('textarea').getAttribute('scrollHeight'),
+  setLineHeight: lineHeight =>
+    browser.executeScript(
+      `document.querySelector('textarea').style.lineHeight = '${lineHeight}'`,
```
shlomitc: I think this should be taken somehow from the `component` and not query on the document, if you will have several textarea, you will not be able to tell which one should be taken

src/InputArea/InputArea.js
```diff
@@ -14,18 +16,24 @@ class InputArea extends WixComponent {
...
   }
 
   state = {
     focus: false,
     counter: (this.props.value || this.props.defaultValue || '').length,
+    computedRows: InputArea.MIN_ROWS,
   };
 
+  // For autoGrow prop min rows is 2 so the textarea doesn't look like an input
+  static MIN_ROWS = 2;
```
shlomitc: can this be unit-tested?

src/InputArea/InputArea.js
```diff
@@ -175,6 +187,65 @@ class InputArea extends WixComponent {
...
+    this.setState({ computedRows: 1 }, () => {
+      const rowsCount = this._getRowsCount();
+      const computedRows = Math.max(InputArea.MIN_ROWS, rowsCount);
+      this.setState({
+        computedRows,
+      });
+    });
+  }
+
+  _updateComputedStyle = debounce(
```
shlomitc: better have all these funcions in a separate file, it adds a lot to the component logic...

src/InputArea/InputArea.e2e.js
```diff
@@ -46,4 +48,56 @@ describe('input area page', () => {
     expect(await inputAreaTestkit.isFocused()).toBeTruthy();
     expect(await inputAreaTestkit.isHovered()).toBeTruthy();
   });
+
+  eyes.it('should grow automatically with input', async () => {
```
shlomitc: please use `eyesInstance`, in order to have only a single screenshot instead of two (single snapshot)
https://github.com/wix/wix-style-react/blob/master/src/Avatar/Avatar.e2e.js#L32

src/InputArea/InputArea.js
```diff
@@ -58,6 +67,8 @@ class InputArea extends WixComponent {
     } = this.props;
 
     const inlineStyle = {};
+    const rowsAttr = rows ? rows : (autoGrow ? this.state.computedRows : undefined);
```
shlomitc: If possible, I prefer to put all `autoGrow` logic into a single place, just so it will not be mixed in the multiple conditions.
Maybe have it in a function that returns the relevant props for the input? WDYT?

src/InputArea/InputArea.e2e.js
```diff
@@ -46,4 +48,56 @@ describe('input area page', () => {
...
+    const longValue = 'asdf\n'.repeat(30);
+
+    await inputAreaTestkit.sendKeys(longValue);
+    const currentOffsetHeight = await inputAreaTestkit.getOffsetHeight();
+
+    expect(currentOffsetHeight).toBe(previousOffsetHeight);
+  });
+
+  eyes.it('should grow according to input when line-height is set to "normal"', async () => {
+    await inputAreaTestkit.setLineHeight('normal');
```
shlomitc: what is the use case a `lineHeight` is set to normal?
how can I simulate it in the storybook?
setting the line-height of the textarea the way the driver does it looks very strange from a user usage prespective.

src/InputArea/InputArea.protractor.driver.js
Tokyros: As far as we've seen there is not other way to set style properties using the `component`.

src/InputArea/InputArea.e2e.js
Tokyros: This test is crucial since it tests an important branch in the logic of the component.
If the line-height is not set, we calculate it based on the browser's default line-height, this is what this test addresses.

src/InputArea/InputArea.js
Tokyros: Added tests for this in e2e, cannot be tested in spec due to lack of height in jsdom

src/InputArea/InputArea.e2e.js
Tokyros: Done

hepiyellow:
Just wondering...
Did you take a look at this very popular package?
https://www.npmjs.com/package/react-textarea-autosize

Tokyros:
@hepiyellow Our implementation should be taking these things into consideration due to the use of `getComputedStyles()`, which should calculate the height based on box-sizing, padding, etc...

regarding the API points, the apis they provide can be easy to implement for out component, so it's mostly a product decision weather we want these features or not (min\max rows, onHeightChanged...)

src/InputArea/InputArea.js
```diff
@@ -104,6 +111,7 @@ class InputArea extends WixComponent {
             onBlur={this._onBlur}
             onKeyDown={this._onKeyDown}
             onChange={this._onChange}
+            onInput={this._getMaybeOnInput()}
```
shlomitc: just call `this._onInput` and inside decide whether do an action or not

src/InputArea/InputArea.js
```diff
@@ -144,6 +152,11 @@ class InputArea extends WixComponent {
     this.textArea && this.textArea.select();
   }
 
+  _onSetRef(ref) {
+    this.textArea = ref;
+    this._computedStyle = window.getComputedStyle(this.textArea);
```
shlomitc: is this calculation done here for better caching?
is this a costly function? if not, I prefer calling it from `_onInput` (or in `calculateRowCount` if possible)

src/InputArea/InputArea.protractor.driver.js
shlomitc: what about `getCssValue`?
https://www.protractortest.org/#/api?view=webdriver.WebElement.prototype.getCssValue

we should find a way to get the unique value and not just a general textarea

src/InputArea/rowCount.utils.js
```diff
@@ -0,0 +1,36 @@
+import isNaN from 'lodash/isNaN';
+
+export default function calculateRowCount(computedStyle, height) {
```
shlomitc: prefer a named function instead of a default one

src/InputArea/rowCount.utils.js
```diff
@@ -0,0 +1,36 @@
+import isNaN from 'lodash/isNaN';
```
shlomitc: just name it `utils` for now

hepiyellow:
@Tokyros 
`react-textarea-autosize` also uses `getComputedStyles` and then does many calculations (more than us) according to it.
Can you go over [their implementation](https://github.com/andreypopp/react-textarea-autosize/blob/c168ae1c4916694b2d480bfa14ea973240811336/src/calculateNodeHeight.js#L129), and see what we are missing?

(for example, they have some IE bug workaround)

src/InputArea/InputArea.js
```diff
@@ -175,6 +188,29 @@ class InputArea extends WixComponent {
     this.props.hasCounter && this.setState({ counter: e.target.value.length });
     this.props.onChange && this.props.onChange(e);
   }
+
+  _getMaybeOnInput() {
```
shlomitc: just call it `shouldAutoGrow` and decide to set the `onInput` based on this value

src/InputArea/InputArea.protractor.driver.js
```diff
@@ -1,11 +1,24 @@
...
+import { browser } from 'protractor';
 
 const toggleSwitchDriverFactory = component => ({
   element: () => component,
   hover: () => mouseEnter(component.$('textarea')),
   click: () => component.$('textarea').click(),
   isFocused: () => isFocused(component.$('textarea')),
   isHovered: () => component.$(':hover').isPresent(),
+  sendKeys: keys => component.$('textarea').sendKeys(keys),
+  getOffsetHeight: () => component.$('textarea').getAttribute('offsetHeight'),
```
shlomitc: should be in private driver

src/InputArea/InputArea.js
Tokyros: moved the grow/shrink logic inside _onChange

src/InputArea/InputArea.js
Tokyros: Seems like input on performance should not be significant, so this moved to the calculation itself

adavidof:
@shlomitc 
yeah, sorry, understood it before your comment, and deleted my comment before you answered :)

PR #2565, opened 2018-12-11
<Proportion/> - create a new aspect ratio component

jonathanadler:
### What changed

Added the new Proportion component

### Why it changed

Feature request

Created by @Tokyros & @jonathanadler 

---

- [ v ] Change is tested
- [ v ] Change is documented
- [ ? ] Build is green


src/Proportion/Proportion.scss
```diff
@@ -0,0 +1,21 @@
...
+    -ms-grid-column: 1;
+    -ms-grid-column-span: 1;
+    grid-area: 1 / 1 / 2 / 2;
+
+    &.svg {
+      z-index: 0;
+    }
+
+    &:not(.svg) {
+      z-index: 1;
```
hepiyellow: since we have hard-coded z-index in css, we might need a zIndex prop on the component, so consumers can put this component in a certain zIndex (we will add 1 to the &:not(.svg))

src/Proportion/Proportion.scss
```diff
@@ -0,0 +1,21 @@
...
+    -ms-grid-row-span: 1;
+    -ms-grid-column: 1;
+    -ms-grid-column-span: 1;
+    grid-area: 1 / 1 / 2 / 2;
+
+    &.svg {
+      z-index: 0;
+    }
+
+    &:not(.svg) {
```
hepiyellow: What if someone passes `svg` as a child ?
Maybe we need our own class name on the first `svg`

src/Proportion/Proportion.js
```diff
@@ -0,0 +1,56 @@
...
+    aspectRatio: 1,
+  };
+
+  render() {
+    const { dataHook, children } = this.props;
+    const ratio = this._getRatio();
+
+    return (
+      <div className={styles.root} data-hook={dataHook}>
+        <img className={styles.svg} src={`data:image/svg+xml,%3Csvg viewBox='0 0 ${ratio[0]} ${ratio[1]}' xmlns='http://www.w3.org/2000/svg'%3E%3C/svg%3E`} />
```
hepiyellow: Can you explain this `img` tag and how it affects the following children?
(For those of us which are not so css proficient)

src/Proportion/Proportion.js
```diff
@@ -0,0 +1,56 @@
...
+  portrait: [3, 4],
+  cinema: [16, 9],
+  landscape: [4, 3],
+};
+
+class Proportion extends React.PureComponent {
+  static displayName = 'Proportion';
+
+  static propTypes = {
+    children: PropTypes.node,
```
hepiyellow: I think `children` should be a required prop, or is it?

src/Proportion/Proportion.js
```diff
@@ -0,0 +1,56 @@
...
+    dataHook: PropTypes.string,
+
+    /** predefined aspectRatios */
```
hepiyellow: State that it overrides the `aspectRatio` prop

src/Proportion/Proportion.js
```diff
@@ -0,0 +1,56 @@
...
+  static defaultProps = {
+    aspectRatio: 1,
+  };
+
+  render() {
+    const { dataHook, children } = this.props;
+    const ratio = this._getRatio();
+
+    return (
+      <div className={styles.root} data-hook={dataHook}>
```
hepiyellow: You should also accept a `className` prop and append it.
See https://github.com/wix/wix-style-react/blob/master/docs/contribution/COMPONENT_GUIDELINES.md#api-props

src/Proportion/Proportion.scss
jonathanadler: This shouldn't really affect the user, since the user's content is on another stacking context.
When we'll add the `className` prop, the user will be able to give the whole component a `z-index` if they'll need one. Other than that, any `z-index` inside the user's content won't be affected.

src/Proportion/Proportion.scss
jonathanadler: It's actually a class 😄 but we should rename it for readability.

argshook:
hello!

i might not know enough details but it appears that this solution is
over complex.

as far as i understand, the purpose of `<Proportion/>` component is to
ensure some div has correct width and height ratio and any given
`children` are rendered in the middle with 100% width and height.


here's how it can be done simply by using `padding-bottom` and `position: absolute`:

https://jsbin.com/sugogid/1/edit?html,css,output

`padding-bottom` value, for 16/9 ratio is `9 / 16 = 56.25`

this solution might have some edge cases that we need to take care of,
but i certainly think it's easier than having to deal with svg, zindex
and such

maybe this technique could be used here too, wdyt?

src/Proportion/Proportion.js
jonathanadler: So we used Wix's Noam Rosenthal's solution for preserving aspect ratio (which can be found [here](https://codeburst.io/keeping-aspect-ratio-with-html-and-no-padding-tricks-40705656808b)), which uses SVG's property to maintain aspect ratio.
However, in IE11, SVG's have a problem setting the correct height, so we had to put the SVG inside an `img` element, which fixed the issue.

src/Proportion/Proportion.js
jonathanadler: Sure, will do

src/Proportion/Proportion.js
argshook: please see my comment [here](https://github.com/wix/wix-style-react/pull/2565#issuecomment-446252483)

the article you linked talks about padding trick too, but for some reason disregards it for svg and css grid (and in this implementation also z-index, and problematic css selectors)

in the end it even says:
> (Of course the padding trick may still seem nicer, a matter of taste…)

so i'm a little dumbfounded, why use multiple technologies if a mere `padding-bottom` does the trick? maybe some edge case or something?

src/Proportion/Proportion.js
jonathanadler: Done

src/Proportion/Proportion.js
jonathanadler: Done

src/Proportion/Proportion.js
jonathanadler: You might be right, both options have their own pros and cons.
@shlomitc what do you think?
Personally we feel this option is more readable and works just as well, and we have complete confidence in it.

argshook: so i would like to know those pros and cons, cause maybe `padding` thing doesn't cover all cases and there really is a need for complexity as in this PR.

from maintenance perspective, i have no confidence in z-index, questionable css selectors, and svg. moreover, svg as data-url inside img, with specific viewbox, specific grid, thinkin of handling IE...

please prove me wrong. not as a challenge, but i might really be missing something here :)

stories/Proportion/ExampleComponent.js
```diff
@@ -0,0 +1,35 @@
+import React, {Component} from 'react';
+import PropTypes from 'prop-types';
+import Proportion from "../../src/Proportion";
```
shlomitc: Because this is an example that consumers see, we have a webpack alias to make imports look nicer.
just do `import Proportion from 'wix-style-react/Proportion';`

do the same for the rest of the imports

stories/Proportion/ExampleComponent.js
```diff
@@ -0,0 +1,35 @@
...
+
+  render() {
+    const { predefined, aspectRatio } = this.props;
+    return (
+      <Container fluid>
+        <div className={styles.wrapper}>
+          <Row stretchViewsVertically className={styles.row}>
+            <Col span={4}>
+              <Proportion predefined={predefined} aspectRatio={aspectRatio}>
+                <AddItem />
```
shlomitc: this looks strange. The + sign should look more like this:
image

this is how it looks now
image

stories/Proportion/ExampleComponent.js
```diff
@@ -0,0 +1,35 @@
...
+  static propTypes = {
+    predefined: PropTypes.string,
+    aspectRatio: PropTypes.number,
+  };
+
+  render() {
+    const { predefined, aspectRatio } = this.props;
+    return (
+      <Container fluid>
+        <div className={styles.wrapper}>
```
shlomitc: this is not how Grid should be used.
It should simply be a Container=>Row=>Col

stories/Proportion/ExampleComponent.js
```diff
@@ -0,0 +1,35 @@
+import React, {Component} from 'react';
+import PropTypes from 'prop-types';
+import Proportion from "../../src/Proportion";
+import { Container, Row, Col } from '../../src/Grid';
+import styles from './ExampleComponent.scss';
```
shlomitc: We shouldn't need to use any additional styles.
if it doesn't work out of the box, let me know

testkit/protractor.js
```diff
@@ -346,6 +346,12 @@ export const carouselTestkitFactory = protractorTestkitFactoryCreator(
   carouselDriverFactory,
 );
 
+import { proportionDriverFactory } from '../src/Proportion/Proportion.protractor.driver';
```
shlomitc: after implementing your function in unidriver, make sure you export it correctly in each one of the `testkit/***.js` file:
```js
import { textButtonDriverFactory } from '../src/TextButton/TextButton.driver';

export const textButtonTestkitFactory = uniTestkitFactoryCreator(
  textButtonDriverFactory,
);

export const textButtonTestkitFactory = enzymeUniTestkitFactoryCreator(
  textButtonDriverFactory,
);

export const textButtonTestkitFactory = protractorUniTestkitFactoryCreator(
  textButtonDriverFactory,
);

export const textButtonTestkitFactory = puppeteerUniTestkitFactoryCreator(
  textButtonDriverFactory,
);
```

we will automate this soon, but currently it's still manual

stories/Proportion/ExampleComponent.js
```diff
@@ -0,0 +1,35 @@
...
+  };
+
+  render() {
+    const { predefined, aspectRatio } = this.props;
+    return (
+      <Container fluid>
+        <div className={styles.wrapper}>
+          <Row stretchViewsVertically className={styles.row}>
+            <Col span={4}>
+              <Proportion predefined={predefined} aspectRatio={aspectRatio}>
```
shlomitc: here again, consumers won't see the actual code of what `predefined` is.

So I suggest putting all examples in the same file currently...

src/Proportion/Proportion.e2e.js
```diff
@@ -0,0 +1,78 @@
...
+  waitForVisibilityOf,
+  scrollToElement,
+} from 'wix-ui-test-utils/protractor';
+import autoExampleDriver from 'wix-storybook-utils/AutoExampleDriver';
+import { eyesItInstance } from '../../test/utils/eyes-it';
+import { proportionTestkitFactory } from '../../testkit/protractor';
+import { storySettings } from '../../stories/Proportion/storySettings';
+import { PREDEFINED_RATIOS } from './Proportion';
+
+const eyes = eyesItInstance();
```
shlomitc: `const eyes = eyesItInstance({ enableSnapshotAtBrowserGet: false });`

stories/Proportion/ExampleComponent.js
jonathanadler: I think it's just the `size` prop the `AddItem` component receives.
The default size seems to be `'tiny'`.

src/Proportion/Proportion.js
jonathanadler: @argshook no problem, this is what pr's are for.
I'll try to break down our solution, I hope i'll be able to make it clearer.

So first, regarding the `padding-top` solution, as you can see in this [article](https://www.bram.us/2017/06/16/aspect-ratios-in-css-are-a-hack/), all aspect-ratio solutions are considered hacks. 
In addition, there is a known issue (which I guess can be addressed and handled) when the component is placed inside a `flex` component, `padding-top` isn't calculated the same way (you can read about it [here](https://www.bram.us/2017/07/30/vertical-marginspaddings-and-flexbox-a-quirky-combination/).

So I think it really just boils down to a matter of taste.

The SVG solution, uses the fact that an svg maintains the aspect ratio of its `viewBox`, and so we considered it a more declarative and readable solution (less "magic").

We did however make a few adjustments the solution, in order to support different browsers and situation.

- The `svg` inside an `img` was used in order to support ie, because the svg wasn't rendered with the correct height (we checked this using browser stack)
- The `z-index` - Since the solution uses two siblings (the svg element and the content) inside one parent container, we wanted to ensure the the svg will not take any pointer events from the content itself. We currently replaced the `z-index` solution to just using `pointer-events: none` on the svg element (however, this is not supported in ie < 11).
However, just to be clear, using z-index this way, has no affect whatsoever on any other elements, as you can see [here](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context)

I hope I was clear and didn't miss anything, feel free to ask anything

PR #2566, opened 2018-12-11
<CardGalleryItem/> - new component

ihork:
### What changed

image

### Why it changed

...

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


ihork:
@shlomitc could you take a look? Also please notice that it's only POC, so there is no test coverage here yet.

src/Card/Card.scss
```diff
@@ -3,6 +3,7 @@
 .card {
   background: $D80;
   border-radius: 8px;
+  overflow: hidden;
```
shlomitc: this can't be applied for every card. it has a massive effect on card content, so it should be applied only when needed.

src/CardGalleryItem/CardGalleryItem.js
```diff
@@ -0,0 +1,99 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import Hover from './Hover';
+import Card from '../Card';
+import Button from '../Button';
+import TextLink from '../TextLink';
```
shlomitc: use `TextButton` instead

src/CardGalleryItem/CardGalleryItem.js
```diff
@@ -0,0 +1,99 @@
...
+    title: PropTypes.string,
+
+    /** Card subtitle */
+    subtitle: PropTypes.string,
+
+    /** Background image of CardGalleryItem */
+    backgroundImageUrl: PropTypes.string,
+
+    /** Properties for primary action button */
+    primaryActionProps: PropTypes.object,
```
shlomitc: please define the actual object structure

src/CardGalleryItem/CardGalleryItem.scss
```diff
@@ -0,0 +1,63 @@
+@import "../common.scss";
+
+$cardSize: 343px;
+$contentHeight: 262px;
+$footerHeight: 80px;
+
+.root {
+    width: $cardSize;
```
shlomitc: size should be 100%. aspect ratio will be defined from the outside (a relevant component is being created here https://github.com/wix/wix-style-react/pull/2565)

src/CardGalleryItem/CardGalleryItem.scss
```diff
@@ -0,0 +1,63 @@
...
+    cursor: pointer;
+}
+
+.content, .hoveredContent {
+    height: $contentHeight;
```
shlomitc: should expand height as well

src/CardGalleryItem/CardGalleryItem.scss
```diff
@@ -0,0 +1,63 @@
...
+    left: 0;
+    right: 0;
+    display: flex;
+    justify-content: center;
+    padding-top: 18px;
+}
+
+.footer {
+    height: $footerHeight;
+    border-top: 1px solid #e2e9f5;
```
shlomitc: colors should be taken from palette

src/CardGalleryItem/CardGalleryItem.js
```diff
@@ -0,0 +1,99 @@
...
+          >
+            <div
+              className={styles.content}
+              style={{
+                backgroundImage: `url(${backgroundImageUrl})`,
+              }}
+            />
+
+            <div className={styles.footer}>
+              <div className={styles.title}>{title}</div>
```
shlomitc: use Text component instead

stories/CardGalleryItem/index.js
```diff
@@ -0,0 +1,20 @@
...
+import { storiesOf } from '@storybook/react';
+import Markdown from 'wix-storybook-utils/Markdown';
+import CodeExample from 'wix-storybook-utils/CodeExample';
+
+import CardReadme from '../../src/CardGalleryItem/README.md';
+
+import ExampleBasic from './ExampleBasic';
+import ExampleBasicRaw from '!raw-loader!./ExampleBasic';
+
+storiesOf('2. Layout', module).add('CardGalleryItem', () => (
```
shlomitc: use the `.story` file structure and place it under the `Components` category

src/CardGalleryItem/CardGalleryItem.js
argshook: can this accept a component and not just object? now it says `primaryActionProps` but in reality it's only `{ label, onClick }`. what will happen when i need a tooltip? or an icon? or popover menu or smthng

src/CardGalleryItem/CardGalleryItem.js
```diff
@@ -0,0 +1,99 @@
...
+import TextLink from '../TextLink';
+
+import styles from './CardGalleryItem.scss';
+import animationStyles from './CardGalleryItemAnimation.scss';
+
+class CardGalleryItem extends React.Component {
+  static displayName = 'CardGalleryItem';
+  static propTypes = {
+    /** Card title */
+    title: PropTypes.string,
```
argshook: `PropTypes.node` please, as node is also a string but don't restrict from using HOCs for translation or experiment or whatever

src/CardGalleryItem/CardGalleryItem.js
```diff
@@ -0,0 +1,99 @@
...
+            </TextLink>
+          </div>
+        </div>
+      </div>
+    );
+
+    return (
+      <div
+        className={styles.root}
+        onClick={primaryActionProps.onClick}
```
argshook: i would define this as dedicated  `onClick` prop, not something from `primaryAction`

src/CardGalleryItem/Hover/Hover.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { CSSTransition } from 'react-transition-group';
```
argshook: please take a look at https://github.com/wix/wix-animations which is already in use in  Page component (https://github.com/wix/wix-style-react/blob/master/src/PageHeader/PageHeader.js#L13-L23) or DataTable (https://github.com/wix/wix-style-react/blob/master/src/DataTable/DataTable.js#L8). you might not need this `Hover` component at all

src/CardGalleryItem/CardGalleryItem.js
```diff
@@ -0,0 +1,99 @@
...
+      subtitle,
+      backgroundImageUrl,
+      primaryActionProps,
+      secondaryActionProps,
+      dataHook,
+    } = this.props;
+
+    const hoveredContent = (
+      <div className={styles.hoveredContent}>
+        <div className={styles.primaryAction}>
```
argshook: is `primaryAction` div really needed? it seems like all it does is `position: relative;` which can be part of `styles.hoveredContent`

src/CardGalleryItem/CardGalleryItem.js
shlomitc: The props are propagated to the primary action button, so it shouldn't be replaced by another component.

src/CardGalleryItem/CardGalleryItem.js
shlomitc: specs of the component described this as the same action, as the entire card should be clickable.

it's similar to what happens in the table, the main action on the row is the same as the primary action cell click.

src/CardGalleryItem/CardGalleryItem.js
ihork: @argshook this container is also responsible for positioning of secondaryAction. The main reason - on mockups primaryAction button is positioned in the center of hovered part. So I used flex to position primaryAction. SecondaryAction has position absolute - to be near firstAction but doesn't take a part in flex calculations

src/CardGalleryItem/Hover/Hover.js
ihork: @argshook thanks for showing me this component. I didn't know it exists. I've made workable POC with Wix-animations, so we can use it here. @shlomitc wdyt? Was there a reason why you suggested to implement our own Hover component?

ihork:
@shlomitc @argshook thanks a lot for detailed review. I've covered all comments you left here. So could you please check them. And let's decide according Wix-animations.
FYI, I'm going to start with test coverage in this PR

src/CardGalleryItem/CardGalleryItem.scss
```diff
@@ -0,0 +1,60 @@
...
+.footer {
+  height: $footerHeight;
+  border-top: $footerHeightBorderHeight solid $D60;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  padding: 0 30px;
+
+  .title {
+    font-size: 18px;
```
shlomitc: see comment in the component

src/CardGalleryItem/CardGalleryItem.scss
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  .subtitle {
```
shlomitc: see comment in the component

src/CardGalleryItem/CardGalleryItem.js
```diff
@@ -0,0 +1,135 @@
...
+              <Text
+                className={styles.title}
+                size="medium"
+                weight="normal"
+                data-hook="title"
+              >
+                {title}
+              </Text>
+              <Text
+                className={styles.subtitle}
```
shlomitc: Use the `Text` component internal styling. in this case, should be:
`<Text size="small" weight="thin" secondary/>`

src/CardGalleryItem/CardGalleryItem.scss
```diff
@@ -0,0 +1,60 @@
...
+  position: absolute;
+  top: 100%;
+  display: flex;
+  justify-content: center;
+  padding-top: 18px;
+}
+
+.footer {
+  height: $footerHeight;
+  border-top: $footerHeightBorderHeight solid $D60;
```
shlomitc: maybe can use `<Card.Divider/>` instead?

src/CardGalleryItem/CardGalleryItem.driver.js
```diff
@@ -0,0 +1,36 @@
...
+    getTitle: () => getTitle(element).textContent,
+    getSubtitle: () =>
+      element.querySelector('[data-hook="subtitle"]').textContent,
+    getBackgroundImageUrl: () =>
+      element
+        .querySelector('[data-hook="background-image"]')
+        .style['background-image'].match(/url\((.*)\)/)[1],
+    isHoveredContentPresent: () =>
+      Boolean(element.querySelector('[data-hook="hovered-content"]')),
+    hover: () =>
```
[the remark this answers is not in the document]
ihork: @shlomitc are you sure we don't want to expose `hover`? How user of this driver will be able to click on secondary action, for example? Do you want `clickOnSecondaryAction` to include hover functionality inside?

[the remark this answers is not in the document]
shlomitc: exactly

src/CardGalleryItem/CardGalleryItem.js
```diff
@@ -0,0 +1,135 @@
...
+      <div className={styles.hoveredContent} data-hook="hovered-content">
+        <div className={styles.primaryAction}>
+          <Button dataHook="primary-action">{primaryActionProps.label}</Button>
+
+          <div className={styles.secondaryAction}>
+            <TextButton
+              skin="light"
+              onClick={event => {
+                secondaryActionProps.onClick(event);
+                event.stopPropagation();
```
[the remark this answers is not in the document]
ihork: Please notice, that `primaryActionProps.onClick` is attached to the whole `CardGalleryItem` and not directly to the `primaryActionButton`. That's why without `stopPropagation` it will actually call two handlers.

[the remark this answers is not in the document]
ihork: I don't attach primaryAction to the button, since click on button will be handled the same way as click on the whole card.
I have tests on each click. Do you suggest to extend that tests and check that only relevant handler called only once?

src/CardGalleryItem/CardGalleryItem.scss
```diff
@@ -0,0 +1,60 @@
...
+
+$footerHeight: 80px;
+$footerHeightBorderHeight: 1px;
+
+.root {
+  width: 100%;
+  height: 100%;
+  cursor: pointer;
+
+  & > * {
```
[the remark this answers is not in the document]
ihork: You're talking about `className` or `style` prop? Please notice that each of them requires changes in `Card` component

src/CardGalleryItem/CardGalleryItem.scss
```diff
@@ -0,0 +1,60 @@
...
+}
+
+.primaryAction {
+  position: relative;
+  width: 100%;
+  display: flex;
+  justify-content: center;
+}
+
+.secondaryAction {
```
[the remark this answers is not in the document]
ihork: As I described here https://github.com/wix/wix-style-react/pull/2566#discussion_r241003520 I've chosen this solution to position primary button in the center. And secondary button is connected to the primary. This solution covers cases with different height correctly and display secondary button near the primary. Do you see the other approach here?

src/CardGalleryItem/CardGalleryItem.scss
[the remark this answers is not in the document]
shlomitc: I think it should be a `Card` feature, by passing a prop called `allowOverflow` or something similar

src/CardGalleryItem/CardGalleryItem.scss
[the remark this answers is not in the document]
shlomitc: just fixed px with some margin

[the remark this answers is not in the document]
ihork: It can't be a fixed px because height is set from the outer component and may be different. But I may suggest you another approach - primary and secondary actions are in the same container which is positioned in the center of hoveredContent. And, since we know height of both actions, I can move that container with `margin-top` down, which will position primaryAction in the center of hoveredContent. Visually it will look the same as now, but it will be created without position: absolute.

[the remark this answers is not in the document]
shlomitc: so just to make sure I understand.
1. position the actions container in the center absolutely
2. make the margin between primary and secondary actions fixed size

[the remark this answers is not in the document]
ihork: 1. Position actions container in the center with flex
2. Add fixed margin-top to move actions container down, which will position button in center of the hovered content
screen shot 2018-12-19 at 12 41 12 pm

ihork:
@shlomitc the latest portion of fixes is here. Could you please take a look again?
According merge conflicts - after your approve of this PR I'll squash all in one commit, rebase it and push with force here. It will be easier to fix conflicts in that case

src/CardGalleryItem/CardGalleryItem.scss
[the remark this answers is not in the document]
shlomitc: Ahhh, now I get it. Good illustration.
I thought everything should be centered (which is the easy solution).

so now the margin-top will be hacky, since it depends on the button size.
given that, I think it's better to leave the first solution. which one do you prefer?

src/CardGalleryItem/CardGalleryItem.protractor.driver.js
```diff
@@ -0,0 +1,39 @@
+import { baseUniDriverFactory } from 'wix-ui-test-utils/base-driver';
```
shlomitc: can be deleted

stories/CardGalleryItem/ExampleBasic.scss
```diff
@@ -0,0 +1,8 @@
+.wrapper {
+  background: #F0F4F7;
+  padding: 30px;
+}
+
+.container {
```
shlomitc: not needed

stories/CardGalleryItem/CardGalleryItem.story.js
```diff
@@ -0,0 +1,100 @@
...
+  height: '343px',
+  width: '343px',
+  backgroundImageUrl: backgroundImageUrl,
+  primaryActionProps: getPrimaryActionProps('Button'),
+  secondaryActionProps: getSecondaryActionProps('Text Link'),
+};
+const exampleProps = {
+  title: [
+    { label: 'Card Title', value: 'Card Title' },
+    { label: 'Long Card Content Title', value: 'Long Card Content Title' },
```
shlomitc: maybe should be longer in order to see the line breaking

stories/CardGalleryItem/CardGalleryItem.story.js
```diff
@@ -0,0 +1,100 @@
...
+    { label: 'Button', value: getPrimaryActionProps('Button') },
+    {
+      label: 'Long label for Button',
+      value: getPrimaryActionProps('Long label for Button'),
+    },
+  ],
+  secondaryActionProps: [
+    { label: 'Text Link', value: getSecondaryActionProps('Text Link') },
+    {
+      label: 'Long label for Text Link',
```
shlomitc: not sure this is necessary

stories/CardGalleryItem/CardGalleryItem.story.js
```diff
@@ -0,0 +1,100 @@
...
+      label: 'Long card content subtitle',
+      value: 'Long card content subtitle',
+    },
+  ],
+  height: ['200px', '343px', '400px'],
+  width: ['200px', '343px', '400px'],
+  primaryActionProps: [
+    { label: 'Button', value: getPrimaryActionProps('Button') },
+    {
+      label: 'Long label for Button',
```
shlomitc: not sure this is necessary

src/CardGalleryItem/CardGalleryItem.protractor.driver.js
```diff
@@ -0,0 +1,37 @@
+import { mouseEnter } from 'wix-ui-test-utils/protractor';
```
shlomitc: remove this file

testkit/protractor.js
```diff
@@ -386,6 +386,12 @@ export const avatarTestkitFactory = protractorUniTestkitFactoryCreator(
   avatarDriverFactory,
 );
 
+import cardGalleryItemDriverFactory from '../src/CardGalleryItem/CardGalleryItem.protractor.driver';
+
+export const cardGalleryItemTestkitFactory = protractorUniTestkitFactoryCreator(
```
shlomitc: include the uni-driver and not the protractor one

src/CardGalleryItem/CardGalleryItem.driver.private.js
```diff
@@ -0,0 +1,14 @@
+import cardGalleryItemDriverFactory from './CardGalleryItem.driver';
+import ReactTestUtils from 'react-dom/test-utils';
+
+const cardGalleryItemPrivateDriverFactory = base => ({
+  ...cardGalleryItemDriverFactory(base),
+  hover: async () =>
+    ReactTestUtils.SimulateNative.mouseOver(
```
shlomitc: please move the hover functionality to the spec, just like you did in the `e2e.js`

src/CardGalleryItem/CardGalleryItem.scss
```diff
@@ -0,0 +1,51 @@
...
+.content {
+  background-size: cover;
+  background-position: center;
+}
+
+.hoveredContent {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  padding: 10px;
```
shlomitc: what is this padding for?

src/CardGalleryItem/Hover/Hover.js
shlomitc: I didn't investigate so I'm not sure about its capabilities. anyway let's keep you solution and maybe refactor later

shlomitc:
@milkyfruit / @LittleTal - can you please approve?

dor-itzhaki:
When is this expected to be released ?

shlomitc:
@dor-itzhaki I merged now, so will prepare the release for tomorrow morning

dor-itzhaki:
Awesome, thanks!

PR #2567, opened 2018-12-11
Add husky `precommit` with lint-staged

hepiyellow:
## Motivation

To avoid annoying lint errors in CI which makes the PR cycle longer.

## Solution
For those of you who say that linting every commit is annoying, 
then :
- It lints only staged files so its fast

the alternative is to lint at `prepush`, but then I think we need to lint ALL the files, and that take a lot of time.


package.json
```diff
@@ -25,6 +25,7 @@
   },
   "scripts": {
     "start": "./scripts/npm-start.sh",
+    "precommit": "lint-staged",
```
argshook: prepush please, precommit is annoying more than failed build

package.json
```diff
@@ -87,8 +88,10 @@
     "gh-pages-auto-release": "^1.1.2",
     "github-markdown-css": "^2.4.1",
     "highlight.js": "^9.13.1",
+    "husky": "~0.14.0",
     "identity-obj-proxy": "^3.0.0",
```
argshook: This is likely already part of yoshi

package.json
```diff
@@ -151,6 +154,9 @@
     "wix-ui-icons-common": "^2.0.9",
     "wix-ui-test-utils": "^1.0.115"
   },
+  "lint-staged": {
+    "*.{js,scss}": "yoshi lint"
```
yardnsm: Why the `scss` files?

package.json
```diff
@@ -87,8 +88,10 @@
     "gh-pages-auto-release": "^1.1.2",
     "github-markdown-css": "^2.4.1",
     "highlight.js": "^9.13.1",
+    "husky": "~0.14.0",
     "identity-obj-proxy": "^3.0.0",
     "import-path": "^1.0.0",
+    "lint-staged": "^7.3.0",
```
argshook: No need, if we use prepush

yardnsm:
I'm totally up for it. Wdyt about also running `--fix`? I think `create-yoshi-app` also does this for generated projects.

shlomitc:
I'd go for pre-push, can't it be done only for the committed files?

package.json
yardnsm: So we'll lint all the files? Maybe we can compare the current branch to master/upstream to get all the changed files since last push?

argshook: Yeah, why not. How often do you push that you cant wait a minute? :)

package.json
hepiyellow: I took it from yoshi generated client project.
I guess it lints scss? didn't verify it.

package.json
hepiyellow: but `prepush` means we need to run on ALL files in project. that takes about a minute (or more).

package.json
hepiyellow: I took it from a yoshi generated client project.

argshook:
i vote no.

already been there with precommit and husky, both are annoying (husky adds hooks everywhere, up to home folder, precommit fucksup my tooling)

it's not a big deal for me to disable it locally, so im not gonna fight

hepiyellow:
@argshook Can you elaborate about what "fucksup my tooling" means?

hepiyellow:
## Vote count

precommit - erez, yarden, nitay
prepush - shlomi, arijus

mykolas ?

argshook:
surely

when invoking `git commit` a certain output to stdin is expected, but
with hooks which do side effects that output is no longer deterministic.

in particular this fucksup my vimagit (https://github.com/jreybert/vimagit) integration.

but as i said, i whish for no precommit hook, but am ok with disabling
it for myself.

see you want to push this to yoshi too. i think we've already been there and it was reverted at some point.

mykas:
I am ok with precommit. I use prettier so I don't have any issues with lint mostly

PR #2568, opened 2018-12-12
Add a default config to eyesItInstance

yardnsm:
Also added some unit tests :)


package.json
```diff
@@ -27,7 +27,7 @@
     "start": "./scripts/npm-start.sh",
     "pretest": "npm run lint && ./scripts/ensure-dist.sh",
     "test": "npm run test:unit && npm run test:e2e",
-    "test:unit": "yoshi test --jest",
+    "test:unit": "EYES_API_KEY= yoshi test --jest",
```
yardnsm: Since `EYES_API_KEY` is set in the CI, when requiring `eyes.it` it'll try to initialize. As it runs in Jest and not in Protractor, some globals (specifically `browser`) are missing, and an error will be thrown.

This change will unset `EYES_API_KEY` only for that specific script, so `eyes.it` won't initialize itself during unit tests.

yardnsm:
@shlomitc can we merge this?

PR #2579, opened 2018-12-13
Docs/notification revamp

argshook:
### What changed

Notification storybook page

### Why it changed

as part of the ongoing effort to make story config more flexible, while still avoiding wiring detail


stories/Notification/index.story.js
```diff
@@ -0,0 +1,106 @@
...
+          text: readme,
+        }),
+
+        description({
+          title: 'Examples',
+        }),
+
+        liveCode({
+          title: 'Themes',
+          compact: true,
```
yardnsm: Just a side note (not related directly to this PR), I think we can make `compact` to be `true` by default.

PR #2581, opened 2018-12-13
allow clicking the back button from a PageHeader driver

dor-itzhaki:
### What changed
- PageHeader driver exposes "clickBackButton" function
- PageHeader puts a dataHook on the Button component instead of its wrapper
- PageHeader "onBackClick" prop is now tested

### Why it changed
- The "onBackClick" prop was not tested
- Driver was this lacking functionality.

---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


src/PageHeader/PageHeader.js
```diff
@@ -147,9 +147,12 @@ export default class PageHeader extends WixComponent {
                     [s.minimized]: minimized,
                     [s.darkTheme]: isDarkTheme(hasBackgroundImage, minimized),
                   })}
-                  data-hook="page-header-backbutton"
                 >
-                  <Button onClick={onBackClicked} theme="icon-white">
+                  <Button
+                    dataHook="page-header-backbutton"
```
dor-itzhaki: i see that the norm is to set the data-hook on the wrapper component, but couldn't really understand why. in my opinion, putting the data-hook on the wrapper only requires the driver later on to be familiar with the internals of the used component. BUT, if that's an issue, i could move the data-hook back and just look for an inner "button" in the driver.

src/PageHeader/PageHeader.spec.js
```diff
@@ -169,6 +170,21 @@ describe('PageHeader', () => {
     expect(driver.isBackButtonExists()).toBeFalsy();
   });
 
+  it('should execute the given back button callback once the back button is clicked', () => {
+    const backButtonCallback = sinon.spy();
```
hepiyellow: Please use `jest.fn()`

hepiyellow:
```suggestion
    const backButtonCallback = jest.fn();
```

src/PageHeader/PageHeader.spec.js
```diff
@@ -169,6 +170,21 @@ describe('PageHeader', () => {
...
+      <PageHeader
+        title={title}
+        showBackButton
+        onBackClicked={backButtonCallback}
+      />
+    );
+    const driver = createDriver(pageHeader);
+    driver.clickBackButton();
+
+    expect(backButtonCallback.calledOnce).toBeTruthy();
```
hepiyellow:
```suggestion
    expect(backButtonCallback).toBeCalledTimes(1);
```

src/PageHeader/PageHeader.spec.js
```diff
@@ -5,6 +5,7 @@ import { createDriverFactory } from 'wix-ui-test-utils/driver-factory';
 import Button from '../Button';
 import Breadcrumbs from '../Breadcrumbs/Breadcrumbs';
 import PropTypes from 'prop-types';
+import sinon from 'sinon';
```
hepiyellow:
```suggestion
```

PR #2584, opened 2018-12-13
<Popover/> - make autodocs work with the drivers

yardnsm:
This PR changes the driver a little bit, specifically the `import` path of the original core's driver and removing the `as` keyword, in order it to play nicely with AutoDocs.


shlomitc:
@argshook - is this the right way to go? we have the same issue with the new Buttons.

argshook:
@shlomitc im not 100%, cause writting from memory, but i think if `drivers/vanilla` were in a shape of `module.exports = { ComponentName: require('..testkit') }` without spreads, it should eitther work out of the box or need a relatively small  tweak

argshook:
@yardnsm

i pushed a fix which imports testkit directly and does not create proxy
function, as it's useless.

unit testkit documentation works ok now, but not one for protractor.
That one is a different case, not related to this PR (tl;dr `mouseLeave` is wrongly parsed, because it does not define a method but reference it from 3rd party)

please check and merge, if ok.

yardnsm:
Looks good to me, I think we can merge this.

PR #2585, opened 2018-12-13
`<Tooltip/>` (Driver) - Add error if driver method while `element===null`

hepiyellow:
## Motivation

give consumers a better error message so they understand that the Tooltip component was not found by the `dataHook`.


src/Tooltip/Tooltip.driver.js
```diff
@@ -111,4 +111,24 @@ const tooltipDriverFactory = ({ element, wrapper }) => {
   };
 };
 
-export default tooltipDriverFactory;
+export default assertExistsWrapper(
+  tooltipDriverFactory,
+  `TooltipDriver: element must be provided in order to use driver methods other than exists(). If you are finding the Tooltip by dataHook,
```
shlomitc: > element must be provided in order to use driver methods

If the error message is for consumers, then `element` doesn't mean anything. They pass `wrapper` and `dataHook` which are eventually translated into `element`.

> If you are finding the Tooltip by dataHook, then it means it was not found

Consumers always use dataHook to initialize the driver

> A common mistake might be that you have passed dataHook-A to the Tooltip and a dataHook-B to the Tooltip's child, and then try to look for the tooltip by dataHook-B.`

This is ok, but might still be confusing.


I would just suggest write:

> The Tooltip driver could not be initialized, please make sure that a `dataHook` is directly applied to the Tooltip component and passed to the driver

src/Tooltip/Tooltip.spec.js
```diff
@@ -730,6 +730,18 @@ describe('Tooltip', () => {
       expect(driver.hasAnimationClass()).toBeFalsy();
     });
   });
+
+  describe('assetExistsWrapper', () => {
```
shlomitc: `assert`

hepiyellow: fixed

src/Tooltip/Tooltip.driver.js
hepiyellow: fixed. thanks !

PR #2590, opened 2018-12-14
`<DatePicker/>` - Fixed Calendar Not Closing When Wrapped In `<label/>` tag.

AtomicAlienZ:
### What changed
`<Calendar/>` now passes an event as an argument to it's `onClose` callback
`<DatePicker/>` `preventDefault`s this event if it's a click.

### Why it changed
When `<DatePicker/>` was wrapped in `<label/>`, selecting a date by clicking on it made it close and open in a quick succession, so that it looked like nothing changed for the user.

*How To Reproduce:*
```JSX
<label>
  <DatePicker onChange={() => {}} shouldCloseOnSelect={true} />
</label>
```
Open the datepicker and click on any selectable date. DatePicker will stay open.
---

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green


hepiyellow:
Can you try and see if this happens when you wrap DatePicker with FormField?

AtomicAlienZ:
`<DatePicker/>` closes OK when accessed through the label, but `<FormField/>` itself is not that good due to not actually wrapping its children in a `<label/>`.

1. It does not make the input icon clickable (which was my initial goal of wrapping picker in label)
2. It requires page-global `id` to work properly

shlomitc:
@AtomicAlienZ 
let me understand - the missing feature is that the input field icons should also be clickable?
if so, then it's not related to the label wrapping at all.

regarding the page-global id, I'm not sure I understand what is the use case

AtomicAlienZ:
The missing feature is that `<DatePicker/>` should behave like the vanilla field when wrapped into `<label>`.

As for the `id`s issue, I need to track ids and keep them unique to avoid collisions. To illustrate:
```html
<div class="SomeComponent">
  <label htmlFor="a">LABEL 1</label>
  <input id="a" />
</div>
<div class="SomeComponent">
  <label htmlFor="a">LABEL 2</label>
  <input id="a" />
</div>
```

If I click on "LABEL 2", first input gets focused.

The only other way I see to achieve same results would be to `Portal` the datepicker's dropdown somewhere else, removing the issue's cause.

hepiyellow:
Codesandbox that reproduces the issue:

https://codesandbox.io/s/9zypvz8p5o

hepiyellow:
## Summary

There are 3 Issues:

### 1. Calendar Icon not clickable
The icon In DatePicker, input prefix - is not opening Calendar on click.

This, is UX agrees, can be resolved within the `<Input/>`.

@AtomicAlienZ Please open an issue, and in it ask UX (Ben Benhorin @wuwa), if he confirms.

@wuwa Do you think that clicking on `<Input/>`'s prefix/suffix elements should delegate click on the `<input/>` (meaning that it would get focus).
(I think it is a good change)

### 2. Calendar Popover not closing when day clicked, given DatePicker is wrapped in a label.

This should be fixed.
I think the correct fix is within **the `<Calendar/>` only**.
Every click in the Calendar that triggers a calendar action, should prevent default. (click day, next_month, prev_month)

@AtomicAlienZ Do you want to try a fix within `<Calendar/> only?

### 3. `<FormField/>` requires an id

It is a legit feature request.

@AtomicAlienZ Please open a "FeatureRequest" Issue.

src/Calendar/Calendar.spec.js
```diff
@@ -115,6 +115,26 @@ describe('Calendar', () => {
     });
   });
 
+  describe('clicking in caption', () => {
+    it('should prevent event default', () => {
+      const eventListenerMock = jest.fn();
+      const driver = createDriver(
+        <div onClick={eventListenerMock}>
+          <Calendar onChange={() => {}} showYearDropdown />
```
hepiyellow: What about `showMonthDropdown`?

src/Calendar/Calendar.spec.js
```diff
@@ -135,6 +155,23 @@ describe('Calendar', () => {
...
+      it('should prevent event default', () => {
+        const date = new Date(2018, 10, 5);
+        const eventListenerMock = jest.fn();
+        const driver = createDriver(
+          <div onClick={eventListenerMock}>
+            <Calendar value={date} onChange={onChange} selectionMode={'day'} />
+          </div>
+        );
+
+        expect(eventListenerMock).toHaveBeenCalledTimes(0);
```
hepiyellow: This is a bit redundant

src/Calendar/Calendar.js
```diff
@@ -95,11 +96,11 @@ export default class Calendar extends WixComponent {
             : { from: value, to: anchor };
 
         onChange(newVal, modifiers);
-        shouldCloseOnSelect && this.props.onClose();
+        shouldCloseOnSelect && this.props.onClose(event);
```
hepiyellow: Can you add a test that `onClose` is called with the event?

hepiyellow:
I pushed some changes.
It is ready for merge. Just waiting for build to be green.

PR #2594, opened 2018-12-16
<Popover/> - improve arrow styles

yardnsm:
This PR improves the `<Popover/>`'s arrow styles to match the `<Tooltip/>` design. It increases its size a bit and adds a shadow to it.

Old design:
screen shot 2018-12-16 at 15 03 09

New design:
screen shot 2018-12-16 at 15 03 20


src/Popover/Popover.st.css
```diff
@@ -13,6 +18,10 @@
   -st-named: shadow30;
 }
 
+:vars {
+  contentArrowSize: 10px;
```
shlomitc: change to 8px (so the cross will be ~12px)

alonbt:
@yardnsm @shlomitc @hepiyellow Can you please resolve this ? I need this fix in order to use the tooltip and deploy a feature...

yardnsm:
@LittleTal The arrow spacing has been fixed.

**Regarding the minimum margin:** this can be easily implemented using a [Popper.js modifier](https://popper.js.org/popper-documentation.html#modifiers..preventOverflow.padding), but I think its better if we won't that to the core's `<Popover/>` directly.

Instead, we can add a `popperModfiers` prop to the core's `<Popover/>`, so we can control the modifiers  inside the WSR `<Popover/>`.

I think we can merge this PR without the minimum margin thing.

@shlomitc wdyt?

shlomitc:
@yardnsm - if we can't achieve this with pure css, then let's merge this one just so it won't block @alonbt. then, let's add the margin capability to the core

PR #2595, opened 2018-12-16
<BadgeSelect/> - migrate to use <Popover/>

yardnsm:
This PR makes the `<BadgeSelect/>` to use the new `<Popover/>` component. It benefits a better positioning mechanism, and removes the css hacks used to manage the `<DropdownLayout/>` positioning.


scripts/npm-start.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
 
 if scripts/ensure-dist.sh; then
-  npm run storybook 
+  npm run storybook & yoshi start
```
yardnsm: It was removed in the `<Carousel/>` PR, so I added it back 😬

shlomitc: It's actually ok, I think it's fine to run the tests separately. it also keeps a clean log when debugging the webpack output

src/BadgeSelect/BadgeSelect.st.css
```diff
@@ -0,0 +1,12 @@
+:import {
+  -st-from: "../Popover/Popover.st.css";
+  -st-named: Popover;
+}
+
+.root {
+  -st-extends: Popover;
+}
+
+.root::popoverContent {
```
shlomitc: Creating a component like this should be possible also to external users.
If they need to use stylable, then it's not that easy.
Might be better if this functionality could live inside Popover instead

yardnsm: Again, we're planning to provide a new `<DropdownPopover/>` component for these kind of cases, and it will apply this style.

IMO, Stylable should be to go-to way to customize the popover styles.

scripts/npm-start.sh
yardnsm: So...

```suggestion
  npm run storybook
```

?

hepiyellow: yes

src/BadgeSelect/BadgeSelect.driver.js
```diff
@@ -1,19 +1,23 @@
...
-  });
-
+const badgeSelectDriverFactory = ({ element }) => {
   const driver = {
     /** Returns 'true' wether the element exists */
     exists: () => !!element,
+
     /** Click on an option */
-    clickAtOption: index => dropdownLayoutDriver.clickAtOption(index),
+    clickAtOption: index => {
```
shlomitc: add click that also opens the popover before calling `clickAtOption`

src/BadgeSelect/BadgeSelect.js
```diff
@@ -129,33 +125,44 @@ export default class BadgeSelect extends React.Component {
...
+
     return (
-      <div className={styles.container} data-hook={dataHook}>
-        <div data-hook="badgeSelect-badge-wrapper" onKeyDown={this.onKeyDown}>
+      <Popover
+        shown={visible}
+        dataHook={dataHook}
+        placement="bottom"
+        onKeyDown={this.onKeyDown}
+        onClickOutside={() => this.hideDropdown()}
```
shlomitc: bind function

src/BadgeSelect/BadgeSelect.js
```diff
@@ -129,33 +125,44 @@ export default class BadgeSelect extends React.Component {
...
           <Badge
             ref={badge => (this.badge = badge)}
-            {...{ type, size, uppercase }}
+            type={type}
+            size={size}
+            uppercase={uppercase}
             suffixIcon={<ChevronDown viewBox="6 6 12 12" />}
-            onClick={() => this.toggleDropdown()}
             skin={this.state.selectedBadge.skin}
+            onClick={() => this.toggleDropdown()}
```
shlomitc: bind function

PR #2597, opened 2018-12-16
Chore/automated testkit exports

argshook:
tl;dr; this PR is part of ongoing effort to automate testkit exports, starting with enzyme first. Also fixes currently broken testkit-smoke.test.js and introduces component-definition

---

ideal component should have only one unidriver testkit in one known
location. currently we have a lot of components that don't conform to
this rule.

we have:
* non unidriver components
* only enzyme or only vanilla testkit
* testkit exported from nondeterministic path
* testkit exported from another lib
* "naked" testkit exported (without `createEnzymeTestkitCreator`) called, for example
* likely more

also, most components have required props. those must be supplied in
order to have automated smoke test.

we want to automate testkit exports but to accomodate so many
differences a config (component definition) is required.

for now this is done in component-definitions.js. ideally they
should live together with component implementation, as that's easier to
manage, but that's for next iteration.

---

i'd like to merge this soon, because it fixes testkit-smoke and is valid step towards total exports automation


testkit/enzyme.js
```diff
@@ -1,535 +1,68 @@
...
+    throw new Error(
+      `ERROR: Unable to require ${testkitPath} Ensure ${name} component has enzyme testkit and it is exported correctly, or set it in component-definitions, ${e}`,
+    );
+  }
+};
 
-export const iconButtonTestkitFactory = enzymeUniTestkitFactoryCreator(
-  iconButtonDriverFactory,
-);
+const isNotIgnored = name =>
```
hepiyellow: I prefer positive variable names. (double negative is confusing)

argshook: it's just extracted function to make `.filter(isNotIgnored)` read
nicely.

where's the double negative?

testkit/component-definitions.js
```diff
@@ -0,0 +1,244 @@
...
+      isOpen: false,
+      contentLabel: 'modal_12345678',
+    },
+  },
+  ContactItemBuilder: { ignore },
+  Carousel: {
+    drivers: ['enzyme'],
+  },
+  Proportion: { ignore, unidriver, drivers: ['enzyme'] },
+  GeneratedTestComponent: { ignore, unidriver, drivers: ['enzyme'] },
```
yardnsm: I would also update the component generator with a codemod that adds the new generated component. Its definitions should be the same as the `<GeneratedTestComponent/>`.
If we are going to have the component definition sitting next to the component soon, we can skip adding the codemod for know (but I still prefer to add it :D).

Also, we need to disable the `testkit-exports` codemod for enzyme:

https://github.com/wix/wix-style-react/blob/144135de05d9a2c8b5738ed750f292aa57ef74a2/scripts/component-generator/src/run-codemods.js#L67-L77

argshook:
@hepiyellow 

> I know it's too late, but I would suggest merging a fix to the smoke-test.
> And then separately do the enzyme automation, so we know that we are covered by the smoke test.

we're not too too late but we're also not too too broken, because most
components have testkit sanity tests within their own specs. Those that don't have them probably didn't have them before smoke test.

i'm on those failing tests, should fix today (fingers crossed)

btw, it's fun working on this because you can see just how many
inconsistencies we have throughout components, and that's just about
testkits. hopefully efforts like this and generator will make some
standards :)

testkit/component-definitions.js
argshook: i vote for skipping codemod for now, cause there's still work to do in component definition, testkit/enzyme.js, testkit-smoke.test.js, fix failing snowflake components etc.

we'll probably wont need any codemod at all.

let's disable enzyme testkit-exports codemod after  this is  merged, so today or tomorrow (i really hope)

argshook:
some FYI so that it's not only in my head :)

here's rough list of goals i'd like to achieve in the area of testkits

1. fix currently failing tests (by some manual configuration, no inventions. It will be more clear what to abstract afterwards)
2. refactor `testkit/enzyme.js`, generalise it all so that it could be reused in `testkit/index.js`, later `testkit/protractor.js`
3. dream about https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/scripts/driver-export.js, maybe this should be the approach here too (create exports file during build). should be easy to do, since testkit-smoke.test.js covers testkit/enzyme.js, meaning that testkit/enzyme.js can be refactored without fear
4. generalise export mechanism, move it to `wix-ui-test-utils`, reuse in other libraries
5. make it so that `import { someTestkit } from 'wix-style-react/testkit'` (or similar) can work (so that we don't need to `import from 'wix-style-react/dist/src/testkit/enzyme`)

number 1. is this PR, it shouldn't be held for long.

other numbers are my eventual goals to move towards, not immediate actions.

pinging you all for additional spam in your inbox :smiling_imp: 
@shlomitc @hepiyellow @yardnsm @mykas

shlomitc:
number 5 is not an issue and should stay as it is right now. Combining all testkits to a single file is problematic.
For example, projects which are not using `puppeteer` will try to import it from the single entry file and their build will fail.
it is not going through tree shaking (because it runs in node), so no static build analysis can remove the unused imports

shlomitc:
however, you can try to create a more good looking imports like `wix-style-react/testkit/enzyme`

PR #2598, opened 2018-12-16
`<Button/>`,`<MessageBoxMarketerialLayout/>` - Fix  deprecationLog

hepiyellow:
## Prop deprecation logs should stop once the suggested alternative is used
- `<Button/>` - avoid logging a deprecation log  when using `upgrade` prop
- `<MessageBoxMarketerialLayout/>` - avoid logging a deprecation log when `fixImagePosition` prop is provided.

## Also in this PR (infrastructure)

- SpyOnHelper - does setup of beforeAndAfterEach
- utils/propTypes `validatorWithSideEffect` - helps adding conditional deprecationLogs on propTypes validation ( instead of at top of file)
  - For testing that the original propTypes validation still works, I had to add a `consoleErrors` API in `wix-ui-test-utils/jest-setup`. See [this PR](https://github.com/wix/wix-ui/pull/1034).


test/utils/deprecationLogTestHelper.js
```diff
@@ -0,0 +1,41 @@
...
+  mockRestore() {
+    this.spy.mockRestore();
+  }
+
+  mock() {
+    this.spy = jest.spyOn(global.console, 'warn');
+  }
+
+  expectLoggedOnceWithMsg(msg) {
+    expect(this.spy).toHaveBeenCalledTimes(1);
```
yardnsm: What would happen if `console.warn` will also be called outside of the deprecation log? I would suggest mocking the `deprecationLog` and not `console.warn`.

src/MessageBox/MessageBoxMarketerialLayout.spec.js
```diff
@@ -195,6 +201,40 @@ describe('MessageBoxMarketerialLayout', () => {
...
+    });
+
+    it('should show deprecationLog and warn of imageUrl not a string', () => {
+      createDriver(
+        <MessageBoxMarketerialLayout {...requiredProps} imageUrl={4} />,
+      );
+      expect(consoleErrors.get()).toHaveLength(1);
+      expect(consoleErrors.get()[0].includes('Invalid')).toBeTruthy();
+
+      consoleErrors.reset(); // prevent test from failing
```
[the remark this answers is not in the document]
hepiyellow: its only for tests that check prop-types errors.
I can put it in a separate describe block.

We don't usually test prop-types errors, but I felt I had to in this case.

src/utils/deprecationLog.spec.js
```diff
@@ -26,4 +27,34 @@ describe('deprecationLog', () => {
...
+      deprecationLog('msg 1');
+      deprecationLog('msg 2');
+
+      let errorThrown = false;
+      try {
+        helper.expectLoggedOnceWithMsg('msg 2');
+      } catch (e) {
+        errorThrown = true;
+      }
+      expect(errorThrown).toBeTruthy();
```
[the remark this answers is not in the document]
hepiyellow: done

test/utils/deprecationLogTestHelper.js
hepiyellow: Removed `DeprecationLogTestHelper`, and using now a `jest.spyOn(depLogger, 'log')`.

hepiyellow:
@yardnsm re-review?

test/utils/unit/jest-helpers.js
```diff
@@ -0,0 +1,33 @@
+export class SpyOnHelper {
```
yardnsm: I don't see the reason for adding more abstractions above jest's mock feature. You can add the `beforeEach` and `afterEach` hooks manually for every test file, this is still fine and may be even better, since maintainers don't need to dig around and figure out what `SpyOnHelper#beforeAndAfterEach` exactly does.

src/utils/deprecationLog.js
```diff
@@ -30,9 +41,8 @@ if (process.env.NODE_ENV !== 'production') {
...
   }
 
-  const logger = new DeprecationLogger();
-
-  deprecationLog = msg => logger.log(msg);
+  depLogger = new DeprecationLogger();
 }
 
-export default deprecationLog;
+export { depLogger };
```
yardnsm: Are you exporting the `depLogger` for the `spy` feature? If so, you can try to implement a [Manual Mock](https://jestjs.io/docs/en/manual-mocks) and spare this additional step.

This way, you can get rid of the `SpyOnHelper` completely.

PR #2606, opened 2018-12-17
`<Search/>` - Allow regexp special chars as search term

ChenRoth:
### What changed

added regex escaping to search term in `<Search/>`

### Why it changed

component would crash when using unescaped regex pattern

---

- [x] Change is tested
- [ ] Change is documented
- [ ] Build is green


shlomitc:
Hey. Thanks.
What exactly is this Regex matching? Need to add test as well.

hepiyellow:
So you are trying to filter the options according to a RegExp which was typed by the user?

This sounds more like a new feature.

ChenRoth:
@shlomitc @hepiyellow using a package and added a test, take a look again

hepiyellow:
@ChenRoth 
Actually , this regex escape package is munch more popular, and seems more simple.

https://www.npmjs.com/package/escape-string-regexp


Usually , at least arijus, prefers that for such small utils, we would better copy the code than add a dependency.
@shlomitc @argshook What do you think?

This is the code of `escape-stirng-regexp`:
```js
'use strict';

var matchOperatorsRe = /[|\\{}()[\]^$+*?.]/g;

module.exports = function (str) {
	if (typeof str !== 'string') {
		throw new TypeError('Expected a string');
	}

	return str.replace(matchOperatorsRe, '\\$&');
};
```

src/Search/Search.js
```diff
@@ -48,7 +49,7 @@ export default class Search extends WixComponent {
 
     return options.filter(option =>
       searchText && searchText.length
-        ? new RegExp(searchText.trim(), 'i').test(option.value)
+        ? new RegExp(escapeRegex(searchText.trim()), 'i').test(option.value)
```
hepiyellow: Why not simply use toLowerCase on both strings?
```js
/**
* a includes b (with case insensitive matching)
*/
function includesCaseInsensitive(a, b) {
 return a && (!b || (a.toLowerCase().indexOf(b.toLowerCase() != -1));
}

return options.filter(option =>
      searchText && searchText.length
        ? includesCaseInsensitive(option.value, searchText.trim())
        : true,
    );
```

We can put the `includesCaseInsensitive` in `src/utils/string.js`

argshook: this should cover 90% of cases:

```js
const isString = a => typeof a === 'string';

const matches = (a = '', b = '') =>
  [a, b].every(isString) && a.toLowerCase().includes(b.toLowerCase());
```

for advanced uses we can expose `matcher` (or similar) prop for custom function.

@hepiyellow you're totally right i'd like to avoid new dependencies :)

ChenRoth: then go ahead and make this change

hepiyellow: @argshook You or me?

PR #2629, opened 2018-12-19
DO NOT MERGE - Eyes Visual Grid Experiment

hepiyellow:
### Multiple checkWindow

`checkWindow` is not async.
So for every checkWindow, we need to do: `eyes.Open`, `checkWindow`, `close`.?
(Verify with Applitools, or test)

### batchDate (batchStartAt)

In eyes.it we do :
`setBatch(name, id, batchStartAt)`

while batchStartAt is either `undefined` or `0`. According to Eyes sdk [documentation](https://applitools.com/docs/api/eyes-sdk/classes-gen/class_eyes/method-eyes-setbatch-selenium-javascript.html) `eyes.setBatch`  check if the `bachDate`( `batchStartAt`) is falsy, it uses the current date.

So it is redundant. And we don't need to do anything with it in the new VisualGrid client.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,8 +68,8 @@
   },
   "homepage": "https://github.com/wix/wix-style-react#readme",
   "devDependencies": {
-    "@applitools/dom-capture": "^3.3.3",
-    "@applitools/visual-grid-client": "^9.0.4",
+    "@applitools/dom-capture": "^6.0.1",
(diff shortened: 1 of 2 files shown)


hepiyellow:
Closing for the weekend
(To save redundant builds)

PR #2632, opened 2018-12-19
call super.componentDidUpdate

### LINKED ISSUE #2538: Unexpected datahook behaviour

yakirn:
### What changed
Components that extends WixComponent and define `componentDidUpdate` will call `super.componentDidUpdate()`
...

### Why it changed

Along with [this PR](https://github.com/wix/wix-ui/pull/1038) it should resolve #2538 .

---

- [X] Change is tested
- [ ] Change is documented
- [ ] Build is green (Once both PRs are merged the code in `testkit-smoke.test.js` can be "unskipped".


shlomitc:
Thanks!

PR #2645, opened 2018-12-21
Unidriver supported getStylableState() method 

mykas:
Created getStylableState() handler that is unidriver friendly. The one from `stylable/dom-test-kit` is only meant for ReactDOMTestUtils.


mykas:
@hepiyellow @shlomitc useful method if we will have stylable states in one of our future components. This method is unidriver friendly and can be used to expose different kind of methods for testkit users.

test/utils/stylable-uni-testkit.js
```diff
@@ -0,0 +1,4 @@
+export const getStylableState = async (base, style, state) => {
+  const stateKey = Object.keys(style.$cssStates({ [state]: true }))[0];
```
hepiyellow: Nice trick.
Difficult to understand though. Could you a little comment.
Or maybe in the StyleMap have this self explaining naming: `{[state]: ARBITRARY_TRUTHY_VALUE}`.

src/Button/Button.driver.js
```diff
@@ -1,14 +1,12 @@
-import { StylableDOMUtil } from '@stylable/dom-test-kit';
 import { baseUniDriverFactory } from 'wix-ui-test-utils/base-driver';
 import button from 'wix-ui-core/dist/src/components/button-next/button-next.st.css';
```
hepiyellow: Could rename to `styles` or `buttonStyles`

PR #2647, opened 2018-12-22
`<DropdownLayout/>` & `<InputWithOptions/>` - add a11y roles

hepiyellow:
Just for fun. (And to learn a little)

Was using this page mainly for reference:
https://www.w3.org/TR/2010/WD-wai-aria-20100916/roles

### DIFF
--- a/src/DropdownLayout/DropdownLayout.js
+++ b/src/DropdownLayout/DropdownLayout.js
@@ -238,6 +238,7 @@ class DropdownLayout extends WixComponent {
       onMouseEnter,
       onMouseLeave,
       inContainer,
+      role,
     } = this.props;
     const contentContainerClassName = classNames({
       [styles.contentContainer]: true,
@@ -258,6 +259,7 @@ class DropdownLayout extends WixComponent {
         onKeyDown={this._onKeyDown}
         onMouseEnter={onMouseEnter}
         onMouseLeave={onMouseLeave}
+        role={role}
       >
         <div
           className={contentContainerClassName}
@@ -451,6 +453,7 @@ DropdownLayout.propTypes = {
   itemHeight: PropTypes.oneOf(['small', 'big']),
   selectedHighlight: PropTypes.bool,
   inContainer: PropTypes.bool,
+  role: PropTypes.string,
 };
 
 DropdownLayout.defaultProps = {
@@ -461,6 +464,7 @@ DropdownLayout.defaultProps = {
   itemHeight: 'small',
   selectedHighlight: true,
   inContainer: false,
+  role: 'list',
 };
 
 DropdownLayout.NONE_SELECTED_ID = NOT_HOVERED_INDEX;
--- a/src/DropdownLayout/DropdownLayout.js
+++ b/src/DropdownLayout/DropdownLayout.js
@@ -318,7 +318,14 @@ class DropdownLayout extends WixComponent {
   }
 
   _renderDivider(idx, dataHook) {
-    return <div key={idx} className={styles.divider} data-hook={dataHook} />;
+    return (
+      <div
+        key={idx}
+        className={styles.divider}
+        data-hook={dataHook}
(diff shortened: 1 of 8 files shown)


src/DropdownLayout/DropdownLayout.js
```diff
@@ -13,6 +13,11 @@ const modulu = (n, m) => {
 const NOT_HOVERED_INDEX = -1;
 export const DIVIDER_OPTION_VALUE = '-';
 
+const itemRoleByContainerRole = {
```
yardnsm: To make the intention more clear, I would use this as a function. For example:

```js
const mapItemRoleByContainerRole = role => {
	const rolesMap = {
		list: 'listitem',
		listbox: 'option',
	};

	return rolesMap[role];
};
```

src/DropdownLayout/DropdownLayout.spec.js
```diff
@@ -609,6 +610,44 @@ describe('DropdownLayout', () => {
...
+      const driver = createDriver(
+        <DropdownLayout options={options} role="listbox" />,
+      );
+      expect(driver.element.getAttribute('role')).toBe('listbox');
+    });
+
+    it(`should have role 'listitem' for all non-divider options`, () => {
+      const driver = createDriver(<DropdownLayout visible options={options} />);
+      options.forEach((option, index) => {
+        if (option.value === DIVIDER_OPTION_VALUE) {
```
yardnsm: You can use the `isOptionADivider` method of the driver instead

src/DropdownLayout/DropdownLayout.spec.js
```diff
@@ -609,6 +610,44 @@ describe('DropdownLayout', () => {
...
+          expect(driver.optionAt(index).getAttribute('role')).toBe('separator');
+        } else {
+          expect(driver.optionAt(index).getAttribute('role')).toBe('listitem');
+        }
+      });
+    });
+
+    it(`should have role 'option' for all options and 'separator' for dividers`, () => {
```
yardnsm: Maybe:

```suggestion
     it(`should have role 'option' for all options and 'separator' for dividers when role="listbox"`, () => {
```

src/InputWithOptions/InputWithOptions.private.driver.js
```diff
@@ -0,0 +1,14 @@
+import inputWithOptionsDriverFactory from './InputWithOptions.driver';
+
+const inputWithOptionsPrivateDriverFactory = ({ element, wrapper }) => {
```
yardnsm: Some for the private `<DropdownLayout/>` driver, all these methods are available in the public driver.

I noticed you're accessing `element` directly in order to use `getAttribute('role')`. Maybe you can add a `getRole()` method?

And as for the `dropdownLayoutElement`, you can access to `<DropdownLayout/>` driver in the `<InputWithOptions/>` driver:

https://github.com/wix/wix-style-react/blob/4573d14b88/src/InputWithOptions/InputWithOptions.driver.js#L75

src/DropdownLayout/DropdownLayout.spec.js
hepiyellow: will do

src/DropdownLayout/DropdownLayout.spec.js
hepiyellow: ok

src/DropdownLayout/DropdownLayout.js
yardnsm: You can still leave it as an object, it's just my preference. And yeah, call it what you want :)

src/InputWithOptions/InputWithOptions.private.driver.js
hepiyellow: done

hepiyellow:
Closing because of nested private driver hell !
Will resume later.

PR #2668, opened 2018-12-25
Fix/automating popover menu docs

nitayneeman:
Resolves wix-private/wsr-issues#302


src/PopoverMenu/PopoverMenu.js
```diff
@@ -25,19 +46,11 @@ class PopoverMenu extends WixComponent {
     buttonTheme: 'icon-greybackground',
     buttonHeight: 'medium',
     maxWidth: '378px',
+    appendToParent: false,
+    zIndex: 0,
```
hepiyellow: @nitayneeman Why did you add default `zIndex: 0`?
This `zIndex` goes to `<Tooltip/>`, which puts it on the `<TooltipContent/>`.
The tooltip has a default zIndex which should be the highest of all components (currently 2000).

Setting it to 0, means it could be (will be) under most other part/components of the page.

nitayneeman: I agree with your point and I guess that part of the code shouldn't be merged (probably was just for checking a specific case momently). Moreover, the whole PR focused on improving the relevant docs, without changing the component implementation. Anyway, this code is redundant.

PR #2563, opened 2018-12-11
Added multi month support to Calendar

malsomnus:
Calendar now has the `numOfMonths` prop. Currently supporting only 1 or 2.

PR #2669, opened 2018-12-26
`<Calendar/>` - Add 2 months view 

hepiyellow:
Replaces this PR #2563 

Continuing from Tal's last commit,
I only wanted to fix a bug that if you click on 2nd month, it should not navigate to that month.
- I added 2 tests for it (day and range cases)
- I also refactored and extracted a CalendarView utils.


src/Calendar/utils.js
```diff
@@ -0,0 +1,48 @@
...
+   * @memberof CalendarView
+   */
+  isContained(date) {
+    return isWithinRange(date, this.startDate, this.endDate);
+  }
+
+  /**
+   * Is the given date after the view's end date
+   */
+  isAfter(date) {
```
dorduch: the function's name is a bit confusing. when you call `view.isAfter(date)` it may seem that the function checks if the entire view is after the given date. 
(the same goes for `isBefore(date)`

hepiyellow: Renamed to `isAfterView` and `isBeforeView`

PR #2673, opened 2018-12-26
<InputArea/> - added "disabled" prop

nitayneeman:
Resolves https://github.com/wix-private/wsr-issues/issues/306


nitayneeman:
Hi @wuwa,
Could you perform a UX review? :)

wuwa:
couple things - 
1. if disabled should not be resizable.
2. why is it transparent (different issue?)

nitayneeman:
> couple things -
> 
> 1. if disabled should not be resizable.
> 2. why is it transparent (different issue?)

Regarding 1 - we can make it be non-resizable in case of disabled mode, but the resize handle will disappear (just in that case), because the current implementation uses the native browser feature. Are we OK with that?

Regarding 2 - do you mean why the background color of the textarea element is transparent? If so, I guess the reason for that was to support different themes, so in that way, its container defines the background color (which is changed by the different themes). Is it correct @shlomitc?

shlomitc:
Regarding 2 - it's an old component with some deprecated themes. we can decide to remove them and make the component cleaner, but I'd say this is a different task.

wuwa:
yes to 1 (hide the resize)

PR #2686, opened 2018-12-28
Fixed the RTL bugs because of #2577

rghorbani:
### What changed

Fixed the bugs happened in RTL because of #2577 
Seems these bugs happened because their was no e2e tests to cover them.

### Why it changed

...

---

- [x] Change is tested
- [x] Change is documented
- [x] Build is green


src/Carousel/Pagination/Pagination.scss
```diff
@@ -1,4 +1,4 @@
-@import '../../../colors.scss';
```
rghorbani: @hepiyellow It happened again!

hepiyellow:
Maybe we can copy the scss files to root only at packaging time...?

src/Carousel/Pagination/Pagination.scss
hepiyellow: Thanks.
I moved the `copy-public-styles` to run at `postbuild`, so it won't happen again (the build will fail)
See this
https://github.com/wix/wix-style-react/pull/2690

src/TableToolbar/Toolbar.scss
```diff
@@ -19,17 +19,16 @@
     margin-right: $margin;
   }
 
-  @include withRTL() {
-    & {
-      &:not(:first-child) {
-        margin-left: 0;
-        margin-right: $margin;
-      }
+  :global(.rtl) &,
```
hepiyellow: Why was this `@include withRTL()` not working?

hepiyellow:
@rghorbani 
I added some e2e tests that should cover the changes you made.
From now on, I would not be able to add these tests, since it's not in our priority.
Please try to add them yourself.
Thanks.

src/TableToolbar/Toolbar.scss
rghorbani: I think it's just limitations of Sass, I could not get it to work so I went with the old way.

src/InputArea/InputArea.scss
```diff
@@ -173,6 +166,11 @@ $gap: 6px;
 
 // RTL layout based on global style set on document body or parent element.
 @include withRTL() {
+  .exclamation {
```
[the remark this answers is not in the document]
rghorbani: I just refactored the code. Did not check the usage.

rghorbani:
@hepiyellow Would it be possible to merge this request?
I would be happy to complete #2691 in very near future.

hepiyellow:
@rghorbani #2691 merged

rghorbani:
@hepiyellow RTL in Popover Menu is not working because of docking problems. Storybook adds rtl direction to button's parent and since the tooltip itself is appended to body, the RTL has no effect on it. I think because of this some tests are failing.

PR #2687, opened 2018-12-29
Drivers Cleanup -Remove `setProps` from ALL drivers

hepiyellow:
### Motivation
Main goal is to allow converting our drivers to UniDrivers.
The `setProps` can not be in a UniDriver. It is actually a responsibility of the consumer's test, not the driver.
It will also remove the need for the `wrapper` and `component` arguments for creating driver,
and so allow us to convert drivers to unidrivers.

### Breaking change?

Searching in wix-private I could not find any specific use of WSR driver's setProps.
It is a bit difficult to search so it's no 100%.
But in this case, we don't care to break tests.
This needs to go!

**And talking with Shlomi, we don't consider this as a breaking change that requires a Major version bump.**


src/ImageViewer/ImageViewer.driver.js
```diff
@@ -1,16 +1,9 @@
...
 import addItemDriverFactory from '../AddItem/AddItem.driver';
 import tooltipDriverFactory from '../Tooltip/Tooltip.driver';
 
-const imageViewerDriverFactory = ({
-  component,
-  wrapper,
-  element,
-  eventTrigger,
-}) => {
+const imageViewerDriverFactory = ({ wrapper, element, eventTrigger }) => {
```
shlomitc: wrapper and eventTrigger are needed here?

hepiyellow: That a different cleanup we need to do.
I don't want to mix them.

Tooltip needs a `wrapper`, but we can remove it (gently) later.

PR #2691, opened 2018-12-30
Add RTL e2e tests to some components

hepiyellow:
Add e2e tests to cover #2686


src/RadioGroup/RadioGroup.e2e.js
```diff
@@ -12,127 +12,151 @@ describe('RadioGroup', () => {
...
 
-  eyes.it('should select the second option in a group', async () => {
-    await waitForVisibilityOf(
-      radioGroupDriver.element(),
-      'Cannot find RadioGroup',
-    );
-    radioGroupDriver.selectByIndex(1).click();
-    expect(radioGroupDriver.isRadioChecked(1)).toBe(true);
-  });
+  describe('LTR', () => {
```
yardnsm: Since LTR is the default behaviour, I think you can remove this wrapping `describe` block. This may also mess Eyes up as it would create new snapshots.

argshook:
@yardnsm we can add it but `autoExampleDriver` is such a hack im not sure we should continue with it. in general doing e2e on storybook was a mistake imo

src/RadioGroup/RadioGroup.e2e.js
hepiyellow: But There is a :
```
beforeAll(() => browser.get(storyUrl));
```
And in thr RTL block, I don't want to get the LTR page.

hepiyellow: And we want to be consistent in having all test suit use a single top level describe.
I think only because it's easy to focus on the test suit (add `fdescribe` at top level)

hepiyellow: I'll remove it anyways. Another single get page is not so bad.

Issue #2559, opened 2018-12-10
Multiselect: paste value


adavidof:
# Relevant Components
#### Multiselect
`import WixMultiselect from 'wix-style-react/MultiSelect';`

Paste text(Cmd + V)

### Current Behavior
Paste to input some text
The `onChange` callback is fired twice with empty value
The `onSelect` callback is fired

### Expected Behavior
Paste to input some text
The `onManuallyInput` callback once is fired with proper value

## Steps to Reproduce
https://codesandbox.io/s/xv0n1zppop

## Severity

Blocker

adavidof:
Regarding https://github.com/wix/wix-style-react/issues/2199 as well

shlomitc:
Just for the reference - the Tags/MultiSelect component is a bit complex regarding what should happen on a new value. it acts both as a "Select" (pick from list) component and as an "Input" (add new items).

The behavior you're describing in the second issue is relevant to the "Input" family, and this is currently achieved with the `onManuallyInput` prop.

I think we can achieve this also with `onBlur` but for now I think you can use `onManuallyInput`

adavidof:
@adavidof yeah, but in the example https://codesandbox.io/s/xv0n1zppop `onManuallyInput` doesn't work :/

shlomitc:
Try the following snippet, it's the most "controlled" way you can work with the component.
```jsx
import React from 'react';
import MultiSelect from 'wix-style-react/MultiSelect';

class App extends React.Component {
  state = {
    inputValue: '',
    tags: [],
  };

  uniqueTagId = 0;

  handleOnSelect = tags => {
    const newTag = { ...tags[0], id: this.uniqueTagId.toString() };
    this.uniqueTagId++;
    return this.setState({ tags: [...this.state.tags, newTag] });
  };

  handleOnRemoveTag = tagId =>
    this.setState({
      tags: this.state.tags.filter(currTag => currTag.id !== tagId),
    });

  handleOnChange = event => this.setState({ inputValue: event.target.value });

  render() {
    return (
      <MultiSelect
        // control the input changes
        value={this.state.inputValue}
        onChange={this.handleOnChange}
        // the actual tags to display
        tags={this.state.tags}
        // manage tags creation and removal
        onSelect={this.handleOnSelect}
        onRemoveTag={this.handleOnRemoveTag}
      />
    );
  }
}

export default App;
```

adavidof:
@shlomitc 
[...]

PR #2692, opened 2018-12-31
`<MultiSelect/>` - Remove special handling of Paste

hepiyellow:
Resolves https://github.com/wix/wix-style-react/issues/2559

## New Behavior

When pasting is done, we won't call nither `onSelect` nor `onManuallyInput`.
After pasting, the user needs to press the "Enter" key or select from the options.

- This means we won't be supporting something like pasting `'alabama,alaska'` and expecting that 2 tags will be automatically created.
- In fact, if `'alabama,alaska'` is pasted and enter is pressed, then `onManuallyInput` would be called with `'alabama,alaska'`.

FYI: @wuwa @adavidof 

## New Callback Summary Table

|#| Options | Type & Enter | onSelect | onManuallyInput |
|---|---|---|---|--|
|1| [] |   foo |foo| foo |
|2| [alabama] |  alabama |alabama| alabama |
|3| [alabama] |  alab |alabama| alab |
|4| [alabama] | foo | |foo |

### Bug Fix 1

Previously we if options included an `'alabama'` optoin, then the user would not be able to add a new tag with `'alab'` value. (A partial of an existing option).

Now it's possible !

### Bug Fix 2

 This change fixes (incidentally) also a bug that after an option is selectedm and the input is clicked, the options are close, and tying in the input box does NOT open the options again.

### Bug Fix 3

I don't know what the specific scenario is, but the `this.state = { pasteDetected: false };` from MultiSelect constructor, was overwriting the initial state initialized by `InputWithOptions`.
So this fixed that as well.


hepiyellow:
See new decission
https://github.com/wix/wix-style-react/issues/2559#issuecomment-450616904

PR #2694, opened 2018-12-31
Modal content label

zivl:
### What changed

1. `src/Modal/Modal.js`
2. `src/Modal/README.md`

### Why it changed
1. the `Modal` prop `contentLabel` should not be required by a WSR consumer. The WSR should take care of it on its own as this implementation is WSR's (consider a scenario in the future where  WSR might replace `react-modal` with something else)
2. update documentation regarding the `contentLabel` prop + added A11Y documentation for those WSR consumers who are interested in this feature

---

- [X] Change is tested
- [X] Change is documented
- [X] Build is green


src/Modal/Modal.js
```diff
@@ -45,6 +45,7 @@ class Modal extends WixComponent {
     height: '100%',
     maxHeight: 'auto',
     overlayPosition: 'fixed',
+    contentLabel: '',
```
hepiyellow: why empty string, why should we have a default value at all?

shlomitc: yes please

zivl: done!

zivl: pls merge

PR #2698, opened 2019-01-01
<FormField/> - allow setting direction for one-unit components

nitayneeman:


stories/CheckBox/ExampleFormField.js
```diff
@@ -0,0 +1,31 @@
...
+          <FormField
+            dataHook="storybook-checkbox-formfield"
+            infoContent="I help you to fill info"
+            required
+          >
+            <Checkbox
+              checked={this.state.value}
+              onChange={e => this.setState({ value: e.target.checked })}
+            >
+              Checkbox
```
nitayneeman: I'm still wondering about this example. The label could be passed through the `FormField` instead, but then we're faced some issues:
image

1. The checkbox isn't aligned vertically
2. There's no pointer cursor over the label
3. We should connect the label to to the checkbox using `for`

Obviously, we can handle them manually - but probably will be better to achieve the behavior through the checkbox component. 

What do you think? @shlomitc

stories/FormField/ExampleWithinGrid.js
```diff
@@ -0,0 +1,65 @@
...
+
+export default () => (
+  <MessageBoxFunctionalLayout
+    title="User Details"
+    confirmText="Save"
+    cancelText="Cancel"
+  >
+    <Container fluid>
+      <Row stretchViewsVertically>
+        <Col span={3}>
```
nitayneeman: @shlomitc 	 Can we use float numbers for columns span (it's possible in Bootstrap indirectly)?

src/FormField/FormField.e2e.js
```diff
@@ -39,6 +39,32 @@ describe('FormField', () => {
...
 
+    eyes.it(
+      'should render with label on the right',
+      async () => {
+        await autoExampleDriver.setProps({ label: 'hello', labelPlacement: 'right' });
+        await waitForVisibilityOf(
+          driver.element(),
+          'Cannot find FormField component',
+        );
+        expect(await driver.getLabel().getText()).toMatch('hello');
```
nitayneeman: Instead of implementing `hasTopLabel`, `hasRightLabel` and `hasLeftLabel` I just used `getLabel`. I had started to implement the three methods, but actually got the same code for them (selecting the element by `formfield-label` data hook) - what made it unnecessary IMO.

stories/FormField/ExampleWithinGrid.js
shlomitc: probably not. we use only integers. @wuwa - what do you think?

src/FormField/FormField.e2e.js
shlomitc: let's keep it and use the class names to verify the right direction

stories/CheckBox/ExampleFormField.js
shlomitc: 1. suggested suggestion - might want to position the FormField label correctly (top-left, middle-left, bottom-left).
2. it might make sense that when you have an `id`, it will make it clickable. should verify with UX.
3. right

stories/FormField/ExampleWithinGrid.js
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
+
+import { MessageBoxFunctionalLayout } from 'wix-style-react/MessageBox';
+import { Container, Row, Col } from 'wix-style-react/Grid';
+import FormField from 'wix-style-react/FormField';
+import Input from 'wix-style-react/Input';
+
+const verticalAlignColStyle = {
```
shlomitc: let's keep it that way currently, but we should have a better alignment solution.
either as part of the Grid's `Col` or a separate `Box` component that takes care of vertical and horizontal alignment

stories/CheckBox/ExampleFormField.js
nitayneeman: Regarding 1 - I think it's a specific issue of the Checkbox. When there's no label, it's not aligned with the element next to it. It makes sense, because the text defines the height of the whole line. Anyway I found a fix for that using `inline-flex` and `min-height`.

Regarding 2 - I talked to UX and they don't think that `pointer` cursor should be displayed over the label for all cases (while `id` is passed), for instance, it's not relevant for Input.

So I see the following options:
1. Using the Checkbox's label instead of FormField's label, but the issue would still exist for ToggleSwitch
2. Adding an API that customizes the cursor of FormField's label
3. Distinguishing somehow between the children components of the FormField, so the label will be styled for the relevant component

shlomitc: Native `html` elements don't have a pointer when wrapping a checkbox, so I don't think we should do that as well, for the sake of consistency.

@milkyfruit - can we ditch the pointer and make the pattern consistent?

milkyfruit: @shlomitc @nitayneeman 
I suggest using default cursor on labels:
image
While on checkbox itself keeping a pointer, as well as text cursor for text input:
image

PR #2702, opened 2019-01-01
`<MultiSelect/>` - New callback API 

hepiyellow:
According to [this spec](https://github.com/wix/wix-style-react/pull/2697).
## `upgrade` prop

When `upgrade` prop is provided, then a new API will be used.

## What is NOT included yet in this PR

- `markFirstMatchingOption` is not implemented yet. The default behavior is `markFirstMatchingOption=false`
- `onSelect` is still called with old argument logic.
- No Documentation or Examples
- No Migration Doc

## DO NOT MERGE YET

Missing features and docs will be in separate PRs (to avoid a huge PR) which will have this branch as the target branch. 
When everything is ready, then we could merge to master.


src/MultiSelect/MultiSelect.js
```diff
@@ -69,32 +77,43 @@ class MultiSelect extends InputWithOptions {
     this.setState({ pasteDetected: true });
   }
 
+  _splitValues(value) {
+    const delimitersRegexp = new RegExp(this.props.delimiters.join('|'), 'g');
+    const valueWithCommas = value.replace(delimitersRegexp, ',');
+    return valueWithCommas
+      .split(',')
```
yardnsm: You can pass a `RegExp` to `split`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split#Parameters

src/MultiSelect/MultiSelect.js
```diff
@@ -69,32 +77,43 @@ class MultiSelect extends InputWithOptions {
...
+      .map(str => str.trim())
+      .filter(str => str);
+  }
+
   _onChange(event) {
-    if (!this.state.pasteDetected) {
+    if (this.state.pasteDetected) {
+      if (this._isNewCallbackApi()) {
+        const value = event.target.value;
```
yardnsm: I've seen `event.value.target` used multiple times in this methods. Why not hoist it to the top?

src/MultiSelect/MultiSelect.js
```diff
@@ -163,25 +197,39 @@ class MultiSelect extends InputWithOptions {
...
-      this.input.blur();
       return;
     }
 
-    if (this.props.onManuallyInput) {
-      this.props.onManuallyInput(
+    const { onManuallyInput, onTagsAdded } = this.props;
+    if (this._isNewCallbackApi()) {
+      const values = this._splitValues(inputValue);
+      onTagsAdded && values.length && onTagsAdded(values);
```
yardnsm: I'm not a fan of this chaining style in general, but you can remove it.

src/MultiSelect/MultiSelect.js
```diff
@@ -163,25 +197,39 @@ class MultiSelect extends InputWithOptions {
     return tag ? { id, ...tag } : { id, label: value, theme };
   }
 
-  onSelect(options) {
+  onSelect(_options) {
```
yardnsm: Why `_options` and not keeping `options`?

src/MultiSelect/MultiSelect.js
hepiyellow: ok. done.

src/MultiSelect/MultiSelect.js
hepiyellow: I want to differentiate it from the `props.options`.
Maybe I should call the argument `selectedOptions`?
(which could also be misleading...)

hepiyellow:
Approved by Shlomi.
Waiting for review on The Examples PR.
https://github.com/wix/wix-style-react/pull/2708

hepiyellow:
Replaced by #2752

Issue #2596, opened 2018-12-16
<Tooltip/> - causes infinite loop on Firefox


liorur:
# Bug Report

### Relevant Components
Tooltip

apparently there is a very unlucky font+string that causes this loop to never exit on FF:
[https://github.com/wix/wix-style-react/blob/master/src/Tooltip/Tooltip.js#L443](https://github.com/wix/wix-style-react/blob/master/src/Tooltip/Tooltip.js#L443)

the `sw` and `fw` begin alternating values over fractions of pixels

i suggest limiting the comparison precision or limiting the do-while iterations

### Steps to Reproduce
very difficult to reproduce

### Severity
Major - rarely occurs but hangs the browser when it does

shlomitc:
@liorur - I couldn't reproduce. can you please add an example?

liorur:
@shlomitc you can open a store in BM and go to settings-> tax then add a country , save, click it for editing and then try to delete it using the popoverMenu

Crembot:
Did repro - required *very* high DPI screen (e.g. macbook retina display). PR incoming.

PR #2704, opened 2019-01-02
<Tooltip /> reposition loop: added iteration guards to avoid hangs (#2596)

Crembot:
### What changed
`<Tooltip />` now stops the layout->reposition loop when the results are "close enough" or after too many attempts.

### Why it changed
`<Tooltip />` used to reapply layout-reposition in a loop until it converged to an *exact* solution but in certain configurations (notably, Firefox + retina DPI) it got locked in an infinite loop and hung (see #2596)

---

- [x] Change is tested
 Locally in storybook
- [ ] ~~Change is documented~~ **N/A**
- [X] Build is green


shlomitc:
@yardnsm - can you please review it?

src/Tooltip/Tooltip.js
```diff
@@ -434,14 +434,22 @@ class Tooltip extends WixComponent {
...
             fw = this._getRect(_tooltipNode).width;
             this._updatePosition(this.tooltipContent);
             sw = this._getRect(_tooltipNode).width;
           }
-        } while (!props.appendToParent && fw !== sw);
+          ++iterations;
+          pixelChange = Math.abs(fw - sw);
+        } while (
+          !props.appendToParent &&
+          pixelChange > 0.1 &&
```
yardnsm: What's the rationale of `pixelChange`? It "just works" rather than `fw !== sw`?

Crembot: Mainly that under the conditions specified (Firefox, retina DPI, this font) it will _never_ actually converge to exactness, therefore we will stop if we are "close enough". A tenth of pixel should be fine.

PR #2705, opened 2019-01-02
`<MultiSelect/>` - Fix `<Input/>` & `<InputWithTags/>` so that clearing input value would actually clear the state.

hepiyellow:
This fixes the following bug in `<MultiSelect/>`, that if you:
-  type a value `foo`
-  press Enter
  - `onManuallyInput` is called with `foo`
  - The input is cleared
- press Enter (again)
  - `onManuallyInput` is called (again) with `foo`

There are a few fixes here:
### `<Input>`
- Change `_onClear()` to a public method `clear()` (that can be called by a consumer with a component instance ref)
- Fix `input.clear()` so it can be called without an event.
- Fix `input.clear()` so it doesn't call `onChange` if the value is already empty. 
- Refactor `clear()` so that it doesn't to `input.focus()`, instead, we do it in a new `handleSuffixOnClear`.

### `<InputWithTags/>`

- Fix `InputWithTags` to call `input.clear()` instead of just setting state. (so that onChange will be called)

### `<MultiSelect/>`

- Now that calling `input.clear()` triggers an `onChange`, then fix `MultiSelect.onChange()` Paste handling, to do the Clear only after setting `pasteDetected` to false. To prevent calling onSelect multiple times.


src/Input/InputSuffix.js
```diff
@@ -45,7 +45,7 @@ const InputSuffix = ({
   onIconClicked,
   magnifyingGlass,
   isClearButtonVisible,
-  onClear,
+  clearInput,
```
argshook: why rename? `onClear` makes sense to me

src/Input/Input.js
```diff
@@ -282,16 +282,35 @@ class Input extends Component {
...
 
+    const prevValue = this.input.value;
     this.input.value = '';
 
-    e.target = {
-      ...e.target,
-      value: '',
-    };
-    this._onChange(e);
+    if (prevValue) {
```
argshook: is it not enough to mutate and simply `e.target.value = ""` and then leave `this.onChange(e)` unchanged?

im not sure but guessing that even `this.input.value = '';` should be enough

src/Input/Input.js
```diff
@@ -282,16 +282,35 @@ class Input extends Component {
...
+      if (!e) {
```
argshook: is it possible to get here without event? `this.clear` is passed to `InputSuffix` which is invoked `onClick`, it appears to always have event, no?

src/Input/Input.js
```diff
@@ -282,16 +282,35 @@ class Input extends Component {
...
+        e = {
+          target: this.input,
+        };
+      }
+      /* FIXME: The event (e) could be any event type, and even it's target may not be the input.
+       * So it would be better to do e.target = this.input.
+       * We don't use `clear` in WSR except in InputWithTags which does not pass an event, so it's ok.
+       * But if some consumer is using <Input/> directly, then this might be a breaking change.
+       */
+      e.target = {
```
argshook: if `e.target.value = ""` is enough, the no need for this spread, or any of the comments

src/Input/InputSuffix.js
hepiyellow: its not a listener, its an action trigger.

hepiyellow: reverted

src/Input/Input.spec.js
```diff
@@ -637,6 +640,47 @@ describe('Input', () => {
...
+      class ControlledInputWithRef extends React.Component {
+        state = {
+          value: '',
+        };
+        constructor(props) {
+          super(props);
+          this.handleChange = this.handleChange.bind(this);
+        }
+
+        handleChange(e) {
```
argshook: just fyi, our yoshi can handle arrow functions here, then you don't need line 651:

`handleChange = e => { ...`

hepiyellow: I known

PR #2706, opened 2019-01-02
`<ButtonWithOptions/>`,` <IconWithOptions/>` - deprecate components

yardnsm:
This PR deprecates the `<ButtonWithOptions/>` and the ` <IconWithOptions/>` components in favour of the `<DropdownPopover/>` components.

The components' stories have been moved to the `Deprecated` section. In addition, this PR adds padding to the `<DropdownLayout/>` (of `<DropdownPopover/>`) when `showArrow` is set.

- [ ] Add a deprecation log to the components


yardnsm:
This PR also add padding to the `<DropdownPopover/>` component when the `showArrow` prop is provided. You can check the playground and the examples in https://wix-wix-style-react-pr-2706.surge.sh/?selectedKind=Components&selectedStory=DropdownPopover&full=0&addons=0&stories=1&panelRight=0.

@wuwa @milkyfruit can you please review this? Do we want to have the padding to every "dropdown-like" component with an arrow, or do we want to disable the padding in some times? What about cases when the placement is `left` or `right` (currently the padding is applied in the `top` and in the `bottom`)?

src/ButtonWithOptions/ButtonWithOptions.js
```diff
@@ -9,6 +9,12 @@ import ChevronDown from '../new-icons/ChevronDown';
 
 import styles from './ButtonWithOptions.scss';
 
+import deprecationLog from '../utils/deprecationLog';
+
+deprecationLog(
```
shlomitc: I think it worth showing the log only when using the component and not just when importing it

src/DropdownPopover/DropdownPopover.st.css
```diff
@@ -20,3 +20,8 @@
   min-width: inherit;
   width: inherit;
 }
+
+/* Add padding when an arrow is used */
+.root:withArrow::popoverContent {
```
shlomitc: This is a bit opinionated approach, so better have this agreed with the UX team (and covered by test)

src/IconWithOptions/IconWithOptions.js
```diff
@@ -5,6 +5,12 @@ import DropdownLayout from '../DropdownLayout/DropdownLayout';
 import styles from './IconWithOptions.scss';
 import classNames from 'classnames';
 
+import deprecationLog from '../utils/deprecationLog';
+
+deprecationLog(
```
shlomitc: same as ButtonWithOptions

src/ButtonWithOptions/ButtonWithOptions.js
yardnsm: I've seen this pattern in `<ButtonLayout/>` and in `<TextLink/>`. So you suggest putting the `deprecationLog` in the `constructor`?

yardnsm:
So just to be clear - we'll apply the padding only if and arrow is show **and** placement is `top` or `bottom`? @wuwa

wuwa:
yes

shlomitc:
@wuwa - let's add this to the specs, otherwise, it will be forever a mystery

milkyfruit:
Specs updated to match 12px paddings:

Popover Menu
zpl.io/Vx4KyWV

List Item Action 
zpl.io/2ZkBO3J

yardnsm:
@wuwa @milkyfruit 

Also take a look at the `<IconWithOption/>` story and the `<DropdownPopover/>` example that tries to replicate it.

You can see that the icon color is changing on hover. 

##### `<IconWithOption/>`:

screen shot 2019-01-08 at 14 51 06

##### `<DropdownPopover/>`:

screen shot 2019-01-08 at 14 50 58

Also notice the focus styles on the latter. Is it needed?
The `<DropdownPopover/>` example uses `<TextButton/>` with `skin="dark"`. Maybe we should define a new `skin` for the **`<IconButton/>`** component so it would match? Or maybe using the `<TextLink/>` is enough.

wuwa:
i think it's not the best semantically to use TextButton for icon...
I guess IconButton needs new skins (and sizes - L,T btw).
focus is needed.

milkyfruit:
I generally don't feel strong with exposing example like plain icon with options on hover.
If some options should be hidden under the icon, I would suggest this pattern from restaurants product:

**Default:**
image

**Hover:**
image

**Click:**
image

In this example `<iconButton />` is used, but the same pattern should be applied if a dropdown mechanism is used with a plain icon.

yardnsm:
@milkyfruit So you're suggesting removing the `<IconWithOptions/>` "thing" completely, including the example in the `<DropdownPopover/>` story?

The examples you've attached looks like the `<PopoverMenu/>`, so I guess we should refer consumers to use it instead.

PR #2707, opened 2019-01-03
Autotools snap POC

danielstro:
### What changed
I added `@ui-autotools/snap` to one component.

---

- [x] Change is tested
- [ ] Change is documented
- [ ] Build is green


scripts/component-generator/templates/component/src/Component/Component.meta.js
```diff
@@ -3,6 +3,7 @@ import Registry from '@ui-autotools/registry';
 
 const metadata = Registry.getComponentMetadata({%ComponentName%});
 
-metadata.exportedFrom({
+metadata.exportInfo = {
   path: 'src/{%ComponentName%}/{%ComponentName%}.js',
-});
+  exportName: '{%ComponentName%}',
```
hepiyellow: Most of our components use `default` export.
Does the registry support that?

The Generated component currently also uses default export, so this is broken.

scripts/generate-autodocs-registry.js
```diff
@@ -21,8 +21,13 @@ async function generate() {
   const autodocsRegistry = {};
   const gatherAllPromises = [];
   Registry.default.metadata.components.forEach(async comp => {
+    const path = comp.exportInfo.path;
+    if (!path) {
+      // eslint-disable-next-line no-console
+      console.error('Export info must contain a path to the component.');
```
hepiyellow: I think we can throw an error and abort.

src/EmptyState/EmptyState.meta.js
```diff
@@ -0,0 +1,31 @@
...
+const meta = Registry.getComponentMetadata(EmptyState);
+meta.addSim({
+  title: 'Normal_example',
+  props: {
+    theme: 'page',
+    title: 'Title!',
+  },
+});
+
+meta.addCustomField(wsrPluginMetaKey, {
```
hepiyellow: What is this?

src/Calendar/Calender.meta.js
```diff
@@ -0,0 +1,34 @@
...
+  },
+});
+
+meta.addCustomField(wsrPluginMetaKey, {
+  compPath: 'src/Calendar',
+});
+
+meta.staticResources = [
+  {
+    path: 'dist/statics/App.css',
```
hepiyellow: Where do we create this `App.css`?

src/Calendar/Calender.meta.js
```diff
@@ -0,0 +1,34 @@
...
+    mimeType: 'text/css',
+  },
+];
+
+meta.exportInfo = {
+  path: 'src/Calendar',
+  exportName: 'Calender',
+};
+
+meta.nonEventListenerTestCompliant = true;
```
hepiyellow: Should this be set for ALL WSR components?
Can we define meta defaults in the configuration? (Or simply call a common function in all meta files)

src/Input/Input.scss
```diff
@@ -507,8 +507,8 @@ $material-gap: 8px;
   padding: 0 $gap;
   color: $paletteD3;
   @include FontRoman();
-  //needs display:block to fix inputs with prefix in ie11
-  //https://github.com/wix/wix-style-react/issues/1691
+  /*needs display:block to fix inputs with prefix in ie11
```
hepiyellow: Why was this fix necessary?
Is it a JSDOM issue?
I recall having a similar issue while trying to use render grid, and trying to use JSDOM to create a DOM snapshot. The case there was with empty class definitionwith only comments:
```css
.myClass {
 /* my comment */
}
```
So stuff like this pass in browser, but fails in JSDOM.

In some cases these css stylesheets are coming from 3rd party libraries, so we can't fix them to fit JSDOM limitation (or bugs).

How do you intend to deal with such cases?

package.json
```diff
@@ -53,6 +53,9 @@
     "proxy-wix-ui-icons": "node scripts/proxy-wix-ui-icons",
     "generate": "./scripts/component-generator/bin/component-generator.js",
     "generate-test-component": "./scripts/component-generator/bin/generate-test-component.js",
+    "a11y": "autotools-a11y -f 'src/**/*.meta.js'",
+    "snap": "autotools-snap -f 'src/**/*.meta.js'",
```
hepiyellow: I would rename to `test:snap`

src/Calendar/Calender.meta.js
```diff
@@ -0,0 +1,34 @@
...
+  props: {
+    theme: 'page',
+    title: 'Title!',
+    onChange: () => {},
+    value: new Date(1, 1, 1),
+  },
+});
+
+meta.addCustomField(wsrPluginMetaKey, {
+  compPath: 'src/Calendar',
```
hepiyellow: What is the difference between this custom `compPath` and the `exportInfo.path`?

.autotools/plugins/wsr-snap-plugin.js
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * We only want to create snapshots for components that have specified the correct info
+     * in their metadata file.
+     */
+    if (!wsrPluginMetadata) { return; }
+
+    // Get the path to the compiled component
+    const compPath = wsrPluginMetadata.compPath;
+    const compName = componentMetadata.exportInfo.exportName;
```
hepiyellow: I think the `compName` should be `getCompName(componentMetadata.component)`.

And then we can also remove the `exportName` you added to the  meta files

hepiyellow:
Remind me how do I workaround this :
```
Error: The environment variable "APPLITOOLS_API_KEY" needs to be defined.
```

I would like to be able to run `npm run snap` without Appliotools.
Only creating the snapshots, and writing them to files (under `test/snap`)

shlomitc:
We decided that in `wix-style-react` we'll use `eyes-storybook` for simple visual testing. https://github.com/wix/wix-style-react/pull/3275

The `snap` tool is great for the advanced usages of the core components and other libraries, so I think it should keep growing there and we'll adapt in the future

PR #2710, opened 2019-01-04
generate enzyme and vanilla testkits exports

argshook:
continuing with testkit automation effort.

---

<details>
  <summary>previous message, pr updated, ignore</summary>

  we already automate enzyme testkits but this is done runtime.
  this PR changes approach so that testkit exports are generated build time.

  runtime is not bad per se but there are cases when our testkits are used
  in browser, meaning no `fs`, no `path` modules.

  ---

  PR includes exports generator as well as generated file. We could
  probably have generated file only after build, but didn't work on that
  here.

  ---

  PR also includes updates for `component-definitions` files:

  * less skipped sanity tests. recent updates in wix-style-react and wix-ui-test-utils allowed that. this is around 16 more sanity tests
  * more control in `component-definitions`. added `manualExport` flag to not automate export for some components. Those are proxied ones like from wix-ui-bakcoffice

  ---

  if build is green, this can be merged, test amount increased, not reduced.
</details>

---

this PR contains few things:

* generator for enzyme and vanilla testkits exports:  
new `npm run generate-testkit-exports` which happens durng `prebuild`

* generated testkits code for enzyme and vanilla testkits  
good old `testkit/enzyme.js` and `testkit/vanilla.js` files but they are now generated in build time

* component-definitions refactored into two:
  1. `testkit/component-definitions.js`  
  object to define required props, before/after test hooks (for sanity test setup & teardown)
  2. `scripts/generate-testkit-exports/testkit-definitions`  
  object to define custom testkit path, is it unidriver etc.

  yes, this split is not ideal but a good step towards some sort of component registry.

* some documentation updates

---

around 100+ tests. We didn't check exports for vanilla testkits, which we now do. Also less skipped sanity tests due to various fixes.

---

i want to merge this soon, so speak now!


test/jest-setup.js
```diff
@@ -1,5 +1,6 @@
 import Enzyme from 'enzyme';
 import Adapter from 'enzyme-adapter-react-15';
+import './generate-testkit-exports';
```
yardnsm: Why add it in the test phase? IMO this is more appropriate ~~but I'm fine with having it at `jest-setup.js`~~.

`jest-setup` will be ran *for each test file*, according to the [docuemntation](https://jestjs.io/docs/en/configuration.html#setupfiles-array), causing this script to run ~140 times. So either use this script before we run `jest`, or use it in the `build` script.

Also, if you'll add it before the `generate-test-component` script, you can get an indication when the latest commit doesn't have the latest version of the generated testkit file.

test/generate-testkit-exports/index.js
```diff
@@ -0,0 +1,96 @@
...
+    ' */',
+  ].join('\n');
+
+export const generateTestkits = ({
+  templatePath,
+  outputPath,
+  factoryCreator,
+  uniFactoryCreator,
+}) => {
+  const loadUtilPath = pathResolve('load.js');
```
yardnsm: Since `load.js` is a template, maybe call it `load.templete.js`?

test/jest-setup.js
argshook: that's good observation and something i want to fix before merging. Yes, script in `jest-setup` is silly, but works.

the reason i do like this is because i need this file to be run in
"yoshi runtime", meaning es6, stylable, scss and friends.
`generate-test-component` can work from `scripts` and es5 is enough
there (no transpillation, nothing) so that's easy.

i need yoshi because `component-definitions` may import components or
use jsx (to set required props). it's possible to go away with it but it
would require to split `component-definitions` into some
`testkit-sanity-config` and `testkit-export-config` which is not
something i would like.

in other words, how can i do this:

`node test/generate-teskit-exports`

but instead of `node` use `yoshi`:

`yoshi test/generate-teskit-exports`

and ideally avoid any `entry` configuration, starting up yoshi cdn server and stuff, just transpile and run.

yardnsm: From a quick search in `jest`'s docs, I found the [`globalSetup`](https://jestjs.io/docs/en/configuration#globalsetup-string) option which seems it can suit our needs but I haven't looked deep further.

argshook: i'm aware of `globalSetup` but it doesn't help as it runs `node script.js` without any of the yoshi

PR #2711, opened 2019-01-06
Make Dropdown controlled again

Wo1v3r:
### What changed

Added a `controlled` prop to <Dropdown/> to make it "really controlled"

### Why it changed

Before, <Dropdown/> was not really controlled, and for example, when `onSelect` does not change the `selectedId` prop (which is mapped to the component's state) the local state of the <Dropdown/> is still altered to the "selected" option.

---

- [x] Change is tested
- [x] Change is documented
- [ ] Build is green

### DIFF
--- a/src/Dropdown/Dropdown.js
+++ b/src/Dropdown/Dropdown.js
@@ -2,7 +2,7 @@ import classNames from 'classnames';
 import isUndefined from 'lodash/isUndefined';
 import InputWithOptions from '../InputWithOptions/InputWithOptions';
 import styles from './Dropdown.scss';
-import { PropTypes } from 'react';
+import PropTypes from 'prop-types';
 
 class Dropdown extends InputWithOptions {
   constructor(props) {
--- a/src/Dropdown/Dropdown.js
+++ b/src/Dropdown/Dropdown.js
@@ -79,7 +79,12 @@ class Dropdown extends InputWithOptions {
   }
 }
 
-Dropdown.propTypes = {...InputWithOptions.propTypes, controlled: PropTypes.bool};
+Dropdown.propTypes = {
+  ...InputWithOptions.propTypes,
+  /** When true, changing `selectedId` prop will determine the selected option and its' parsed value */
+  controlled: PropTypes.bool
+};
+
 Dropdown.defaultProps = InputWithOptions.defaultProps;
 Dropdown.displayName = 'Dropdown';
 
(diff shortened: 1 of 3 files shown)


src/Dropdown/Dropdown.js
```diff
@@ -68,15 +69,17 @@ class Dropdown extends InputWithOptions {
   }
 
   _onSelect(option) {
-    this.setState({
-      value: this.props.valueParser(option),
-      selectedId: option.id,
-    });
+    if (!this.props.controlled) {
```
hepiyellow: This means that `value ` is also controlled?
It should affect only `selectedId`

src/Dropdown/Dropdown.js
```diff
@@ -2,6 +2,7 @@ import classNames from 'classnames';
 import isUndefined from 'lodash/isUndefined';
 import InputWithOptions from '../InputWithOptions/InputWithOptions';
 import styles from './Dropdown.scss';
+import { PropTypes } from 'react';
```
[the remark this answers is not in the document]
Wo1v3r: Fixing

src/Dropdown/Dropdown.js
Wo1v3r: Of-course, otherwise, the uncontrolled option would be rendered as the input's value.
I'll add a test to make sure it does *not* happen.

src/Dropdown/Dropdown.js
```diff
@@ -68,15 +69,17 @@ class Dropdown extends InputWithOptions {
...
+        value: this.props.valueParser(option),
+        selectedId: option.id,
+      });
+    }
     super._onSelect(option);
   }
 }
 
-Dropdown.propTypes = InputWithOptions.propTypes;
+Dropdown.propTypes = {...InputWithOptions.propTypes, controlled: PropTypes.bool};
```
hepiyellow: Add documentation on the `controlled` prop. By adding JSDoc comments above it.
```js
Dropdown.propTypes = {
   ...InputWithOptions.propTypes,
   /** When true then `selectedId` and `value` are controlled */
   controlled: PropTypes.bool
};
```

You may think of better phrasing for the docs.

src/Dropdown/Dropdown.js
hepiyellow: If `value` is also going to be controlled, then you need in `dropdownAdditionalProps` to use the `props.value` instead of from the state (when in `controlled` mode)

Wo1v3r: value is inferred from selectedId in Dropdown so IMO no reason to change the current behaviour (works correctly on story example)

src/Dropdown/Dropdown.js
Wo1v3r: fixed

Wo1v3r:
@hepiyellow  Any other changes you wish to make ?

hepiyellow:
Replaced by #2719

PR #2716, opened 2019-01-07
`<MultiSelect/>` - Disable dropdownLayout selection

hepiyellow:
When an option is clicked, we add it to the tags.

We either remove the option from the list of options, or leave it, but the option should not be marked as selected.

>I encountered this while using ContactItemBuilder as an option builder, when clicking on an option, it got selected,  and the blue background color didn't work for some reason, so the text got WHITE (selected mode) and I couldn't see it.


src/MultiSelect/MultiSelect.js
```diff
@@ -245,8 +246,19 @@ class MultiSelect extends InputWithOptions {
...
+    closeOnSelect,
+    selectedHighlight,
+    ...rest
+  } = InputWithOptions.defaultProps;
+  return rest;
+}
+
 MultiSelect.propTypes = {
-  ...InputWithOptions.propTypes,
+  ...inputWithOptionsPropTypes,
```
yardnsm:
```suggestion
  ...inputWithOptionsPropTypes(),
```

src/MultiSelect/MultiSelect.js
```diff
@@ -245,8 +246,19 @@ class MultiSelect extends InputWithOptions {
...
 }
 
+function inputWithOptionsPropTypes() {
+  const {
+    // The following props are overriden in dropdownAdditionalProps()
+    selectedId,
+    closeOnSelect,
+    selectedHighlight,
+    ...rest
+  } = InputWithOptions.defaultProps;
```
yardnsm:
```suggestion
  } = InputWithOptions.propTypes;
```

PR #2719, opened 2019-01-07
`<Dropdown/>` - Add support for Controlled `selectedId`

hepiyellow:
Replaces #2711 

## `upgrade` prop

New `upgrade` prop needs to be added so that Dropdown will support Controlled and Uncontrolled modes (for `selectedId`).
The `upgrade` prop is needed, since we also fix the Uncontrolled behaior.

## Modes
 - Controlled - `selectedId` provided
 - Uncontrolled - `selectedId` is NOT provided. `initiallySelectedId` may be used.

## Tests

I copied ALL the previous tests into a new describe Block called "Upgrade", so that we do regression tests in the NEW APi.


src/Dropdown/Dropdown.js
```diff
@@ -57,26 +132,74 @@ class Dropdown extends InputWithOptions {
...
+        );
+      }
+      if (props[propName] && props['initiallySelectedId']) {
+        deprecationLog(
+          `'selectedId' and 'initiallySelectedId' cannot both be used at the same time.`,
+        );
+      }
+    },
+  ),
+  initiallySelectedId: validatorWithSideEffect(
```
yardnsm: IMO `initialSelectedId` is a better name, also used in `<BadgeSelect/>` and `<DropdownPopover/>` and defined in the `CONTROLLED_VS_UNCONTROLLED.md` doc.

src/Dropdown/Dropdown.js
```diff
@@ -1,14 +1,55 @@
 import classNames from 'classnames';
 import isUndefined from 'lodash/isUndefined';
+import defaultTo from 'lodash/defaultTo';
```
yardnsm: Why do we have to use it?

src/Dropdown/Dropdown.js
```diff
@@ -1,14 +1,55 @@
 import classNames from 'classnames';
 import isUndefined from 'lodash/isUndefined';
```
yardnsm: Can we remove it? It only check if `value === undefined` (but a better way is probably `typeof value === 'undefined'`).

src/Dropdown/Dropdown.js
```diff
@@ -1,14 +1,55 @@
...
+    const { upgrade, selectedId } = this.props;
+    return upgrade && !isUndefined(selectedId);
+  }
+
+  isControlledSupported() {
+    return this.props.upgrade;
+  }
+
+  static isOptionsEqual(optionsA, optionsB) {
+    return differenceBy(optionsA, optionsB, o => o.id).length === 0;
```
yardnsm: Why? Can't we use `optionA.id === optionB.id`?

src/Dropdown/Dropdown.js
```diff
@@ -1,14 +1,55 @@
...
+      this.deprecatedUpdate(props, { isFirstTime: true });
+    }
+  }
+
+  isSelectedIdControlled() {
+    const { upgrade, selectedId } = this.props;
+    return upgrade && !isUndefined(selectedId);
+  }
+
+  isControlledSupported() {
```
yardnsm: You can remove it. `isSelectedIdControlled` does the same check so it will suffice.

src/Dropdown/Dropdown.js
```diff
@@ -1,14 +1,55 @@
...
+    return this.props.upgrade;
+  }
+
+  static isOptionsEqual(optionsA, optionsB) {
+    return differenceBy(optionsA, optionsB, o => o.id).length === 0;
   }
 
+  getSelectedId() {
```
yardnsm: This can simply be changed to

```js
getSelectedId() {
  return this.isSelectedIdControlled()
    ? this.props.selectedId
    : this.state.selectedId;
}
```

src/Dropdown/Dropdown.js
```diff
@@ -24,7 +65,27 @@ class Dropdown extends InputWithOptions {
     }
   }
 
-  update(props, { isFirstTime }) {
+  /**
+   * Updates the value by the selectedId.
+   * If selectedId is not found in options, then value is NOT changed.
+   */
+  static getNextState(props, selectedId) {
```
yardnsm: Please put all `static` methods at the beginning of the clas.

src/Dropdown/Dropdown.js
```diff
@@ -46,7 +107,21 @@ class Dropdown extends InputWithOptions {
   }
 
   componentWillReceiveProps(nextProps) {
-    this.update(nextProps, { isFirstTime: false });
+    if (this.props.upgrade) {
```
yardnsm: `componentWillReceiveProps` is considered legacy as in React 16. Use `getDerivedStateFromProps` to replace it (using `getDerivedStateFromProps` may also turn `getNextState` redundant).

src/Dropdown/Dropdown.js
yardnsm: It seems that `getNextState` can be replaced by `getDerivedStateFromProps` (see my comment on `componentWillReceiveProps`).

src/Dropdown/Dropdown.js
```diff
@@ -57,26 +132,74 @@ class Dropdown extends InputWithOptions {
...
+    if (!this.isControlledSupported() || !this.isSelectedIdControlled()) {
+      this.setState({
+        value: this.props.valueParser(option),
+        selectedId: option.id,
+      });
+    }
     super._onSelect(option);
   }
+
+  _onChange(event) {
```
yardnsm: Why is this required?

src/Dropdown/Dropdown.js
```diff
@@ -1,14 +1,55 @@
...
+import defaultTo from 'lodash/defaultTo';
+import differenceBy from 'lodash/differenceBy';
+import { validatorWithSideEffect } from '../utils/propTypes';
+import deprecationLog from '../utils/deprecationLog';
 import InputWithOptions from '../InputWithOptions/InputWithOptions';
 import styles from './Dropdown.scss';
+import PropTypes from 'prop-types';
+
+const NO_SELECTED_ID = null;
+const UPGRADE_PROP_NAME = 'upgrade';
```
yardnsm: Why?

src/Dropdown/Dropdown.js
```diff
@@ -57,26 +132,74 @@ class Dropdown extends InputWithOptions {
...
+          `'selectedId' and 'initiallySelectedId' cannot both be used at the same time.`,
+        );
+      }
+    },
+  ),
+  initiallySelectedId: validatorWithSideEffect(
+    InputWithOptions.propTypes.selectedId,
+    (props, propName) => {
+      if (props[propName] && !props[UPGRADE_PROP_NAME]) {
+        deprecationLog(
```
yardnsm: Maybe it'll be better to return an `Error`.

stories/components/Dropdown/Dropdown.story.js
```diff
@@ -60,6 +60,7 @@ export default {
     options,
     onSelect: option =>
       console.log(`option selected. ${JSON.stringify(option)}.`),
+    upgrade: true,
```
yardnsm: I would like to prevent users from toggling this option. Maybe you can move it to `exampleProps` and use something like:

```js
    upgrade: [
		{ label: 'true', value: true },
	],
```

src/Dropdown/Dropdown.spec.js
```diff
@@ -89,6 +89,150 @@ describe('Dropdown', () => {
     expect(driver.isReadOnly()).toBeTruthy();
   });
 
+  describe('Upgrade', () => {
+    const NewDropdown = props => <Dropdown {...props} upgrade />;
```
yardnsm: It seems like we're testing a different component, so I would to the following:

```js
const dropdownProps = {
	upgrade: true,
	options: getOptions(),
};

// In you test
const { inputDriver, dropdownLayoutDriver } = createDriver(
	<Dropdown {...dropdownProps} options={createOverridenOption()} />
);
```

src/Dropdown/Dropdown.spec.js
```diff
@@ -89,6 +89,150 @@ describe('Dropdown', () => {
...
+
+        expect(dropdownLayoutDriver.optionById(0).isSelected()).toBeTruthy();
+        expect(inputDriver.getValue()).toBe('Option 1');
+        rerender(<NewDropdown options={[{ id: 1, value: 'Option 2' }]} />);
+
+        expect(
+          dropdownLayoutDriver
+            .optionDrivers()
+            .some(option => option.isSelected()),
+        ).toBeFalsy();
```
yardnsm: You're only testing a single option, which makes this iteration redundant IMO. Use instead

```js
expect(dropdownLayoutDriver.optionById(1).isSelected()).toBeFalsy();
```

src/DropdownLayout/DropdownLayout.driver.js
```diff
@@ -78,11 +80,18 @@ const dropdownLayoutDriverFactory = ({ element }) => {
...
       doIfOptionExists(position, () => optionAt(position).textContent),
-    /** @deprecated Use optionAtDriver*/
+    /** @deprecated Use optionDriver*/
     optionAt,
     /** Get option driver given an option index */
-    optionDriver: position =>
-      doIfOptionExists(position, () => optionDriver(optionAt(position))),
+    optionDriver: createOptionDriver,
+    /** Get an array of all option drivers */
+    optionDrivers: () => {
```
yardnsm: Why?

src/Dropdown/Dropdown.js
hepiyellow: ok

src/Dropdown/Dropdown.js
hepiyellow: I like it

src/Dropdown/Dropdown.js
hepiyellow: Let's do a research and add to Contributing Best practices, a white list or/and black-list of `lodash` functions. Until then, "To each his own"

src/Dropdown/Dropdown.js
hepiyellow: This is a comparator of 2 arrays of options, not 2 options.

src/Dropdown/Dropdown.js
argshook: please, such "helpers" are useless in these scenarios, just use the language:

`isUndefined` === `typeof thing === 'undefined'`

`defaultTo` === `thing || defaultThing`

src/Dropdown/Dropdown.js
hepiyellow: No. `isSelectedIdControlled` checks if it `selectedId` was passed (hence Controlled).

`isControlledSupported` means that we are in upgrade mode. So we can change the behaviour of Uncontrolled mode as well.

src/Dropdown/Dropdown.js
hepiyellow: Fixed

stories/components/Dropdown/Dropdown.story.js
hepiyellow: cool

src/Dropdown/Dropdown.js
argshook: no, research, contribution list for such things are waste of time. it's as simple as "if problem trivial then use language; otherwise use helper" that's it. `isUndefined` is trivial.

src/Dropdown/Dropdown.js
```diff
@@ -1,14 +1,54 @@
...
 import isUndefined from 'lodash/isUndefined';
+import defaultTo from 'lodash/defaultTo';
+import differenceBy from 'lodash/differenceBy';
+import { allValidators } from '../utils/propTypes';
 import InputWithOptions from '../InputWithOptions/InputWithOptions';
 import styles from './Dropdown.scss';
+import PropTypes from 'prop-types';
+
+const NO_SELECTED_ID = null;
+export const UPGRADE_PROP_NAME = 'upgrade';
```
argshook: i wouldn't do this, just write regular `upgrade` where necessary.

usage of `UPGRADE_PROP_NAME` is inconsistent, so either use it always or never.

also, autodocs wont parse `[UPGRADE_PROP_NAME]` on line 166

---

on a side note, i wish there was no `upgrade` prop at all, but maybe
that's not possible.

i wish, because feature toggles in wix-style-react were never a good
idea, we always suffered one way or another (e.g. `new-icons`, how long will they stay new?)

src/Dropdown/Dropdown.js
```diff
@@ -1,14 +1,54 @@
...
 
 class Dropdown extends InputWithOptions {
   constructor(props) {
     super(props);
-    this.update(props, { isFirstTime: true });
+    if (props.upgrade) {
+      this.state = {
+        value: '',
+        selectedId: NO_SELECTED_ID,
+        ...Dropdown.getNextState(
```
argshook: i would put this to be first, because it seems like you want to override `value` and `selectedId` but if some day those will come from `Dropdown`, there could be issue here

src/Dropdown/Dropdown.spec.js
hepiyellow: In some tests we do `rerender`, we don't use `createDriver` always.
Furthermore, I am going to extract all the old tests to a separate file.

src/Dropdown/Dropdown.js
yardnsm: But you always check for `isSelectedIdControlled` **and** ``isControlledSupported`.

src/Dropdown/Dropdown.spec.js
yardnsm: > In some tests we do rerender, we don't use createDriver always.

Ok, but I was talking about dropping the `<NewDropdown/>` thing and instead use `<Dropdown {...dropdownProps} />`.

> Furthermore, I am going to extract all the old tests to a separate file.

Why? It's the same component.

src/Dropdown/Dropdown.spec.js
hepiyellow: This iteration is instead of asking `!isAnyOptionSelected`,
I don't want it to be dependent on the ids in the test.
Maybe option id=0 was not removed and it is still selected. So asking about option id=1 might be misleading.

src/DropdownLayout/DropdownLayout.driver.js
hepiyellow: To allow iterating over all options. It would make `optionsLength` deprecated in favor of `optionDrivers().length`.

src/Dropdown/Dropdown.js
hepiyellow: Will change to simple `upgrade`.
About feature toggles. I also wish it, but we are to "heavy" with major versions.
I am trying to figure a way to make it as painless as possible, to do the "cleanup" later.
Like:
- separating the spec tests to different files (e.g `Dropdown.deprecated.spec.js`)
- creating `deprecatedXXXXX` functions, instead of having feature toggles inside functions' logic.

We can also do like the new Button, the has totally 2 separate components.

Would you like to have all "Feature toggles" (API changes) as totaly new components (that is, if not a major version)?

src/Dropdown/Dropdown.js
hepiyellow: I want the value to have a default value of `''`.
And `getNextState` can either return a new value or NOT.

src/Dropdown/Dropdown.js
argshook: > Would you like to have all "Feature toggles" (API changes) as totaly new components (that is, if not a major version)?

ideally there should be a way to opt-in to new api without dedicated prop, maybe a new usage of prop
or new value for prop (something like what you did in MultiSelect). But i understand it's not always
possible, just in general saying i don't like feature toggles in components:)

> as totaly new components

that's sometimes possible, for example new `Layout` instead of `Grid`
but that's not the case with Dropdown.

src/Dropdown/Dropdown.spec.js
hepiyellow: I'd rather see it as a pseudo separate component.
This whole `upgrade` feature toggle, is actually a cheap Major version bump.
The idea is to make it as easy as possible to cleanup this mess when we DO bump the major version.
So I separated all tests which we should remove after Major version bump, to a separate file.
And try to use new methods with `deprecatedXXX` names in the component code, to emphasise that we should NOT touch deprecated code, or do fixes in deprecated code.

Furthermore, about your idea to do `<Dropdown {...dropdownProps} />`, I don't want someone to overwrite and pass `upgrade=false`. So if you use the `NewDropdown`, you can NOT make `upgrade=false`.

src/Dropdown/Dropdown.js
hepiyellow: But I don't do :  `isSelectedIdControlled` && `isControlledSupported`

PR #2724, opened 2019-01-08
WIP - <SelectableList/> - new component

mykas:
#### A wrapper that makes a list clickable.

```jsx
import React from 'react';
import Thumbnail from 'wix-style-react/Thumbnail';

class SelectableExample extends React.Component {
  _onSelect = (event, item) => {};

  _onDeselect = (event, item) => {};

  render() {
    return (
      <SelectableList
        limit={2}
        onSelect={this._onSelect}
        onDeselect={this._onDeselect}
        trigger="selected"
      >
        <Thumbnail src="small" />
        <Thumbnail src="medium" />
        <Thumbnail src="large" />
      </SelectableList>
    );
  }
}
```

### DIFF
--- a/src/SelectableList/SelectableList.js
+++ b/src/SelectableList/SelectableList.js
@@ -1,0 +1,60 @@
+import React, { Children } from 'react';
+import { generateID } from '../utils/generateId';
+
+const mapChildrenToId = children =>
+  Children.toArray(children).reduce(
+    (a, c) => ({
+      ...a,
+      [generateID()]: c,
+    }),
+    {},
+  );
+
+class SelectableList extends React.Component {
+  static displayName = 'SelectableList';
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      items: mapChildrenToId(props.children),
+      selected: [],
+    };
+  }
+  _change = (e, key) => {
+    const { selected, items } = this.state;
+
+    if (selected.includes(key)) {
+      this.setState(
+        state => ({
+          selected: state.selected.filter(value => value !== key),
+        }),
+        this.props.onDeselect(items[key]),
+      );
+    } else {
+      this.setState(
+        state => ({
+          selected: [...state.selected, key],
+        }),
+        this.props.onSelect(items[key]),
+      );
+    }
+  };
+
+  render() {
+    const { dataHook } = this.props;
+    const { items, selected } = this.state;
+    return (
+      <div data-hook={dataHook}>
+        {Object.entries(items).map(([key, value]) =>
+          React.cloneElement(value, {
+            key,
+            onChange: e => this._change(e, key),
+            checked: selected.includes(key),
+          }),
+        )}
+      </div>
+    );
+  }
+}
+
+export default SelectableList;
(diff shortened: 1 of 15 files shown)


src/SelectableList/SelectableList.js
```diff
@@ -0,0 +1,60 @@
...
+  static displayName = 'SelectableList';
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      items: mapChildrenToId(props.children),
+      selected: [],
+    };
+  }
+  _change = (e, key) => {
```
argshook: let `e` be `event` cause what if i think `e` stands for `error`

src/SelectableList/SelectableList.spec.js
```diff
@@ -0,0 +1,90 @@
...
+
+import Checkbox from '../Checkbox';
+import { checkboxTestkitFactory } from '../../testkit';
+
+describe('SelectableList', () => {
+  const createDriver = createUniDriverFactory(
+    selectableListPrivateDriverFactory,
+  );
+
+  it('should render', async () => {
```
argshook: no need for this, smoke test should cover it. `should render` is more important in e2e with screenshot

src/ColorPicker/ColorPicker.js
```diff
@@ -66,12 +66,7 @@ export default class ColorPicker extends WixComponent {
   }
 
   render() {
-    const {
```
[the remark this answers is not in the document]
mykas: Yes, I will remove these.

src/SelectableList/SelectableList.spec.js
mykas: ok

src/SelectableList/SelectableList.js
mykas: ok

hepiyellow:
## Where is the `README.DESIGN-SPEC.md` ?

First of all, there has been an effort in core3 to develop a super-generic selection list.
Here is the [PR](https://github.com/wix/wix-ui/pull/802) and the [design-spec](https://github.com/wix/wix-ui/blob/d3fcbdcbfa390cfc3e6544f7e86088730137a10c/packages/wix-ui-core/src/components/selection-list/spec.md)

We don't have to build such a full solution, but let's have a design spec here.
Decide on the scope (what features we want now, and which ones could be added later), and have all use-cases in mind, so we can find an API which is forward compatible.

argshook:
## it does not exist

@hepiyellow thanks for your input. this component is a WIP and part of effort together with Promote team to help them create Shoutout Composer 2.0

we don't do it how core3 did because:

1. we want to move gradually forward, not get into endless discussions to make everyone happy about the API
2. we don't want to solve everything immediately
3. we don't want to hold Promote team
4. we have different requirements
5. we want to make flexible reusable component, not something that knows about specifics (like `typeAheadValue` or something)

mykas:
Dismissed component development. Component is not as useful as just writing your own handlers.  User use case example: https://gist.github.com/mykas/3abc8eff86d1fed50e9d259b108a010e

PR #2726, opened 2019-01-09
<Box/> - create a layout component that helps aligning content

nitayneeman:
This component is implemented because of https://github.com/wix-private/wsr-issues/issues/359


src/Input/Input.scss
```diff
@@ -24,7 +24,7 @@ $material-gap: 8px;
 .input {
   flex-grow: 1;
   flex-shrink: 1;
-  min-width: 20px;
+  width: 100%;
```
nitayneeman: As part of this task, I refactored TableToolbar to use the box, but then I got the following issue:
image

It turns out that the Input should specify a defined width in order to not exceed the parent flex container. I think it doesn't matter because it already gets a full width from its parent (`input-wrapper`).

src/Box/Box.js
```diff
@@ -0,0 +1,137 @@
...
+  const rootClassNames = classNames(styles.root, {
+    [styles.inline]: inline,
+  });
+  const rootStyles = {
+    // Alignment
+    justifyContent: horizontalAlignmentValues[align],
+    alignItems: verticalAlignmentValues[verticalAlign],
+
+    // Spacing
+    padding: formatSpacingValue(padding),
```
nitayneeman: Perhaps it worths accessing the props directly, instead of using destructuring assignment for each prop, for example, using `props.padding` here. In that way, we'll not have to maintain the props in two places, which means, less code :)

Either way, we need the prop variables we extracted in only one place (`rootStyles `).

src/Box/Box.js
```diff
@@ -0,0 +1,137 @@
...
+Box.propTypes = {
+  /** Any element to be rendered inside */
+  children: PropTypes.node.isRequired,
+  /** Defines if the box behaves as an inline element */
+  inline: PropTypes.bool,
+  /** Defines how the children are aligned according to the X axis */
+  align: PropTypes.string,
+  /** Defines how the children are aligned according to the Y axis */
+  verticalAlign: PropTypes.string,
+  /** Sets padding on all sides */
```
nitayneeman: Part of these comments seems to be completely unnecessary. But I kept this in order to be consistent.

src/Input/Input.scss
argshook: it'd be better to not touch unrelated components in one PR. if there's a fix to be done in Input, let that be separate PR. easier to review, merge, revert

src/Box/Box.js
argshook: if comment does not explain anything, don't add it, not worth it :)

src/Box/Box.js
```diff
@@ -0,0 +1,137 @@
...
+  isFinite(value) ? value * spacingUnit : `${value}`;
+
+/**
+ * 📦 Box is a wrapper component that provides a way to align, space and resize its children - easily and straightforwardly.
+ */
+const Box = props => {
+  const {
+    dataHook,
+    children,
+    inline,
```
argshook: most of these props are awfully similar to `styles={{ ...inline styles }}` except that they 
1. introduce new names, like align -> justifyContent, verticalAlign -> alignItems. they also have new values, like left and not flex-start
2. since it's so similar to inline styles, why not use inline styles then? Sure theres `spacingUnit` but that can go within inline styles too
3. this component seems to do something that `Layout` can already do

what would be the fate of `VBox` and `HBox`? when they were introduced i suggested to do just `Box` instead, but lost the debate. But here we are adding `Box` now. @shlomitc

src/Box/Box.e2e.js
```diff
@@ -0,0 +1,73 @@
...
+const storyUrl = createStoryUrl({
+  ...baseStorySettings,
+  withExamples: false,
+});
+const storyUrlWithExamples = createStoryUrl({
+  ...baseStorySettings,
+  withExamples: true,
+});
+
+const eyes = eyesItInstance();
```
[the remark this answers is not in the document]
hepiyellow: Yes. There is a defaultConfig.

src/Box/Box.js
nitayneeman: Absolutely agree

src/Box/Box.e2e.js
```diff
@@ -0,0 +1,73 @@
...
+const byDataHook = dataHook => $(`[data-hook="${dataHook}"]`);
+
+describe('Box', () => {
+  describe('AutoExample', () => {
+    beforeAll(async () => await browser.get(storyUrl));
+
+    afterEach(async () => await autoExampleDriver.remount());
+
+    eyes.it('should be rendered', () => {
+      expect(true).toBeTruthy();
```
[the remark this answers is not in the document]
nitayneeman: The whole PR is WIP :)

src/Box/Box.e2e.js
```diff
@@ -0,0 +1,73 @@
...
+const baseStorySettings = {
+  kind: storySettings.category,
+  story: storySettings.storyName,
+};
+
+const storyUrl = createStoryUrl({
+  ...baseStorySettings,
+  withExamples: false,
+});
+const storyUrlWithExamples = createStoryUrl({
```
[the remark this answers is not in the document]
nitayneeman: Actually I was inspired by other spec files (like in `FormField` and `Text`). I thought you did it in order to separate the screenshots so it'll be easy to visual diff in Applitools. Isn't it?

src/Box/Box.e2e.js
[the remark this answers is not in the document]
nitayneeman: Here I was also inspired by other spec files. I saw it in more than 20 files so I assumed that's your convention...

src/Box/Box.e2e.js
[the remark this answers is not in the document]
argshook: whoops, somehow i missed the label :)

src/Box/Box.scss
```diff
@@ -0,0 +1,7 @@
+.root {
+  display: flex;
+
+  &.inline {
```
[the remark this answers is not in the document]
nitayneeman: In this case, it doesn't really matter - maybe it's neater (but that's a matter of personal taste).

src/Box/README.md
```diff
@@ -0,0 +1,3 @@
+# `<Box/>`
+
+📦 Box is a wrapper component that provides a way to align, space and resize its children - easily and straightforwardly.
```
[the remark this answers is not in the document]
nitayneeman: I'd remove the comment above the class instead of the README file because the README's content appears on the generated story page, which's a nice benefit. BTW in the `Avatar` component includes both of cases (here as well I thought that's your convention)

src/Box/Box.js
```diff
@@ -0,0 +1,137 @@
...
+  /** Sets the maximum height of the box */
+  maxHeight: PropTypes.string,
+  /** Sets the height of the box */
+  height: PropTypes.string,
+  /** Applied as data-hook HTML attribute that can be used in the tests */
+  dataHook: PropTypes.string,
+};
+
+Box.defaultProps = {
+  inline: false,
```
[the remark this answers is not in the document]
nitayneeman: @shlomitc 	 What's your opinion?

src/Box/Box.e2e.js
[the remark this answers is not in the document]
argshook: that's not big issue at all, was just wondering why do it this way. If component is small enough and examples don't interfere with what you test, then i wouldn't bother with splitting.

applitools catches all diffs good, personally i don't care if diffs are in separate screenshots or multiple in one

src/Box/README.md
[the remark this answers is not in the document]
argshook: comment above code should also appear in storybook. does it not show up for you?

src/Box/Box.js
nitayneeman: Regarding 1 - the motivation was to not be tightly-coupled to the flex values by mapping them with semantic prop names and values.

Regarding 2 - if I understood correctly, that's an option, but then we'd accept every inline property which's passed and probably we've to filter it out somehow only for the accepted properties. The question is, would we want to enable using any valid style property, without limiting them specifically? @shlomitc 	

Regarding 3 - good to know, I'm not familiar with the `Layout`. Could that component assist easily with spacing issues as well?

I didn't know you had a debate about this subject... Maybe it wasn't prioritized in the past? @sholim

src/Box/README.md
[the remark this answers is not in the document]
nitayneeman: I've just checked it once again, but it seems not to be working. Should we open a bug? :P

src/Box/Box.scss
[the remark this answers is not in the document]
argshook: if it doesn't matter, why nest then?

1. nesting makes it slightly harder to read, cause you need to know the `&` operator. sure, this usage is easy enough, but still you need to know it
2. nesting generates longer classnames. this is `.root.inline` as opposed to just `.inline`

this is very minor but im pointing anyway, cause nesting in sass is way overrated

src/Box/README.md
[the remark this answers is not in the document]
argshook: you're right, it's a bug. description from source should work but it gone missing at some point. I'll patch it up.

src/Box/Box.js
argshook: > Regarding 1 - the motivation was to not be tightly-coupled to the flex values by mapping them with semantic prop names and values.

why not? css is well known among our users, but our custom properties
and values will be new. moreover, things like `flex-start` and
`flex-end` are named like that for a reason: consider what is left and
right in RTL? we also use `start` and `end` in popovers, iirc.

also, setting `padding={3}` but getting `padding: '18px'` would be hella confusing. I understand UX works in divisions of 6, but that's not how devs work. they see sketch, see the gap, say, `36` and will try `padding={36}`, not `padding={6}`. but then, `width={6}` would be actually `6px` and not `36px`. confussion intensifies

> Regarding 2 - if I understood correctly, that's an option, but then we'd accept every inline property which's passed and probably we've to filter it out somehow only for the accepted properties. The question is, would we want to enable using any valid style property, without limiting them specifically? @shlomitc

also, why not? why should we filter anything? if user expects css properties to work, then let them use it.
but then i'd argue that `Box` is basically `div`, except that it's not
`<div style={{something: value}}/>` but `<Box something="value"/>`. Is that really useful?

> Regarding 3 - good to know, I'm not familiar with the Layout. Could that component assist easily with spacing issues as well?

`Layout` has `gap` prop which lets define even spacing between cells.
it's based on css grid, you can play with component here https://wix-wix-style-react.surge.sh/?selectedKind=Components%2FLayout&selectedStory=Usage&full=0&addons=0&stories=1&panelRight=0

---

all in all it looks like `Box` is trying to be a combination of `<Cell/>` (from Layout) and `div` with custom properties. Not sure how useful is that

stories/Box/ExampleBasic.js
```diff
@@ -0,0 +1,61 @@
...
+          align="center"
+          verticalAlign="middle"
+        >
+          Green
+        </Box>
+      </Box>
+    );
+  }
+
+  // This is a workaround in order to remove the parent, so that the children will be contained directly inside the storybook preview container.
```
argshook: @nitayneeman wait what why? what's the issue?

nitayneeman: The example is aimed to demonstrate how the Box behaves and effects on its children. It's pretty cool when you play and see it's reflected in the preview. 

In React 15.x, AFAIK, it's not possible to render directly a group of elements without an extra wrapper DOM node. React 16.x solves that issue with the syntax of `React.Fragment`.

Anyway, Shlomi helped me to find a better solution which doesn't add an extra parent element, so this hacky workaround is no longer needed :)

BTW, I'm still working on this PR - but once in a while, I push.

stories/Box/index.story.js
```diff
@@ -0,0 +1,99 @@
...
+import Box from '../../src/Box';
+import ExampleMultipleBoxes from './ExampleMultipleBoxes';
+import ExampleMultipleBoxesRaw from '!raw-loader!./ExampleMultipleBoxes';
+
+const childrenExamples = [
+  {
+    label: 'Text',
+    value: (
+      <div style={{ textAlign: 'center' }}>
+        <Text tagName="div" size="medium" style={{ color: 'inherit' }}>
```
shlomitc: `tagName` and `style` should be removed

stories/Box/ExampleMultipleBoxes.js
```diff
@@ -0,0 +1,77 @@
...
+import Button from 'wix-style-react/Button';
+import IconButton from 'wix-style-react/IconButton';
+import Image from 'wix-style-react/new-icons/Image';
+import More from 'wix-style-react/new-icons/More';
+import Hint from 'wix-style-react/new-icons/Hint';
+
+export default () => (
+  <Box
+    minHeight={200}
+    padding={1.5}
```
shlomitc: can this be just a single unit, it's bummer to see this half unit already as the first example

stories/Box/ExampleMultipleBoxes.js
```diff
@@ -0,0 +1,77 @@
...
+      >
+        <Image/>
+      </Box>
+    </Box>
+    <Card stretchVertically>
+      <Card.Content>
+        <Box
+          direction="vertical"
+          verticalAlign="space-between"
+          minHeight={145}
```
shlomitc: shouldn't it just spread to the container height instead of specifying a minHeight?

src/Box/Box.js
argshook: disagree, i think such abstractions are not only useless but will bring confusion for us and consumers. E.g. why `padding=1` is multiplied, but `width=1` is not?

> Setting unit-less values that are transformed into 6px units is actually good in my opinion and similar to how the Grid component works.

unitless values already can be passed through `style`, react adds `px` to them. moreover, if this is similar to how Grid works, why make new component then? 

> Yes, it is possible just to use a div with styles/classnames, but we do want to give this kind of abstractions. we will open both className and style for all regular values.

why do we want to abstract style properties? I think developers are clever enough to add `padding: 30`, when they see that padding is 30px in zeplin. furthermore, developers will see padding 30px in zeplin, but will actually need to use `padding={5}` or something, how is that helping?

if we'll open `className` and `style` then what's gonna happen with `<Box padding={2} style={{ padding: 30}}/>`?

> if a gap will be requested, I will happily like to add it.

Bb..but `Layout` already has it

---

I would understand if this were onboarding task to familiarize with the library. But i wouldn't release such component for the sake of releasing it.

currently for layout users can:

1. use just `div`s (native API)
1. use `VBox` (custom API)
1. use `HBox` (custom API)
1. use `Grid` (custom API)
1. use `Layout` (custom API)

i thought the idea was to get `Layout` do it all, so that it's one stop shop. We already have 5 stop shops and want to introduce another one.

feels like we focus with quantity over quality

https://imgs.xkcd.com/comics/standards.png

:man_shrugging:

src/Box/README.md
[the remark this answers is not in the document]
nitayneeman: I just opened a bug in order to not miss it: https://github.com/wix/wix-ui/issues/1077.

scripts/generate-testkit-exports/testkit-definitions.js
```diff
@@ -322,4 +322,9 @@ module.exports = {
     testkitPath: '../src/BaseComponents/TextLinkLayout/TextLinkLayout.driver',
     skipSanityTest,
   },
+
+  Box: {
+    skipSanityTest,
+    noTestkit,
```
yardnsm: Why?

src/Box/Box.scss
```diff
@@ -0,0 +1,69 @@
+.root {
```
yardnsm: Why not stylable? It could spare some the inline styles calculations, such as:

```js
// Styling
color: colors[color] || color,
backgroundColor: colors[backgroundColor] || backgroundColor,
```

src/Box/Box.spec.js
```diff
@@ -0,0 +1,61 @@
...
+describe('Box', () => {
+  describe('API', () => {
+    it('should pass the inline style', async () => {
+      const expectedOpacity = '0.5';
+      const children = <span>Children</span>;
+      const wrapper = mount(
+        <Box style={{ opacity: expectedOpacity }}>{children}</Box>,
+      );
+
+      expect(getStyle(wrapper)).toHaveProperty('opacity', expectedOpacity);
```
yardnsm: I'm missing a driver 🙃

src/Box/Box.js
```diff
@@ -0,0 +1,181 @@
...
+  /** Sets the width of the box (pixels) */
+  width: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+  /** Sets the minimum height of the box (pixels) */
+  minHeight: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+  /** Sets the maximum height of the box (pixels) */
+  maxHeight: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+  /** Sets the height of the box (pixels) */
+  height: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+  /** Sets a text color */
+  color: PropTypes.oneOfType([PropTypes.string, PropTypes.oneOf(Object.keys(colors))]),
```
yardnsm: I might've missed something, but why the `<Box/>` component should handle text color? Is it really it's concern?

stories/Box/index.story.js
```diff
@@ -0,0 +1,89 @@
...
+
+import { storySettings } from './storySettings';
+import Box from '../../src/Box';
+import ExampleMultipleBoxes from './ExampleMultipleBoxes';
+import ExampleMultipleBoxesRaw from '!raw-loader!./ExampleMultipleBoxes';
+
+const childrenExamples = [
+  {
+    label: 'Text',
+    value: <span>Here is a simple text</span>,
```
yardnsm: Maybe use the `<Text/>` component to have the right font? Unless you'd wanted to demonstrate the text color set by `<Box/>`.

stories/Box/index.story.js
```diff
@@ -0,0 +1,89 @@
...
+        width={80}
+        height={80}
+        color="D70"
+        backgroundColor="B10"
+        borderRadius={3}
+        key={1}
+      >
+        Blue
+      </Box>,
+      <Box
```
yardnsm: Maybe add a `margin` to each `<Box/>` so the example would look a bit nicer?

stories/Box/index.story.js
```diff
@@ -0,0 +1,89 @@
...
+    dataHook: storySettings.dataHook,
+  }),
+
+  exampleProps: {
+    children: childrenExamples,
+  },
+
+  examples: (
+    <div>
+      <CodeExample title="Multiple Boxes" code={ExampleMultipleBoxesRaw}>
```
yardnsm: I suggest using a `<LiveCodeExample/>` for this

stories/Box/storySettings.js
```diff
@@ -0,0 +1,5 @@
+export const storySettings = {
+  category: 'Components',
```
yardnsm: You can import:

```js
import { Category } from '../storiesHierarchy';
```

And then:

```js
export const storySettings = {
  category: Category.COMPONENTS,
  // ...
};
```

scripts/generate-testkit-exports/testkit-definitions.js
nitayneeman: Because there's no driver :)

src/Box/Box.scss
nitayneeman: @shlomitc When we discussed about implementing Stylable, we reached a conclusion that it doesn't really benefit this component and even overcomplicates a bit.
But now, we've these use-cases so maybe it's a justification for implementing it?

src/Box/Box.spec.js
nitayneeman: I thought about implementing a driver, but @shlomitc convinced me to out of it. It's said that's a utility component, pretty much like the `<Grid/>`, which assists layouting other components so there's no point at exposing an external driver. Perhaps `getStyle` justifies reconsideration of a private driver?

src/Box/Box.js
nitayneeman: Box is aimed to ease basic styling stuff, basically for `backgroundColor` and `color` with built-in support of the color palette keys. When the child is plain text (probably not so often because of `Text`, but could happen sometimes) - it's pretty nice to be able to customize it using the palette.

stories/Box/index.story.js
nitayneeman: Actually I used `<Text/>` before and replaced it with a `span` exactly for that reason (demonstrating the color changes by `<Box/>`) :P

stories/Box/index.story.js
nitayneeman: Good idea, will be done

stories/Box/index.story.js
nitayneeman: I'll be done as well :)

stories/Box/storySettings.js
nitayneeman: I didn't know there's a constant for that, nice!

src/Box/Box.js
shlomitc: 1. There's no correct answer, just two opinions, probably there are even more opinions but this was the decision taken when creating the API.
2. When Box is out, HBox and VBox can be removed. Regarding Layout, if you think it can replace the Grid component, then I'll add it as a task to the backlog

argshook: 1. well then it's not a good decision.
2. again, the idea with `Layout` was to make it replace all other lay out mechanisms and be one thing for either small or big areas. if it doesn't handle all cases, i would focus on that and improve it, certainly not on implementing yet another way of laying things out.

but whatever, i see my points are not valid enough

Issue #2181, opened 2018-09-02
V6 Breaking Changes Suggestions


shlomitc:
Please edit this comment with things to go into wix-style-react@6.0.0

Branch name: [`version_6`](https://github.com/wix/wix-style-react/tree/version_6)

- [x] Build is green, ready to release

| # | scope | description | merged to v6 branch | in migration doc |
| --- | --- | --- | :---: | :--: |
| 1 | `<MessageBoxFixedHeaderFooter/>` | Remove component | 👍 #2734 | ✅ |
| 2 | `Typography` folder | Cleanup and remove old styles | 👍 #2747 | ✅ |
| 3 | `Card` | Remove deprecated headers (LinkHeader, ButtonHeader, CollapsedHeader) | 👍 #2735 | ✅ |
| 4 | Polyfills | Remove old polyfills and keep only the new `testkit/polyfills` folder | 👍 #2736 | ✅ |
| 5 | `Tag` | make `useOldMargins` default behavior to `true` and remove flag | 👍 #2737 | ✅ |
| 6 | `Text` | remove `bold` prop | 👍 #2738 | ✅ |
| 7 | `Table/TableToolbar` | remove the nested folder | 👍 #2739 | ✅ |
| 8 | `TextField`, `DropdownComposite`, `TextArea` | remove components | 👍 #2740 | ✅ |
| 9 | `InputWithOptions` testkit | remove pressing methods | 👍 #2741 | ✅ |
| 10 | `DataTable`, `Table` | remove `infoTooltip` props | 👍 #2742 | ✅ |
| ~~11~~ | ~~`Button`~~ | ~~remove deprecated component~~ | ~~no~~ |
| 12 | `MessageBoxMarketerialLayout` | make fixImagePosition true and remove prop | 👍 #2743 | ✅ |
| 13 | TPA | remove folder and relevant components | 👍 #2744 | ✅ |
| 14 | Backoffice | remove folder and relevant components (**Edit:** only `<Tooltip/>` and `<TextLink/>` were removed) | 👍 #2745 | ✅ |
|15 | Dropdown | Add separate Controlled / Uncontrolled modes (`upgrade` prop) | 👍  #2806 | ✅ |
| 16 | `<ButtonWithOptions/>`,` <IconWithOptions/>` | Remove components | 👍 #2746 | ✅ |
| 17 | `<DataTable/>` | Remove `newDesign` prop | 👍 #2764 | ✅ |
[...]

PR #2736, opened 2019-01-13
Remove old polyfills

yardnsm:
As suggested for version 6 in https://github.com/wix/wix-style-react/issues/2181.


yardnsm:
@shlomitc I'm planning to write to migration doc before our release

PR #2741, opened 2019-01-13
<InputWithOptions/> - remove key presses methods from the driver

yardnsm:


src/InputWithOptions/InputWithOptions.driver.js
```diff
@@ -35,36 +35,6 @@ const inputWithOptionsDriverFactory = ({ element }) => {
       const { element: optionElm } = dropdownLayoutDriver.optionById(optionId);
       return !!optionElm().querySelector(`[data-hook=highlighter-${optionId}]`);
     },
-
```
hepiyellow: Also need to remove the `createDeprecationMessageForKeyMethod` function

scripts/component-generator/templates/component/src/Component/Component.driver.js
```diff
@@ -5,15 +5,12 @@ export const {%componentName%}DriverFactory = base => {
     ...baseUniDriverFactory(base),
 
     /** Get the current count */
-    getCountText: async () =>
-      base.$('[data-hook="{%componentName%}-count"]').text(),
+    getCountText: async () => base.$('[data-hook="{%componentName%}-count"]').text(),
```
shlomitc: doesn't look right

yardnsm: Yeah, the this specific template file is changing some times because of the lint, I'll revert it.

PR #2742, opened 2019-01-13
<DataTable/> - remove the deprecated `infoTooltip` prop

yardnsm:


scripts/component-generator/templates/component/src/Component/Component.driver.js
```diff
@@ -5,15 +5,12 @@ export const {%componentName%}DriverFactory = base => {
     ...baseUniDriverFactory(base),
 
     /** Get the current count */
-    getCountText: async () =>
-      base.$('[data-hook="{%componentName%}-count"]').text(),
+    getCountText: async () => base.$('[data-hook="{%componentName%}-count"]').text(),
```
hepiyellow: When I run lint on this it gets broken into a new line.
I pushed a fix for this.
and it seems that you reverted it now.

Issue #2450, opened 2018-11-11
tree-shaking


lbelinsk:
**Background**
Today we already support tree shaking, but we have one problem which is a blocking lint rule which we get from yoshi.

**Issue**
We can't simply remove this lint rule due to the possible scenario where we have a project with advanced yoshi version, and old wix-style-react version. In that case a user will not have lint error although he will bring the entire bundle due to the unsupported tree shaking feature.

**Solution**
The lint rule should be smart enough to check for the wsr version and to apply the rule only if the version is >= 5.9.0.

- Tslint - https://github.com/wix/wix-ui/tree/master/packages/tslint-plugin-wix-style-react
- Eslint - https://github.com/wix/wix-ui/tree/master/packages/eslint-plugin-wix-style-react

PR #2756, opened 2019-01-14
Remove unused, nonexistent module dependencies. Fix ModalSelectorLayout export

yavorsky:
### What changed

Removed a few unused and nonexistent imports and fixed `ModalSelectorLayout` export for index.js.

### Why it changed

Currently, I'm working on ES exports (with https://github.com/wix/wix-style-react/issues/2450) and faced some warnings by `webpack` about nonexistent modules we want to import in some files.

Also, ModalSelectorLayout is broken for `index.js` file:
`import { ModalSelectorLayout } from 'wix-style-react' // Error, module not found`.
and here we are fixing it.

- [ ] Change is tested
- [ ] Change is documented
- [ ] Build is green

Before:
screen shot 2019-01-14 at 1 18 40 pm
After:
screen shot 2019-01-14 at 1 15 45 pm


src/Grid/index.js
```diff
@@ -6,5 +6,4 @@ export {
   AutoAdjustedColumns,
   AutoAdjustedColumns as AutoAdjustedRow,
   Col,
-  Card,
```
shlomitc: this is a bit of a breaking change, but it wasn't supposed to be here anyway, so I think it's ok to let people break and make a quick fix

PR #2764, opened 2019-01-15
<DataTable/> - remove the newDesign prop, make it the default

yardnsm:


src/DataTable/DataTable.driver.js
```diff
@@ -80,13 +80,7 @@ const dataTableDriverFactory = ({ element }) => {
     getRowDetailsText: index => getRowDetails(index).textContent,
     hasSortableTitle: index => !!getSortableTitle(index),
     hasInfoIcon: index => !!getTitleInfoIcon(index),
-    hasNewSortDescending: index => !!getSortableTitleArrowDesc(index),
```
hepiyellow: We didn't have a deprecationLog for this... )-:

PR #2767, opened 2019-01-15
Update DEPRECATION_GUIDE

yardnsm:


docs/internal/DEPRECATION_GUIDE.md
```diff
@@ -1,89 +1,119 @@
...
+
+You can see an example for this kind of logic [in the following link](https://github.com/wix/wix-style-react/blob/cc58ef88aff66efc5278ffc5adff115ca7264650/src/Button/index.js).
+
+In addition, you may also move the deprecated component's story to the `Deprecated` section and to
+the `stories/Deprecated` folder accordingly, and update the story with a deprecation message.
+
+When releasing a new major version, you can remove the deprecated component and the added feature
+flag accordingly.
+
+**Do not forget to find all of the internal usages and update them accordingly**
```
hepiyellow: What about the component's driver?
We need to export both old and new drivers? How do we do that?

docs/internal/DEPRECATION_GUIDE.md
```diff
@@ -1,89 +1,119 @@
...
+
+## Deprecating props values / types
```
hepiyellow: This should be only about deprecating a prop value.

- We should have another example about deprecating a prop (deleting a prop).
- And an example of the `upgrade` mechanism (Which is in discussion, since it has a serious issue)

yardnsm:
@hepiyellow 

> It is too "story-telling", and could be more TL;DR-ish, with examples

"Story telling" is arguably better than a TL;DR version. I wanted to explain to our contributors the reasons behind all of these steps.

But yeah, I definitely would like to _also_ have some short TL;DR sections for each case.

hepiyellow:
Just merge it and we'll improve it later.

PR #2777, opened 2019-01-16
<ButtonWithOptions/> - remove internal usages

yardnsm:
This PR removes internal usages of the deprecated `<ButtonWithOptions/>` component, that has been replaced by the `<DropdownBase/>` component.

### `<EditableSelector/>`

The `<EditableSelector/>` used <ButtonWithOptions.Button/>` internally, with no apparent reason. It was changed to the upgraded `<Button/>`.

### `<StatsWidget/>`

The `<StatsWidget/>` used it pretty heavily, as a way to add a "filter" dropdown. For example:

```jsx
<StatsWidget
	title="Let's see what's going on with your store"
	statistics={statistics}
>
	<StatsWidget.Filter {...ButtonWithOptionsProps}>
		<ButtonWithOptions.Button />
		<ButtonWithOptions.Option key={1} id={1}>
			Look at me!
		</ButtonWithOptions.Option>
	</StatsWidget.Filter>
</StatsWidget>
```

As we don't want to break consumers, a new component API has been introduced, `<StatsWidget.Filter`**`Button`**`/>`, while the `<StatsWidget.Filter/>` has been deprecated. Internally, it uses the `<DropdownBase/>` and its usages is as follows:

```jsx
<StatsWidget
	title="Let's see what's going on with your store"
	statistics={statistics}
>
	<StatsWidget.FilterButton
		dataHook="StatsWidgetFilter"
		initialSelectedId={1}
		options={[
			{ id: 1, value: 'Look at me!' },
		]}
		onSelect={selected => console.log(selected)}
	/>
</StatsWidget>
```

---

#### Issues with the new `<StatsWidget.FilterButton/>`

It uses the new `<TextButton/>` under the hood, which has different styles than the old `<Button/>` with the `'no-border'` theme.

`<TextButton/>` (idle, hover and spacing):

screen shot 2019-01-16 at 16 30 58
screen shot 2019-01-16 at 16 31 06
screen shot 2019-01-16 at 16 33 10

Old `<Button/>` (idle, hover and spacing):

screen shot 2019-01-16 at 16 31 37
screen shot 2019-01-16 at 16 31 43
screen shot 2019-01-16 at 16 33 26

I think we should add a new theme to the `<Button/>` (not the `<TextButton/>`, as the new theme should have a similar padding to the `<Button/>` and does behave like one) that would resemble the _4.1a Dropdown + No Border_.

@wuwa @LittleTal @milkyfruit wdyt regarding the theme?


src/StatsWidget/StatsWidget.js
```diff
@@ -123,7 +148,20 @@ class StatsWidget extends WixComponent {
   }
 }
 
+const FilterButton = props => (
+  <DropdownBase minWidth={160} {...props}>
```
hepiyellow: This means that `props` can also include ` children` prop. Which means that the consumer can override the whole `<TextButton/>` and do something else.

I am ok with that.

src/StatsWidget/StatsWidget.js
```diff
@@ -49,27 +57,44 @@ class StatsWidget extends WixComponent {
...
-      if (!propValue || propValue.length > 3) {
-        return new Error(
-          `StatsWidget: Invalid Prop children, maximum amount of filters are 3`,
-        );
-      }
+    /**
+     * Filters for statistics (will be shown in right top corner). Accepts an array of
+     * `<StatsWidget.FilterButton>` which accepts all `<DropdownBase/>` props.
+     */
+    children: allValidators(
```
hepiyellow: In order for AutoDocs to show proper type of the prop, I started doing it like this:
- In Component.propTypes I write the propper "new" PtopType validator.
- Later I use the `extendPropTypes` (from `src/utils/propTypes/`) to overwrite the prop validation with the `allValidators` validator.

src/StatsWidget/StatsWidget.js
```diff
@@ -49,27 +57,44 @@ class StatsWidget extends WixComponent {
...
+      (props, propName) => {
+        if (
+          props[propName] !== undefined &&
+          !Array.isArray(propName[propName]) &&
```
hepiyellow: So if `children` is an array of `<StatsWidget.Filter/>`, then we would NOT have a deprecationLog?

src/StatsWidget/StatsWidget.js
yardnsm: Well, yeah. The deprecated implementation with the `<ButtonWithOption/>` allowed the same thing, so I think it's okay.

src/StatsWidget/StatsWidget.js
```diff
@@ -49,27 +57,44 @@ class StatsWidget extends WixComponent {
...
+          props[propName].type !== StatsWidget.FilterButton
+        ) {
+          deprecationLog(deprecationMessage);
+        }
+      },
+      PropTypes.arrayOf((propValue, key) => {
+        if (!propValue || propValue.length > 3) {
+          return new Error(
+            `StatsWidget: Invalid Prop children, maximum amount of filters are 3`,
```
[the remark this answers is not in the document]
yardnsm: The limitations has also been there before I made the changes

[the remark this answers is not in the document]
yardnsm: And apparently we do need the prefix, the component name does not shows in the error (I checked)

milkyfruit:
@yardnsm @wuwa 
I think we can omit the padding area and not include it in a future theme.
Regarding the color style, as a temporary solution current styles would work too:
image

Off-course, the dropdown style looks more emphasising, I'm just afraid what would be the best name for the new theme, so it wouldn't alienate existing themes and would keep organised:
image

Maybe, smth like `DarkToStandard` would be clear and right? @wuwa what do you think?

PR #2779, opened 2019-01-16
chore(testkit-definitions): relocate file to be under `testkit` folder

argshook:


argshook:
closing, will do this together with generator update

PR #2786, opened 2019-01-20
Drivers cleanup/remove wrapper

hepiyellow:
Remove the `wrapper` argument from most component drivers.

In general, no driver functions were removed or changed.

### Calendar
- For some reason `wrapper was used, but it seems redundant.

### Tooltip
The tooltip driver searches for the tooltip content from the `document.body` OR the `wrapper`.
If we assume that consumers ALWAYS render to the body. Then we can remove the `wrapper. If not, then we need to keep it.


src/Tooltip/Tooltip.driver.js
```diff
@@ -9,27 +9,15 @@ const arrowDirectionToPlacement = {
...
+const tooltipDriverFactory = ({ element }) => {
   const getContentRoot = () => {
     const contentRootHook = element.getAttribute('data-content-hook');
     if (!contentRootHook) {
       throw new Error(
         `Tooltip.driver: contentRootHook attribute must exist on the Toolrip's root element`,
       );
     }
-    return bodyOrWrapper.querySelector(`[data-hook="${contentRootHook}"]`);
+    return document.body.querySelector(`[data-hook="${contentRootHook}"]`);
```
hepiyellow: There is no need to use the wrapper. querying from body is good enough because each tooltip content has it's own id.
The only difference now is that if the component was rendered NOT into the document body, then it wouldn't work now?

PR #2790, opened 2019-01-20
`<Page/>` - Add min/max width according to Grid

hepiyellow:
Resolves this private issue: https://github.com/wix-private/wsr-issues/issues/345

Added `min-width` and `max-width` to Page root element.

Technically, instead of adding a `<Grid/>` wrapper to the `<Page/>` (like I talked with @wuwa ),
I have now the min/max width defined in `<Page/>`, and the `<Grid/>` component imports those variable values.


hepiyellow:
I just realized, that I need to add the min/max width to the HEader content, NOT the whole page in order to be able to to this:

screen shot 2019-01-20 at 21 59 54

milkyfruit:
@hepiyellow by the current design guidelines, when business manager interface is with a sidebar, it doesn't support smaller browser viewports than 1188px. In this size 12 columns cards are 870px width.
Link to zeplin: https://zpl.io/2j46Wex

To keep it consistent, I suggest applying same rules in the `page` component. If someone decides to customise the `min-width` then should contact us and work on a solution together.

Currently without a specific example, I'm not sure whether that 1060px comes from a legacy project or someone wants to override our current solutions.

hepiyellow:
@milkyfruit I spoke with @wuwa and he agreed to make the `minWidth` a prop, with a default value of `870px`.

stories/Page/index.story.js
```diff
@@ -18,7 +17,7 @@ export default {
   componentPath: '../../src/Page',
 
   componentProps: {
-    children: [header(Breadcrumbs), tail, content(false)],
+    children: [header(), tail, content(false)],
```
yardnsm: Why removing the breadcrumbs?

stories/Page/PageExample.scss
```diff
@@ -15,11 +15,12 @@
 
 .pageContainer {
   @include PageConteiner;
-  margin: 0px 50px;
 }
 
 [data-hook="body-content"] {
-  @include PageConteiner;
+  height: calc(100vh - 16px); // -16px account for chrome default body margin
```
yardnsm: Isn't there a better way for this?

src/Page/Page.js
```diff
@@ -49,6 +50,8 @@ import {
 class Page extends WixComponent {
   static defaultProps = {
     gradientCoverTail: true,
+    minWidth: 798, // 894 - 48*2,- Should correspond to the Grid min-width
```
yardnsm: Use the constants from `constants.js` (for example `PAGE_SIDE_PADDING_PX`) and add some if needed, it'll be clearer.

stories/Page/index.story.js
hepiyellow: I didn't remove it. I added it as default for the `header()` function, since it is used in every occurence of calling `header()`.

stories/Page/PageExample.scss
hepiyellow: Better way for what?
BTW: This is only an example.
This `calc` this is not new, I just inlined it from the PageContainer.

hepiyellow:
@wuwa @milkyfruit @LittleTal 
Can someone review?
See thes Test stories:
- [Default min/max values](https://wix-wix-style-react-pr-2790.surge.sh/?selectedKind=Tests%2F2.%20Layout%2F2.5%20Page&selectedStory=5.%20Default%20%5Bmin%2Fmax%5D-width&full=0&addons=0&stories=1&panelRight=0)
- [Custom min/max values](https://wix-wix-style-react-pr-2790.surge.sh/?selectedKind=Tests%2F2.%20Layout%2F2.5%20Page&selectedStory=6.%20Custom%20%5Bmin%2Fmax%5D-width&full=0&addons=0&stories=1&panelRight=0)

Notice that currently the Grid `min-width` is 894 and not 870 like the specs.

Do you want to change it now? Or in separate PR?
Is it a breaking change for consumers?

hepiyellow:
@milkyfruit We discussed the 894/870 issue. The fix will be part of a separate PR, since it is a potentially breaking change.

src/Page/Page.e2e.js
```diff
@@ -29,6 +32,11 @@ describe('Page', () => {
   };
 
   const runTestCases = initTestConfig => {
+    const eyes = eyesItInstance({
+      enableSnapshotAtBrowserGet: false,
```
shlomitc: this is false by default

src/Grid/constants.js
```diff
@@ -0,0 +1,2 @@
+export const mainContainerMinWidthPx = 894;
```
shlomitc: instead of duplicating to a js file, just export it from scss like this:
https://github.com/wix/wix-style-react/blob/3b11ddbc4aefd839ac1dde239f1e91854d86ea9e/src/colors.scss#L77-L78

and then consume in code like this
https://github.com/wix/wix-style-react/blob/master/src/Box/Box.js#L111

src/Page/Page.js
```diff
@@ -359,8 +371,10 @@ Page.Tail = Tail;
 Page.propTypes = {
   /** Background image url of the header beackground */
   backgroundImageUrl: PropTypes.string,
-  /** Sets the max width of the header and the content */
+  /** Sets the max width of the content (Both in header and body) NOT including the page padding */
```
shlomitc: body or content?

PR #2792, opened 2019-01-21
<Box/> - add explanations with examples

nitayneeman:
This PR creates a dedicated entry section for descriptions (including explanations and examples).


stories/Box/propExplanations.js
```diff
@@ -0,0 +1,142 @@
...
+    Button
+  </Button>
+</Box>
+      `,
+  },
+  spacing: {
+    description: `
+### 👌 Spacing
+
+The component extends the values for \`padding\` and \`margin\`, by accepting the following:
```
nitayneeman: Instead of escaping we could export each description to a separate Markdown file, but that feels a bit overkill.

So, an additional little reason to consider MDX :P

argshook: no one is against MDX, it's just not trivial to incorporate into what we currently have

stories/Box/index.story.js
```diff
@@ -79,13 +99,64 @@ export default {
...
+
+        // Styling
+        description({ text: propExplanations.styling.description }),
+        propExplanationLiveExample(propExplanations.styling.example),
+
+        description({
+          title: 'Examples',
+        }),
+
+        liveCode({
```
nitayneeman: Is there any way to render this example when running the E2E using `liveCode`? @shlomitc @argshook

Previously, I used `examples` setting and accessed to the relevent storyUrl, with `withExamples` URL param, and that made the attached examples be shown.

But now, the example is not shown when accessing to the same URL:
http://localhost:6006/iframe.html?selectedKind=Components&selectedStory=Box&withExamples=

PR #2797, opened 2019-01-22
<SegmentedToggle/> - new component

mykas:
Shoutout composer initiative. 

Initial API.

```jsx
<SegmentedToggle defaultChecked="short" onChange={onChange}>
  <SegmentedToggle.Button value="short">Short Option</SegmentedToggle.Button>
  <SegmentedToggle.Button value="medium">Medium Option</SegmentedToggle.Button>
  <SegmentedToggle.Button value="long">Long Option</SegmentedToggle.Button>
<SegmentedToggle>
```


src/SegmentedToggle/ToggleButton/ToggleButton.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
```
argshook: would it ever be a valid case to use `ToggleButton` as standalone component not, and likely
`SegmentedToggle` is a compound component, meaning, ToggleButton should always be together with
`SegmentedToggle`.

if so, then i'd go with something like `SegmentedToggle.Button` and avoid
separate export/import.

src/SegmentedToggle/SegmentedToggle.js
```diff
@@ -0,0 +1,36 @@
...
+
+class SegmentedToggle extends React.Component {
+  static displayName = 'SegmentedToggle';
+
+  static propTypes = {
+    dataHook: PropTypes.string,
+  };
+
+  state = {
+    checked: this.props.initiallyChecked,
```
argshook: i understand this is a WIP but please don't forget about multiple options, cause this component
should support it.

it will likely look something very similar to what we already brainstormed with `SelectableList`

src/SegmentedToggle/ToggleButton/ToggleButton.js
mykas: Moved to `<SegmentedToggle/>` composite.

src/SegmentedToggle/SegmentedToggle.js
mykas: Changed to "defaultChecked" and agreed to limit component to support only single selection.

src/SegmentedToggle/ToggleButton/ToggleButton.spec.js
```diff
@@ -0,0 +1,53 @@
...
+
+describe('ToggleButton', () => {
+  const createDriver = createUniDriverFactory(toggleButtonPrivateDriverFactory);
+
+  it('should render', async () => {
+    const driver = createDriver(<ToggleButton />);
+    expect(await driver.exists()).toBeTruthy();
+  });
+
+  it('should be controlled', async () => {
```
argshook: controlled is that which is controlled from the outside, as in, has props passed
uncontrolled is that which is not controlled from outside, and controls it's own state itself

here, since state changes after click but outside world doesn't change, component is `uncontrolled`.

these are confusing as hell, i know

src/SegmentedToggle/ToggleButton/ToggleButton.js
```diff
@@ -0,0 +1,50 @@
...
+  >
+    {_addPrefix(prefixIcon)}
+    <Text ellipsis size="medium" weight="normal">
+      {children}
+    </Text>
+  </button>
+);
+
+ToggleButton.PropTypes = {
+  children: string,
```
argshook: it should be `node`, cause what if i want `<Translate key={my.translation.whatever}/>` as children?

rule of thumb is: when you think `string` for prop, you actually think `node`

src/SegmentedToggle/ToggleIcon/ToggleIcon.st.css
```diff
@@ -0,0 +1,83 @@
...
+  -st-default: Focusable;
+}
+
+:import {
+  -st-from: "wix-ui-core/dist/src/themes/backoffice/colors.st.css";
+  -st-named: F00, B20, B50, D55, D80;
+}
+
+
+.root {
```
argshook: how different is this file from `ToggleButton`? Maybe you can reuse stuff?

stories/SegmentedToggle/index.story.js
```diff
@@ -0,0 +1,95 @@
...
+      sections: [
+        description({
+          text: `🍫 SegmentedToggle is a view group containing typically three or more buttons that can be toggled on and off. These buttons visibly change to indicate whether an option is active or inactive.`,
+        }),
+        importExample({
+          source:
+            "import SegmentedToggle from 'wix-style-react/SegmentedToggle';",
+        }),
+        description({
+          text: `🙋 Component includes compound components: Button and Icon.  With following components user can build 3 types of SegmentedToggles:`,
```
argshook: you can use markdown here, so no need for multiple `descriptions` just for new lines

src/SegmentedToggle/SegmentedToggle.driver.private.js
```diff
@@ -0,0 +1,7 @@
+import { segmentedToggleDriverFactory as publicDriverFactory } from './SegmentedToggle.driver';
+
+export const segmentedTogglePrivateDriverFactory = base => {
```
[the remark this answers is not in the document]
mykas: Sure

src/SegmentedToggle/SegmentedToggle.spec.js
```diff
@@ -0,0 +1,59 @@
...
+  const Segmented = props => (
+    <SegmentedToggle {...props}>
+      <SegmentedToggle.Button dataHook={dataHook} value="short">
+        Short
+      </SegmentedToggle.Button>
+    </SegmentedToggle>
+  );
+
+  describe('`onClick` handler should return', () => {
+    it('when clicked with a mouse', async () => {
```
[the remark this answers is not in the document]
mykas: No need, children are based on `button`.

src/SegmentedToggle/ToggleButton/ToggleButton.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import { string, node, bool } from 'prop-types';
+import { withFocusable } from 'wix-ui-core/dist/src/hocs/Focusable/FocusableHOC';
+
+import styles from './ToggleButton.st.css';
+
+import Text from '../../Text';
+
+const _addPrefix = icon =>
```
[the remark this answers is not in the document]
mykas: true

src/SegmentedToggle/ToggleButton/ToggleButton.js
mykas: Oh yes yes..

src/SegmentedToggle/ToggleButton/ToggleButton.spec.js
mykas: But ToggleButton is controlled from outside and it doesn't have its own state.

src/SegmentedToggle/SegmentedToggle.spec.js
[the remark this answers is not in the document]
mykas: Added

src/SegmentedToggle/ToggleIcon/ToggleIcon.st.css
mykas: welll I could use stylable mixins :trollface:

mykas:
@LittleTal Hey, can you check https://wix-wix-style-react-pr-2797.surge.sh/?selectedKind=Components&selectedStory=SegmentedToggle&full=0&addons=0&stories=1&panelRight=0 ? and

mykas:
Also, check these test pages which shows behaviour: https://wix-wix-style-react-pr-2797.surge.sh/?selectedKind=Tests%2FComponents%2FSegmentedToggle&selectedStory=Text%20%26%20Prefix&full=0&addons=0&stories=1&panelRight=0 and http://localhost:6006/?selectedKind=Tests%2FComponents%2FSegmentedToggle&selectedStory=Icon&full=0&addons=0&stories=1&panelRight=0

mykas:
thanks! 

* Fixed.
* You gonna need to show me this case. 
* Fixed. 
* Fixed.
* I am not sure if this is possible right now, because text ellipsis is bound to text and in your defined case I need to extract ellipsis from Text component and detect when text gets too little space that it gets `...`. But you can open an issue or feature request for this.

LittleTal:
@mykas you should keep the hover effect, only without covering the toggle border while hovering

src/SegmentedToggle/SegmentedToggle.js
```diff
@@ -0,0 +1,65 @@
...
+    return (
+      <div
+        {...rest}
+        data-hook={dataHook}
+        {...styles('root', { disabled }, rest)}
+      >
+        {React.Children.map(children, (child, index) =>
+          React.cloneElement(child, {
+            disabled,
+            'data-click': `segmented-toggle-${index + 1}`,
```
argshook: use `data-hook`

src/SegmentedToggle/SegmentedToggle.js
```diff
@@ -0,0 +1,65 @@
...
+
+  state = {
+    selected: this.props.defaultSelected,
+  };
+
+  _onClick = evt => {
+    const { onClick } = this.props;
+    const { value } = evt.currentTarget;
+    this.setState({ selected: value }, () =>
+      onClick ? onClick(evt, value) : null,
```
argshook: why `null`?
```suggestion
      onClick && onClick(evt, value),
```

src/SegmentedToggle/ToggleIcon/ToggleIcon.js
```diff
@@ -0,0 +1,83 @@
...
+class ToggleIcon extends React.Component {
+  static displayName = 'SegmentedToggle.Icon';
+
+  state = {
+    shown: false,
+  };
+
+  open = () => {
+    const { disabled } = this.props;
+    this.setState({ shown: disabled ? false : true });
```
argshook:
```suggestion
    this.setState({ shown: !disabled });
```

src/SegmentedToggle/ToggleIcon/ToggleIcon.js
```diff
@@ -0,0 +1,83 @@
...
+import Text from '../../Text';
+import Popover from '../../Popover';
+
+import styles from './ToggleIcon.st.css';
+
+class ToggleIcon extends React.Component {
+  static displayName = 'SegmentedToggle.Icon';
+
+  state = {
+    shown: false,
```
argshook: what `shown`? maybe `tooltipVisible` or something?

src/SegmentedToggle/SegmentedToggle.js
mykas: This case is different. I need to change how data-hook is passed in order to be able to test it in SegmentedToggle with method selectChild. By using custom `data` attribute I don't need to reorder how ToggleIcon handles data-hook.

argshook: what i mean is that `ToggleButton` and `ToggleIcon` should both support dataHook and you should use
them to target elements in tests. That's exactly what dataHook is for, how is new name
(`data-click`) helping?

argshook: also, why `+ 1`?

mykas: ok, let me explain. ToggleIcon can be clicked only if I would use testkit which handles click handler. The reason for this is because Popover wraps the `button` component. So simple base.click() would not just work.

I want to expose for users selectChild() method in SegmentedToggle where they could click either ToggleButton or ToggleIcon. So here is the problem: how should I handle click method inside SegmentedToggle.driver if one can be clicked like this: base.click() and another one needs to be used with testkit? this data attribute way was the easiest for me to implement without changing the code.

P.S. `+1` seems to be easier for users to selectChild. It is documented in api docs too.

src/SegmentedToggle/ToggleIcon/ToggleIcon.js
mykas: its a copy paste from Yardens example in Popover :trollface:

src/SegmentedToggle/SegmentedToggle.driver.js
```diff
@@ -0,0 +1,11 @@
+export const segmentedToggleDriverFactory = base => {
+  return {
+    /* fulfilled if element in the DOM */
+    exists: async () => await base.exists(),
+    /* get the actual element */
+    element: async () => await base.getNative(),
+    /* selects child by given number (selection begins with 1) */
+    selectChild: async number =>
+      base.$(`[data-click="segmented-toggle-${number}"]`).click(),
```
argshook: https://github.com/wix/wix-style-react/pull/2797/files#r252577365

```suggestion
      base.$(`[data-hook="segmented-toggle-${number + 1}"]`).click(),
```

PR #2811, opened 2019-01-23
<TextLink/>, <TextLinkLayout/>, Backoffice <Button/> , <ButtonLayout/> - deprecated components

yardnsm:
This is a big one. I suggest to take a look at each commit separately.

-----

	* Moved `<TextLink/>` and `<TextLinkLayout/>` from `BaseComponents` to `Deprecated`
	* Moved the old backoffice `<Button/>` and `<ButtonLayout/>` to `Deprecated`
	* Removed stories for the above-mentioned components, as well as e2e tests and protractor drivers
	* Removed the `export`s for above-mentioned components, as well for the testkits
	* Some minor cleanups
	* Removed the `upgrade` prop for the `<Button/>` usages.

These components are now being used internally. There are some uses of `<TextLink/>` that we can remove, but it'll take some time so I suggest adding it to the backlog.


testkit/index.js
```diff
@@ -22,18 +21,6 @@ export {
   circularProgressBarTestkitFactory,
 } from 'wix-ui-backoffice/dist/src/testkit';
 
-import buttonDriverFactory from '../src/Backoffice/Button/Button.driver';
```
hepiyellow: We need to still keep providing a sync driver for button. I talked with shlomi about it, and we don't want to stop providing sync drivers yet. but they should have a deprecationLog if there is an async driver alternative.
For the new button.
So we need to write it.
@yardnsm you or me?
And run the Button tests with both drivers.
That could be tricky since the Button tests are in wix-ui-core.

yardnsm: Why? I though we'll also provide a sync driver after a potentially breaking change, and then remove it in the next major version.

We're breaking consumers anyway with the removal of the old button, so why not break them more? 🙃

src/GoogleAddressInputWithLabel/GoogleAddressInputWithLabel.driver.js
```diff
@@ -1,6 +1,6 @@
 import inputAreaWithLabelCompositeDriverFactory from '../Composite/InputAreaWithLabelComposite/InputAreaWithLabelComposite.driver';
 
-const GoogleAddressInputWithLabelDriverFactory = ({ element }) => {
+const googleAddressInputWithLabelDriverFactory = ({ element }) => {
```
hepiyellow: So this driver was broken?

yardnsm: It's because of a change made in the `version_6` branch, I fixed it here.

src/TextLink/TextLink.driver.js
```diff
@@ -1,6 +1,7 @@
 import ReactTestUtils from 'react-dom/test-utils';
-import { ThemeOptions } from '../BaseComponents/TextLinkLayout/TextLinkLayout';
-import textLinkLayoutDriverFactory from '../BaseComponents/TextLinkLayout/TextLinkLayout.driver';
+
+import { ThemeOptions } from '../Deprecated/TextLinkLayout/TextLinkLayout';
+import textLinkLayoutDriverFactory from '../Deprecated/TextLinkLayout/TextLinkLayout.driver';
```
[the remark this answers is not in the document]
yardnsm: This driver is only used internally, since the `<TextLink/>` is an internal component now.

src/TextLink/TextLink.driver.js
```diff
@@ -1,6 +1,7 @@
 import ReactTestUtils from 'react-dom/test-utils';
-import { ThemeOptions } from '../BaseComponents/TextLinkLayout/TextLinkLayout';
-import textLinkLayoutDriverFactory from '../BaseComponents/TextLinkLayout/TextLinkLayout.driver';
+
+import { ThemeOptions } from '../Deprecated/TextLinkLayout/TextLinkLayout';
```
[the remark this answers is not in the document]
yardnsm: This driver is only used internally, since the `<TextLink/>` is an internal component now.

PR #2815, opened 2019-01-24
<FormField/> - apply min-height on inline label

nitayneeman:
This PR implements https://github.com/wix-private/wsr-issues/issues/374


src/FormField/FormField.js
```diff
@@ -178,7 +180,11 @@ class FormField extends React.Component {
         })}
       >
         {label && labelPlacement === labelPlacements.top && (
-          <div className={styles.label}>
+          <div
+            className={classnames(styles.label, {
+              [styles.minLabelHeight]: !children,
```
shlomitc: Please add a test for this logic

PR #2826, opened 2019-01-27
`<Page/>` - Add ellipsis to title and subtitle 

hepiyellow:
Adds ellipsis to title and subtitle, only when they are simple strings.

Title can be a ReactNode.
Valid UX examples are a title with a Dropdown or Input.
I those cases we should NOT apply the ellipsis text.


hepiyellow:
Replaced by #2831

PR #2827, opened 2019-01-27
 `<Input />` migrate to Autodocs

NitayRabi:
### What changed

- Convert `Input`'s Story Book to use auto-docs. 

### Why it changed

Using auto docs for most components (standardisation).

---

- [x] Change is tested
- [x] Change is documented
- [ ] Build is green
- [ ] Change of UX is approved by [wuwa](https://github.com/wuwa) or [milkyfruit](https://github.com/milkyfruit) - No changes to UX.


stories/Input/Input.story.js
```diff
@@ -0,0 +1,52 @@
...
+  category: storySettings.category,
+  storyName: storySettings.storyName,
+  component: Input,
+  componentPath: '../../src/Input/Input.js',
+
+  componentProps: setState => ({
+    dataHook: storySettings.dataHook,
+    value: '',
+    onChange: e => setState({ value: e.target.value }),
+    size: 'small',
```
shlomitc: change to normal

src/Input/Input.e2e.js
```diff
@@ -0,0 +1,46 @@
...
+} from 'wix-ui-test-utils/protractor';
+import { createStoryUrl } from '../../test/utils/storybook-helpers';
+import autoExampleDriver from 'wix-storybook-utils/AutoExampleDriver';
+
+const storyUrl = createStoryUrl({
+  kind: 'Components',
+  story: 'Input',
+  withExamples: false,
+});
+const storyUrlWithExamples = createStoryUrl({
```
shlomitc: redundant

src/Input/Input.e2e.js
```diff
@@ -0,0 +1,46 @@
...
+  kind: 'Components',
+  story: 'FormField',
+  withExamples: true,
+});
+const driver = inputTestkitFactory({ dataHook: 'storybook-input' });
+
+describe('Input', () => {
+  describe('[AutoExample]', () => {
```
shlomitc: you can remove this describe block

src/Input/Input.e2e.js
```diff
@@ -0,0 +1,46 @@
...
+      'should render with correct text',
+      async () => {
+        await autoExampleDriver.setProps({ value: 'hello' });
+        await waitForVisibilityOf(
+          driver.element(),
+          'Cannot find Input component',
+        );
+        expect(await driver.getText()).toEqual('hello');
+      },
+      { version: 'no maxlength' },
```
shlomitc: remove this version annotation

stories/Input/examples/Commands.js
```diff
@@ -0,0 +1,51 @@
...
+        <TextButton style={{ padding: '0 1rem' }} onClick={handleClick3}>Select text</TextButton>
+      </div>
+    );
+  }
+}`;
+
+export default () => {
+  return (
+    <div>
+      <Markdown source="## Commands" />
```
yardnsm: I would call this `Using ref methods`

stories/index.js
```diff
@@ -116,7 +116,7 @@ import './FormField/FormField.story.js';
 import './GenericModalLayout/GenericModalLayout.story.js';
 import './GoogleAddressInput/index.story.js';
 import './Heading/index.story';
-import './Input';
+import './Input/Input.story';
```
yardnsm: Maybe the file should be `index.story.js`? Does it matter? @shlomitc

src/Input/README.md
```diff
@@ -1,7 +1,3 @@
-## Functions
+# `<Input/>`
 
-| function name   | description                   |
-| ---             | ---                           |
-| focus           | Focuses on the input          |
-| blur            | Blurs the input (loses focus) |
-| select          | Selects all text in the input |
+General input container
```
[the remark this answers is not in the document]
NitayRabi: For some reason without it didn't create the `<Input />` title

stories/index.js
shlomitc: Its ok like that in my opinion. also makes it easier to search file in IDEs by name and not by `index`

src/Input/README.md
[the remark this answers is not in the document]
shlomitc: @argshook - are you aware of it?

stories/Input/examples/Commands.js
```diff
@@ -0,0 +1,51 @@
...
+
+    const handleClick3 = () => {
+      this.refs.inputtest.focus();
+      this.refs.inputtest.select();
+    };
+
+    return (
+      <div>
+        <div style={{ width: '400px' }}>
+          <Input theme={this.props.theme} ref="inputtest" />
```
shlomitc:
```suggestion
          <Input theme={this.props.theme} ref={r => this.inputRef= r} />
```

and then use it like this
```js
const handleClick1 = () => this.inputReft.focus();
```

src/Input/README.md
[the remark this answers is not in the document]
argshook: here's current logic (writing from memory):
```
if (readme.md || comment) exists:
 don't add title; // assume readme.md or comment has it
otherwise:
  add title (`component.displayName`)
```

in this case readme.md includes heading at the top of file. if contents of readme.md would be moved
to comment in source, then heading should also go there.

otherwise it's possible to get duplicated titles and i don't really want to parse markdown and try
to guess if it has a heading or not

stories/Input/examples/Commands.js
```diff
@@ -0,0 +1,51 @@
...
+
+export default () => {
+  return (
+    <div>
+      <Markdown source="## Using ref methods" />
+      <Layout>
+        <Cell span={4}>
+          <LiveCodeExample
+            compact
+            title="Click to test commands API"
```
hepiyellow:
```suggestion
            title="Click buttons to trigger the Input's instance methods"
```

stories/Input/examples/Commands.js
```diff
@@ -0,0 +1,51 @@
...
+      this.inputRef.focus();
+      this.inputRef.select();
+    };
+
+    return (
+      <div>
+        <div style={{ width: '400px' }}>
+          <Input theme={this.props.theme} ref={r => this.inputRef = r} />
+        </div>
+        <TextButton style={{ padding: '0 1rem' }} onClick={handleClick1}>Focus</TextButton>
```
hepiyellow: I would use `<Button secondary/>` instead of `<TextButton/>`.
The TextButton doesn't look good, it is confusing, you don't know when a button starts and when it ends.

src/Input/Input.e2e.js
```diff
@@ -0,0 +1,27 @@
+import eyes from 'eyes.it';
+import { inputTestkitFactory } from '../../testkit/protractor';
+import { waitForVisibilityOf } from 'wix-ui-test-utils/protractor';
+import { createStoryUrl } from '../../test/utils/storybook-helpers';
+import autoExampleDriver from 'wix-storybook-utils/AutoExampleDriver';
+
+const storyUrl = createStoryUrl({
```
hepiyellow: I like to put everything inside the describe block.
That way when for example you run jest with some other focused test (`fdescribe`, `fit`),
which means that this test would not even run,
Then this code (outside your describe) would also not run.

If you do it outside the describe block, then it would be redundant code running.

But its a minor thing.

src/Input/Input.e2e.js
```diff
@@ -0,0 +1,27 @@
+import eyes from 'eyes.it';
```
hepiyellow: Don't use eyes.it directly.
```js
import {eyesItInstance} from '../../test/utils/eyes-it';
...
describe(...
  const eyes = eyesItInstance();
```

It applies default eyes config.

src/Input/Input.e2e.js
```diff
@@ -0,0 +1,27 @@
...
+  kind: 'Components',
+  story: 'Input',
+  withExamples: false,
+});
+
+const driver = inputTestkitFactory({ dataHook: 'storybook-input' });
+
+describe('Input', () => {
+  beforeAll(() => browser.get(storyUrl));
+
```
hepiyellow: Add :
```js
afterEach(()=> autoExampleDriver.remount());
```

Although you don't have more tests here, it will help not to forget it when future tests are added.

src/Input/Input.e2e.js
hepiyellow: I actually think it is a good practice.

stories/Input/examples/Error.js
```diff
@@ -0,0 +1,72 @@
...
+            title="Tooltip"
+            initialCode={createInputExample({
+              status: 'error',
+              statusMessage: 'I am a message',
+            })}
+          />
+        </Cell>
+      </Layout>
+      <Layout>
+        <Cell span={4}>
```
hepiyellow: I would take this opportunity and remove the `forceFocus` and `forceHover` examples.
We should deprecate these props anyways.

src/Input/Input.protractor.driver.js
```diff
@@ -1,8 +1,8 @@
 import { isFocused } from 'wix-ui-test-utils/protractor';
 
 const inputDriverFactory = component => {
-  const input = component.$('input');
-  const clearButton = component.$('[data-hook="input-clear-button"]');
+  const input = component.$('input'),
```
hepiyellow: Really?!
You like it this way?
I don't )-:

stories/Input/examples/Unit.js
```diff
@@ -0,0 +1,40 @@
...
+import Markdown from 'wix-storybook-utils/dist/src/Markdown';
+
+import { Cell, Layout } from '../../../src';
+import LiveCodeExample from '../../utils/Components/LiveCodeExample';
+import createInputExample from './createInputExample';
+
+export default () => {
+  return (
+    <div>
+      <Markdown source="## Unit" />
```
hepiyellow: This markdown with `##` heading is repeating itself.
We should extract it to one place, so we can change it later.

I started a while ago this [UXStoryTemplate](https://github.com/wix/wix-style-react/blob/master/stories/UXStoryTemplate.js), where I collect reusable story components.

It is kind of unofficial, since we will be moving soon (hopefuly) to using `wix-storybook-utils` for story page layout, but I think it's a good idea to use it.
So you can add there a `<ExampleGroup title="whatever">` component.

src/Input/Input.protractor.driver.js
NitayRabi: You mean the general `input` element selector ? - haven't noticed that - I'll set up a data hook

stories/Input/examples/Error.js
NitayRabi: NP removing them from the example - @shlomitc - should I open an issue for deprecating officially or perform it in this scope?

PR #2865, opened 2019-01-31
update(storybook): use babel-loader:7 in storybook config and update deps

mykas:
Stable storybook. Finally.


shlomitc:
as we're not transpiling with babel-7, wouldn't be a problem installing it alongside yoshi?
@yavorsky - any thoughts about this?

mykas:
babel-loader is not babel-7. Yoshi uses babel-loader 7. Issues started to happened with stable storybook, because storybook 4.1 has dependency of babel-loader@8

mykas:
screenshot 2019-01-31 at 17 05 59

mykas:
What I did is basically forced storybook to use babel-loader@7

yavorsky:
Yeah, babel-loader@7 works with babel@6. Anyway, it shouldn't break something, since `yoshi` will use own babel-loader.

yavorsky:
I'm preparing PR to migrate to babel@7 the whole project. But we need to update yoshi to 4 (since `yoshi test --protractor` requires babel-register@6) or run e2e w/o yoshi (which is also good option in long-term perspective).

PR #2868, opened 2019-01-31
Improve build process. Migrate to babel@7.

yavorsky:
### 🔦 Summary
`npm run build` command takes a lot of time. It's actually processing redunant operations for current stage. For example, it uses `yoshi build`, which not just transpilying code + copying assets, but also creating bundle with webpack, copying assets in different ways, using petri, etc (https://github.com/wix/yoshi/blob/version_4.x/packages/yoshi/src/commands/build.js#L36-L293)

I think for current operation we just need to transpile ESNext -> ES5 + ES5(+ imports).
So I got rid of `yoshi build` in favor of [babel-cli](https://babeljs.io/docs/en/babel-cli) tool.

Before
screen shot 2019-01-31 at 3 57 33 pm

After
screen shot 2019-01-31 at 3 48 12 pm

*Update after custom script with same AST*
screen shot 2019-02-05 at 1 56 37 am

It's wip since e2e is not working. We need to wait yoshi@4 release or moving out from yoshi and run e2e tests by our own.

### ✅ Checklist
- [x] ⏳ Fix e2e tests (wait for yoshi@4 release or run protractor manualy).
- [x] 😬 Make sure all components are present and transpiled correctly.


shlomitc:
maybe we should add a real e2e step to make sure components can be imported?
kind of a sanity test.

shlomitc:
@yavorsky - waiting with merge until the build is fixed. ping me when it's ready. thanks

yavorsky:
@shlomitc  sure. But I would merge it after https://github.com/wix/wix-style-react/pull/2944 to be sure `dist` is ok.

yavorsky:
I refactored a bit warnings tests here to use asymmetric matchers.
We dan't want to test the second part of the warnings - location. We are responsible for just our part - main warning message.
With babel 7, we're adding _source to the components, so each warning message includes warning coordinates: `in Breadcrumbs (at PageHeader.js:37)`. We don't need to handle this at all, so with asymmetric matchers we will check only our part - `Warning: Failed prop type: The prop `items[0].value` is marked as required in `Breadcrumbs`, but its value is 'undefined'.`

package.json
```diff
@@ -76,37 +80,47 @@
...
+    "@babel/core": "^7.2.2",
+    "@babel/parser": "^7.3.1",
+    "@babel/plugin-proposal-class-properties": "^7.3.0",
+    "@babel/plugin-proposal-export-namespace-from": "^7.2.0",
+    "@babel/plugin-transform-modules-commonjs": "^7.2.0",
+    "@babel/plugin-transform-runtime": "7.2.0",
+    "@babel/polyfill": "^7.2.5",
+    "@babel/preset-env": "^7.3.1",
+    "@babel/preset-react": "^7.0.0",
+    "@babel/preset-stage-2": "^7.0.0",
```
hepiyellow: @yavorsky 
I think preset-stage-2 is not in use. (?)

yavorsky: Thanks!

Issue #1545, opened 2018-03-13
<RadioGroup> - Deprecating 'button' type (and spacing bugs)


hepiyellow:
Issues:
1. spacing prop is applied only for radioButtons not for buttons.
  - I think it sohuld apply to buttons as well.
2. spacing is applied only for vertical alignment not for horizontal. 
  - We should suuport it for horizontal as well.
3. spacing adds margin after the last item.
  - I think is should be only between the items.
4. only button type gets a a wrapping element style of 
```
 {
  border: solid 1px $B40;
  border-radius: 6px;
}
```
  - I think the border style should be the same for radios and buttons. Which one shell it be? with a border? or Without?
5. While we're at it (breaking change): we can change the `type=default` to `type=radio` (and set it as the default is not specified)

These fixes are needed in order to comply with the spec for FieldWithSelection.
https://app.zeplin.io/project/5864e02695b5754a69f56150/screen/5871f730af5dd9d6276a4ff9

**NOTE: Fixing this might be considered a breaking change, since users might already have workarounds in place.**

argshook:
`button` type should be removed entirely, no need to fix it. there's `4.5a Advanced Toggle Switch` and `4.6 Thumbnails` to be used instead i think we should focus on implementing those.

https://github.com/wix/wix-style-react/issues/1497

hepiyellow:
@shlomitc wrote in #1497 (which i a duplicate and was closed):

> We will deprecate this component once the Thumbnails component will be created.

hepiyellow:
About deprecating `button` type. 
In FieldWithSelection (3.3b) there is use of RadioGroup in button mode.
And the new spec for 4.5 AdvancedToggle is not usable to get the required result for 3.3b.

So we might want to change the spec for 3.3b?

milkyfruit:
@hepiyellow either change 3.3b specs or update 4.5 Advance Toggle specs by adding new theme.
Frankly, when 3.3b was designed there wasn't much technical thinking involved.
So, let's just remove / hide 3.3b for now, until we find a working solution.

hepiyellow:
@milkyfruit please update the Zeplin spec by removing 3.3b.
BTW: Do you have an issue tracking system? (For design specs)

hepiyellow:
I will remove the example (with Buttons) from the storybook (Of FieldWithSelection 3.3)

milkyfruit:
@hepiyellow updated the [specs](https://zpl.io/a7jKmvV)

[...]

PR #2874, opened 2019-02-01
<RadioGroup/> - add deprecation message for type="button"

mykas:
This PR removes RadioGroup `type="button"` styling option. To achieve this kind of styling use - `<SegmentedToggle/>`.


shlomitc:
Maybe we should deprecate instead of removing? I guess there are some people using it

mykas:
Yes, I will probably deprecate it instead. Thought  that maybe I could still jump in to wsr 6 breaking changes wagon.

shlomitc:
Better to leave it to next version. In the meanwhile, just remove proptypes (so it will be removed from story) and add deprecation log.

shlomitc:
this PR will close this issue as well https://github.com/wix/wix-style-react/issues/1545

shlomitc:
@mykas - are you taking care of this?

mykas:
yes, I will continue later this week

mykas:
Hey @shlomitc , I have reverted my previous changes and added deprecation log instead.

PR #2880, opened 2019-02-02
<Button/> - improve and tweak documentation

mykas:
Improve Button documentation and remove `upgrade` values from api docs and proptypes.


mykas:
Yes

stories/Button/index.story.js
```diff
@@ -28,5 +51,104 @@ export default {
...
+        description({
+          title: `Primary (Filled Buttons)`,
+          text:
+            'Use primary buttons to give more prominence to actions in layouts with a lot of varying content.',
+        }),
+
+        liveCode({
+          ...defaultLiveCodeProps,
+          source: examples.primary,
+          components: { Layout, Cell, Button },
```
shlomitc: Instead of manually applying the components, it should use all components
Take the baseScope from here
https://github.com/wix/wix-style-react/blob/master/stories/utils/Components/LiveCodeExample/index.js

stories/Button/index.story.js
```diff
@@ -28,5 +51,104 @@ export default {
     as: ['button', 'a', 'span', 'div'],
   },
 
-  examples: <ButtonStory />,
+  sections: [
+    tab({
+      title: 'Description',
+      sections: [
+        description({
```
shlomitc: I think emojis are cool in general, but aren’t they making the docs a bit not professional?

stories/Button/index.story.js
```diff
@@ -28,5 +51,104 @@ export default {
...
+          text: `🔨 To trigger an operation.`,
+        }),
+
+        importExample({
+          source: "import Button from 'wix-style-react/Button';",
```
shlomitc: Isn’t there a default for that?

stories/Button/index.story.js
mykas: In general I agree, but I think that when used moderately - it should be ok. I would say that emojis and especially the one that symbolises, describes the meaning of section - are pretty great way to visually make docs more readable.

stories/Button/index.story.js
mykas: It was before. I am not sure about this. Maybe its default if you don't specify anything.

argshook: there is not default when working with sections, cause what if your sections describe multiple components with different imports.

stories/Button/index.story.js
argshook: im aware about this, since `liveCode` comes from wix-storybook-utils
which cannot have wix-style-react as dependency, for now there must be
`components` set. it can be `baseComponents` for now, in future perhaps
this can be set in webpack loader in order to not repeat it everywhere.

mykas: resolved. Made sure that `defaultLiveCodeProps` includes `baseProps` from LiveCodeExample source.

mykas:
@hepiyellow hey, can you review the testkit tab? I have tried to document async await methods, but maybe there is a better way to represent it.

src/Button/README.TESTKIT.md
```diff
@@ -4,13 +4,13 @@
...
-| method               | arguments | returned value | description                        |
-| -------------------- | --------- | -------------- | ---------------------------------- |
-| exists               | -         | boolean        | returns true if element in the DOM |
-| element              | -         | element        | returns the component element      |
-| click                | -         | -              | clicks on the button               |
-| getButtonTextContent | -         | string         | returns the button text            |
-| isButtonDisabled     | -         | boolean        | returns true if button is disabled |
+| method                                | arguments | returned value | description                        |
+| ------------------------------------- | --------- | -------------- | ---------------------------------- |
+| async => await exists()               | -         | boolean        | returns true if element in the DOM |
```
hepiyellow:
```suggestion
| exists               | -         | Promise<boolean>        | Returns a promise that will be resolved with true if the element in the DOM |
```

src/Button/Button.js
```diff
@@ -39,15 +39,12 @@ class Button extends Component {
     children: node,
     /** String based data hook */
     dataHook: string,
-    /** Flag to enable new button API. IMPORTANT! - After upgrading, when you import the react/enzyme "buttonTestkitFactory", you will get an async testkit (all methods are async)*/
-    upgrade: bool,
   };
```
hepiyellow: Can you also add `as` and `href` props

mykas: I think that it is pretty clear that you can render as something else and use href, but I will add `as` prop.

PR #2886, opened 2019-02-03
`<Page/>` - Add  `stretchContentVertically` prop

### LINKED ISSUE #2810: Page.Content does not fill all available height
WixStores UX requires a <Page> component whose <Page.Content> should occupy all the available vertical space.

Currently the only way I found, was to attach CSS properties to the inner [data-class=page-scrollable-content] node, as in:

https://github.com/wix-private/ecom/blob/4bace2504c7718f24d2269dee857d802b6c90cf0/client/wixstores-dashboard/wixstores-dashboard-abandoned-carts/src/components/Main/Main.scss#L17

but it is unstable because it relies on private implementation.

Possible to add a public property? or maybe a way to specify Content's class?

Thank you!

# Bug Report

### Relevant Components
`<Page/>`, `<Page.Content/>`

### Current Behavior
`<Page.Content/>` does not fill all available height, there is no easy way to achieve that

### Expected Behavior
`<Page.Content/>` should either always fill all available height (as in, have `height: 100%`) or have ability to do so without any hacks

### Steps to Reproduce
https://github.com/wix-private/ecom/blob/4bace2504c7718f24d2269dee857d802b6c90cf0/client/wixstores-dashboard/wixstores-dashboard-abandoned-carts/src/components/Main/Main.scss#L17

### Severity
Major

originally reported by @zzgab

hepiyellow:
Replaces #2839 - Rebase on master, lint fix, move hidden examples to `PageTestStories.js`
Resolves #2810
@zzgab 
### stretchVertically prop

I think we don't really need the `stretchVertically` prop.
The page wrapper (most outer div) is already stretching using `flex-grow: 1`.
We can add all those `height 100%` to all the `div`'s in the heirarchy, and keep them there always.

Then child of `<PageContet/>` can still decide if to stretch or not using it's own styles (height:100%).

Having said that, I am not sure it is NOT a breaking change to make the `Page.Content`'s parent fill the height, since if someone added `height: 100%` and it didn't have any effect, now it might suddenly have an effect and break some layout. SO...

Let add the prop, just to be careful.
I would call it `stretchContentVertically`.

### Tests

- [ ] Need to add e2e tests
- [ ] Make sure it doesn't break Page when FixedContent is provided.

### Documentation

- [ ] Add a propper example with `stretchContentVertically`. Add a content with :
```js
<Page stretchContentVertically>
  <Page.Header>
  <Page.Content>
    <Container>
      <Row>
        <Card>
          <Card.Header/>
          <Card.Content>Hello</Card.Content>
        </Card>
      </Row>
    </Container>
  </Page.Content>
</Page>
```

I suggest adding a `2.5 Page Examples` folder in the storybook.

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test

### DIFF
--- a/src/Page/Page.js
+++ b/src/Page/Page.js
@@ -253,20 +253,35 @@ class Page extends WixComponent {
       minimizedFixedContainerHeight,
     } = this._calculateHeaderMeasurements({ PageTail });
 
-    const classNameStretchVertically = stretchVertically ? s.stretchVertically: '';
+    const classNameStretchVertically = stretchVertically
+      ? s.stretchVertically
+      : '';
 
     const contentLayoutProps = {
-      className: classNames(s.content, {
-        [s.contentFullScreen]: contentFullScreen,
-      }, classNameStretchVertically),
+      className: classNames(
+        s.content,
+        {
+          [s.contentFullScreen]: contentFullScreen,
+        },
+        classNameStretchVertically,
+      ),
       style: contentFullScreen ? null : pageDimensionsStyle,
     };
 
     return (
-      <div className={classNames(s.pageWrapper, className, classNameStretchVertically)}>
+      <div
+        className={classNames(
+          s.pageWrapper,
+          className,
+          classNameStretchVertically,
+        )}
+      >
         <div
           className={classNames(s.page, classNameStretchVertically)}
-          style={{ minWidth: minWidth + 2 * PAGE_SIDE_PADDING_PX, paddingBottom: bottomPadding }}
+          style={{
+            minWidth: minWidth + 2 * PAGE_SIDE_PADDING_PX,
+            paddingBottom: bottomPadding,
+          }}
         >
           <div
             data-hook="page-fixed-container"
@@ -340,7 +355,12 @@ class Page extends WixComponent {
                 style={{ height: gradientHeight }}
               />
             )}
-            <div className={classNames(s.contentContainer, classNameStretchVertically)}>
+            <div
+              className={classNames(
+                s.contentContainer,
+                classNameStretchVertically,
+              )}
+            >
               <div {...contentLayoutProps}>
                 {this._safeGetChildren(PageContent)}
               </div>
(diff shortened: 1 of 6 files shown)


src/Page/Page.js
```diff
@@ -251,18 +253,35 @@ class Page extends WixComponent {
...
     const contentLayoutProps = {
-      className: classNames(s.content, {
-        [s.contentFullScreen]: contentFullScreen,
-      }),
+      className: classNames(
+        s.content,
+        {
+          [s.contentFullScreen]: contentFullScreen,
+        },
+        classNameStretchVertically,
```
hepiyellow: `classNameStretchVertically` should go only on the scrollable content not the fixed content

src/Page/Page.js
```diff
@@ -233,6 +233,8 @@ class Page extends WixComponent {
       className,
       children,
       minWidth,
+      stretchVertically,
+      bottomPadding,
```
hepiyellow: Can you share UX specs? why can't you add padding in the child of `<Page.Content/>`?

stories/Page/SomeContentComponent.js
```diff
@@ -12,7 +12,7 @@ export default class SomeContentComponent extends React.Component {
 
   render() {
     return (
-      <div style={{ backgroundColor: 'white' }}>
+      <div style={{ backgroundColor: 'white', minHeight: '100%' }}>
```
hepiyellow: Maybe minHeight should be a prop `stretchVertically`

hepiyellow:
This PR might not be relevant, 
@zzgab reported that " Stores UX considers an alternate design"...

Waiting for confirmation.

hepiyellow:
@zzgab Anyways, we are thinking of making a bigger breaking change, and include this change as part of it.
See this PR (fork of your PR) https://github.com/wix/wix-style-react/issues/2888

hepiyellow:
Replaced by Breaking Big Change #2915

hepiyellow:
Replaced by #2929

PR #2889, opened 2019-02-03
`<Page/>` - Remove dependency on a PageContainer with flex. And stretch vertically the Page.Content container. (`upgrade` prop)

### LINKED ISSUE #2888: `<Page/>` - Use `height: 100%`, and deprecate requirements for page container styles
# ✨ Feature Request

### 🔦 Scope

...

### 💁 Explanation

```
height: 100vh;
display: flex;
flex-flow: column;
min-height: 0;
```

### 🐾 Possible solution 

### 👀 Severity

Major

 - slow dev velocity
 - Bugs
 - high maintenance

hepiyellow:
### Overview
- It resolves #2888 and more.
- Adds an `upgrade` prop, that makes the page use `height: 100%` instead of depending on a parent with `display: flex;flex-flow: column`
- Page.Content's container is always 100% height.

> Originally branched from the PR that adds `stretchVertically` (#2886)

### Includes:

### Issues
- Storybook: Can't have `2.5 Page Example` folder next to the `2.5 Page`
- How do we show the code of the examples?


boriskar:
I think you added me by mistake... no?

hepiyellow:
Replaced by #2915

PR #2887, opened 2019-02-03
<Input /> - Use compound components for prefix/suffix rendering

NitayRabi:
### 🔦 Summary
1. Add `CustomAffix` component to replace rendering of `unit` and support text rendering. (used when suffix/prefix is string only)
2. Add `IconAffix` for correct rendering of icons (replacing magnifyingGlass prop)
3. Match design with definition in specs (Remove separator, change paddings to fit constraints)
4. Deprecate (log only) magnifyingGlass, unit prop and Input.Unit component.
5. Remove redundant stories, create Test page and e2e tests for new Input Applitool baseline.

### ✅ Checklist
- [ ] 👨‍💻 API change is approved by the UI Infra Developers 
- [x] 👨‍🎨 UX change is approved by the Design System UX 
- 📚 Change is documented
  - [x] Story
  - [x] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [x] Visual test

PR #2898, opened 2019-02-04
<ColorInput/> - new component

mykas:
### 🔦 Summary

Input component with ColorPicker popover.

### 👌API

```js
<ColorInput 
      value="string"
      onChange={func} // (parsedColor) => () 
      placeholder={node}
      error={boolean}
      errorMessage={node}
      size="small | medium | large"
      popoverPlacement='bottom',
      popoverAppendTo='parent',
      disabled={boolean}
/>
```


mykas:
Some features are blocked until this PR is resolved: https://github.com/wix/wix-style-react/pull/2887

LittleTal:
- Color viewer should be clickable and open color picker
- Color picker should have 8px border radius
- Confirm button should be disabled when no input was entered
- Input value is not vertically center aligned in small size
- Selecting a new color from color picker and then clicking the cancel button should revert to the previous color value. currently clicking cancel has the same result as approve button
- Disabled color is not aligned - currently D60
- Esc should be the same as clicking the cancel button
- Tab - approve and move to the next component
- Button icons in color picker should be 24px, currently 18px

UX changes:
- Change color viewer border to D30 with 30% opacity locally (shouldn’t be defined as color variable)
- Noticed that I can’t enter # char in the input, let’s solve all invalid chars in the same way- allowing user to enter only 0-F numbers.

mykas:
- [ ] Color viewer should be clickable and open color picker
- [ ] Color picker should have 8px border radius
- [ ] Confirm button should be disabled when no input was entered
- [ ] Input value is not vertically center aligned in small size
- [ ] Selecting a new color from color picker and then clicking the cancel button should revert to the previous color value. currently clicking cancel has the same result as approve button
- [ ] Disabled color is not aligned - currently D60
- [ ] Esc should be the same as clicking the cancel button
- [ ] Tab - approve and move to the next component
- [ ] Button icons in color picker should be 24px, currently 18px

UX changes:

- [ ] Change color viewer border to D30 with 30% opacity locally (shouldn’t be defined as color variable)
- [ ] Noticed that I can’t enter # char in the input, let’s solve all invalid chars in the same way- allowing user to enter only 0-F numbers.

stories/ColorInput/Usage.md
```diff
@@ -0,0 +1,5 @@
+ColorInput is semi-controlled which means that consumer will only get changed value through `onConfirm` handler when one of the following confirm events occurs:
```
argshook: can't do suggestions for multiple lines, i'd write something like this:

    `<ColorInput/>` is an input which allows to write HEX color manually or pick it from color picker.

    Value is received through `onConfirm` callback, which is called after:

    - clicking <kbd>Enter</kbd> key
    - clicking confirm button
    - clicking outside of `<ColorInput/>`

---

i wouldn't mention controlled/semi/whatever as these are internal
details not really needed for user.

src/ColorInput/components/index.js
```diff
@@ -0,0 +1,2 @@
+export { Hash } from './Hash';
+export { ColorViewer } from './ColorViewer';
```
argshook: maybe this file is not needed at all?

src/ColorInput/ColorInput.spec.js
```diff
@@ -0,0 +1,286 @@
...
+import { extractHex } from './hex-helpers';
+
+describe('ColorInput', () => {
+  beforeAll(() => {
+    requestAnimationFramePolyfill.install();
+  });
+
+  const renderColorInput = ({
+    value = '',
+    onConfirm = () => {},
```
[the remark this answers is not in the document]
mykas: value is required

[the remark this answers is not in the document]
argshook: i mean default value for `onConfirm`, not `value` :)

src/ColorInput/ColorInput.js
```diff
@@ -47,10 +47,10 @@ class ColorInput extends React.Component {
     value: string.isRequired,
     /** returns confirmed value */
     onConfirm: func,
-    /** returns last confirmed value which is props.value */
+    /** returns last confirmed value which is from user prop - value */
     onCancel: func,
-    /** returns colorpicker change value */
-    onPreview: func,
+    /** returns either input's or colorpicker's changed value */
```
argshook: i'm not gonna block though i don't really agree with this decision.

this change means we no longer have a solid contract with user.

instead of

`onChange has valid color string, always`

we're now saying

`onChange has maybe valid color string, maybe not, good luck`

but anyway just a comment, not gonna block

mykas: Thanks, so as we talked before I don't think that now 'onChange' is returning REALLY invalid cases. This is more about: does user module that receives `#12` kind of values support it or not? If not I have added the example which will always return valid cases so there is a fallback in such cases.

PR #2899, opened 2019-02-04
<Autocomplete /> - empty state

giladartzi:
### 🔦 Summary
Introducing the `emptyStateMessage` to `<Autocomplete />`. It renders a predefined message in case all options are filtered out.

A few questions:
- At the moment, when focusing on the input, it displays the entire unfiltered suggestion list. This is a defined behaviour - is it relevant to our use case as well?
- How does this feature behave when the option list is empty to begin with?
- How should we handle initialValue?
- Is the empty state selectable? Maybe some kind of a call to action?

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


stories/AutoComplete/ExampleComplex.js
```diff
@@ -35,20 +35,32 @@ class ControlledAutoComplete extends React.Component {
 
   render() {
     console.log('@@@@@@@@@@', this.state.suggestions);
+
```
shlomitc: I'm not sure what was the intention of this very much complex example, but I would really appreciate if you can create a simple standalone example that explains how empty message works

You can take a look in our "example-oriented-documentation" in the `Input` [examples section](https://wix-wix-style-react.surge.sh/?selectedKind=Components&selectedStory=Input&full=0&addons=0&stories=1&panelRight=0).
The LiveCode displays a simple scenario with the very minimal code needed to achieve it.

PR #2902, opened 2019-02-05
Accordion Component

mozesoron:
### 🔦 Summary
...

### ✅ Checklist
- 📚 Change is documented
  - [x] Story
  - [x] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [x] Visual test


src/Accordion/Accordion.spec.js
```diff
@@ -0,0 +1,114 @@
...
+    const driver = createDriver(<Accordion data={towItemsData}/>);
+    expect(await driver.isContentOpen(0)).toBeFalsy();
+    expect(await driver.isContentOpen(1)).toBeFalsy();
+
+    await driver.clickAtRow(0);
+    expect(await driver.isContentOpen(0)).toBeTruthy();
+    expect(await driver.isContentOpen(1)).toBeFalsy();
+  })
+
+  it('should check the default more less button labels', async () => {
```
shlomitc: it should verify the expand and collapse labels

src/Accordion/Accordion.spec.js
```diff
@@ -0,0 +1,114 @@
...
+    const driver = createDriver(<Accordion data={towItemsData}/>);
+    expect(await driver.getToggleButtonLabelAt(0)).toEqual('More');
+    await driver.clickAtRow(0);
+    expect(await driver.getToggleButtonLabelAt(0)).toEqual('Less');
+  });
+
+  it('should override default more less button labels', async () => {
```
shlomitc: there shouldn't be defaults

src/Accordion/Accordion.spec.js
```diff
@@ -0,0 +1,114 @@
...
+        lessLabel: 'subtract'
+      }
+    ];
+    const driver = createDriver(<Accordion data={data}/>);
+    expect(await driver.getToggleButtonLabelAt(0)).toEqual(data[0].moreLabel);
+    await driver.clickAtRow(0);
+    expect(await driver.getToggleButtonLabelAt(0)).toEqual(data[0].lessLabel);
+  });
+
+  it('should toggle the open between rows', async () => {
```
shlomitc: I would split it into two tests
1. should collapse an expended row on click
2. should expend a row on click and close an already expanded row (click on 0, click on 1, verify 0 is collapsed and 1 is expanded.

src/Accordion/AccordionHeader/Component.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
```
shlomitc: remove

src/Accordion/README.md
```diff
@@ -0,0 +1,11 @@
+# `<Accordion/>`
+
+Accordion with collapse
+
+```xml
```
shlomitc: remove API from readme and move to comment on the propTypes

src/Accordion/Accordion.spec.js
mozesoron: but in that case I will need to open another row and only then check to see that the new clicked on row has expanded so I will have to redo the steps.

shlomitc: you're testing both collapsing of a single row and the side effect of collapsing a row when another row is expanded.

Just for the sake of clear tests - split it into two different test cases

mozesoron: done

src/Accordion/Accordion.js
```diff
@@ -0,0 +1,54 @@
...
+  static displayName = 'Accordion';
+
+  static propTypes = {
+    dataHook: PropTypes.string,
+
+    /** Row Data: {title: string, icon: node, content: node, expandLabel: string, collapseLabel: string}*/
+    data: PropTypes.array,
+  };
+
+  static defaultProps = {
```
shlomitc: remove

src/Accordion/README.md
```diff
@@ -0,0 +1,7 @@
...
+<Accordion title="Accordion title" data="[]"/>
```
shlomitc: also this should be removed

stories/Accordion/storySettings.js
```diff
@@ -0,0 +1,9 @@
+import { Category } from '../storiesHierarchy';
+
+export const storySettings = {
+  kind: Category.COMPONENTS,
```
shlomitc: change to WIP folder

src/Accordion/AccordionRow/row.scss
```diff
@@ -0,0 +1,53 @@
+@import '../../common.scss';
```
shlomitc: rename file to AccordionRow.scss

src/Accordion/Accordion.js
```diff
@@ -0,0 +1,54 @@
...
+    const { dataHook, data} = this.props;
+
+    return (
+      <div className={styles.root} data-hook={dataHook}>
+        {data && data.map((row, index) => (
+          <div className={styles.rowWrapper} key={index}>
+            <AccordionRow
+              dataHook={'accordion-row'}
+              id={index}
+              toggleView={this.toggle}
```
shlomitc: this prop is now called `toggleOpenClsoe` so your code and tests break

src/Accordion/Accordion.js
```diff
@@ -0,0 +1,54 @@
...
+              {...row}
+              open={openId === index}
```
shlomitc: isOpen

src/Accordion/Accordion.spec.js
```diff
@@ -0,0 +1,94 @@
...
+  })
+
+  it('should verify the expand and collapse labels', async () => {
+    const driver = createDriver(<Accordion data={towItemsData}/>);
+    expect(await driver.getToggleButtonLabelAt(0)).toEqual('More');
+    await driver.clickAtRow(0);
+    expect(await driver.getToggleButtonLabelAt(0)).toEqual('Less');
+  });
+
+  it('should collapse an expended row on click', async () => {
```
shlomitc: this is not testing the collapsed behavior, but rather the expanding behavior.

1. ditch the first expectation (since it is collapsed at first)
2. click on the row (also no need to expect it is expended since you already tested it)
3. click on the row and expect it to be closed

src/Accordion/Accordion.driver.js
```diff
@@ -0,0 +1,27 @@
...
+  const getRowAtIndex = (idx) =>  base.$$('[data-hook="accordion-row"]').get(idx);
+
+  return {
+    ...baseUniDriverFactory(base),
+
+    getTitleOfRow: async (idx) => getRowAtIndex(idx).$('[data-hook="title"]').text(),
+    getIconAt: async (idx) => getRowAtIndex(idx).$('[data-hook="icon"]').exists(),
+    isContentOpen: async (idx) => getRowAtIndex(idx).$('[data-hook="content"]').exists(),
+    clickAtRow: async (idx) => getRowAtIndex(idx).$('[data-hook="toggle-accordion-wrapper"]').click(),
+    isRowOpenAt: async (idx) => getRowAtIndex(idx).$('[data-hook="toggle-accordion-wrapper"]').hasClass('open'),
```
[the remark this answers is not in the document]
mozesoron: isContentOpen = to see that the content only got expand
isRowOpenAt = that the complete component has the open class.

Row includes all the elements with the content

shlomitc:
1. blue hover should be removed when expanded.
2. hover (and "more" show) should be displayed on row hover
3. click on the row should expand as well (but not close)

shlomitc:
- Background color should be applied only for the content that is opened
- Row height should be 72px
- Title typography should be normal (55-16)
- More-Less button should be sticked to the row and not at the bottom
- the entire row should be clickable and effect the button hover
    - When expanded - only the button is clickable and hoverable
- The title should apply `Text` component for string and nothing for node

@LittleTal - please add the zeplin spec once it refactored from the collapsable table

gilisho:
Hey, we made changes according to your request. Would be glad if you could check it out.

shlomitc:
@LittleTal - please make a UX review again.
@gilisho - can you please fix the failing test and conflicts? thanks.

shlomitc:
1. keyboard focus might be problematic so I say not to handle this at the moment
2. icons existence/non-existence is already supported, maybe let's just add an example for that in the story
3. text size - are you sure? it uses the `Text` component, so no differences should be present

LittleTal:
image

shlomitc:
these styles are applied directly to any element in the `storybook`-  they are bad and should be removed anyway, but affects some of our visual tests so it's in our backlog

since the content of the accordion item is a simple `div`, it inherits these styles.

should the item content define specific styles? I don't think we want to restrict.

LittleTal:
Yes, it should contain any content.

gilisho:
Hey, I finished to fix the component and improved the storybook examples - there's now examples of different icons and a non-icon item + node content and not just text content.

src/Accordion/AccordionItem/AccordionItem.scss
```diff
@@ -0,0 +1,93 @@
+@import '../../common.scss';
+$row-hover-vertical-padding: 21px;
+$row-padding-left-no-icon: 30px;
+$row-sides-padding: 24px;
+$row-vertical-padding: 24px;
+$icon-accordion-height: 24px;
+$icon-margin-right: 24px;
+$moreLessPosition: 30px;
```
argshook: @gilisho please review these styles, e.g. `moreLessPosition` looks unused.

moreover, things like `moreLess-width` seem very restrictive. What will happen if `expandLabel` is a long german word, will 72px hold it?

if variable is used only once, perhaps it's not enough to become a variable? :)

src/Accordion/MoreLessButton/MoreLessButton.scss
```diff
@@ -0,0 +1,26 @@
+@import '../../common.scss';
+
+.toggleButtons {
+  position: relative;
+  display: flex;
+  justify-content: flex-end;
+
+  .button,
```
argshook: are these nestings necessary?

stories/Accordion/index.story.js
```diff
@@ -0,0 +1,101 @@
...
+    label: 'Simple data',
+    value: [
+      {
+        title: 'First Item',
+        icon: <InfoCircle />,
+        expandLabel: 'More',
+        collapseLabel: 'Less',
+        content: (
+          <Text>
+            lauren ipsum lauren ipsum lauren ipsum lauren ipsum lauren ipsum
```
argshook: it's `Lorem ipsum` :)
you can put example text in variable to avoid repetition, e.g. `const text = 'lorem ipsum dolor sit amet'.repeat(20)`

src/Accordion/Accordion.scss
```diff
@@ -0,0 +1,10 @@
+@import '../common.scss';
```
argshook:
```suggestion
@import '../colors.scss';
```

src/Accordion/Accordion.spec.js
```diff
@@ -0,0 +1,115 @@
...
+    it('should render item with icon', async () => {
+      const data = [
+        {
+          title: 'first item',
+          icon: <FakeIcon />,
+          content: 'first item content',
+        },
+      ];
+      const driver = createDriver(<Accordion data={data} />);
+      expect(await driver.isIconExistsAt(0)).toBeTruthy();
```
argshook: please be explicit throughout this file. `toBeTruthy` and `toBeFalsy` can be false positive
```suggestion
      expect(await driver.isIconExistsAt(0)).toBe(true);
```

src/Accordion/AccordionItem/AccordionItem.js
```diff
@@ -0,0 +1,111 @@
...
+import { BUTTON_TYPES } from '../constants';
+
+class AccordionItem extends React.PureComponent {
+  static displayName = 'AccordionItem';
+
+  static propTypes = {
+    dataHook: PropTypes.string,
+    buttonType: PropTypes.oneOf(Object.values(BUTTON_TYPES)),
+    id: PropTypes.number,
+    title: PropTypes.string,
```
argshook: Accordion accepts items with title as `node`, but here it's limited to `string`, that's no good.

title, expandLabel and collapseLabel should all be `node`

src/Accordion/AccordionItem/AccordionItem.js
```diff
@@ -0,0 +1,111 @@
...
+      [styles.open]: isOpen,
+    });
+    const headerTextStyle = classNames(styles.headerText, {
+      [styles.headerTextNoIcon]: !icon,
+    });
+    return (
+      <div
+        className={styles.wrapper}
+        data-hook={dataHook}
+        onMouseEnter={this.toggleHoverIn}
```
argshook: why not remain consistent with event and name handler `this._onMouseEnter`

src/Accordion/MoreLessButton/MoreLessButton.js
```diff
@@ -0,0 +1,73 @@
...
+
+  const shouldRenderButton =
+    hover && !isOpen && buttonType === BUTTON_TYPES.BUTTON;
+  const shouldRenderTextButton =
+    hover && !isOpen && buttonType === BUTTON_TYPES.TEXT_BUTTON;
+  return (
+    <div
+      data-hook={dataHook}
+      className={toggleStyle}
+      onClick={e => {
```
argshook: `handleClick` prop should be named `onClick`, then

```suggestion
      onClick={onClick}
```

the `stopPropagation`, if really needed, should be handled by consumer of `MoreLessButton` component

src/Accordion/MoreLessButton/MoreLessButton.scss
```diff
@@ -0,0 +1,25 @@
+@import '../../common.scss';
+
+.toggleButtons {
+  position: relative;
+  display: flex;
+  justify-content: flex-end;
+}
+
+.button,
+.expand-collapse-arrow {
```
argshook: please remain consistent, some class names are in kebab-case, some in camelCase

src/Accordion/MoreLessButton/MoreLessButton.scss
```diff
@@ -0,0 +1,25 @@
+@import '../../common.scss';
```
argshook:
```suggestion
@import '../../colors.scss';
```

src/Accordion/MoreLessButton/MoreLessButton.scss
```diff
@@ -0,0 +1,25 @@
...
+  justify-content: flex-end;
+}
+
+.button,
+.expand-collapse-arrow {
+  transition: all .2s;
+}
+
+.expand-collapse-arrow {
+  opacity: 1;
```
argshook: is this really needed?

src/Accordion/MoreLessButton/MoreLessButton.scss
```diff
@@ -0,0 +1,25 @@
...
+  fill: $B10;
```
argshook:
```suggestion
  color: $B10;
```

src/Accordion/MoreLessButton/MoreLessButton.scss
```diff
@@ -0,0 +1,25 @@
...
+}
+
+.button {
+  pointer-events: none;
```
argshook: are these styles really needed? perhaps we should use `disabled` prop on the button?

src/Accordion/docs/storySettings.js
```diff
@@ -0,0 +1,9 @@
+import { Category } from '../../../stories/storiesHierarchy';
+
+export const storySettings = {
+  kind: Category.WIP,
```
argshook: please use only `category`

this may have been generated for you but we want to get rid of duplication

src/Accordion/Accordion.spec.js
gilisho: Could you clarify on this one? I'm not really aware that it can be false positive and I see it's being used in other files in wsr.

src/Accordion/MoreLessButton/MoreLessButton.js
gilisho: If we remove the `stopPropagation` then the Less button won't work. Therefore I think it should be inside the `MoreLessButton` and not handled by the consumer.

argshook: ok, ignore it for now

src/Accordion/MoreLessButton/MoreLessButton.scss
gilisho: It was needed so that when hovering on the button, it will interpret it as we are hovering on the whole row - and that's how the color of the row is not changed to white.
Weirdly, I see that it works also without it... I remembered that it didn't work without it. Anyway, I removed it as it's not necessary...

argshook:
This PR is ready to be merged but due to urgency it has a few leftovers.

1. unit tests do not cover all corners
2. keyboard a11y works well but is not 100% according to spec. spec displays changing focused area after `<AccordionItem/>` has been opened. this currently is not possible because ButtonNext lacks `.focus` method. Fixing this is out of scope of this PR

@gilisho thanks for your hard work and patience with me :)

PR #2905, opened 2019-02-05
Add transform-runtime and use babel’s helpers + generator from the babel-runtime package.

yavorsky:
### 🔦 Summary
This PR adds `transform-runtime` + `babel-runtime` package.

Since we are exporting some stuff that requires additional helpers (async/generators requires regeneratorRuntime helpers to be in bundle). But with current approach we assume that our users will export additional babel stuff on their side (`import 'regenerator-runtime/runtime'`).
User's shouldn't know about WSR language and technology implementation.
Better is to export everything ready-to-use with needed helpers.

So it's better to have `babel-runtime` package and export some helpers + regenerator things from it. (classCallCheck, regeneratorRuntime, _isObject, etc). It should reduce bundle size, since we won't have each helper implemented in each file.

app.bundle.js comparison

*Before (using babel helpers in each module + not using regeneratorRuntime helper). Not working for TS.*
screen shot 2019-02-06 at 2 30 28 am
screen shot 2019-02-06 at 2 30 36 am

*After (using babel-runtime for helpers + regeneratorRuntime). Works with TS.*
screen shot 2019-02-06 at 2 30 47 am
screen shot 2019-02-06 at 2 30 59 am

### ✅ Checklist
- [ ] 👨‍💻 Run smoke tests against `dist` (also would be good for https://github.com/wix/wix-style-react/pull/2868).


package.json
```diff
@@ -118,11 +119,11 @@
     "simple-git": "^1.107.0",
     "sinon": "^1.17.6",
     "tempy": "^0.2.1",
+    "terser": "3.14.0",
```
yairhaimo: https://github.com/terser-js/terser/pull/254 was merged.
Can remove this fixed version IMO

yavorsky: Oops

yavorsky:
I've added node_modules to ignore transpilation. `module.exports + imports` are not working in the same scope. Some of node_modules packages are using babel inlined helpers, which is being override with import's in the top of the these modules. But on the same time, these modules were already transpiled and have `module.exports` inside. So we had an error.

We can specify explicitly npm packages we don't want to transform, but ignoring all node_modules could be also useful for build time.

Issue #2532, opened 2018-12-03
<Search /> doesn't handle element options


sthuck:
# Bug Report

### Relevant Components
`<Search>`

### Current Behavior
When passing elements to search `<Search options={{id: '1', value: <div>111</div>}}/>`, the component is broken. 
the component tries to run a regex over the element (which I guess has toString() of '[object Object]' or something.) 
https://github.com/wix/wix-style-react/blob/master/src/Search/Search.js#L51

### Expected Behavior
Fix the behavior, so that elements could be passed.
Either don't filter for controlled components, or or some other magic I can't think of 😄 

if not/at the meanwhile, fixing the documentation will be appreciated:
https://wix-wix-style-react.surge.sh/?selectedKind=3.%20Inputs&selectedStory=3.9%20Search&full=0&addons=0&stories=1&panelRight=0

### Steps to Reproduce
https://codesandbox.io/s/mo9q3wonrj

### Severity
Low

shlomitc:
Fix was merged, will be released in next version (6.5.0)

PR #2907, opened 2019-02-05
<Search /> - allow filtering options using custom predicate

NitayRabi:
### 🔦 Summary
Fixs - #2532. 
Allows consumers to state predicate function for custom filtering of options (using metadata/other custom logic).

### ✅ Checklist
- 📚 Change is documented
  - [x] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [ ] Visual test


src/Search/Search.js
```diff
@@ -18,6 +18,7 @@ export default class Search extends WixComponent {
     ...InputWithOptions.propTypes,
     placeholder: PropTypes.string,
     expandable: PropTypes.bool,
+    predicate: PropTypes.func,
```
shlomitc: add comment

src/Search/Search.spec.js
```diff
@@ -240,22 +255,21 @@ describe('Search', () => {
       expect(driver.isCollapsed()).toBeFalsy();
     });
   });
-});
-
-describe('Testkits', () => {
```
shlomitc: remove and make sure it is part of the testkit-smoke script (in testkit folder)

stories/Search/examples/Expandable.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import * as wsrScope from '../../../src';
```
shlomitc: `import {Layout, Cell}  from '../../../src/Layout';`

stories/Search/examples/Predicate.js
```diff
@@ -0,0 +1,50 @@
...
+import LiveCodeExample from '../../utils/Components/LiveCodeExample';
+import { Section } from '../../UXStoryTemplate';
+
+const Example = `
+class SearchPredicate extends React.Component {
+  
+  constructor() {
+    this.state = {
+      options: [
+        { id: 1, value: <div style={{color: 'blue'}}>111</div>, metadata: { keywords: ['one', '1', 'first'] }},
```
shlomitc: since `metadata` concept is important, yet not documented, I think it's worth adding a short paragraph about it above the example (like in the buttons)

stories/Search/examples/Predicate.js
```diff
@@ -0,0 +1,50 @@
...
+        { id: 2, value: <div style={{color: 'orange'}}>222</div>, metadata: { keywords: ['two', '2', 'second'] }},
```
shlomitc: prefer not to color it (it's not a common nor advised use case)

src/Search/Search.spec.js
NitayRabi: I removed and validated smoke-test includes this component.

Issue #1693, opened 2018-04-23
Cut Date Picker icon in ie11


mariafeldman:
### Priority:
- [ ] Blocker
- [ ] Urgent
- [V ] Normal
- [ ] Minor

### Steps to reproduce:
 1.Add Date Picker to a page
2.Open the page in IE11

### Expected behavior:
DatePicker icon should be not cut
### Relevant Components:
DatePicker

### Additional notes:
![dashboard](https://user-images.githubusercontent.com/2092879/39114074-6869f274-46e6-11e8-8a64-c5c13724bf03.png)

mariafeldman:
Fixed

PR #2911, opened 2019-02-06
<Input/> - make sure that prefix is aligned vertically

mykas:
### 🔦 Summary

Prefix wrapper is defined with styles `display:block` and so the direct child is not aligned vertically.

The `display: block` value was used because of ie11 issues: https://github.com/wix/wix-style-react/issues/1693 https://github.com/wix/wix-style-react/issues/1691

I'll make sure that by changing display:block to display: flex - icon will still work properly in ie11.


mykas:
Hey @NitayRabi, can you check if its aligned correctly in examples? I couldn't make it work with 
```js
export default props => `
<Input ${createPropsArray(props).join('\n       ')} />
`;
```

I get circular JSON issues.

NitayRabi:
Hey @mykas, I'm pretty sure it'll be fixed via #2887

mykas:
Oh great! Then I am closing this PR.

PR #2912, opened 2019-02-06
Item on drag between lists has wrong width - bug fix

Dariia:
### 🔦 Summary
...

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [x] Other (explain)
No need documenting, bug fix of existing functionality
- 🔬 Change is tested
  - [ ] Component
  - [x] Visual test

### DIFF
--- a/src/DragAndDrop/Draggable/components/DraggableSource.js
+++ b/src/DragAndDrop/Draggable/components/DraggableSource.js
@@ -212,6 +212,7 @@ export default class DraggableSource extends React.Component {
   }
 
   _setRootNode = node => {
+    // Don't need to reset the values if node remains the same
     if (node && this.rootNode !== node) {
       this.rootNode = node;
       this.setState({itemWidth: this.rootNode.getBoundingClientRect().width});
--- a/src/DragAndDrop/Draggable/components/DraggableSource.js
+++ b/src/DragAndDrop/Draggable/components/DraggableSource.js
@@ -223,7 +223,7 @@ export default class DraggableSource extends React.Component {
     return renderItem({
       id,
       item,
-      previewStyles,
+      previewStyles: {width: this.state.itemWidth, ...previewStyles},
       isPreview: true,
       connectHandle: noop,
     });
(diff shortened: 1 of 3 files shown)


src/DragAndDrop/Draggable/components/DraggableSource.js
```diff
@@ -237,6 +237,8 @@ export default class DraggableSource extends React.Component {
         usePortal={usePortal}
         renderPreview={this._renderPreview}
         id={id}
+        // pass 'width' this prop to rerender element with changed width from state
+        width={this.state.itemWidth}
```
shlomitc: Will it support ‘100%’?

src/DragAndDrop/Draggable/components/DraggableSource.js
```diff
@@ -208,12 +211,19 @@ export default class DraggableSource extends React.Component {
     );
   }
 
+  _setRootNode = node => {
+    if (node && this.rootNode !== node) {
+      this.rootNode = node;
+      this.setState({itemWidth: this.rootNode.getBoundingClientRect().width});
```
Faradey27: a little scare about this, `if (node && this.rootNode !== node) {`, maybe let left some comment to explain why we need to do this and why it safe

src/SortableList/API.md
```diff
@@ -89,6 +89,7 @@ Some details about complex props
   so you able to style your item by cheking isPreview.
   - `id` - an id from item that you render
   - `previewStyles` - styles that coming from SortableList, `you always need to apply` them on your root div, inside of renderItem
+  You can add default item width as style={{...previewStyles, width: your_width }}
```
Faradey27: maybe let mention somewhere that we are passing width inside of previewStyles

src/DragAndDrop/Draggable/components/DraggableSource.js
xmityaz: It won't recognize `100%`, because the draggable preview is `position: fixed`.
But actually the value in state represents `width: 100%` in absolute value of pixels.

PR #2915, opened 2019-02-07
`<Page/>` - Change PageContainer constraints. Add stretch-vertically. Add bottom-padding. Fix gap. (`upgrade` prop)

### LINKED ISSUE #2888: `<Page/>` - Use `height: 100%`, and deprecate requirements for page container styles

hepiyellow:
### General
- It resolves #2888 and more.
- Adds an `upgrade` prop since there are a few breaking changes.

### Changes Overview

- Makes the page use `height: 100%` instead of depending on a parent with `display: flex;flex-flow: column`
- Consumer content may stretch vertically to fill the viewport, by setting `min-height: inherit`
- Minimum bottom padding (of 48px) is always applied (previously there was padding only in the minimized state)

### Changes

#### Examples

####

#### Code

- [ ] Investigate Stratrched Table in Page Example. Weird Gap on refresh, dissapears when navigating stories.
  - [ ] In order to make `<Card/>` stretch, need that `stretchVertically` prop will apply `min-height: inherit`
- [ ] Add window resize listener
- [ ] FixedContent is Jumping (Need to use the sticky technique)
- [ ] minimizedHeader calculation is naive(wrong), need to calculate by headerContainer reference after minimization

#### Missing Tests

- [ ] Window resize
- [ ] Test Stories 10 - 15

#### Documentation

- [ ] Storybook: Can't have `2.5 Page Example` folder next to the `2.5 Page`
- [ ] How do we show the code of the examples?


hepiyellow:
Replaced by #2929

PR #2917, opened 2019-02-07
#EE-16832| changes in Page (background opacity, classNames, dataHooks)

varzager:
changes in `Page`: 
- fixing page background to be opaque 
- added classNames and dataHooks to nested components of Page (fixedContent, content, tail)
- Page accepts also inline conditional rendering

- [ ] 👨‍💻 API change is approved by the UI Infra Developers @hepiyellow 
- [✅] 👨‍🎨 UX change is approved by the Design System UX @donatas 
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [✅] 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test

### DIFF
--- a/src/Page/FixedContent.js
+++ b/src/Page/FixedContent.js
@@ -1,13 +1,13 @@
-import PropTypes from 'prop-types';
+import PropTypes from "prop-types";
 
-const Content = props => props.children;
-Content.displayName = 'Page.Content';
-Content.propTypes = {
+const FixedContent = props => props.children;
+FixedContent.displayName = 'Page.FixedContent';
+FixedContent.propTypes = {
   children: PropTypes.element.isRequired,
-  fullScreen: PropTypes.bool,
   /** A css class to be applied to the component's root element */
   className: PropTypes.string,
   /** A data-hook to be applied to the component's root element */
   dataHook: PropTypes.string,
 };
-export default Content;
+
+export default FixedContent;
--- a/src/Page/Page.driver.js
+++ b/src/Page/Page.driver.js
@@ -1,26 +1,38 @@
-import { isClassExists } from '../../test/utils';
+import {isClassExists, findByHook} from '../../test/utils';
 
-export default ({ element }) => ({
+
+export default ({element}) => {
+  const fixedContent = () => findByHook(element, 'page-fixed-content');
+  const content = () => findByHook(element, 'page-scrollable-content');
+  const tail = () => findByHook(element,'page-tail');
+
   /** fulfilled if element in the DOM */
-  exists: () => !!element,
-  hasClass: className => isClassExists(element, className),
+  return {
+    exists: () => !!element,
+    hasClass: className => isClassExists(element, className),
 
-  /** true if header background image exist */
-  backgroundImageExists: () =>
-    !!element.querySelector('[data-hook="page-background-image"]'),
+    /** true if header background image exist */
+    backgroundImageExists: () =>
+      !!element.querySelector('[data-hook="page-background-image"]'),
 
-  /** true if gradient class name exist */
-  gradientClassNameExists: () =>
-    !!element.querySelector('[data-hook="page-gradient-class-name"]'),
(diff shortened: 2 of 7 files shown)


hepiyellow:
@varzager 
1. I have re-written the `<Page/>` component, so it's difficult to resolve conflicts now, so I am putting Page in freeze until I merge my [WIP PR](https://github.com/wix/wix-style-react/pull/2929).
2. I don't agree with the `className` props that's you've added, and where you apply them to internal Page elements. Please bring a use-case that you need! (Where adding class to your own `FixedContent` doesn't work for you, and adding it to the `FixedContent` wrapper does)
3. I am not sure I understand what you are trying to test exactly. If you want to test if `tailExists`, you can pass it with a `data-hook` and check if your data-hook exists. The only use-case I can think of is that if you're passing a Tail and a FixedContent and you want to verify that the `Page` component doesn't mix between them. This, you can't test, but also you should "trust" us.

varzager:
@hepiyellow thanks for the response. here r my comments:

- `className` - we have a few elements in `fixedContent`, and want to set it some `background-color`, so i either add a redundant div to give it a `className`.
- `tailExists` was a refactor i made for ur code
- `dataHooks`  - to my knowledge u have a HOC that adds `dataHooks` to all `elements`, and here it seemed it was just missing.
- `className` also seems to be in all components, and give us the freedom to stye our view to fit our needs.
- i added also the opaque bug fix
- and added inline conditional rendering - making page accept `false` as a child node - the react way. also needed for `fixedContent` to show it only in some condition

why would u encourage PR's if u'r planning to make a big refactors and dont want to accept them?

hepiyellow:
@varzager Hi,
Actually, the opacity change can be merged.
Can you remove all other stuff? so we can merge it independently?
Or you can open a separated PR for the opacity only. And we'll close this one.

You are welcome to call me and discuss again the data-hooks and classes use-cases.

hepiyellow:
@varzager 
About your comments:
- *className*: You may add a redundant div. It is best-practice not to mix internal layouting and consumer styling on the same div, in render-slots. There could be collisions of CSS, and then CSS load order starts to kick in, and you get unpredicated behaviour. However, we could open a `fixedContentWrapperStyle` prop, for inline styles, but that seems an overcomplication of API.

- *tailExists*: oops, you are right.
- *data-hooks*: The `Page.FixedContent` and `Page.Content` are no ordinary components. They are "named-render-slots" you should imagine that they are simple Props, but are passed as children (to enjoy the JSX syntax). So... they don't render anything of their own to the DOM, they just render their children "as-is". Which means, that YOU can pass any data-hook you want on any element you want, and fetch it in your test.
- *opaque*: please make a separated PR for the opaque change
- *conditional rendering*: This looks good, if you can make a PR only for that,  we can add it.

### Encouraging PRs

We are always encouraging PR and keen on merging them.
My re-write of `<Page/>` should not block you. 

- Most of this PR includes things I didn't agree with, that's why I closed it.
- The opaque change is good, but it's so small, I already did it in my re-write, and it will be available soon. And it doesn't seem urgent.
- The conditional rendering fix I simply missed. Sorry.

> Please for future note, it is always recommended to open separate PRs for independent issues. Even if it is a PR with one line, eventually it makes things merge faster.

varzager:
as i said on the other PR.
if the new `page` is will come out at the near time, and it includes the `opaque` and the `conditional rendering`, so its ok to wait.

btw in my opinion classNames are better than adding redundant dom elements. and many times we find are self in need for changing css that dont necesarly align exactly with ur designs.
same for dataHooks, we use this for our tests for some specific needs.

but in general, i think that it should be consistent. at the moment u usually support `className` and `dataHook`, i will always assume i will have them if i need

hepiyellow:
Closing for inactivity.
@varzager You are welcome to make a new PR adding driver methods to get content of `Page.FixedContent`, `Page.Content`, `Page.Tail`

PR #2919, opened 2019-02-07
fix lint warnings

argshook:
1. fix some recent lint warnings (how do they go through if there's `precommit` hook?)
2. ensure generated component is linted correctly


mykas:
Precommit handles only errors I think.

argshook:
right, it only does `yoshi lint` which fails on errors only https://github.com/wix/wix-style-react/blob/master/package.json#L174

if it were `prepush` it could be `yoshi lint --fix` and prevent from pushing if there're were fixed chnages.

PR #2929, opened 2019-02-09
`<Page/>` - New Generation! (`upgrade` prop)

### LINKED ISSUE #2888: `<Page/>` - Use `height: 100%`, and deprecate requirements for page container styles

hepiyellow:
### General
- It resolves #2888 and much more.
- Adds an `upgrade` prop since there are a few breaking changes.

### Changes Overview

- **PageContainer**: Makes the page use `height: 100%` instead of depending on a parent with `display: flex;flex-flow: column`
- **stretch-vertically**: Consumer content may stretch vertically to fill the viewport, by setting `min-height: inherit`
- **Bottom Padding**: Minimum bottom padding (of 48px) is always applied (previously there was padding only in the minimized state)
- **Sticky**: FixedContent now scrolls with the content until it gets sticky (instead of jumping with the minimized Header
- **minimized Header DOM calculation**:  Fix HeaderContainer hard-coded minimized calculations - Now it measures the DOM after minimization.

### Changes Details

#### Examples

#### UX
@milkyfruit @wuwa @LittleTal 

- **Small Scroll** - When the scroll is small, but enough to trigger the minimization, and not enough to reach the bottom of the minimized header. **Should we allow extra scroll to get to top**? (Or all the way with some stop -padding, since minimized Header has no bottom padding)
- Are there 2 stretch options: 
  - Stretch till the end of the viewport? (Filtered Table case)
  - Stretch until bottom padding starts? (EmptyState)

#### Code

- [ ] In order to make `<Card/>` stretch, need that `stretchVertically` prop will apply `min-height: inherit`
  - We might want to provide a `stretchVertically` class/style, so that we have freedom in the future to change 
- [ ] *minimizedHeader calculation*  - I disblaed PageHeader transition. Need to Use `<CSSTransition/>`
- [ ] *Stretch Vertically* - I added window resize eventListener, but Page height may change regardless of window (Not relevant for current BM layout)

#### Missing Tests

- [ ] Window resize
- [ ] Test Stories 11 - 15 (Also with FixedContent, 21-25)
- [ ] Scroll tests with Tail, or different Header combinations, sizes.

#### Documentation

- [ ] Storybook: Can't have `2.5 Page Example` folder next to the `2.5 Page`
- [ ] How do we show the code of the examples?

Replaces #2915


hepiyellow:
Replaced by #2935

PR #2931, opened 2019-02-10
 #EE-16891|page| add transition on scroll to improve experience

varzager:
add transition to header minimize  to improve scroll down experience. 

### ✅ Checklist
- [✅ ] 👨‍💻 API change is approved by the UI Infra Developers @hepiyellow 
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


hepiyellow:
@varzager 
Hi, 
thanks for the PR !
However, I have re-written the Page component ([WIP](https://github.com/wix/wix-style-react/pull/2929)), and removed animations.
It seems that it is necessary to use `<CSSTransition/>` if we want to apply any animations.
This is because for minimized `headerContainer` height measurements to be accurate, we need to measure them **only after** it was minimized (transition ended).

Please open an issue requesting animation.
Let's have UX guys define the exact transitions.
And we'll add it later.

varzager:
@hepiyellow 
i thought we agreed on this solution on Thursday when we met...
js animation really gives a good solution to avoid race conditions - what happens first the css change, or the call of the js function that calculates the page.
but we found out there is no need to recalculate when the header changes from minimize to maximize..
sent a video with the solution to @milkyfruit

src/Page/Page.scss
```diff
@@ -34,6 +34,7 @@ $image-container-offset: 30px;
 }
 
 .page-header-container {
+  transition: padding .2s ease-in-out;
```
hepiyellow: I think we want to use `wix-animation` here like we use in `<PageHeader/>`.
https://github.com/wix/wix-style-react/blob/8b77dd51ba741f8e779942474d2074a00fd39249/src/PageHeader/PageHeader.js#L13

https://wix-wix-animations.surge.sh/?selectedKind=Animator&selectedStory=Basic%20Animation&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

src/Page/Page.js
```diff
@@ -83,14 +83,18 @@ class Page extends WixComponent {
...
-    if (!this.state.minimized) {
+  calculateHeightIfHeaderIsntMinimizedAndWasntUpdated(prevState) {
+    if (!this.state.minimized && this.state.minimized === prevState.minimized) {
       this._calculateComponentsHeights();
     }
   }
 
+  componentDidUpdate(prevProps, prevState) {
+    super.componentDidUpdate(prevProps);
+    this.calculateHeightIfHeaderIsntMinimizedAndWasntUpdated(prevState);
```
hepiyellow: You don't need to extract a function if the name is so long (-:

Just write a comment.

hepiyellow:
@varzager 
I agree with you that this transition you added helps.

I am working on a Page NextGeneration PR, and it makes your changes obsolete.
So if you are capable of waiting a week with this, and then upgrade to the new Page, you will get a lot of goodies !

Here is a demo:
There is no minimization of the header, the header scrolls up, and a new minimized-header is coming instead of it. It should have a transition (not yet implemented), probably gliding from the top.

https://wix-wix-style-react-pr-2935.surge.sh/?selectedKind=2.%20Layout%2F2.5%20Page%20Examples&selectedStory=5.%20Sticky%20Table%20With%20Gap&full=0&addons=0&stories=1&panelRight=0

If you still want to merge this PR, please follow the change requests, but I think it is not worth your effort.

milkyfruit:
@hepiyellow is this PR still relevant? I shared a rough header transition prototype with Erez via Slack.

hepiyellow:
@milkyfruit if guy decides they cant wait he may finish this PR for the old design, with your approval. 
So answer us, ask guy.

src/Page/Page.js
varzager: actually this is the exact opposite of our principles.. but its ur code, so sure

varzager:
if ur going a head with the other design, and it suppose to go out at the near time.. 
so i guess there is no need for this solution

PR #2932, opened 2019-02-10
<RichTextInputArea/> - add e2e tests

nitayneeman:
This PR adds e2e for basic behavior and design.

The issue:
https://github.com/wix-private/wsr-issues/issues/35


stories/RichTextInputArea/examples/ExampleRichTextInputArea.js
```diff
@@ -0,0 +1 @@
+export default `<p>Hello <strong>World</strong></p><p>The text goes here</p>`;
```
shlomitc: this should be the entire component and not just the string

stories/RichTextInputArea/examples/ExampleRichTextInputArea.js
```diff
@@ -0,0 +1,6 @@
+export default `
```
shlomitc: in order to maintain a js syntax, you can just write the component in a js and import it using `raw-loader`
https://github.com/wix/wix-style-react/blob/master/stories/Thumbnail/index.story.js#L17

nitayneeman:
Hi @argshook,

Maybe do you have a guess why ReactTestUtils & Enzyme sanity tests are failing?
I'm getting the following error for both:
```
Expected received Promise to resolve, instead it rejected to value
[TypeError: driverFactory is not a function]
```

Also, I've executed `npm run build` so all testkits are added.

argshook:
@nitayneeman the reason is two exports in `src/RichTextInputArea/RichTextInputArea.driver.js`

this would fix it:

```diff
src/RichTextInputArea/RichTextInputArea.driver.js
@@ -3,7 +3,7 @@ import ReactTestUtils from 'react-dom/test-utils';
 
 export const getTextArea = base => base.$('.public-DraftEditor-content');
 
-export const richTextInputAreaDriverFactory = base => {
+export default base => {
   return {
     ...baseUniDriverFactory(base),
     getContent: () => base.text(),
```

with two exports sanity fails, because currently there's an assumption that
`driver.js` files export just one thing - the testkit.

logic is: `if export.default then use export.default; otherwise use
first named export`

according to this logic, given your case the teskit is assumed to be
`getTextArea`, which is not correct and thus smoke test fails.

This can be circumvented by either:
* never having multiple exports in `driver.js` file
* upgrading `testkit-definitions` to support `namedExport` or similar, and set it there
* upgrading `smoke-test` and `generate-testkits-exports` to assume that exports named with `TestkitFactory` ending are actual testkits

for now i would go with simplest approach and just have single export
in testkit file. i'd argue that all exported functions in `.driver.js`
become public API which means we need to maintain it with care. i'd try
to avoid that.

PR #2934, opened 2019-02-10
<ColorPicker/> - add methods onMouseDownConfirm and onMouseDownCancel

mykas:
### 🔦 Summary

Added additional mouse handlers for `Confirm` and `Cancel` buttons. The reason for this is to enable `onMouseDown` functionality for popover based `ColorPickers`.

Usually when ColorPicker is used inside Popover - very often onClickOutside functionality is used .For example in `Input` used together with ColorPicker. 

I.e. if `Input` is depending on `onBlur` or `onClickOutside` events then users will not able to click on `ColorPickers` confirm or cancel buttons. The point is the onBlur is triggering a re-render which seems to lead the browser to do not follow up with the onClick. 

Additionally I have removed `onCancel` and `onConfirm` proptypes required rule. In this case if two kind of mouse event handlers can be used - user needs to choose which one to use. And also I don't think that `oCancel` and `onConfirm` is seriously needed to be passed to render the component. It would be strange that user would not handle those by default. 

### Solution

Add onMouseDown event handlers which will not trigger popover childs `onBlur`  or `onClickOutside`methods.

### ✅ Checklist
- [ ] 👨‍💻 API change is approved by the UI Infra Developers  @hepiyellow @argshook @shlomitc  
- 📚 Change is documented
  - [x] Story
  - [x] API description
- 🔬 Change is tested
  - [x] Component

### DIFF
--- a/src/ColorPicker/ColorPicker.driver.js
+++ b/src/ColorPicker/ColorPicker.driver.js
@@ -7,8 +7,16 @@ export const colorPickerDriverFactory = ({ element }) => {
       ReactTestUtils.Simulate.click(
         element.querySelector('[data-hook="color-picker-confirm-button"]'),
       ),
+    mouseDownConfirm: () =>
+      ReactTestUtils.Simulate.mouseDown(
+        element.querySelector('[data-hook="color-picker-confirm-button"]'),
+      ),
     cancel: () =>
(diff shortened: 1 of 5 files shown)


mykas:
Not relevant anymore.

PR #2944, opened 2019-02-12
Add e2e runtime tests against `dist` with GC and yoshi app

yavorsky:
### 🔦 Summary
**TLDR; This is an implementation of e2e runtime tests with real yoshi app + real chrome browser + components from `dist` (not `src` like our current tests)**

**What's the problem?**
Unit tests are working against `src` directory. We are not checking actual content we provide to users. If something happens with our babel configuration, dist content, files to be copied presence - we'll face some problems.

Current e2e tests are working inside react-storybook bundle generated by it's own config. This bundle have nothing in common with bundle user's will use. Moreover, all components included in RS bundle are from the `src` dir also.

So seems like `dist` part remains **uncovered at all**.
This PR is an intention to solve it.

Here we have new dead-simple project with basic configuration. This project contains 3 routes.
This routes should represent and cover all ways to use WSR.

1. **Commonjs**: Covers `dist/src` with `module.exports = WSR`. So we'll use `const Text = require('wix-style-react/Text')`.
2. **ES**: Covers `dist/es/src` with `export const Text`. So we'll use `import Text from 'wix-style-react/Text'`.
3. **ES Named `import`s** The same as ES exports, but we are tying to use `import { Text } from 'wix-style-react'`. 

#### How it's working?
We have project with 3 routes.  `express` server checks `type` query param and pass it to VM. VM is rendering page with needed bundle. Each test suite is navigating to it's route and check needed bundle.
`localhost:3005/?type=es` -> `es.bundle.js`, etc.
If there are errors, then puppeteer will fail with some needed selectors.

So here we covered all possible ways to use WSR Components with real yoshi app + real chrome browser.


test/e2e-runtime/imports/src/es-named/App.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { Text } from '../../../../../dist/es/src';
```
shlomitc: shouldn't this be taken from the `index` file instead?

PR #2945, opened 2019-02-12
on show added to PopoverMenu

adirou:
### 🔦 Summary
onShow prop added to popover menu 

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test

### DIFF
--- a/src/PopoverMenu/PopoverMenu.js
+++ b/src/PopoverMenu/PopoverMenu.js
@@ -40,6 +40,7 @@ class PopoverMenu extends WixComponent {
     zIndex: number,
     showArrow: bool,
     onShow: func,
+    onHide: func,
   };
 
   static defaultProps = {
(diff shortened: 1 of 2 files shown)


src/PopoverMenu/PopoverMenu.js
```diff
@@ -105,6 +107,8 @@ class PopoverMenu extends WixComponent {
         appendToParent={this.props.appendToParent}
         zIndex={this.props.zIndex}
         showArrow={this.props.showArrow}
+        onShow={this.props.onShow}
```
shlomitc: please add two tests for the new props - when `onShow` is passed, it is called, when `onHide` is passed it is called.

also, if prop doesn't exist, it should not call the function (otherwise it will throw an error

PR #2952, opened 2019-02-13
<DropdownLayout/> - Fixed two scrollbars in dropdownLayout

adirou:
fix two scrollbars in dropdownLayout when there is fixed footer or header

### 🔦 Summary
...

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


mykas:
Hi @adirou , I would like to review your PR. Can you tell me the reason why you do height calculation in this case? I though that making dropdown container with `display:flex; flex-direction: column` is easier solution? But maybe I am missing something here.

adirou:
its looks like you are right , would you like me to fix the PR?

mykas:
Yes, please. Then we can check if any visual tests are breaking.

PR #2960, opened 2019-02-14
<CloseButton/> -  add support for custom icon

ipanasenko:
### 🔦 Summary

PR adds custom icon support for <CloseButton/>. Previously <CloseButton/> was only meant for X icon.

### ✅ Checklist
- [x] 👨‍🎨 UX change is approved by the Design System UX  @milkyfruit 
- 📚 Change is documented
  - [x] Story
  - [x] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [x] Visual test


src/CloseButton/CloseButton.spec.js
```diff
@@ -9,13 +10,36 @@ describe('CloseButton', () => {
 
   describe('Icon', () => {
     it('when given small or none should have size <Close/>', () => {
-      const wrapper = mount(<CloseButton size="small" />);
+      const wrapper = shallow(<CloseButton size="small" />);
```
mykas: Why do we need shallow for this?

ipanasenko: It doesn't make a difference here. I needed `shallow` for new tests, decided to replace it here. I can revert this line if needed

mykas: Nevermind, I think its ok. We have some places where we use it anyway.

mykas:
Lets wait until CI runs builds and visual tests and after visual approve I will merge this. Thanks a lot for the PR by the way!

PR #2973, opened 2019-02-18
Autodocs/registry

hepiyellow:


.autotools/node-require-hooks.js
```diff
@@ -0,0 +1,3 @@
+require('yoshi-helpers').setupRequireHooks();
```
yavorsky: we don't have it as a dependency in `package.json`.

.autotools/node-require-hooks.js
```diff
@@ -0,0 +1,3 @@
...
+require('yoshi-runtime').wixCssModulesRequireHook('./src');
```
yavorsky: we don't have it as a dependency in `package.json`.

PR #2976, opened 2019-02-18
WIP: Funnel chart component migration

Faradey27:
TODO:
- [ ] jsdom and protractor drivers(unidriver?)
- [ ] Add jsdom tests
- [ ] Add protractor tests
- [ ] Replace Fragment polyfil with dom parent

### DIFF
--- a/src/FunnelChart/UnitsHeading.scss
+++ b/src/FunnelChart/UnitsHeading.scss
@@ -1,0 +1,4 @@
+.heading {
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
--- a/src/FunnelChart/_variables.scss
+++ b/src/FunnelChart/_variables.scss
@@ -1,0 +1,5 @@
+$bar-width: 65;
+$color-B05: #308ddd;
+$color-B40: #daeffe;
+$color-D55: #cbd3dc;
+$color-D70: #f0f4f7;
--- a/src/FunnelChart/_variables.scss
+++ b/src/FunnelChart/_variables.scss
@@ -1,5 +1,1 @@
 $bar-width: 65;
-$color-B05: #308ddd;
-$color-B40: #daeffe;
-$color-D55: #cbd3dc;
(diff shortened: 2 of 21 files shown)


src/FunnelChart/index.ts
```diff
@@ -0,0 +1,3 @@
+export { default } from './FunnelChart';
```
shlomitc: `index.js`

src/FunnelChart/index.ts
```diff
@@ -0,0 +1,3 @@
...
+
+export { default as EmptyFunnelChart } from './EmptyFunnelChart';
```
shlomitc: Please export only a single component from an index file and not multiple ones.

Maybe instead of a new component, this empty state should be part of the FunnelChart component and displayed in a case that no data was supplied?

src/FunnelChart/_variables.scss
```diff
@@ -0,0 +1,5 @@
+$bar-width: 65;
+$color-B05: #308ddd;
```
shlomitc: these colors should be taken from the `colors` pallette

src/FunnelChart/UnitsHeading.scss
```diff
@@ -0,0 +1,4 @@
+.heading {
```
shlomitc: can't this be achieved with the `Text` component in ellipsis mode?

src/FunnelChart/index.ts
Faradey27: fixed

src/FunnelChart/index.ts
Faradey27: i like this idea and pushed such logic

@BrooklynKing wdyt?

src/FunnelChart/UnitsHeading.scss
Faradey27: @BrooklynKing can we?

PR #2981, opened 2019-02-19
`<Page/>` (Docs): Write new UX Page Layout story

hepiyellow:


milkyfruit:
@hepiyellow 
To fill description with at least some info here's a short message:

> **Description**
> Compositions from multiple components that help to setup a page.

Everything else, besides the incoherent vertical spacing, I think looks ok:
image

PR #2989, opened 2019-02-19
<RichTextArea> - separate current story page to index, api and test pages

nitayneeman:
This PR separates the current RichTextArea's story page into three different pages (index, API and tests).

Note: A PR with deprecation for `3.2b RichTextAreaComposite` will be opened separately.


mosheKerbel:
plz take a look at the two screenshots
1. like the spec
2. we put the top part in 100%
to us option 2 looks better but it's different than the spec
WDYT?

image

image

milkyfruit:
@mosheKerbel

- Let's remove the example that is above the description. Then this page would be aligned with 3.1 and 3.2 stories.

- Remove the second sentence from the general description, the example with styling options will be added at the bottom.

- I see two "import" titles, looks like a bug:
image

- Rename first example Form Field's label to "Rich Text Area" and remove "infoContent" attribute. It will be demonstrated in the second example. Change first example's description to:
**Title**: Plain Example
**Description**: Default rich text area setup.

- Let's add additional example, that demonstrates styling:
image
**Tile**: Text Styling
**Description**: Rich text area supports two types of bullet point styles, basic text styling and hyperlinks.

PR #3005, opened 2019-02-21
<NoBorderInput/> - add support for focus props

Jefoss:
Added the support for the focus props - autoFocus, onFocus and onBlur


src/NoBorderInput/NoBorderInput.js
```diff
@@ -78,8 +80,18 @@ class NoBorderInput extends React.Component {
           className={styles.nbinput}
           {...wsrInputProps}
           ref={wsrInput => (this.wsrInput = wsrInput)}
-          onFocus={() => this.setState({ focus: true })}
-          onBlur={() => this.setState({ focus: false })}
+          onFocus={() => {
```
shlomitc: isn't this logic implemented already in `Input.js`? why not just to delegate it?

stories/NoBorderInput/NoBorderInputTestStories.js
```diff
@@ -0,0 +1,24 @@
...
+
+const kind = getTestStoryKind({
+  category: storySettings.category,
+  storyName: storySettings.storyName,
+});
+
+storiesOf(kind, module).add('1. NoBorderInput with auto focus', () => {
+  return (
+    <div>
+      <h1>NoBorderInput with auto focus</h1>
```
shlomitc: I would remove all text and keep example only

src/NoBorderInput/NoBorderInput.e2e.js
```diff
@@ -29,15 +30,27 @@ describe('NoBorderInput', () => {
...
 
-  beforeAll(async () => {
+  eyes.it('should render', async () => {
     await browser.get(storyUrl);
+    await createDriver();
   });
 
-  eyes.it('should render', async () => {
-    await createDriver();
+  eyes.it('should auto focus the input', async () => {
```
shlomitc: in case this is actually needed (see the previous comment), can't it be tested in jsdom?

src/NoBorderInput/NoBorderInput.js
Jefoss: What will be passed to the `Input` will always be a function, but the prop passed to the component could be something else and invoking it will fail

src/NoBorderInput/NoBorderInput.e2e.js
```diff
@@ -29,7 +29,7 @@ describe('NoBorderInput', () => {
     return driver;
   };
 
-  beforeAll(async () => {
+  beforeEach(async () => {
```
shlomitc: revert

src/NoBorderInput/NoBorderInput.spec.js
```diff
@@ -0,0 +1,76 @@
...
+    const onBlur = jest.fn();
+    const driver = createDriver(renderNoBorderInputWithProps({ onBlur }));
+
+    driver.focus();
+    driver.blur();
+
+    expect(onBlur).toHaveBeenCalled();
+  });
+
+  describe('testkit', () => {
```
shlomitc: remove

src/NoBorderInput/NoBorderInput.spec.js
```diff
@@ -0,0 +1,76 @@
...
+    it('should exist', () => {
+      const component = renderNoBorderInputWithProps();
+      expect(isTestkitExists(component, noBorderInputTestkitFactory)).toBe(
+        true,
+      );
+    });
+  });
+
+  describe('enzyme testkit', () => {
```
shlomitc: remove

PR #3008, opened 2019-02-21
Add Generic Unidriver Enhancer

hepiyellow:
What do you think?

We can use it to add missing Unidriver features, or convenience methods like `byDataHook`

### DIFF
--- a/test/utils/unidriver/enhance-unidriver.js
+++ b/test/utils/unidriver/enhance-unidriver.js
@@ -2,7 +2,7 @@
  * Adds extra methods to a Unidriver
  *
  * @param {*} baseUniDriver
- * @param {*} createExtraMethods receives (baseDriver, enhacen, enhanceList), returns an object with extra methods. Extra methods may override base unidriver methods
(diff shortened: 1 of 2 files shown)


test/utils/unidriver/enhance-unidriver.js
```diff
@@ -0,0 +1,31 @@
...
+      map: baseUniDriverList.map,
+      filter: predicate => enhanceList(baseUniDriverList.filter(predicate)),
+    };
+  }
+
+  return {
+    ...rest,
+    $: selector => enhance(baseUniDriver.$(selector)),
+    $$: selector => enhanceList(baseUniDriver.$$(selector)),
+    ...createExtraMethods(baseUniDriver, enhance, enhanceList),
```
nitayneeman: I'm wondering what will happen if a consumer would provide a method from UniDriver API in `createExtraMethods` accidentally (or not?), for instance: `$`, `exists` and so on. Does it mean we allow the consumer to override the existing API?

hepiyellow: But you might be right that it's better NOT to allow it.

hepiyellow:
Not for now. Keeping branch for later.

PR #3011, opened 2019-02-21
<NoBorderInput/> - style adjustments

Jefoss:
all updates according to the agreed [design](https://app.zeplin.io/project/5bd9d21f585e2b0a697870f4/screen/5c458cd0c50a21377f0da1a8)

- Added auto selection when focusing the input
- Placeholder style fixes
- Label style fixes


src/NoBorderInput/NoBorderInput.spec.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import noBorderInputDriverFactory from './NoBorderInput.driver';
 import { createRendererWithDriver, cleanup } from '../../test/utils/unit';
-
+import Input from '../Input/Input';
```
shlomitc: this looks incorrect. it should be taken from the `NoBorderInput` component instead

src/NoBorderInput/NoBorderInput.js
```diff
@@ -108,14 +111,16 @@ class NoBorderInput extends React.Component {
...
+  autoSelect: true,
   size: 'normal',
   statusMessage: '',
   textOverflow: 'clip',
   maxLength: 524288,
 };
 
 NoBorderInput.propTypes = {
+  /** The label displayed above the input when focused and as the input text when there is none */
+  label: PropTypes.string,
```
shlomitc: please add to the `story` page some value. so it will appear in the `primary props` and not just `html props`
image

Jefoss: added

PR #3017, opened 2019-02-22
Add `.npmignore` with original dirs and dev configs.

yavorsky:
It should reduce `npm install wix-style-react` time. We don't need to publish original directories, .md files, development configs, etc.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -17,5 +17,3 @@ CHANGELOG.md
 protractor.conf.js
 babel.config.js
 depmonkey.ignore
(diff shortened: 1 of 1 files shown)


.npmignore
```diff
@@ -0,0 +1,19 @@
+# original dirs
```
argshook: this file is not needed, because we define what to include in module through `files` in package.json https://github.com/wix/wix-style-react/blob/master/package.json#L13-L21

PR #3021, opened 2019-02-24
<Tooltip/> - unidriver migration

mykas:
This PR adds unidriver for Tooltip except for hoverAndGetContent method which will be done later.


src/Tooltip/Tooltip.unidriver.js
```diff
@@ -0,0 +1,89 @@
...
+    getTooltipWrapper: async () =>
+      await (await getContentRoot()).$('.tooltip').getNative(),
+    mouseEnter: async () => await base.hover(base),
+    mouseLeave: async () => await ReactBase(base).mouseLeave(),
+    hasErrorTheme: async () => await hasContentClassName('error'),
+    hasDarkTheme: async () => await hasContentClassName('dark'),
+    hasLightTheme: async () => await hasContentClassName('light'),
+    hasAnimationClass: async () =>
+      await (await getContentRoot()).$('.fadeIn').exists(),
+    getChildren: async () => await base.text(),
```
hepiyellow: The original driver returns `innerHTML`. please use `reactBase.innerHTML`

src/Tooltip/Tooltip.unidriver.js
```diff
@@ -0,0 +1,89 @@
...
+    getContent: async () => await (await getContentRoot()).$('.tooltip').text(),
```
hepiyellow: getContent used to find deepest child (which has no more children) and return it's `innerHTML`.
- Please extract a `getTooltipContent` with same implementation as before , and use it in other places.

src/Tooltip/Tooltip.unidriver.js
```diff
@@ -0,0 +1,89 @@
...
+    getPlacement: async () => await getArrowPlacement(),
+    getMaxWidth: async () => await getContentStyleValue('max-width'),
+    getMinWidth: async () => await getContentStyleValue('min-width'),
+    getAlignment: async () => await getContentStyleValue('text-align'),
+    getPadding: async () => await getContentStyleValue('padding'),
+    hasArrow: async () =>
+      await (await getContentRoot()).$('[data-hook="tooltip-arrow"]').exists(),
+    getContentHook: async () =>
```
hepiyellow: `getContentHook` should not be exposed. This is a new method, did you add it by mistake? if it is for private test then move it to a private driver.

src/Tooltip/Tooltip.unidriver.js
```diff
@@ -0,0 +1,89 @@
...
+      .attr('class');
+
+    return arrowDirection[classes.split(' ')[2]];
+  };
+
+  const hoverAndGetContent = async (timeout, interval) => {
+    await base.hover(base);
+    return eventually(
+      async () => {
+        if (!(await getContentRoot()).exists()) {
```
hepiyellow: `exists` returns a promise. Need to add `await` for it.
I think its better how it is in the original driver, where you extract an `isShown` function and use it both here and expose it.

src/Tooltip/Tooltip.unidriver.js
```diff
@@ -0,0 +1,89 @@
...
+    hasArrow: async () =>
+      await (await getContentRoot()).$('[data-hook="tooltip-arrow"]').exists(),
+    getContentHook: async () =>
+      await body
+        .$(`[data-hook="${await base.attr('data-content-hook')}"]`)
+        .getNative(),
+    hoverAndGetContent: async ({ timeout = 1000, interval = 50 } = {}) =>
+      await hoverAndGetContent(timeout, interval),
+  };
+};
```
hepiyellow: Why didn't you add this `assertExistsWrapper`?

https://github.com/wix/wix-style-react/blob/4bfa02b285a81ac1de2f1e54d1e0882458a4eb1f/src/Tooltip/Tooltip.driver.js#L107

src/Tooltip/Tooltip.unidriver.js
```diff
@@ -0,0 +1,89 @@
...
+  };
+
+  const hoverAndGetContent = async (timeout, interval) => {
+    await base.hover(base);
+    return eventually(
+      async () => {
+        if (!(await getContentRoot()).exists()) {
+          throw new Error('Tooltip not visible');
+        }
+        const content = await (await getContentRoot()).$('.tooltip').text();
```
hepiyellow: Use the missing `new ReactBase(await getTooltipContent()).innerHTML`

> `getTooltipContent` is a missing abstraction

src/Tooltip/Tooltip.unidriver.js
```diff
@@ -0,0 +1,89 @@
...
+
+export const teskitTooltip = (base, body) => {
+  const getContentRoot = async () => {
+    const contentRootHook = await base.attr('data-content-hook');
+    return body.$(`[data-hook="${contentRootHook}"]`);
+  };
+
+  const hasContentClassName = async cls =>
+    new RegExp(cls).test(
+      await (await getContentRoot()).$('.tooltip').attr('class'),
```
hepiyellow: Use extracted `getTooltipContent`

src/Tooltip/Tooltip.unidriver.js
```diff
@@ -0,0 +1,89 @@
...
+      : undefined;
+  };
+
+  return {
+    click: async () => await base.click(),
+    focus: async () => await ReactBase(base).focus(),
+    blur: async () => await ReactBase(base).blur(),
+    isShown: async () => (await getContentRoot()).exists(),
+    getTooltipWrapper: async () =>
+      await (await getContentRoot()).$('.tooltip').getNative(),
```
hepiyellow: Use extracted `getTooltipContent`

src/Tooltip/Tooltip.unidriver.js
mykas: By mistake.

src/Tooltip/Tooltip.unidriver.js
mykas: Because I have no idea what a hell this does.

src/Tooltip/Tooltip.unidriver.js
mykas: what do you mean no need to add `await`? I have added it for getContentRoot which is async

PR #3024, opened 2019-02-24
Component-Generator: Updates

hepiyellow:
- Rename driver files:
  - Component.driver.js -> Component.uni.driver.js
  - Component.driver.private.js -> Component.private.uni.driver.js
- Add Component.meta.js : For autodocs-registry
- Remove `kind` prop from generated storySettings

> Component.private.uni.driver.js is better than Component.driver.private.js, since the suffix `uni.driver.js` tells us the file content's type so we can use it for eslint rules. `private` is an semantic addition so it should't even be in the file suffix. See #3025


package.json
```diff
@@ -202,7 +202,7 @@
     "wix-ui-test-utils": "^1.0.115"
   },
   "lint-staged": {
-    "*.{js,scss}": "yoshi lint"
+    "*.{js,scss},.eslintignore": "yoshi lint"
```
shlomitc: are you sure?

docs/contribution/TEST_DRIVERS.md
```diff
@@ -24,7 +24,7 @@ If your component's `unidriver` is composing other components that don’t use u
 
 ## Public and Private drivers
 
-1. The **Public** drivers (`component.driver.js`) are the ones that exposed to the consumers of the components. They should be simple abstractions over common actions (for example, selecting the third element in the dropdown).
+1. The **Public** drivers (`component.uni.driver.js`) are the ones that exposed to the consumers of the components. They should be simple abstractions over common actions (for example, selecting the third element in the dropdown).
 2. The **Private** drivers (`component.driver.private.js`) are used for actions on a component that should not be exposed to the user. For example, asserting a class name existance on some component.
```
shlomitc: should also rename private drivers

package.json
hepiyellow: The comp generator's `templates` folder has an `.eslintignore` file. So the we would try to lint the template files which are not valid JS files.
`lint-staged` runs `yoshi lint` only on the staged files, so it "doesn't see" the `.eslintignore` files, and throws errors on the template files. This fixes it.

hepiyellow:
We are using the ui-autotools meta files to configure a path for each component.
This path is used to build the autodocs-registry, so stories can access multiple parsed-source of different components. Such as displaying multiple API prop table (e,g, SegmentedToggle API tab)

package.json
hepiyellow: I opened a Question" issue: 
https://github.com/okonet/lint-staged/issues/584

hepiyellow: reverted

PR #3033, opened 2019-02-25
<NumberInput /> - new component

NitayRabi:
### 🔦 Summary
Add new component `<NumberInput />`
Supports `min` `max` `step` props with ticker.
- UX Story (Composite of `<FormField />` and `<NumberInput />`)
- Component Story
- TestPage

Followup - 

https://github.com/wix/wix-style-react/issues/3057

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [x] Visual test


src/Input/Ticker/Ticker.driver.js
```diff
@@ -0,0 +1,14 @@
...
+import styles from './Ticker.scss';
+
+export const tickerDriverFactory = base => {
+  return {
+    ...baseUniDriverFactory(base),
+    clickUp: () => base.$(`.${styles.up}`).click(),
+    clickDown: () => base.$(`.${styles.down}`).click(),
+    isUpDisabled: () => base.$(`.${styles.up}`).hasClass(styles.disabled),
+    isDownDisabled: () => base.$(`.${styles.down}`).hasClass(styles.disabled),
+    exists: () => base.exists(),
```
hepiyellow: No need for `exists()`, the `baseUniDriverFactory` adds it.

src/NumberInput/NumberInput.js
```diff
@@ -0,0 +1,97 @@
+import React from 'react';
+import Input from '../Input';
+import WixComponent from '../BaseComponents/WixComponent';
+
```
hepiyellow: You can remove `NumberInput.scss` file, since it is not in use.

stories/NumberInput.ux/storySettings.js
```diff
@@ -0,0 +1,7 @@
+import { Category } from '../storiesHierarchy';
+
+export const storySettings = {
+  kind: Category.INPUTS,
+  category: Category.INPUTS,
+  storyName: '3.4 Number Input',
```
hepiyellow: Last I;ve seen 3.4 is Time Input.
UX is refactoring this whole seciton.
Did you talk to any of them about the indexing of this component?
Maybe you should put it as `3.x  Number Input` for now

stories/NumberInput.ux/index.story.js
```diff
@@ -0,0 +1,115 @@
+import React from 'react';
```
hepiyellow: I actually like the `NumberInput.ux` folder name.
Did you discuss this with someone?
The convention until now was to put component stories under `stories/components`,
and UX stories under `stories`.
I like your solution better 😄 

Lets discuss this in next weekly, if you'll remember.

src/Input/Ticker/Ticker.driver.js
```diff
@@ -0,0 +1,14 @@
+import { baseUniDriverFactory } from 'wix-ui-test-utils/base-driver';
```
hepiyellow: Rename to `Ticker.uni.driver`

src/Input/Ticker/Ticker.driver.private.js
```diff
@@ -0,0 +1,9 @@
+import { tickerDriverFactory as publicDriverFactory } from './Ticker.driver';
```
hepiyellow: rename to `Ticker.private.uni.driver.js`

src/NumberInput/NumberInput.js
```diff
@@ -0,0 +1,97 @@
...
+class NumberInput extends WixComponent {
```
hepiyellow: I think we can stop using `WixComponent`.
I see you apply the data-hook yourself. thats good.
@shlomitc ?

src/NumberInput/NumberInput.js
```diff
@@ -0,0 +1,97 @@
...
+    );
+  }
+}
+
+NumberInput.propTypes = {
+  ...Input.propTypes,
+};
+
+NumberInput.defaultProps = {
+  ...Input.defaultProps,
```
hepiyellow: Is this redundant?

stories/NumberInput.ux/index.story.js
argshook: I donno, i wouldn't do that, it can look like NumberInput.ux is a file and not a folder.

src/NumberInput/NumberInput.js
```diff
@@ -0,0 +1,97 @@
...
+    this._applyChange((value, step) => value + step);
+  };
+
+  _decrement = () => {
+    this._applyChange((value, step) => value - step);
+  };
+
+  _applyChange(operator) {
+    const { value, onChange, step } = this.props,
+      numberValue = parseFloat(value || this.inputDOM.value) || 0,
```
hepiyellow: `parseFloat` does NOT handle localization.
Looking at `react-numeric-input`, they provide a [parse prop](https://github.com/vlad-ignatov/react-numeric-input/blob/79874ccbe2293873300953f32d91f2d70d32a871/src/NumericInput.jsx#L632), and call it before invoking `parseFloat`.

I am guessing that consumers may use a localized parser like [this one](https://github.com/AndreasPizsa/parse-decimal-number).

You can leave this, and open an Issue on it, and we'll add it later.
I usually collect all open issues in the PR's description, and process them after I merge.

NitayRabi: @shlomitc @argshook - What's your opinion ?

src/NumberInput/NumberInput.js
```diff
@@ -0,0 +1,97 @@
...
+      numberStep = step,
+      updatedValue = operator(numberValue, numberStep);
+    if (onChange && this._isInRange(updatedValue)) {
+      this._triggerOnChange(updatedValue);
+    }
+  }
+
+  _triggerOnChange(value) {
+    const { onChange } = this.props;
+    const nativeInputValueSetter = Object.getOwnPropertyDescriptor(
```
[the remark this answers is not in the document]
NitayRabi: @shlomitc @argshook - What's your opinion ?

src/NumberInput/NumberInput.js
NitayRabi: I Don't think so - it also creates the API table in the docs.

stories/NumberInput.ux/index.story.js
NitayRabi: Overall keeping root folder names different (NumberInput/NumberInput.ux) make it easier (for me at least) to keep track on where i'm at.

stories/NumberInput.ux/storySettings.js
NitayRabi: Discussed with Domas - will be 3.4. we do need to create a task for rearranging the indices once they're done refactoring.

src/NumberInput/NumberInput.js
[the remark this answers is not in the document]
argshook: i agree with @hepiyellow all the object descriptor trickery may bring more headache than solving the legacy mistake

stories/NumberInput.ux/index.story.js
argshook: i'd say let's not create new convention and let's remain consistent with other folders.

src/NumberInput/NumberInput.js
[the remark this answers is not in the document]
NitayRabi: NP then I'll refactor - At this stage I'll only output value, and when I'll implement formatting I'll add more params to the CB.

stories/NumberInput.ux/index.story.js
NitayRabi: Well its debatable IMO but for the time being I agree we should keep convention

hepiyellow: I too agree that consistency should be kept.
Right now the convention is to put component stories under `stories/components/CompName`.

Whenever we have a new / better convention, let's add it by refactoring ALL components to the new convention. If you don't have time to do the Refactor, please open an Issue.

stories/components/NumberInput/storySettings.js
```diff
@@ -0,0 +1,13 @@
+import { Category } from '../../storiesHierarchy';
+
+export const storySettings = {
+  kind: Category.COMPONENTS,
```
hepiyellow: Please remove `kind` and use only `category`

stories/NumberInput.ux/index.story.js
NitayRabi: Moved it into components.

stories/components/NumberInput/storySettings.js
NitayRabi: This breaks storybook.

src/NumberInput/NumberInput.uni.driver.js
```diff
@@ -0,0 +1,21 @@
+import { baseUniDriverFactory } from 'wix-ui-test-utils/base-driver';
+import { testkit } from '../Input/Input.uni.driver';
```
hepiyellow:
```suggestion
import { testkit : inputTestkit } from '../Input/Input.uni.driver';
```

src/NumberInput/NumberInput.js
```diff
@@ -0,0 +1,111 @@
...
+
+  state = {
+    value: '',
+  };
+
+  componentWillReceiveProps({ value }) {
+    this.setState(currentState => {
+      return {
+        ...currentState,
+        value: value || '',
```
hepiyellow: ### Bug
`props.value` could be a number.
It could be the number `0`.
Which means that the state would be set to empty string (since 0 is falsy)

I guess your missing a test here.

src/NumberInput/NumberInput.js
```diff
@@ -0,0 +1,111 @@
...
+  static displayName = 'NumberInput';
+
+  state = {
+    value: '',
+  };
+
+  componentWillReceiveProps({ value }) {
+    this.setState(currentState => {
+      return {
+        ...currentState,
```
hepiyellow: The state has only a value, so why do you need to spread it, and then overwrite it?

src/NumberInput/NumberInput.js
```diff
@@ -0,0 +1,111 @@
...
+
+class NumberInput extends React.PureComponent {
+  static displayName = 'NumberInput';
+
+  state = {
+    value: '',
+  };
+
+  componentWillReceiveProps({ value }) {
+    this.setState(currentState => {
```
hepiyellow: You are not using anything from the `currentState` so you don't need this function, you can simply do:
```
this.setState({value: value === undefined? '': value});

File: src/NumberInput/NumberInput.js [hunk 16]
```diff
@@ -0,0 +1,111 @@
...
+      updatedValue = operator(numberValue, numberStep);
+    if (this._isInRange(updatedValue)) {
+      this._triggerOnChange(updatedValue);
+    }
+  }
+
+  _triggerOnChange(value) {
+    const { onChange } = this.props;
+    this.setState(
+      currentState => {
```
hepiyellow: Same as before, I think this is enough:
```
this.setState({value: value === undefined? '': value});
```

stories/components/NumberInput/storySettings.js
```diff
@@ -0,0 +1,12 @@
+import { Category } from '../../storiesHierarchy';
+
+export const storySettings = {
+  kind: Category.INPUTS,
```
hepiyellow: I think `kind` is redundant

stories/NumberInput/storySettings.js
```diff
@@ -0,0 +1,7 @@
+import { Category } from '../storiesHierarchy';
+
+export const storySettings = {
+  kind: Category.INPUTS,
```
hepiyellow: I think `kind` is redundant

src/NumberInput/NumberInput.js
```diff
@@ -0,0 +1,111 @@
...
+          }
+        />
+      </div>
+    );
+  }
+}
+
+NumberInput.propTypes = {
+  ...Input.propTypes,
+};
```
hepiyellow: Aren't you missing these props?
```
  /** Minimum value for input */
  min: PropTypes.number,
  /** Maximum value for input */
  max: PropTypes.number,
  /** Step for each ticker click  */
  step: PropTypes.number,
```

src/NumberInput/NumberInput.spec.js
```diff
@@ -0,0 +1,107 @@
...
+import { numberInputPrivateDriverFactory } from './NumberInput.private.uni.driver';
+
+describe('NumberInput', () => {
+  const createDriver = createUniDriverFactory(numberInputPrivateDriverFactory);
+
+  it('should render', async () => {
+    const driver = createDriver(<NumberInput />);
+    expect(await driver.exists()).toBeTruthy();
+  });
+
```
hepiyellow: Missing tests:
- [ ] should render given value is an empty string
- [ ] should increment to 1 when the initial value is an empty string
- [ ] should increment to 1 when initial value is `undefined`

src/NumberInput/NumberInput.js
```diff
@@ -0,0 +1,111 @@
...
+    const { onChange } = this.props;
+    this.setState(
+      currentState => {
+        return {
+          ...currentState,
+          value,
+        };
+      },
+      () => {
+        onChange && onChange(this.state.value);
```
hepiyellow: ### Type Inconsistency
There is an inconsistency in the `onChange()` argument type.
`this.state.value` could be a `number` if it was increased/decreased.
and it could be a `string` if it was changed by manually typing in the input

I think we should stick to one type.

Which one do you prefer?

Looking at the popular `react-numeric-input` as reference, they use `number`.
https://github.com/vlad-ignatov/react-numeric-input/blob/79874ccbe2293873300953f32d91f2d70d32a871/src/NumericInput.jsx#L1076

src/NumberInput/NumberInput.uni.driver.js
NitayRabi: Named imports do not destructure.

src/NumberInput/NumberInput.js
NitayRabi: Well they're in Input's prop types because generally you could pass these to Input component with type=number and they'll still work like HTML5 props.

src/NumberInput/NumberInput.spec.js
NitayRabi: `should render given value is an empty string` - ? if your'e not given a value I don't to render a value (allow placeholder)

src/NumberInput/NumberInput.js
NitayRabi: Correct - Will be fixed.

stories/components/NumberInput/storySettings.js
NitayRabi: image
This is what i get when I remove it - ideas?

stories/NumberInput/storySettings.js
NitayRabi: image
This is what i get when I remove it - ideas?

argshook: check the where `kind` is used before just removing it: https://github.com/wix/wix-style-react/pull/3033/files#diff-e9d757c7f1e747b1343c1dd9ea16efbeR27

src/NumberInput/NumberInput.uni.driver.js
argshook: Erez suggests an alias, these do work :)

NitayRabi: image

argshook: sorry, i thought it was `as`, not `:`

`import { testkit as inputTestkit } from '../Input/Input.uni.driver';`

src/NumberInput/NumberInput.js
hepiyellow: >they're in Input's prop types

They are not in Input's propTypes:
https://github.com/wix/wix-style-react/blob/51e8a6a8043320f8ae4baaeb7a46248d21c30187/src/Input/Input.js#L391

src/NumberInput/NumberInput.spec.js
hepiyellow: You have code that initializes `this.state.value` from props.
Then you use `this.state.value` to set `<Input value={this.state.value/>`.
So even without the increment/decremenet feature, this code needs a test.
And empty string is a case, since you have logic for that in the constructor:
```
constructor(props) {
    super(props);
    const { value } = props;
    this.state = {
      value: isNaN(value) ? '' : value,
    };
  }
```

src/NumberInput/NumberInput.js
NitayRabi: screen shot 2019-03-07 at 14 58 12
Oh but they are.

src/NumberInput/NumberInput.spec.js
```diff
@@ -0,0 +1,143 @@
...
+    expect(await driver.getValue()).toEqual(-1);
+  });
+
+  it('changing input value should call onChange with number value', async () => {
+    const onChange = jest.fn();
+    const driver = createDriver(<NumberInput onChange={onChange} />);
+    await driver.enterValue('1234');
+    expect(onChange).toHaveBeenCalledWith(1234);
+  });
+});
```
hepiyellow: Missing test:
- [ ] should update value when rerendered with different value

src/NumberInput/NumberInput.js
```diff
@@ -8,12 +8,17 @@ class NumberInput extends React.PureComponent {
...
+  }
+
   componentWillReceiveProps({ value }) {
-    this.setState(currentState => {
-      return {
-        ...currentState,
-        value: value || '',
-      };
+    this.setState({
+      value: isNaN(value) ? '' : value,
```
hepiyellow: ### Bug
This could be problematic in Controlled mode.
If the user is typing a negative number, and currently the input has a value `-`,
Then `-` is a NaN, so you will delete the `-`, and display empty string.
Meaning, that the user can not enter negative numbers manually.

Need a test here !
- [ ] `should render '-' (minus) when value is '-' minus` 

It might be good to see what [react-numeric-input](https://github.com/vlad-ignatov/react-numeric-input/blob/master/src/NumericInput.jsx) is doing

src/NumberInput/NumberInput.js
```diff
@@ -0,0 +1,115 @@
...
+
+  state = {
+    value: '',
+  };
+
+  constructor(props) {
+    super(props);
+    const { value } = props;
+    this.state = {
+      value: isNaN(value) ? '' : value,
```
hepiyellow: I would extract a `parse` function here which is used in:
- constructor
- compWillReceiveProps
- _applyChange (onChange from the native input)

src/NumberInput/NumberInput.js
```diff
@@ -62,7 +64,9 @@ class NumberInput extends React.PureComponent {
   }
 
   _inputValueChanged = e => {
-    return this._triggerOnChange(e.target.value);
+    return isNaN(e.target.value)
```
hepiyellow: I would rather have 1 `parse` function,
instead of checking for `NaN` in many places.

src/NumberInput/NumberInput.uni.driver.js
```diff
@@ -15,7 +15,7 @@ export const numberInputDriverFactory = base => {
     /** Click on ticker down */
     clickOnDecrement: () => getTickerDriver().clickDown(),
     /** Input value to component */
-    enterValue: value => getInputDriver().enterValue(value),
-    getValue: () => getInputDriver().getValue(),
+    enterValue: value => getInputDriver().enterText(value),
+    getValue: async () => +(await getInputDriver().getValue()),
```
hepiyellow: - What is this `+`?
- Redundant `async/await`

package.json
```diff
@@ -25,7 +25,7 @@
   },
   "scripts": {
     "start": "./scripts/npm-start.sh",
-    "precommit": "lint-staged",
+    "_precommit": "lint-staged",
```
hepiyellow: This is probably a temp thing that someone pushed by mistake.
Make sure it is reverted before you push.

package.json
```diff
@@ -25,7 +25,7 @@
   },
   "scripts": {
     "start": "./scripts/npm-start.sh",
-    "_precommit": "lint-staged",
+    "precommit": "lint-staged",
```
hepiyellow: 👍

src/NumberInput/NumberInput.uni.driver.js
NitayRabi: Extract number

src/NumberInput/NumberInput.js
```diff
@@ -12,16 +12,20 @@ class NumberInput extends React.PureComponent {
...
   }
 
   componentWillReceiveProps({ value }) {
     this.setState({
-      value: isNaN(value) ? '' : value,
+      value: this._extractNumberValue(value),
     });
   }
 
+  _extractNumberValue(value) {
```
hepiyellow: Well this hardly is an appropriate name, since it is no extracting a number value, it simply does `defaultToStirngIfUndefined`

src/NumberInput/NumberInput.e2e.js
```diff
@@ -0,0 +1,14 @@
...
+import { eyesItInstance } from '../../test/utils/eyes-it';
+import { storySettings, testStories } from './docs/storySettings';
+
+describe('NumberInput', () => {
+  const eyes = eyesItInstance();
+  const testStoryUrl = testName =>
+    createTestStoryUrl({ ...storySettings, testName });
+  eyes.it('should render NumberInput with variations', async () => {
+    await browser.get(testStoryUrl(testStories.numberInputVariations));
+    await eyes.checkWindow(testStories.numberInputVariations);
```
hepiyellow: You should NOT need to call `eyes.checkWindow()`, since `eyes.it` already takes a screenshot at the end of each test.
If you do have a good reason to take it manually, then you should configure:
```js
const eyes = eyesItInstance({enableSnapshotAtBrowserGet: false, enableSnapshotAtEnd: false});
```

src/NumberInput/NumberInput.spec.js
```diff
@@ -0,0 +1,162 @@
...
+    rerender(<NumberInput value={5} />);
+    expect(await driver.getInputValue()).toEqual('5');
+  });
+
+  it('should render `-` (minus) when value is `-`', async () => {
+    const driver = createDriver(<NumberInput />);
+    await driver.enterValue('-15');
+    expect(await driver.getInputValue()).toEqual('-15');
+  });
+});
```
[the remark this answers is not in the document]
NitayRabi: I spoke to you about this in Slack - '-' is rendered both in value and both in onChange as an empty string - how would you test it?

src/NumberInput/NumberInput.js
```diff
@@ -0,0 +1,115 @@
...
+    }
+  }
+
+  _triggerOnChange(value) {
+    const { onChange } = this.props;
+    this.setState(
+      {
+        value,
+      },
+      () => onChange && onChange(this._extractNumberValue(value)),
```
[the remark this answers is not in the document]
NitayRabi: Yeah I saw their implementation - the `String | Number` duality is used for value of input and the current number value (they can be totally different based on format) - I don't think that at this time we should output both of these values - In your opinion what should be the equivalent number value for `''` ? - not `0`, how about `null`?

src/NumberInput/NumberInput.js
```diff
@@ -0,0 +1,115 @@
...
+    if (inputRef) {
+      inputRef(ref);
+    }
+  };
+
+  render() {
+    const { dataHook, suffix, ...props } = this.props;
+
+    return (
+      <div data-hook={dataHook}>
```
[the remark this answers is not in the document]
NitayRabi: Cool nice Idea.

src/NumberInput/NumberInput.e2e.js
```diff
@@ -0,0 +1,14 @@
+import { createTestStoryUrl } from '../../test/utils/storybook-helpers';
```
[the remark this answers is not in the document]
NitayRabi: This was not part of the refactor (from what i saw) - Can we send some clarification on the new expected structure? instead of guessing it during the PR?

src/NumberInput/NumberInput.spec.js
```diff
@@ -0,0 +1,162 @@
+import React from 'react';
```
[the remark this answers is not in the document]
NitayRabi: This was not part of the refactor (from what i saw) - Can we send some clarification on the new expected structure? instead of guessing it during the PR?

src/NumberInput/docs/storySettings.js
```diff
@@ -0,0 +1,10 @@
+import {Category} from '../../../stories/storiesHierarchy';
+
+export const storySettings = {
+  category: Category.COMPONENTS,
+  storyName: 'NumberInput',
+};
+
+export const testStories = {
```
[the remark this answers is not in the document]
NitayRabi: This was not part of the refactor (from what i saw) - Can we send some clarification on the new expected structure? instead of guessing it during the PR?

package.json
```diff
@@ -320,4 +320,4 @@
       }
     ]
   }
-}
+}
```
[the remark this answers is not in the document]
NitayRabi: ?

src/NumberInput/NumberInput.js
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  componentWillReceiveProps({ value }) {
+    this.setState({
+      value: this._defaultValueToStringIfEmpty(value),
+    });
+  }
+
+  _defaultValueToStringIfEmpty(value) {
+    return value === undefined || value === '' ? '' : +value;
```
hepiyellow: ### Type Consistency of `this.state.value`
This function still returns `string | number`. So `this.state.value` is `stinrg | number`.
I think its better to have it as a single type. You can even rename `this.state.value` to show it's type. (e.g. `numberValue` or `stringValue`

src/NumberInput/NumberInput.js
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  _defaultValueToNullIfEmpty(value) {
+    return value === undefined || value === '' ? null : +value;
```
hepiyellow: ### Missing test
There is no test for an `onChange` called with  `null`

src/NumberInput/NumberInput.js
```diff
@@ -8,99 +8,97 @@ class NumberInput extends React.PureComponent {
...
-                onUp={this._increment}
-                onDown={this._decrement}
-                dataHook="number-input-ticker"
-              />
-            </Input.Group>
-          }
-        />
-      </div>
+      <Input
+        dataHook={dataHook}
```
hepiyellow: You don't need to deconstruct `dataHook` from `this.props`. the `props` are spreaded one line after this one.

src/NumberInput/NumberInput.js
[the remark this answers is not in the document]
hepiyellow: So currently, "Controlled" mode is NOT supported.

```js
class {
  state = {value}
  render() {
     return (<NumberInput 
         value={this.state.value}
         onChange(value=>this.setState({value})
      />)
  }
}
```
With the above code, it isn't possible to enter negative numbers.
If the user types '-', then `onChange` is called with `NaN`

src/NumberInput/NumberInput.js
```diff
@@ -0,0 +1,117 @@
...
+    }
+  }
+
+  _triggerOnChange(value) {
+    const { onChange } = this.props;
+    this.setState(
+      {
+        value,
+      },
+      () => onChange && onChange(this._defaultValueToNullIfEmpty(value)),
```
hepiyellow: Problematic with "Controlled" mode.
if `value = '-'` (The user is starting to enter a negative value, in Contorled mode),
then `onChange` is called with `NaN`

src/NumberInput/NumberInput.spec.js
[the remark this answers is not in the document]
hepiyellow: Lets meet and discuss this. There seems to be a misunderstanding.

src/NumberInput/NumberInput.uni.driver.js
```diff
@@ -0,0 +1,22 @@
...
+
+    /** Click on ticker up */
+    clickOnIncrement: () => getTickerDriver().clickUp(),
+    /** Click on ticker down */
+    clickOnDecrement: () => getTickerDriver().clickDown(),
+    /** Input value to component */
+    enterValue: value => getInputDriver().enterText(value),
+    getNumberValue: async () => +(await getInputDriver().getValue()),
+    getInputValue: () => getInputDriver().getValue(),
+  };
```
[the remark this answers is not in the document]
hepiyellow: ping?

src/NumberInput/docs/testStory.js
```diff
@@ -0,0 +1,65 @@
...
+
+storiesOf(kind, module).add(testStories.numberInputVariations, () => {
+  return (
+    <div>
+      <br />
+      <NumberInput dataHook="storybook-numberinput" {...defaultProps} />
+      <br />
+      <NumberInput
+        size="small"
+        dataHook="storybook-numberinput"
```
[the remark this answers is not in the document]
hepiyellow: ping?

src/NumberInput/docs/testStory.js
```diff
@@ -0,0 +1,65 @@
...
+import NumberInput from 'wix-style-react/NumberInput';
+import { testStories, storySettings } from './storySettings';
+
+const defaultProps = {
+  value: '12345',
+};
+
+const kind = getTestStoryKind(storySettings);
+
+storiesOf(kind, module).add(testStories.numberInputVariations, () => {
```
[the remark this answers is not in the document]
hepiyellow: ping?

src/NumberInput/NumberInput.e2e.js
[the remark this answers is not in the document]
hepiyellow: I wrote "You can..." , meaning that it is optional.
I will send on slack.

src/NumberInput/NumberInput.spec.js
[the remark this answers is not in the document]
NitayRabi: Added tests to prevent '-'/null values to turn into '0' - covers this and controlled case

src/NumberInput/NumberInput.js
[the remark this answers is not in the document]
NitayRabi: Added a test and example for controlled mode.

src/NumberInput/NumberInput.js
NitayRabi: Called with NaN in JSDom - covered by latest changes.

src/NumberInput/NumberInput.js
NitayRabi: Added tests to cover these cases

src/NumberInput/NumberInput.js
NitayRabi: At this time and in this PR's scope I don't want to manage 2 state values (Will be implemented once we implement formatting/masking). 

I fixed it to use default `null` value - then the type is `number | null`;

PR #3045, opened 2019-02-27
Constributing (Docs): Add `TEST_DRIVERS-GUIDELINES.md`

hepiyellow:


docs/contribution/TEST_DRIVERS_GUIDELINES.md
```diff
@@ -0,0 +1,102 @@
...
+<div>
+  <button data-hook="cancel">{props.cancelText}</button>
+  <button data-hook="confirm">{props.confirmText}</button>
+</div>
+```
+
+### Try To Keep `data-hook`'s Values Short
+
+`data-hook`'s inflate the DOM, and are NOT removed in production.
+This is because `Automation` team, is conducting e2e tests on production sites, and are using any `data-hook` attribute they can find.
```
argshook:
```suggestion
This is to allow testing applications in production.
```

we shouldn't mention any teams imo

docs/contribution/TEST_DRIVERS_GUIDELINES.md
```diff
@@ -0,0 +1,102 @@
...
+But... avoid using `data-hook`'s if there is another "Natural" way to locate an element.
+For example, if a component renders:
+
+```js
+<div>
+  <button>Confirm</button>
+</div>
+```
+
+In this simple case you can locate the button with a tagName selector ( e.g `querySelector('button')`).
```
argshook: are you sure this is the better way?

in one place you state:
>  Test drivers have two purposes:
> - Provide an abstraction of component implementation. Making Tests independent of implementation changes.

but here you seem to state the opposite. I understand `querySelector` is easy enough, but data-hook is exactly to avoid entanglement between test and implementation code

docs/contribution/TEST_DRIVERS_GUIDELINES.md
```diff
@@ -0,0 +1,102 @@
...
+
+### Locating By Class Name
+
+TBD: Do we encourage using class name or data-hook?
```
argshook: `data-hook`

docs/contribution/TEST_DRIVERS_GUIDELINES.md
```diff
@@ -0,0 +1,102 @@
...
+|-----|------|
+| clickClearButton | clear |
+
+### Example 2
+
+A `<Dropdown/>` allows selection from options.
+
+| Bad | Good |
+|-----|------|
+| clickOption(index) | selectOption(index) |
```
shlomitc: actually the bad pattern here is to expose implementation instead of abstraction and not just the naming.
what we have today which is bad:
```
focus()
keyDown()
clickOption(index)
```

instead of
```
selectOption(index)
```

docs/contribution/TEST_DRIVERS.md
```diff
@@ -1,45 +1,65 @@
...
+## Writing Drivers
+
+See [Test Drivers Guidelines](./TEST_DRIVERS_GUIDELINES.md) for best-practices when writing test drivers.
+
+## Testing Drivers
+
+Public drivers are also published code that needs to be tested.
+We test them simply by using them in our component tests.
+
+### Just a small but very common Gatch
```
shlomitc: Gatchas

docs/contribution/TEST_DRIVERS_GUIDELINES.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Test the DOM
+
+Every React prop set by the consumer has an affect on the DOM, or the behavior of the component (which affects the DOM).
+
+Don't assert on React props or inner state.
+
+## `data-hook`'s
```
shlomitc:
```suggestion
## `data-hook`s
```

docs/contribution/TEST_DRIVERS_GUIDELINES.md
shlomitc: I wouldn't suggest implementing with `querySelector` because of `unidriver`. just suggestion to take by tag name is good enough

docs/contribution/TEST_DRIVERS_GUIDELINES.md
shlomitc: data-hook

docs/contribution/TEST_DRIVERS.md
argshook: maybe it's intentional? :) https://www.urbandictionary.com/define.php?term=gatch

docs/contribution/TEST_DRIVERS_GUIDELINES.md
hepiyellow: why?

shlomitc: More predictable, no hashing issues, doesn’t intervene with styling or technology

docs/contribution/TEST_DRIVERS_GUIDELINES.md
hepiyellow: The driver IS coupled with implementation. 
I don't care if driver does:

`element.children[3]` to locate some inner element.
There is almost no difference between that and using a data-hook.

The idea to avoid data-hooks, is actually more for Application testing.
I just want to avoid having redundant data attributes in production.
But if you think this claim is not strong enough, then we can remove this section.

docs/contribution/TEST_DRIVERS_GUIDELINES.md
hepiyellow: We still have a lot of places where we identify styling state (like theme, size...) by classes, not data-hooks. So should we refactor?
Also stylable is now switching to work with classes and not data attributes (I don't exactly know why).
It seems to me that classes work.

docs/contribution/TEST_DRIVERS_GUIDELINES.md
argshook: although i'd like having strong selectors (like data hooks) as opposed to vague ones (like `button` or `span > div` etc), i don't have strong opinion, either is fine

docs/contribution/TEST_DRIVERS_GUIDELINES.md
argshook: i guess the answer as usual is "it depends". both approaches have ups and downs. just from memory i know that working with class names as identifiers is prone for breaking (someone updates  styling, but forgets driver and similar issues).

if we must choose what we encourage, i'd say data hook

docs/contribution/TEST_DRIVERS.md
hepiyellow: So single form is "Gatcha"?

docs/contribution/TEST_DRIVERS_GUIDELINES.md
shlomitc: for selecting strategies, let's stick with data-hook.
for verifying the existence of classes for styling - we usually do visual testing. for those who already have it, I'd still query with data-hook and verify existence of class name on the element

PR #3046, opened 2019-02-27
Try re-enabling e2e tests - see . if we get all timeouts still.

hepiyellow:


mykas:
So maybe we should enable this already? I'd like to release `<ColorInput/>` and need to approve applitools changes

PR #3048, opened 2019-02-27
<RichTextInputArea/> - add missing component tests

nitayneeman:
This PR adds missing component tests for the toolbar and implements the relevant driver methods.

Related to issue:
[wix-private/wsr-issues#35](https://github.com/wix-private/wsr-issues/issues/35)


src/RichTextInputArea/RichTextInputArea.spec.js
```diff
@@ -9,35 +9,135 @@ describe('RichTextInputArea', () => {
...
+
+      expect(currentValue).toBe(expectedText);
+    });
+
+    it('should render text as bulleted list after clicking the bulleted list button', async () => {
+      const driver = createDriver(
+        <RichTextInputArea onChange={value => (currentValue = value)} />,
+      );
+      const typedText = 'Text';
+      const expectedText = `<ul>\n<li>${typedText}</li>\n</ul>\n`;
```
nitayneeman: I don't really like the `\n`s, but these are added by `draftjs-to-html`. I'm starting to consider implementing the content converter for plain HTML on our own.

src/RichTextInputArea/RichTextInputArea.private.driver.js
```diff
@@ -1,8 +1,13 @@
 import publicDriverFactory, { getTextArea } from './RichTextInputArea.driver';
 
+import richTextToolbarPrivateDriverFactory from './RichTextToolbar.private.driver';
+
 export default base => {
   return {
     ...publicDriverFactory(base),
+    ...richTextToolbarPrivateDriverFactory(
```
nitayneeman: Is it the correct way to import the toolbar driver? Maybe would be better to do it in the spec file and spreading the methods of both drivers to a "merged" object and use it as the driver?

Issue #2657, opened 2018-12-24
Multiselect: click enter and submit form


davydof:
# Bug Report

### Relevant Components
Multiselect

### Current Behavior
After enter value to multiselect and push `Enter` button the form which wrap the multiselect component start to submit.

### Expected Behavior
At the first `Enter` button push add the value to tag.

### Severity

Major

adavidof:
@shlomitc could you please prioritize this task?
I really need it

shlomitc:
@adavidof - I'm sorry but I don't think it could be pushed in the near future. PRs are also welcome

hepiyellow:
Isn't simply calling having `e.preventDefault()` when Enter is pressed on an option?

shlomitc:
@adavidof - does other component behave correctly? For example when using the `Autocomplete` component and pressing Enter for selection from the list?

adavidof:
@shlomitc @hepiyellow 
the problem is:
1. I fill the input and click `Enter` button
2. value is added
3. the remove tag button clicked and remove added tag

like after add the tag I have the focus on the remove tag button.

![image](https://user-images.githubusercontent.com/10529679/53256008-935d3480-36cf-11e9-847b-3d1cce53c56a.png)
![image](https://user-images.githubusercontent.com/10529679/53256134-d5867600-36cf-11e9-898c-1c9764b15202.png)

So we need to fix this part

shlomitc:
I couldn't get the focus on the remove button in any of the following examples in the storybook:
https://wix-wix-style-react.surge.sh/?selectedKind=3.%20Inputs&selectedStory=3.8%20Tags&full=0&addons=0&stories=1&panelRight=0

maybe you have some custom behavior? or reproducible example?

adavidof:
@shlomitc https://github.com/wix-private/media-platform-client/tree/fix/multiselect branch
![image](https://user-images.githubusercontent.com/10529679/53338086-73b25000-390b-11e9-9b45-7d793c2f99e7.png)

gilmeir:
**Reproducing**
I've created a [codesandbox](https://codesandbox.io/s/5zj7wj50kn) with MultiSelect and Input components inside a form.
See the behaviour when clicking Enter inside the component after selection.

**Prevent default**
If "Prevent Default" is unchecked an Enter key press will refresh the page after submitting the form - this is default browser behaviour for forms. This is the behaviour @davydof ran into.

[...]

PR #3052, opened 2019-02-27
fix: add button type to tag close button

adavidof:
https://github.com/wix/wix-style-react/issues/2657

### 🔦 Summary
Add `button` type to close button in Tag component

The reason is: when Multiselect was wrapped with html element `form` after push Enter the button is submitted.
So I should change the button type to `button`

probably the better solution is to change the default button type to `button` in core ui lib

### DIFF
--- a/src/CloseButton/CloseButton.js
+++ b/src/CloseButton/CloseButton.js
@@ -26,6 +26,8 @@ class CloseButton extends Component {
     disabled: bool,
     /** string based data hook for testing */
     dataHook: string,
+    /** Button type */
+    type: oneOf(['button', 'submit', 'reset']),
   };
 
   static defaultProps = {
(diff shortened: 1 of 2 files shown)


shlomitc:
@mykas that's quite interesting actually - what do you think about changing the core button default type?

mykas:
@shlomitc I thought I have added it already! It was removed in wix-ui-core because of proptypes: https://github.com/wix/wix-ui/commit/ad9293c25569750c135d042589f55af059410ebf#diff-4935ea9c14524abde8e424905ee9254c

mykas:
@adavidof I will add it to the core button. We can close this PR :) thanks for reporting the issue!

mykas:
https://github.com/wix/wix-ui/pull/1176

PR #3053, opened 2019-02-28
<TextButton/> - migrate to sections documentation

mykas:
This PR is to tranform TextButton docs to the new design.


mykas:
@milkyfruit so as we talked please review the examples of TextButton. What I missing now is:
- background color for dark skin
- background color for light skin
- descriptions for examples
- maybe different description for TextButton

milkyfruit:
@mykas 

### Component Descriptions
Text Button works as a text link in a paragraph or a neutral button for some command, like "add item", "filter", etc. It shouldn't be used as a cancel action.

### Example Descriptions

- **Skins**
TextButton supports 4 styles of different skins.
Light skin is designed to be used on dark backgrounds while dark skin is for coloured backgrounds.
- **Size**
This action supports two sizes – small and medium.
- **Affix**
TextButton allows using any icon before or after the text.
- **Weight**
TextButton supports two text weights – thin and normal.
Thin weight should be used when component acts as a text link in a paragraph that share same text weight.
- **Underline**
To emphasise an action, it is allowed to make underline always visible.
While to make it more neutral, it's allowed to disable underline too.
- **Custom rendering**
This component can appear in different HTML tag names – button, a, span, div.

### Background colors
`skin:dark` – bg color: `Y30`
`skin:light` – bg color: `D10`

shlomitc:
`Text Button works as a text link in a paragraph` - that's incorrect.
For inline links, we would advise using the `Text` component, where every link is styled accordingly

mykas:
@shlomitc I am also wondering if we should add an example of using `<Text/>` component when multiline link is needed. I got several requests from channel for this functionality and everybody thought that TextButton suppose to support this.

mykas:
@milkyfruit please double check everything again

hepiyellow:
@milkyfruit @mykas What's holding this from being merged?

milkyfruit:
@shlomitc @mykas very good point, let's correct the "weight" example then and add "inline text link" example:

**Weight**
Text weight can be adjusted to match adjacent components. TextButton supports two text weights – thin and normal.

**Inline text link**
For text inline text link used in a paragraph with single or multiple lines, please use <Text/> component.

I guess, example would be something like this:
`<Text>
  In literary theory, a text is any object that can be "read", <a href="#">whether this object is a work of literature,</a> a street sign, an arrangement of buildings on a city block, or styles of clothing. It is a coherent set of signs that transmits some kind of informative message.
</Text>`

PR #3061, opened 2019-03-04
ComponentGenerator: Move story under Component's folder. Add `ComponentTestStories`

hepiyellow:
- Moved component story to be under `src/CompName/docs`
- Move tests and test stories to be under `src/CompName/test`
- Added `ComponentTestStories`
- Update e2e to run on the test stories and NOT on the Documentation story.
- Remove README, add `description` to `Component.js`'s JSDoc
- Add GeneratedTestComponent to LiveCode scope - So that we can have live example in test-component's story
- transpile: ignore files that don't need to be published (Reducing `dist` side from 28MB to 24MB)
  ```
   './src/**/*.story.js',
   './src/**/test/**/*',
   './src/**/docs/**/*',
   './src/**/*.meta.js',
   './src/**/*.spec.js',
   './src/**/*.e2e.js',
   ```

### Weird Issue

Stories that I put under src,
And imported from stories/index.js

Import ‘../src/…'

These stories did not get loaded when building the storybook statics (npm run storybook:build)

(I run then npm run storybook:static, and the stories were missing)
They do appear if I run 
npm run storybook (dev mode)

#### Workaround
I fixed it by using require

```
// This import was added by the component generator
require('../src/MyNewComponent/docs/index.story');

// This import was added by the component generator
require('../src/MyNewComponent/test/MyNewComponentStories');
```

I don’t understand why it works.


argshook:
> Moved component story to be under `stories/components`

why? i thought we agreed to eventually move `.story.js` and friends to be together with component itself

hepiyellow:
@argshook But I don't know how exactly to exclude a file from the build(so it doesn't end up in `dist`).

And as I mentioned about new conventions...
We'll do the Big refactor later.
This PR is more about adding a proper example for how to do an e2e test on a TestStory (and not on the Documentation Story).

argshook:
> @argshook But I don't know how exactly to exclude a file from the build(so it doesn't end up in dist).

is this a problem though? so what if it ends up in the dist?

hepiyellow:
Waiting for build

scripts/component-generator/src/transforms/stories-file.js
```diff
@@ -7,7 +7,10 @@ module.exports = (file, api, options) => {
 
   j(imports[imports.length - 1]).insertAfter(
     `// This import was added by the component generator
-import './${ComponentName}/index.story';`,
+import './components/${ComponentName}/index.story';
```
argshook: shouldn't these point to `src`?
```suggestion
import '../src/${ComponentName}/docs/index.story';
```

src/GeneratedTestComponent/docs/index.story.js
```diff
@@ -1,14 +1,15 @@
...
-import GeneratedTestComponent from '../../src/GeneratedTestComponent';
+import GeneratedTestComponent from '..';
 
 export default {
   category: storySettings.category,
   storyName: storySettings.storyName,
 
   component: GeneratedTestComponent,
-  componentPath: '../../src/GeneratedTestComponent/GeneratedTestComponent.js',
+  componentPath: '../GeneratedTestComponent.js',
```
argshook:
```suggestion
  componentPath: '..',
```
this should be enough

hepiyellow: fixed

scripts/component-generator/src/transforms/stories-file.js
hepiyellow: fixed

PR #3064, opened 2019-03-04
`<PopoverMenu/>` - add `moveBy` prop

Jefoss:
### 🔦 Summary
We have an edge case that requires moving the popover tooltip a bit, so it would be nice if it would pass the `moveBy` prop to the tooltip component


hepiyellow:
@Jefoss 
`<PopoverMenu/>` is about to be deprecated.
I wouldn't want to expand the API now so openly at this point, it will make deprecating it more difficult.
Please open explicitly any props you need for the tooltip. (There are not many left 😄  )
Sorry for the inconvenience.

- Also: We do not encourage re-using Other component propTypes.
- Also : using `omit` messes-up our AutoDocs

src/PopoverMenu/PopoverMenu.js
```diff
@@ -6,15 +6,8 @@ import Button from '../Deprecated/Button';
...
-  oneOf,
-  oneOfType,
-  bool,
-  element,
-  number,
-  string,
-  func,
-} from 'prop-types';
+import { oneOf } from 'prop-types';
+import omit from 'omit';
```
hepiyellow: Please dont use omit, AutoDocs does know how to handle it, and the Tooltip's props wont show in the Docs.
You can Copy&Paster the prop Types from tooltip

PR #3066, opened 2019-03-04
<Button/>, <TextButton/>, <IconButton/> and <CloseButton/> - proper as proptype

mykas:
This PR fixes thrown error by proptypes when react component is passed instead of reference.


src/Button/Button.js
```diff
@@ -2,14 +2,27 @@ import React, { Component } from 'react';
...
+  oneOf,
+  element,
+  object,
+  bool,
+  func,
+  symbol,
+} from 'prop-types';
+
+export const as = (...args) =>
+  oneOfType([func, object, string, symbol])(...args);
```
argshook: @mykas can this be set directly within `propTypes`?

argshook: also, isn't it enough to have just `node` and `string`?

mykas: It is actually not enough for some reason. I think that when we pass `class` then it treats it as an object.

argshook: i suppose you should update `wix-ui-core/button-next` if we want to support a function?

mykas: Yes

mykas: Writing unit tests for this one

src/Button/Button.spec.js
```diff
@@ -1,10 +1,39 @@
...
   it('should have correct displayName', async () => {
     expect(Button.displayName).toEqual('Button');
   });
+
+  describe(`'as' prop`, () => {
+    const Link = ({ children }) => <a>{children}</a>;
+
+    class LinkClass extends React.Component {
+      render() {
+        return <a>{this.propschildren}</a>;
```
mykas: Syntax. Fixing

src/Button/Button.spec.js
```diff
@@ -1,10 +1,39 @@
...
 describe('Button', () => {
   afterEach(() => cleanup());
 
+  const render = createRendererWithUniDriver(buttonDriverFactory);
+
   it('should have correct displayName', async () => {
     expect(Button.displayName).toEqual('Button');
   });
+
+  describe(`'as' prop`, () => {
```
argshook: maybe these tests should be in wix-ui-core?
then you wouldn't have to repeat them multiple times in wsr

src/Button/Button.spec.js
```diff
@@ -1,10 +1,39 @@
...
+    }
+
+    it('should render without errors when html element is passed', async () => {
+      const { driver } = render(<Button as="a" />);
+      expect(await driver.exists()).toBe(true);
+    });
+
+    it('should render without errors when function reference is passed', async () => {
+      const { driver } = render(<Button as={Link} />);
+      expect(await driver.exists()).toBe(true);
```
argshook: is `.exists` enough to really check that `Link` has been rendered?

src/TextButton/TextButton.spec.js
```diff
@@ -1,7 +1,39 @@
...
   it('should have correct displayName', async () => {
     expect(TextButton.displayName).toEqual('TextButton');
   });
+
+  describe(`'as' prop`, () => {
+    const Link = ({ children }) => <a>{children}</a>;
+
+    class LinkClass extends React.Component {
+      render() {
+        return <a>{this.propschildren}</a>;
```
mykas: Syntax.

src/CloseButton/CloseButton.driver.private.js
```diff
@@ -0,0 +1,12 @@
+import { closeButtonDriverFactory as publicDriverFactory } from './CloseButton.driver';
+
+export const closeButtonPrivateDriverFactory = base => {
+  return {
+    ...publicDriverFactory(base),
+    closeIconExist: async () => await base.$('[data-hook="close"]').exists(),
+    largeCloseIconExist: async () =>
```
mykas: Syntax. Should be `**Exists()`

src/IconButton/IconButton.js
```diff
@@ -33,6 +44,16 @@ class IconButton extends Component {
     disabled: false,
   };
 
+  renderIcon() {
```
argshook: @mykas is it worthy enough to extract to `renderIcon`? i'd just put this within regular `render`. also, `renderIcon` will be considered public api by autodocs

src/Button/Button.spec.js
mykas: I think that in this case we need them here. This one is not testing cores type interfaces, but rather proptype of this component. I want to make sure that whoever changes proptype - wil get error here.

src/Button/Button.spec.js
mykas: Yes, because proptype will throw error anyway.

src/IconButton/IconButton.js
mykas: maybe you're right. I'll revert it

src/Button/Button.spec.js
mykas: It can be expect(true).toBe(true)

src/Button/Button.spec.js
argshook: for proptypes that's ok, but don't you think we should ensure that given function/class has actually been rendered?

what if i change implementation one day and break it so that type checking is still ok, but there's nothing rendered on screen?

mykas: the question is: if we rely on button core and its implementations - should we really test it to make sure that it really works or should we trust it?

src/Button/Button.spec.js
argshook: if you wish to really test for proptypes, make a test so that it would fail when proptypes change.

in this case, if i remove proptypes, test would still pass. so what are we asserting here?

mykas: I am not sure how do we test it to fail if proptypes will throw and error anyway and test will always fail?

src/Button/Button.spec.js
argshook: good question! i say we should trust wix-ui-core to be well tested and then don't repeat same tests here.

mykas: true, but wix-ui-core does not have proptypes. I think this proptype is kind of special as the  react renderer relys on it. What I would rather do better in near future is to actually have a separate function for this proptype which is tested separately. Or maybe I should even do it now.

PR #3068, opened 2019-03-05
docs(Thumbnail): relocate story files

argshook:
a proposal of relocating `story.js` and friends to live together with component implementation files.

example with Thumbnail component.

before:
  1. `src/Thumbnail/Thumbnail.js`
  2. `stories/Thumbnail/index.story.js`

after:
  1. `src/Thumbnail/Thumbnail.js`
  2. `src/Thumbnail/docs/index.story.js`

stories usually have more than one `.story.js` file, also may have
`testStory.js`. In such cases component folder can have `story` folder
with all relevant story files in there

---

# what's important

considered to be convention:

1. component has `.story.js` file nearby
2. any additional story files go to `story` folder.

that's it, any other specific details are non important at this stage

---

# possible cons

dist will contain story files. this is a non issue imo, cause `npm
install` time currently is not a bottleneck. consistency/maintenance
cost easily outweight few more seconds for `npm install` time.

later on this can be fixed by filtering out files during build.


hepiyellow:
Just remember to add an ignore in `transpileSrc` to stories are not ending up in `dist` folder.

argshook:
we had a chat and agreed to move stories contents to be under `src/Component/docs`

will be done in separate PRs, closing this one

PR #3078, opened 2019-03-07
Add new languages

henrykuzmick:
Add support for Thai, Chinese, and Chech.
All supported in BO.
The only language we still miss is Hindi.


mykas:
Looks great! Thanks a lot for this PR.

mykas:
Lets wait until CI passes and then we can merge it.

PR #3079, opened 2019-03-08
stories cleanup

argshook:
contiuing with documentation standardization

moved from `stories/components` to `src/ComponentName/docs`:

* PageHeader
* Grid
* Page
* NewIcons
* ContactsItemBuilder
* BadgeItemBuilder

removed duplicate stories for

* EmptyState
* CalendarPanel
* Calendar
* Card

these were showed up twice, either identical or even wrong (like Calendar)

removed usage of custom sections components and use only wix-storybook-utils

---

next steps

* rename root `stories` to `docs`
* update component generator to include section scaffold


hepiyellow:
Calendar and friends...
Should have both a ux story and a component story.

argshook:
No, if the story is exact duplicate, then it shouldn't, its nit useful at all. Especially if duplicate is wrong, which already was the case. It just outlines the cost of maintenance

hepiyellow:
1. If the story is wrong, then open an issue. 
2. Components category should include all components
3. UX stories should include whatever UX team tells us to include.

argshook:
We can discuss during weekly, cause i disagree with 1 and 2. Its a waste of
effort and investment to confusion, there's no need at all to duplicate
stuff

hepiyellow:
I only want there to be an index of all component stories. 
E.g. Components category in the storybook should list all components.

PR #3092, opened 2019-03-12
Tweak ImageViewer image scaling styles

### LINKED ISSUE #3091: ImageViewer - picture fit is different Chrome vs Firefox
# 🐛 Bug Report

### 🏗 Relevant Components
`<ImageViewer />`

### 😯 Current Behavior
Consider the following:
```
<MediaPicker
  dataHook="setup-business-logo-picker"
  width={132}
  height={132}
  value="https://static.wixstatic.com/media/7175d1_bc0f6b3c4f43485aa99f0bcceaaea740~mv2.png"
/>
```

#### Chrome:
"Locks" the `` tag inside a 132x132 box, and the `` tag has no `object-fit` definition, so the image gets stretched to fill the box as seen here:
image

#### Firefox:
The `` is contained in a 132x132 box, but is only constrained by height, so width grows to accomodate the image's aspect ratio, so the expected behavior of the image filling the box but without distortion as seen here:
image

### 🤔 Expected Behavior
I believe Firefox shows the expected behavior.

### 👣 Steps to Reproduce
  - **WSR version**: `6.7.3`
  - **Yoshi version**: `4.alpha.6`

### 👀 Severity
Seems to be **Low**, but we had a bug opened for this.

### ✅ Proposed solution
I experimented with normalizing behavior across browsers and found that setting the following styles on the `` tag achieves consistent behavior:
```
img {
  width: 100%; // previously only height: 100% was set
  object-fit: cover;
}
```

amitdahan:
### 🔦 Summary
Closes #3091 

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test

Do I need to document & test this change?


argshook:
> Do I need to document & test this change?

this is a relatively small bug fix, don't think you  need explicit test for it.
screenshot diff would complain if anything were broken.

thanks for the fix!

PR #3094, opened 2019-03-12
Fix bug where chrome autofill brakes input borders

shl3vi:
### 🔦 Summary
Chrome autofill for form is breaking the input layout

more on : https://jira.wixpress.com/browse/EE-17186

### DIFF
--- a/src/Input/Input.scss
+++ b/src/Input/Input.scss
@@ -536,6 +536,10 @@ $material-gap: 8px;
   padding: 0 $double-gap;
 }
 
+.themedWrapper {
+	overflow-y: hidden;
+}
+
 .materialTitle {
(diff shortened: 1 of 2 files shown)


src/Input/Input.scss
```diff
@@ -536,6 +536,10 @@ $material-gap: 8px;
   padding: 0 $double-gap;
 }
 
+.themedWrapper {
+	overflow-y: hidden;
```
argshook: unfortunately this cannot be accepted as it is breaking other components using Input. please see CI failures with eyes tests

shlomitc:
closing due to inactivity - please re-open if you can work on fixing it. thanks!

PR #3096, opened 2019-03-12
<Input/> - add styles for read-only state

### LINKED ISSUE #2469: <input /> ReadOnly prop styling
### Relevant Components
[`<input />`](https://wix-wix-style-react.surge.sh/?selectedKind=Components&selectedStory=Input&full=0&addons=0&stories=1&panelRight=0) 

### Current Behavior
`ReadOnly` changes behaviour, but visually remains same as in default state

### Expected Behavior
Changing input's visual style when `ReadOnly` property is enabled.
Link to specs: https://zpl.io/2Egjola

image

### Severity
Low

TalZa1810:
### 🔦 Summary
Resolves #2469 

### ✅ Checklist
- 📚 Change is documented
  - [x] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [x] Visual test


shlomitc:
please fix the `MultiSelectCheckbox` readOnly and add the readOnly example to the `Input` component story

src/Dropdown/Dropdown.js
```diff
@@ -87,7 +87,7 @@ class Dropdown extends InputWithOptions {
   }
 
   inputClasses() {
-    const classes = { [styles.readonly]: true };
+    const classes = { [styles.showPointer]: true };
```
argshook: the whole `inputClasses` can be just

```js
classNames(styles.showPointer, { [styles.noBorder]: this.props.noBorder  })`
```

src/Dropdown/Dropdown.js
```diff
@@ -102,7 +102,8 @@ class Dropdown extends InputWithOptions {
 
   inputAdditionalProps() {
     return {
-      readOnly: true,
+      readOnly: false,
+      disableEditing: true,
```
argshook: what's the purpose of this new prop? is `readOnly` not enough?

src/Dropdown/Dropdown.spec.js
```diff
@@ -66,9 +66,9 @@ describe('Dropdown', () => {
       });
     });
 
-    it('should be read only', () => {
+    it('should not be editable ', () => {
       const { driver } = createDriver(<Dropdown options={getOptions()} />);
-      expect(driver.isReadOnly()).toBeTruthy();
+      expect(driver.isEditable()).toBeFalsy();
```
argshook: please be explicit
```suggestion
      expect(driver.isEditable()).toBe(false);
```

src/Input/Input.spec.js
```diff
@@ -217,6 +217,13 @@ describe('Input', () => {
       });
     });
 
+    describe('disableEditing attribute', () => {
+      it('should pass down to the wrapped input', async () => {
+        const { driver } = render(<Input disableEditing />);
+        expect(await driver.getReadOnly()).toBeTruthy();
```
argshook: please be explicit
```suggestion
        expect(await driver.getReadOnly()).toBe(true);
```

TalZa1810: Hi @argshook ,
All the tests in this file are using this convention. Should I convert all the tests in this file?

src/Dropdown/Dropdown.js
TalZa1810: The readOnly prop was used for the logic and the styling of the Input component. However, it changed also the Dropdown and the MultiSelectCheckbox components so that they get the readOnly styling even though it is not defined in the Zeplin file. Therefore, there was a need to separate the logic and the styling to two different props- readOnly for the styling and disableEditing for the logic. I hope it's clear.

src/Input/Input.spec.js
argshook: if it's not too much work then yeah, sure, why not

the `toBeTruthy` and `toBeFalsy` are no good because they can bring false positives. for example, if someone would forget `await`, then `expect( driver.getReadOnly()).toBeTruthy()` would always pass, because `driver.getReadOnly` returns a promise, which is truthy (but maybe it resolves to false)

this happened multiple times already, that's why i'm nit picking :)

src/Dropdown/Dropdown.spec.js
TalZa1810: Hi @argshook , 
The same here, all the tests in this file are using this convention. Should I convert all the tests in this file too?

PR #3097, opened 2019-03-12
Add dynamic imports support

yavorsky:
WIP. Let's merge it after wix/wix-ui#1184 and https://github.com/wix/wix-ui-backoffice/pull/263 will be merged.

This PR is a part of the work made to implement dynamic imports and tree shaking work.
I'll also open PRs for wix-style-react and wix-ui-core.

Here we want to: add logic to lazy-load Loader's Tooltip (~270KB) only when statusMessage prop is present and reuse ES components form wix-ui-backoffice only for `dist/es/src` directory.
Since we have auto-deploy enabled for wix-ui-core and backoffice, and some users can rely on src with commonjs stuff, we created `babel-plugin-src-to-es.js`, that updates paths inside ES files to use content from ES directory inside `wix-ui-core` and `wix-ui-backoffice`. We can safely migrate to ES6 modules + dynamic imports for all 3 libs (wsr, wix-ui-core, wix-ui-backoffice). Only users who use `import { Text } from 'wix-style-react'` or `import Text from 'wix-style-react/dist/es/src/components/Text' `


yavorsky:
@shlomitc b/c it's how almost all libs for lazy loading I saw implemented this idea.
For example, pretty popular react-loadable https://github.com/jamiebuilds/react-loadable/blob/master/src/index.js#L163

I see 2 benefits here:
- Backward compatibility. We have code splitting only for `dist/es` directory. So users only imported `import { Text } from 'w-s-r'` will have `import(X)` in their code. For guys, who are using commonjs stuff, it will be the same as before - we'll have simple `require` in code and webpack will include it in the main bundle. If we'll make `setState: Tooltip` in didMount, then immediately after render Tooltip won't be rendered. It could break a lot of tests, so users will wrap some checks it in setTimeout, etc.
- Performance. The earlier the Tooltip will start to load, the faster it will be loaded.

I think it's not the best practice for current supported version of React since it will have 1 render for sync loading using `require` and for async version with reacting lazy loading - it will have faster load - 2+ version of `import()` will take it from the cache and it should put this call on the next tick.

But if it's ok to break users with next release and expect tooltip will be rendered only on the next tick, we can move it to `didMount`.

PR #3105, opened 2019-03-14
add minRows prop - works with autoGrow + test it

jonathanadler:
### 🔦 Summary
We wanted to be able to have a growing input area, that starts with 1 row.
So We introduced a new prop called minRows that allows this to be possible.

### ✅ Checklist
- [ ✅] 👨‍🎨 UX change is approved by the Design System UX 
- 📚 Change is documented
  - [✅ ] Story
  - [✅ ] API description
- 🔬 Change is tested
  - [ ] Component
  - [✅ ] Visual test

### DIFF
--- a/src/InputArea/InputArea.e2e.js
+++ b/src/InputArea/InputArea.e2e.js
@@ -109,10 +109,16 @@ describe('input area page', () => {
     },
   );
 
-  eyes.it('should begin with miminum ammount of rows', async () => {
+  eyes.it('should begin with minimum amount of rows', async () => {
     await autoExampleDriver.setProps({ autoGrow: true });
     const rowCount = await inputAreaTestkit.getRowCount();
     expect(rowCount).toBe(InputArea.MIN_ROWS);
+  });
+
+  eyes.it('should begin with minRow amount of rows', async () => {
+    await autoExampleDriver.setProps({ autoGrow: true, minRows: 1 });
+    const rowCount = await inputAreaTestkit.getRowCount();
+    expect(rowCount).toBe(1);
   });
 
   eyes.it('it does not shrink below allowed minimum rows', async () => {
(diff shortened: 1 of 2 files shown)


src/InputArea/InputArea.e2e.js
```diff
@@ -109,12 +109,18 @@ describe('input area page', () => {
     },
   );
 
-  eyes.it('should begin with miminum ammount of rows', async () => {
+  eyes.it('should begin with minimum amount of rows', async () => {
```
shlomitc: I wouldn't change the typo because baseline's of applitools are using the name, so this means it's a new baseline (and wouldn't know if something is broken)

src/InputArea/InputArea.js
```diff
@@ -334,6 +345,9 @@ InputArea.propTypes = {
   /** Will cause the Input Area to grow and shrink according to user input */
   autoGrow: PropTypes.bool,
 
+  /** Sets the minimum amount of rows the component can have when in autoGrow mode */
+  minRows: PropTypes.number,
```
shlomitc: so now we have both `rows` and `minRows`, and it doesn't work if the autoGrow is not applied.

one option will be renaming it `minRowsAutoGrow`, so it will be clear.

what do you think just to apply the logic using the `rows` prop and act as a `minRows`?

src/InputArea/InputArea.e2e.js
jonathanadler: Right, didn't think about that.
It's my OCD.

src/InputArea/InputArea.js
jonathanadler: The problem is that `rows` is an html attribute, that freezes the amount of rows, even when we have autoGrow.
So it won't work with that.

Renaming could be the best solution in this case.

Issue #2754, opened 2019-01-14
<Tooltip> fails at positioning


hitpopdimestop:
# Bug Report

### Relevant Components
`<Tooltip />`

### Current Behavior
When we hover the icon or text tooltip sometimes appears somewhere at the bottom of the screen and then moves to the top position.

### Expected Behavior
Tooltip should appear at the expected position.

### Steps to Reproduce
I made a screencast to show the behavior:
https://drive.google.com/open?id=18Vu4ra-27728asC_vGThdqpXmJz51VLP

My code for Tooltip:
```
    <Tooltip
      content={props.typeDetails}
      theme="dark"
      appendToParent={true}
    >
      <div className={styles.icon}>
        <InfoCircle size="20px" />
      </div>
    </Tooltip>
```
And CSS (if it might be needed):
```
.icon {
  margin-left: 12px;
  position: relative;
  top: 3px;
}

.icon:hover {
  cursor: pointer;
  color: #3899ec;
}
```

### Severity
Major

shlomitc:
Hey. it's a long time existing bug. We plan on rewriting the component to a far better implemented one where this issue and more will be fixed.

I would happily accept PRs with a possible fix if you can create one

sidoruk-sv:
Also got into this issue: when tooltip is first-rendered it's TooltipContent has `position: absolute`, but as there is no `top/left/bottom/right` coordinates it's behaves like `position: relative`.

And there is a bug when we use tooltip with `iframe-resizer` and we got triggering resize for such case:
* video: https://take.ms/mbrHS
* wrong positioning that produce resizing: https://take.ms/4HpTa
* correct position after all styles applied: https://take.ms/KqHWa

Also you can see "blinking" of tooltip in the left corner on this storybook (hover over overflowed text):
https://my-domains.surge.sh/?selectedKind=Modals%2FConfigure%20Mailbox%20Records&selectedStory=Default&full=0&addons=0&stories=1&panelRight=0

For me it is fixed by specifying `top:0; left: -9000px;` (not to be shown in the view on "intermediate" faze), after that `react-tooltip` sets it's own `top, left` properties and tooltips shows to the user in correct position;

@mykas @shlomitc
So for now don't know where to fix - `master` branch, or in scope of #3110?

shlomitc:
@sidoruk-sv - a fix to `master` will be great

mykas:
[...]

Issue #2582, opened 2018-12-13
`<Tooltip/>` - Does not support SFC as children


hepiyellow:
## Cause
This is because we try to take a `ref` of the child
https://github.com/wix/wix-style-react/blob/0a76689f6fca8a0402719136528ba6c066e90356/src/Tooltip/Tooltip.js#L333

## Solution
We could instead do :
```js
componentDidMount() {
    super.componentDidMount && super.componentDidMount();
    this._childeNode = ReactDOM.findDOMNode(this);
    this._setContentDataHook();
  }

  componentDidUpdate(prevProps) {
    this.childeNode = ReactDOM.findDOMNode(this);
    if (prevProps.dataHook !== this.props.dataHook) {
      this.contentHook = this._createContentHook();
      this._setContentDataHook();
    }
    this.renderTooltipIntoContainer();
  }

```

mykas:
Fixed. https://wix-wix-style-react.surge.sh/?selectedKind=7.%20Tooltip&selectedStory=7.1%20Tooltip&full=0&addons=0&stories=1&panelRight=0

Issue #2069, opened 2018-08-01
<Tooltip/> - onClickOutside triggers un-needed rendering


katmai7:
### Summary:
hide() is calling even if a tooltip is hidden which leads to component rendering,
because it calls setState

### Priority:
Normal

### Steps to reproduce:
 1. Add 2 tooltips(or component based on it)
 2. Set up breakpoint  https://github.com/wix/wix-style-react/blob/9c309cddb8d580bcd96fe688a173acb57992428e/src/Tooltip/Tooltip.js#L297
 3. Click somewhere on the page
 4. Check that hide() method is called even if the tooltip is hidden

### Expected behavior:
Do not call hide() if the tooltip is hidden

### Additional notes:
Check other components(like calendar, input with options, dropdown) which are using `onOutsideClick`

shlomitc:
Can you please supply a simple code snippet that shows the issue?
you can you this playground:
https://codesandbox.io/s/mymk2zn1mp

katmai7:
@shlomitc Hi, it's not a bug with tooltip `hide` callback, but more perf optimization for a  tooltip. 
For now, if you have a lot of tooltips on a page, every click on it will call `setState` inside `hide` method
https://github.com/wix/wix-style-react/blob/9c309cddb8d580bcd96fe688a173acb57992428e/src/Tooltip/Tooltip.js#L297
even if a tooltip is hidden.

shlomitc:
how do you suggest fixing it?

katmai7:
@shlomitc I think we can add simple if statement at the beginning of the `hide` like:
```
if(!this.state.visible) {
  return;
}
```

Or if it possible, subscribe to the 'click outside' when a tooltip is shown and unsubscribe when a tooltip is hidden.

shlomitc:
checking the state in ‘onClickOutside’ hook sounds good to me

hepiyellow:
How about if we do :  
```js
!this.state.hidden && this.setState({hidden: true}); 
```

hepiyellow:
Looking into it a bit further, I realized that there is no re-render when clicking outside.
That is, because Tooptip extends WixComponent which extends PureComponent.
PureComponent have a built in optimization that re-renders only if state or props are changed,
and calling `hide()` when the Tooltip is not shown, it does not change the state. (`hidden` stays `true`)

PR #3110, opened 2019-03-15
<Tooltip/> - major refactor

mykas:
# `<Tooltip/>`

## Issues with the current Tooltip

- Uses our own naive postioning mechanism, [which](https://github.com/wix/wix-style-react/issues/2754) [has](https://github.com/wix/wix-style-react/issues/2596) [several](https://github.com/wix/wix-style-react/issues/2582) [issues](https://github.com/wix/wix-style-react/issues/2069). The mechanism is not at mature and community backed as [`Popper.js`](https://github.com/FezVrasta/popper.js).
- Can be both Controlled and Uncontrolled, leads to confusion.
- Does not follow the UX guidelines:
  - A tooltip should only be opened on HOVER, currently the consumers can set the triggering
    events.
  - A tooltip should only contains textual content, currently it can accept any node.
  - Our `<Tooltip/>` could be used as Popover (opens on click), leads to confusion.
  - The Tooltip should has a `dark` theme, currently it's defaulted to a `light` theme.

## The solution

> Update the `<Tooltip/>` component according to the UX guidelines and improve its positioning mechanism.

This "mechanism" is ready, and it's called `<Popover/>`. This is a controlled component that uses
the `Popper.js` library to handle positioning.

### What's next about `<Tooltip/>`

The plan is to create a **new** `<Tooltip/>`. As discussed, it should:

- Use the `<Popper/>` component for positioning;
- Be **Uncontrolled**, and **only open on HOVER**, as per the UX guidelines;
- Has a `dark` theme by default (and the theme cannot be changed);
- Has a slimmer API;
- Provides clear documentation on how to achieve various way of positioning.
- Uses UniDriver.

### General Guidelines for Users

The `Tooltip` pattern is intended to contain text-only content,
and the `Popover` pattern is intended for tooltips with interactive content.

Avoid putting complex structured text content (e.g. with headings, bullet lists, etc) in a Tooltip. For anything mildly complex/structured, it'd be best for authors to go for a Popover.

### Basic usage

```jsx
import React from 'react';
import Tooltip from 'wix-style-react/Tooltip';

const MyComponent = () => (
  <div>
    <Tooltip placement="top" content="This content explains buttons meaning.">
      <Button>Hover on me to see my meaning</Button>
    </Tooltip>
  </div>
);

export default MyComponent;
```

### API

| Name          | Type                        | Default value | Required | Description                                                                                                                                               |
| ------------- | --------------------------- | ------------- | -------- | --------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `dataHook`    | `string`                    |               |          | Used for testing purposes. DataHook is applied on tooltip content container.                                                                              |
| `content`     | `string`, `ReactNode`       |               |          | The content to be inside the `<Tooltip/>`. If `ReactNode`, should be a textual component (such as `<Text/>` and `<TextButton/>` as per the UX guidelines. |
| `appendTo`    | `string`                    | `parent`      |          | Delegated to `<Popover/>` appendTo.                                                                                                                       |
| `children`    | `ReactNode`, `string`       |               |          | The trigger element.                                                                                                                                      |
| `exitDelay`   | `Number`                    |   0             |          | The delay for tooltip to dissapear after mouse out event happens.                                                                                         |
| `enterDelay`  | `Number`                    |    0           |          | The delay for tooltip to appear after mouse down event happens.                                                                                           |
| `onHide`      | `Function`                  |               |          | Callback function to be called when hidden                                                                                                                |
| `onShow`      | `Function`                  |               |          | Callback function to be called when shown                                                                                                                 |
| `placement`   | `string`                    | `'top'`       |          | Delegated to `<Popover>` placement.                                                                                                                       |
| `maxWidth`    | `Number`                    | `204`         |          | Sets maximum width of tooltip content.                                                                                                                    |
| `textAlign`   | `oneOf ['center', 'start']` | `start`       |          | To align content text.                                                                                                                                    |
| `size`   | `oneOf['small', 'medium']`                   |               |          | Toltip size                                                                                                                        |
| `moveTo`      | `{ x: Number, y: Number }`  |               |          |    Moves tooltip relative to the parent                                                                                                                                                                                                                                                                                                      
| `flip`        | `Boolean`                   | `true`        |          |                                                                                                                                                           |
| `fixed`       | `Boolean`                   | `false`       |          |                                                                                                                                                           |
| `zIndex`       | `string`                   | ``       |          |                                                                                                                                                           |



### Accessibility & Keyboard Navigation

A tooltip is a popup that displays information related to an element when the element receives keyboard focus or the mouse hovers over it. It typically appears after a small delay and disappears on blur or mouse leave.

Neil example: https://codepen.io/WW3/pen/BQzEKE?editors=1000

- Pressing TAB will focus Button and open Tooltip.
   - Pressing TAB again will blur the component and close the Tooltip.


### Migration plan

We'll use the `upgrade` prop method like we did for the new `<Button/>`. We'll deprecate the "old" `<Tooltip/>` and remove it completley in the next major Version (V7? 🙏).


hepiyellow:
@mykas Can you share the goal/plan of this refactor?

mykas:
@hepiyellow I will add api and plan to the description soon.

mykas:
> @mykas Can you share the goal/plan of this refactor?

Spec is added to the description.

mykas:
@hepiyellow @shlomitc @argshook @nitayneeman @NitayRabi @boriskor @TalZa1810 @eyaleizenberg @mosheKerbel Lets start a discussion here regarding Tooltip refactor. If you have any ideas or suggestions: feel free to state them here.

argshook:
> Presing TAB will focus Button.
Pressing ENTER or SPACE opens/closes the tooltip.

this is unexpected, because once button is focused <kbd>enter</kbd> should invoke `onClick`

imo once button (or any other Tooltip children) is focused, then tooltip should appear without additional action

mykas:
> > Presing TAB will focus Button.
> > Pressing ENTER or SPACE opens/closes the tooltip.
> 
> this is unexpected, because once button is focused enter should invoke `onClick`
> 
> imo once button (or any other Tooltip children) is focused, then tooltip should appear without additional action

The advantage of automatically opening the tooltip when the element receives focus is that this makes the tooltip very discoverable for keyboard users, but the (big) disadvantage is that keyboard power users very quickly become tired of them, even when there is a delay.

mykas:
So regarding a11y the main question is: Which keyboard key is the trigger of Tooltip component? Or should it open automatically open focus?

mykas:
Another big question is: should we use wsr `Popover` component or wix-ui-core to proxy and restrict?

Pros for wsr:
- User can refer to Popover api in wix-style-react though Tooltip would only use `appendTo` , `placement`,  `flip`, `fixed`, `moveTo` apis which are the same on core Popover. So they could refer to wsr Popover and api would not be different to core Popover.
- Unified dark theme styling for Popover and Tooltip.

Cons for wsr:
- Component weight increase. We would have: Popover(core) -> Popover(wsr) -> Tooltip. Instead of Popover(core) -> Tooltip.
- Debugging. Three layers could be quite complex for junior devs to grasp.

carmacleod:
@mykas We use `SHIFT+F1` to focus the "current" tooltip in our web app. All browsers, all platforms.
The "current" tooltip (in our app) is the one for either the currently-focused control, or the caret location in our editor. `SHIFT+F1` seems to be a relatively common shortcut for this purpose... it's listed in this [Table of Keyboard Shortcuts](https://en.wikipedia.org/wiki/Table_of_keyboard_shortcuts)... :)

image

mykas:
#### After conversation with Neil (WIX a11y expert) here are some notes:
- There are two types of tooltips: tool-tip & toggle-tip (example here: https://codepen.io/WW3/pen/BQzEKE?editors=1000)
- Tooltip by Neil suppose to open on focus. Though he agrees that power users can get irritated by tooltips opening on focus. He will do a research on keyboard shortcuts as alternative and see what fits us.

@carmacleod thanks for your response! `SHIFT + F1` sounds reasonable to invoke the tooltip, but I am just wondering if `F1` is available for web too. This could shorten the key combination for Mac os users (`FN + F1` instead of `FN + SHIFT + F1`)

hepiyellow:
### Copying text
What about the use-case where a user wants to copy-to-clipboard the full text of a `<Text/>` with ellipsis.
If the tooltip closes when you hover on the tooltip content, then the user can not select the text.

### String Content Only

Why not restrict the content to a `string`?
And use `<Popover>` for any other content.

### Arrow

I always wondered what is the use-case for NOT showing an arrow? @wuwa @milkyfruit ?

### zIndex

Tooltip has a default zIndex which is bounded to BM's zIndex layers definitions.
But it should have a `zIndex` prop that sets the zIndex of the content.

### Style API - content

Does this content pseudo element (selector) allow to set the backgroud and the border of the Tooltip content?

### Focus Opens Tooltip
Regarding you previous discussion with @argshook 

> Presing TAB will focus Button.
Pressing ENTER or SPACE opens/closes the tooltip.

I also had a meeting with Neil (A11y) and he confirmed this behavior. (Opens on focus, I guess closes on blur)

carmacleod:
@mykas `F1` in Chrome browser opens "Chrome Help" in a new tab. I guess we didn't want to override that, although looking at it now, it's probably not all that different than googling.  ;)
The other consideration is that `F1` is semantically more appropriate for "global" help, i.e. maybe open your app's Help pages, or similar. `SHIFT+F1` has traditionally been for "context sensitive help", so that's why we went with it for tooltips. At any rate, I'd be curious to know what Neil's research suggests. :)

mykas:
After speaking with @milkyfruit (UX) we decided that tooltip "boredom" on focus is not a case for us in WIX. We believe that this "boredom" might happen when multiple icon buttons appear in the ui (one by one) and it that case this might be true, but even in this case for example in Google Docs tooltip is shown on focus too. 

So in general this is what we will have:

- Pressing TAB will focus Button and open Tooltip.
    - Pressing TAB again will blur the component and close the Tooltip.

carmacleod:
That makes sense. Always good to discuss with the UX folks.  ;)

Our use case is a bit different - we have a code editor where tooltips can appear for comments, errors, warnings, notations, etc. Which tooltip to show depends on the position of the caret in the text. It would be insane for us to show tooltips automatically while a keyboard user is arrowing through their code.  :)
So we added the `Shift+F1` behavior because of the editor, and then we used the same mechanism on the toolbar buttons, etc. for consistency. Users got used to typing `Shift+F1` in the editor, so that made it discoverable in the rest of the app. (Just thought I'd explain more.  :)

mykas:
Thanks @carmacleod for your input here!

milkyfruit:
Copying text from a tooltip
It is not possible either with a mouse either with a keyboard. We don't support it.

Arrow
It's a styling thing. Now for a tooltip, instead of giving a prop show/hide an arrow, it should have size: regular/compact. It needs a research and specs first.

@mykas is this update planned as a breaking change? Or props will remain the same?

mykas:
@milkyfruit this will be a breaking change. Props can change

src/Tooltip/TooltipNext/Tooltip.js
```diff
@@ -0,0 +1,99 @@
...
+    maxWidth: 204,
+    textAlign: 'center',
+    size: 'medium',
+    zIndex: 2000,
+  };
+
+  _renderContent = () => {
+    const { content, maxWidth, zIndex, textAlign } = this.props;
+
+    const style = { maxWidth: `${maxWidth}px`, zIndex, textAlign };
```
argshook: `px` shouldn't be needed, react adds it itself, iirc

src/Tooltip/TooltipNext/Tooltip.js
```diff
@@ -0,0 +1,99 @@
...
+    const text = (
+      <Container>
+        <Text dataHook="tooltip-text" size="small" weight="normal" light>
+          {content}
+        </Text>
+      </Container>
+    );
+    const node = <Container>{content}</Container>;
+
+    return typeof content === 'string' ? text : node;
```
argshook:
```js
_renderContent = () => {
  const { content, maxWidth, zIndex, textAlign } = this.props;

  return (
    <div style={{ maxWidth, zIndex, textAlign }}>
      { typeof content === 'string' ?
        <Text dataHook="tooltip-text" size="small" weight="normal" light>
          {content}
        </Text> :
        node
      }
    </div>
  );
};
```

src/Tooltip/TooltipNext/Tooltip.st.css
```diff
@@ -0,0 +1,161 @@
+:import {
+  -st-from: "wix-ui-core/index.st.css";
+-st-named: Tooltip;
```
argshook:
```suggestion
  -st-named: Tooltip;
```

src/Tooltip/TooltipNext/Tooltip.st.css
```diff
@@ -0,0 +1,161 @@
...
+
+.root::popover[data-placement]::arrow {
+  border-width: 5px;
+  border-color: transparent value(D10) value(D10) transparent;
+  margin: 0;
+}
+
+
+/* Right side */
+.root:size(small)::popover[data-placement*="right"]::arrow  {
```
argshook: are you not able to derive position as stylable state from props to style different positions?

src/Tooltip/TooltipNext/Tooltip.js
```diff
@@ -0,0 +1,99 @@
...
+import Text from '../../Text';
+import styles from './Tooltip.st.css';
+
+/**
+ * Next Tooltip
+ */
+class Tooltip extends React.PureComponent {
+  static displayName = 'Tooltip';
+
+  state = { isOpen: false };
```
argshook: unused?

src/Tooltip/TooltipNext/Tooltip.js
mykas: 💅 so beautifull

src/Tooltip/TooltipNext/Tooltip.st.css
mykas: Nope, because Popover calculation mechanism adds style tags with calculations and it will ignore any classnames

src/Tooltip/TooltipNext/Tooltip.js
mykas: true

mykas:
FIY: The delay for showing and hiding has been aligned to what UX has defined. According to them the delay for enter and exit should be: 350 ms

mykas:
@milkyfruit please review the docs

src/Tooltip/README.SPEC.md
```diff
@@ -1,79 +1,90 @@
-vim:nowrap:
```
argshook: :)

src/Tooltip/TooltipNext/docs/index.story.js
```diff
@@ -0,0 +1,216 @@
...
+import Tooltip from '..';
+import TextButton from '../../../TextButton';
+import Text from '../../../Text';
+import AddItem from '../../../AddItem';
+
+import { placements } from '../../../Popover/Popover';
+import SectionHelper from '../../../SectionHelper';
+import { Layout, Cell } from '../../../Layout';
+import { baseScope } from '../../../../stories/utils/LiveCodeExample';
+import usage from '!raw-loader!./Usage.md';
```
argshook:
```suggestion
import usage from './Usage.md';
```
this also works

src/Tooltip/TooltipNext/docs/index.story.js
```diff
@@ -0,0 +1,216 @@
...
+        sections: [
+          columns([
+            description({
+              text:
+                'A tooltip is a popup that displays information related to an element when the element receives keyboard focus or the mouse hovers over it.',
+            }),
+          ]),
+
+          importExample({
+            source: "import Tooltip from 'wix-style-react/Tooltip';",
```
argshook: no need for `source` anymore, `importExample('string')` is enough

src/Tooltip/TooltipNext/docs/index.story.js
```diff
@@ -0,0 +1,216 @@
...
+      ),
+    }),
+
+    tabs([
+      tab({
+        title: 'Description',
+        sections: [
+          columns([
+            description({
+              text:
```
argshook: if `description` has no `title`, then you can `description('string right here')`

src/Tooltip/README.SPEC.md
mykas: :)

PR #3115, opened 2019-03-18
<MultiSelect/> - Fix line-break issue

mosheKerbel:
https://github.com/wix-private/wsr-issues/issues/412
fixing it by setting a boolean flag of autoSize for the actual input

### DIFF
--- a/src/Input/Input.js
+++ b/src/Input/Input.js
@@ -396,7 +396,7 @@ const ActualInput = props => {
 
   if (autoSize) {
     return (
-      <AutosizeInput {...rest} inputClassName={className} inputRef={reactRef} />
+      <AutosizeInput {...rest} inputClassName={className} inputRef={reactRef} inputStyle={{minWidth: "0px"}} />
     );
   } else {
     return <input {...rest} className={className} ref={reactRef} />;
--- a/src/Input/Input.js
+++ b/src/Input/Input.js
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import AutosizeInput from 'react-input-autosize';
 
 import Ticker from './Ticker';
 import Unit from './Unit';
@@ -116,6 +117,7 @@ class Input extends Component {
       required,
       error,
       errorMessage,
+      autoSize
     } = this.props;
     const onIconClicked = e => {
       if (!disabled) {
@@ -170,15 +172,14 @@ class Input extends Component {
 
     /* eslint-disable no-unused-vars */
     const { className, ...inputElementProps } = props;
-
     const inputElement = (
-      <input
+      <ActualInput
         min={min}
         max={max}
         step={step}
         data-hook="wsr-input"
         style={{ textOverflow }}
-        ref={this.extractRef}
+        reactRef={this.extractRef}
         className={inputClassNames}
         id={id}
         name={name}
@@ -204,6 +205,7 @@ class Input extends Component {
         autoComplete={autocomplete}
         onCompositionStart={() => this.onCompositionChange(true)}
         onCompositionEnd={() => this.onCompositionChange(false)}
+        autoSize={autoSize}
         {...ariaAttribute}
         {...inputElementProps}
       />
@@ -386,6 +388,19 @@ Input.defaultProps = {
   maxLength: 524288,
   withSelection: false,
   clearButton: false,
+  autoSize: false
+};
+
+const ActualInput = props => {
+  const { autoSize, className, reactRef, ...rest } = props;
+
+  if (autoSize) {
+    return (
+      <AutosizeInput {...rest} inputClassName={className} inputRef={reactRef} />
+    );
+  } else {
+    return <input {...rest} className={className} ref={reactRef} />;
+  }
 };
 
 const borderRadiusValidator = (props, propName) => {
@@ -562,6 +577,9 @@ Input.propTypes = {
 
   /** Step steps to increment/decrement - similar to html5 step attribute */
   step: PropTypes.number,
+
+  /** Use AutosizeInput instead of the deafult html input tag (for cases where we need to start with a minimal input size and resize according to the input */
+  autoSize: PropTypes.bool
 };
 
 export default Input;
--- a/src/Input/Input.js
+++ b/src/Input/Input.js
@@ -1,7 +1,6 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
-import AutosizeInput from 'react-input-autosize';
 
 import Ticker from './Ticker';
 import Unit from './Unit';
@@ -117,7 +116,7 @@ class Input extends Component {
       required,
       error,
       errorMessage,
-      autoSize
+      customizedInput
     } = this.props;
     const onIconClicked = e => {
       if (!disabled) {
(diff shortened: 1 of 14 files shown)


mosheKerbel:
@shlomitc i wonder how to test it
what do you think about a visual testing that will take a snapshot of the muiltiselect with a few options selected, to verify that the input visualisation occupies 1 line and not two lines?

src/Input/Input.js
```diff
@@ -370,6 +372,16 @@ class Input extends Component {
...
+
+  _actualInput = props => {
+    const { customInput, reactRef, ...rest } = props;
+    if (customInput) {
+      const CustomInputComponent = customInput;
+      return <CustomInputComponent reactRef={reactRef} {...rest} />;
+    } else {
+      return <input {...rest} ref={reactRef} />;
+    }
+  };
```
NitayRabi: You can use 
```
React.cloneElement(customInput || <input />, props);
```
instead (e.g- https://github.com/wix/wix-style-react/blob/master/src/DateInput/DateInput.js#L41)

src/Input/Input.spec.js
```diff
@@ -873,5 +873,23 @@ describe('Input', () => {
...
     });
+
+    describe('input render', () => {
+      it('should render customized input', async () => {
+        const className = 'foo';
+        const customInput = () => <input className={className} />;
+        const { driver } = render(<Input customInput={customInput} />);
+        const classes = await driver.getInputElementClasses();
+        expect(classes.contains(className)).toEqual(true);
+        expect(classes.length).toEqual(1);
```
NitayRabi: I would use `data-hook` set on the custom component to validate it exists instead of class.

src/Input/Input.spec.js
```diff
@@ -873,5 +873,23 @@ describe('Input', () => {
...
+      });
+
+      it('should render default input', async () => {
+        const { driver } = render(<Input />);
+        const classes = await driver.getInputElementClasses();
+        expect(classes.contains('input')).toEqual(true);
+        expect(classes.length).toEqual(1);
```
NitayRabi: I would validate `input.exists()` (using the driver)

src/MultiSelect/MultiSelect.e2e.js
```diff
@@ -117,6 +117,24 @@ describe('MultiSelect', () => {
...
+
+    eyes.it('should fill input with tags without breaking a line', async () => {
+      const url = createTestStoryUrl({
+        category: storySettings.category,
+        storyName: storySettings.storyName,
+        testName: testStories.lineNotBraking,
+      });
+      await browser.get(url);
+      const driver = multiSelectTestkitFactory({
+        dataHook: 'multi-select-line-not-braking',
```
NitayRabi: use data hook from `storySettings`

src/MultiSelect/docs/tests/TestLineNotBraking.js
```diff
@@ -0,0 +1,19 @@
+import MultiSelect from '../../MultiSelect';
+import React from 'react';
+
+export default () => (
+  <div style={{ width: '400px' }}>
+    <MultiSelect
+      dataHook="multi-select-line-not-braking"
```
NitayRabi: use data hook from `storySettings`

src/MultiSelect/MultiSelect.spec.js
```diff
@@ -85,6 +85,13 @@ describe('MultiSelect', () => {
     expect(dropdownLayoutDriver.optionsLength()).toBe(options.length);
   });
 
+  it('should render AutoSizeInput as the input element', () => {
+    const { inputDriver } = createDriver(<MultiSelect options={options} />);
+
+    const inputElementClasses = inputDriver.getInputElementClasses();
+    expect(inputElementClasses.contains('autoSizeInput')).toEqual(true);
```
NitayRabi: use `styles.autoSizeInput` instead

src/Input/Input.js
mosheKerbel: indeed looks better but i have to use a different prop name for the ref

src/Input/Input.spec.js
mosheKerbel: done

src/Input/Input.spec.js
```diff
@@ -873,5 +873,23 @@ describe('Input', () => {
...
+      it('should render customized input', async () => {
+        const className = 'foo';
+        const customInput = () => <input className={className} />;
+        const { driver } = render(<Input customInput={customInput} />);
+        const classes = await driver.getInputElementClasses();
+        expect(classes.contains(className)).toEqual(true);
+        expect(classes.length).toEqual(1);
+      });
+
+      it('should render default input', async () => {
```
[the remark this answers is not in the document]
mosheKerbel: done

src/Input/Input.spec.js
mosheKerbel: replaced classes check with data-hook

src/MultiSelect/MultiSelect.e2e.js
mosheKerbel: done

src/MultiSelect/docs/tests/TestLineNotBraking.js
mosheKerbel: done

src/MultiSelect/MultiSelect.spec.js
mosheKerbel: done

src/Input/Input.js
mosheKerbel: we are using react ref to get a reference to the dom element.
so far we used the native html input, which get the "ref"
but when we use a customized input, like autoSizeInput, we must supply the ref with an alias, otherwise react will mess it up.
u can take at this [link](https://github.com/JedWatson/react-input-autosize/blob/master/src/AutosizeInput.js#L195) - it expects ref from outside.

mosheKerbel: i can not pass it as a "ref" prop because react will recognize it and use it, so i use an alias - dataRef.

src/MultiSelect/MultiSelect.spec.js
```diff
@@ -85,6 +87,13 @@ describe('MultiSelect', () => {
     expect(dropdownLayoutDriver.optionsLength()).toBe(options.length);
   });
 
+  it('should render AutoSizeInput as the input element', () => {
+    const { inputDriver } = createDriver(<MultiSelect options={options} />);
+
+    const inputElementClasses = inputDriver.getInputElementClasses();
```
shlomitc: instead of verifying classes - verify the data-hook of the AutoSizeInput

src/MultiSelect/MultiSelect.e2e.js
```diff
@@ -117,6 +117,24 @@ describe('MultiSelect', () => {
...
+        testName: testStories.lineNotBraking,
+      });
+      await browser.get(url);
+      const driver = multiSelectTestkitFactory({
+        dataHook: storySettings.dataHook,
+      });
+      const element = driver.element();
+      await waitForVisibilityOf(element, 'Cannot find <MultiSelect/>');
+
+      const height = await driver.getHeight();
```
shlomitc: can be removed, just a simple screenshot

mosheKerbel:
https://github.com/wix/wix-style-react/pull/3146

PR #3119, opened 2019-03-18
Set default top/left css properties for `TooltipContainer`

### LINKED ISSUE #2754: <Tooltip> fails at positioning

sidoruk-sv:
Fixes #2754


sidoruk-sv:
@mykas can you review this, please)

sidoruk-sv:
Guys, please review this pull request, I have blocked production issue because of the #2754 (that is fixed by this short PR).

src/Tooltip/TooltipContent.scss
```diff
@@ -5,6 +5,8 @@
 .root {
   position: absolute;
   z-index: $zIndex_TOOLTIP;
+  top: 0;
+  left: -9000px;
```
mykas: hey @sidoruk-sv , why this number `-9000` ? why not just `-1`?

sidoruk-sv: @mykas cause this tooltip is still showing in wrong place, while component is initialized.

That's why it is moved out of viewport (for any cases, TV or very large screen I think `-9000` is enough)

mykas: oh I see. So this is sort of hack before it gets rendered properly, right?

sidoruk-sv: @mykas yep

mykas: While I be live this hack fixes your issue, but I see now another kind of issue introduced with this change. Check this out how it moves to the right after fraction of a second.
Tooltip

mykas: I saw in the issue link where you described this bug - you said that this happens when `iframe-resizer` is triggered, right?

sidoruk-sv: Yes, ifra-resizer is triggered cause this parameters are not setted at all, and this produced "static" behaviour for <Tooltip/>

mykas: What if you trigger `shouldUpdatePosition` method on Tooltip? Have you tried that?

sidoruk-sv: also don't realize how this fix is produce that behaviour.

Checked on master, and this behaviour with arrow is introduced earlier, you can check that on master:
https://wix-wix-style-react.surge.sh/?selectedKind=7.%20Tooltip%20and%20Popover&selectedStory=7.1.%20Tooltip&full=0&addons=0&stories=1&panelRight=0

mykas: it does not happen on master. At least on my mac + chrome.

sidoruk-sv: @mykas recorded several seconds ego:
https://take.ms/fdlmi

sidoruk-sv: > What if you trigger `shouldUpdatePosition` method on Tooltip? Have you tried that?

Don't think that this would help, cause tooltip is static by default, before top/left parameters are set. And after that parameters are set there are no any  position changing logic that can appear.

My fix is only about setting "default" values that are overridden in a milliseconds after tooltip is rendered. But this tiny bug produced more visible `iframe-resizer`'s twitching.

mykas: Well we can say that this is `acceptable` hack until we resolve the new Tooltip with proper calculation mechanism.

mykas:
Lets wait until this gets green and then we can merge.

sidoruk-sv:
@mykas don't know what wrong with `continuous-integration/teamcity` but it feels too flicky to not merge this PR (or I will try to re-start and re-start tests until I got into green state).

mykas:
@sidoruk-sv its a bit flaky today with timeouts, but it is important to see at least once that this PR does not break tooltip usage on our other components.

sidoruk-sv:
> @sidoruk-sv its a bit flaky today with timeouts, but it is important to see at least once that this PR does not break tooltip usage on our other components.

that was some time ago)
https://take.ms/WvTZ9w

mykas:
ok, lets merge then

PR #3121, opened 2019-03-18
`<Page/>` - Add `z-index: 1` to `PageWrapper`

### LINKED ISSUE #3083: `<Page.Sticky/>` - Has higher `zIndex` then the BM's User Popover
### The Issue

@varzager Can you also share a link to production (with feature-toggle or static override details on how to reproduce)

>The lovely Island with birds and a palm tree in the below image, is a `<Page/>` content with `position: sticky` and a very high `zIndex` (11000).

image (2)

### Solution

Either we coordinate `zIndex`-s with BM.
OR
Is there a way to isolate Stacking context so that BM will always be on top of the Page's content?

hepiyellow:
Resolves #3083 

### What Changed
- Extracted a `Zindex.scss` and `Zindex.js` - This is just a refactor "start" of having Zindexs in one place.
- Add zIndex to `Page` (This affects only `<Page upgrade/>`)

### Temp Fix
This is actually a temporary fix since the real fix should come from the BusinessManager.
See this [Jira Ticket](https://jira.wixpress.com/browse/WOS2-1721)

Once BM applies the `z-index: 1` to the content-area, then we can remove the zIndex from Page.
> The `ZIndex` infrastructure will still be needed for Modals and other Layers.

### The Big Picture
See this [WIP Issue](https://github.com/wix-a/wsr-issues-erez/issues/34)

### DIFF
--- a/package.json
+++ b/package.json
@@ -208,7 +208,8 @@
       "<rootDir>/test/jest-setup.js"
     ],
     "setupFilesAfterEnv": [
-      "wix-ui-test-utils/jest-setup"
+      "wix-ui-test-utils/jest-setup",
+      "<rootDir>/test/jest-setup-after-env.js"
     ],
     "moduleNameMapper": {
       "\\.(?!\\st).(css|less|scss)$": "identity-obj-proxy"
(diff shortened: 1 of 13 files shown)


hepiyellow:
Replaced by #3136

PR #3130, opened 2019-03-20
`<Modal>`: Fix- large content gets the header cut

### LINKED ISSUE #2344: <Modal/> -  not scrollable to to if content is too high
# Bug Report
If modal doesn't fit the window top part of modal is hidden and it's not possible to scroll to it. And in our product user cannot close modal and stays with our wizard forever.
### Relevant Components

Modal
### Current Behavior

If modal is too high and doesn't fit window user can scroll to bottom but not to the top of modal
### Expected Behavior

All modal content should be scrollable, visible and acceceble
### Steps to Reproduce
[https://codesandbox.io/s/xr6rq60oxo](https://codesandbox.io/s/xr6rq60oxo)

www wix com_dashboard_9a70d363-f222-4512-a385-6032257d292f_store_products_product_59bdc5db-787c-2e4a-4bf9-70609769f62d

### Severity
- [ ] Critical
- [x] Major
- [ ] Low

TalZa1810:
### 🔦 Summary
Resolves #2344 

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test

### DIFF
--- a/src/MessageBox/HeaderLayout.scss
+++ b/src/MessageBox/HeaderLayout.scss
@@ -14,6 +14,7 @@
     justify-content : space-between;
     border-radius   : 8px 8px 0px 0px;
     line-height     : 24px;
+    margin-bottom: 30px;
 }
 .title-label{
     display: inline-block;
--- a/src/MessageBox/MessageBoxFunctionalLayout.scss
+++ b/src/MessageBox/MessageBoxFunctionalLayout.scss
@@ -2,17 +2,22 @@
 @import '../mixins';
 
 $minimum-modal-margin: 48px;
+$footer-height: 90px;
+// Header component is 60px height with 30px margin-bottom
+$header-height: 90px;
 
 .content {
   box-shadow: $Shadow40;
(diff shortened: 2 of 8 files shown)


shlomitc:
There are a lot of changes which are not just fixing the modal cut.
please describe the changes, so we can decide if they're mandatory or not.

1. Everything related to the overflowing is not required for this PR, in my opinion.
2. also, changes to the `.fullscreen-content` class - is this changing the behavior? what's the difference now betwen the fullscreen and not fullscreen?
3. I don't see any change to the actual `Modal`, so I guess some fix must be done there as well

TalZa1810:
I've created another Modal with a div with a height which is higher than the viewport. 
The changes of the Modal css works fine on this div- meaning the div is scrollable and there's a 48px margin. I still need to check whether the css changes of the MessageBox (both Marketerial and Functional ) are necessary or not.

PR #3133, opened 2019-03-21
`<PopoverMenu/>` - Remove default zIndex

hepiyellow:
default zIndex of `0` was introduces by this PR #2668

@nitayneeman 
Why did you add default zIndex: 0?
This zIndex goes to <Tooltip/>, which puts it on the <TooltipContent/>.
The tooltip has a default zIndex which should be the highest of all components (currently 2000).

Setting it to 0, means it could be (will be) under most other part/components of the page.


nitayneeman:
Added comment here: https://github.com/wix/wix-style-react/pull/2668#discussion_r267646841

hepiyellow:
@nitayneeman Commented:
>I agree with your point and I guess that part of the code shouldn't be merged (probably was just for checking a specific case momently). Moreover, the whole PR focused on improving the relevant docs, without changing the component implementation. Anyway, this code is redundant.

PR #3155, opened 2019-03-24
add onOptionFocused to DropdownLayout

morsh:
### 🔦 Summary
I have a requirement where I need to show more information regarding a hovered item on a dropdown list (like a small description / image / etc.) to do that, I need to be able to discern the currently active/hovered item.

The event I created is called whenever the `hovered` state is updated which will send the currently active option to the event. in case of `-1`, the event will be called with `null`.

not sure how to add documentation on this.

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [X] Component
  - [X] Visual test


hepiyellow:
@morsh
How do you intend to show this "more info", with a Tooltip/Popover?
I think that WSR-UX guideline discourages showing "double-popovers" (The dropdown-list is already a popover). We don't even allow simple ellipsis-tooltip for dropdown-list options.

@wuwa @milkyfruit @LittleTal Can you confirm?

src/DropdownLayout/DropdownLayout.js
```diff
@@ -82,20 +94,18 @@ class DropdownLayout extends WixComponent {
     if (!this._isControlled()) {
       newState.selectedId = chosenOption && chosenOption.id;
     }
-    this.setState(newState);
+    this._onOptionFocused(NOT_HOVERED_INDEX, newState);
```
hepiyellow: ### Controlled Mode

If we are in the controlled mode, it may be that the consumer is prompting the user, and maybe blocking the selection of the option. In that case we shouldn't reset the markedIndex.
This seems like an existing bug.
Which makes me think that `selectedId` and `markedId` (aka `hovered` index) are related, and maybe both should have Controlled mode.
What do you think?
Worth opening an issue about it?

src/DropdownLayout/DropdownLayout.spec.js
```diff
@@ -178,6 +178,50 @@ describe('DropdownLayout', () => {
       });
     });
 
+    describe('onOptionFocused', () => {
```
hepiyellow:
```suggestion
    describe('onOptionMarked', () => {
```

src/DropdownLayout/DropdownLayout.spec.js
```diff
@@ -178,6 +178,50 @@ describe('DropdownLayout', () => {
...
+        const onOptionFocused = jest.fn();
+        const driver = createDriver(
+          <DropdownLayout
+            visible
+            options={options}
+            onOptionFocused={onOptionFocused}
+          />,
+        );
+        driver.mouseEnterAtOption(1);
+        expect(onOptionFocused).toBeCalledWith(options[1]);
```
hepiyellow: Check that it is called only one time

src/DropdownLayout/DropdownLayout.spec.js
```diff
@@ -178,6 +178,50 @@ describe('DropdownLayout', () => {
       });
     });
 
+    describe('onOptionFocused', () => {
+      it('should focus an option some option', () => {
```
hepiyellow:
```suggestion
      it('should call onOptionMarked when an option is hovered', () => {
```

src/DropdownLayout/DropdownLayout.spec.js
```diff
@@ -178,6 +178,50 @@ describe('DropdownLayout', () => {
...
+            visible
+            options={options}
+            onOptionFocused={onOptionFocused}
+          />,
+        );
+        driver.mouseEnterAtOption(1);
+        expect(onOptionFocused).toBeCalledWith(options[1]);
+      });
+
+      it('should focus on null when no option is hovered', () => {
```
hepiyellow:
```suggestion
      it('should call onOptionMarked with null when mouse leaves a hovered option', () => {
```

src/DropdownLayout/DropdownLayout.spec.js
```diff
@@ -178,6 +178,50 @@ describe('DropdownLayout', () => {
...
+        const onOptionFocused = jest.fn();
+        const driver = createDriver(
+          <DropdownLayout
+            visible
+            options={options}
+            onOptionFocused={onOptionFocused}
+          />,
+        );
+        driver.mouseLeaveAtOption(1);
```
hepiyellow: might be nice to hover the option first

src/DropdownLayout/DropdownLayout.spec.js
```diff
@@ -178,6 +178,50 @@ describe('DropdownLayout', () => {
...
+        expect(onOptionFocused).toBeCalledWith(null);
+      });
+
+      it('should focus on option with down key', () => {
```
hepiyellow:
```suggestion
      it('should call onOptionMarkedfocus when down key is pressed', () => {
```

src/DropdownLayout/DropdownLayout.spec.js
```diff
@@ -178,6 +178,50 @@ describe('DropdownLayout', () => {
...
+            onOptionFocused={onOptionFocused}
+          />,
+        );
+        driver.pressDownKey();
+        driver.pressDownKey();
+        expect(onOptionFocused).toHaveBeenCalledTimes(2);
+        expect(onOptionFocused.mock.calls[0]).toEqual([options[0]]);
+        expect(onOptionFocused.mock.calls[1]).toEqual([options[1]]);
+      });
+    });
```
hepiyellow: ### Missing Tests
- [ ] onSelect
- [ ] onClose
- [ ] props change

milkyfruit:
@hepiyellow yes, overlapping tooltips/popovers isn't a good UX practice.
@morsh could you share what you're trying to achieve and share the name of your team's UX?

morsh:
@milkyfruit @hepiyellow - this is an idea we had with product (not UX) I'm not yet aligned with processed in Wix I might jumped ahead too quickly, but I wanted to have a demo to play with.

The idea:
1. The drop down would be divided into left & right sections
2. The left section would be the usual options
3. The right side would be constant section that would not scroll (achievable using fixedHeader + some styling changes to make it stick to the right instead of top)
4. Whenever hovering on an option, you'd be able to display rich information like: long description, image, read more links etc...)

The specifics of our usage:
1. Display search results
2. Search results sometimes require some rich information (like is this a premium feature, preview, where is this result located in the business manager, etc...)

Again, this is an idea we're playing with and not yet an approved UX requirement. It's main usage it to help conceive a richer experience - From the conversation here, I guess this has to go through UX to be approved (again sorry, but I'm new to the processed in Wix)

hepiyellow:
@morsh Yeah, definitely needs a UX review.
Please set a meeting with either: @wuwa or @milkyfruit .

- Do you have UX specs? User flow?
- How does the screen with your  Dropdown look?
(You don't have to answer here, just bring the answeres to the meeting)

hepiyellow:
@morsh Nevertheless, we can continue with this PR, since its a legitimate feature. If you have time to continue with it, even if you won't end up using it because of UX decisions.

morsh:
@hepiyellow - ok, I'll go over the comments this week and get this PR through

src/DropdownLayout/DropdownLayout.spec.js
morsh: What exactly is missing in each of them?

src/DropdownLayout/DropdownLayout.js
morsh: In general I think it would be nice to enable a controlled hovered state - in which case the developer can mark a specific option for the user without selecting it.

It's might be a weird scenario, but if you're adding a `selectedId` controlled mode, then `markedId` would be a nice way to expose functionality.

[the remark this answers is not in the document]
morsh: I don't really understand the comments here or their relation to the code.
If this is regarding adding this functionality to storybook - I think that using a fixed header with a changing color can be a nice way to display the functionality of onOptionMarked (Where fixed header is controlled)

hepiyellow:
@morsh 
You know that with your implementation, `onOptionMarked` will get called also whenever the user hovers with the mouse over each of the options.
Is that what you really want?

You might want to be able to separate mouse hovers from keyboard navigation....(?)

If so, then you can:
- Add a 2nd argument to `onOptionMarked` called `source`, which can have a value of either `hover | keyboard | initial`.

src/DropdownLayout/DropdownLayout.spec.js
hepiyellow: - `should call onOptionMarked with undefined when an option is selected`
- `should call onOptionMarked with undefined when the options are closed`
- `should call onOptionMarked with undefined when options change and the marked option is removed`

hepiyellow: I went ahead and added these tests

hepiyellow:
Ready for merge. waiting for build.

morsh:
@hepiyellow - 

> You know that with your implementation, onOptionMarked will get called also whenever the user hovers with the mouse over each of the options.
> Is that what you really want?

Yes, that was my intention - I currently don't have a been to differentiate between the event types, but if someone needs it, this can easily be added in the future.

Thanks for merging!

PR #3157, opened 2019-03-24
<RichTextInputArea/> - make texts to be configurable

nitayneeman:
This PR makes the texts that appear in `<RichTextInputArea>` configurable through a new prop, called `texts`.

Notes:

- Component tests for rendering tooltips (hover buttons, input placeholders) are missing. A separate PR will be opened with these tests (checking default & custom translations using the `texts` prop).
- The `buttons` array is still placed inside the `render` function. We can make it static partially, but it complicates the code a bit.

Related to issue:
wix-private/wsr-issues#35


src/RichTextInputArea/RichTextInputAreaTexts.js
```diff
@@ -0,0 +1,33 @@
...
+export const defaultTexts = {
+  toolbarButtons: {
+    boldButtonLabel: 'Bold',
+    italicButtonLabel: 'Italic',
+    underlineButtonLabel: 'Underline',
+    linkButtonLabel: 'Insert link',
+    bulletedListButtonLabel: 'Bulleted List',
+    numberedListButtonLabel: 'Numbered List',
+  },
+  insertionForm: {
```
nitayneeman: I couldn't think of a better name, but `insertionForm` actually represents the domain pretty well.

src/RichTextInputArea/RichTextInputAreaTexts.js
```diff
@@ -0,0 +1,33 @@
...
+    confirmButtonLabel: 'Confirm',
+    cancelButtonLabel: 'Cancel',
+    link: {
```
nitayneeman: Probably in the future we'll have an additional `urlInputPlaceholder` for an embedded image, so that sounds right to attach another level of nesting.

src/RichTextInputArea/RichTextInputAreaTexts.js
```diff
@@ -0,0 +1,33 @@
...
+      textInputPlaceholder: 'Text to display',
+      urlInputPlaceholder: 'URL this link should go',
+    },
+  },
+};
+
+export const textsPropType = PropTypes.shape({
```
nitayneeman: I'm wondering if we would prefer it explicit, or even more automated (mapping each string to `PropTypes.string` and object to `PropTypes.shape` recursively, for instance)?

PR #3162, opened 2019-03-25
Modal height

### LINKED ISSUE #2344: <Modal/> -  not scrollable to to if content is too high

TalZa1810:
### 🔦 Summary
Resolves #2344

I've created an example of a regular high div, making sure it fits the viewport and making it scrollable in case the div is too high.

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [x] Visual test

### DIFF
--- a/src/Modal/Modal.e2e.js
+++ b/src/Modal/Modal.e2e.js
@@ -1,5 +1,9 @@
 import { eyesItInstance } from '../../test/utils/eyes-it';
-import { storySettings, testStories, testPageDataHooks } from './test/storySettings';
+import {
+  storySettings,
+  testStories,
+  testPageDataHooks,
+} from './test/storySettings';
 import { createTestStoryUrl } from '../../test/utils/storybook-helpers';
 import {
   waitForVisibilityOf,
@@ -7,7 +11,7 @@ import {
 } from 'wix-ui-test-utils/protractor';
 
 //we want to take snapshots both at the beginning and end of the test
-const eyes = eyesItInstance({enableSnapshotAtBrowserGet: true});
+const eyes = eyesItInstance({ enableSnapshotAtBrowserGet: true });
 
 const { category, storyName } = storySettings;
 
@@ -28,10 +32,12 @@ describe('Modal', () => {
   });
 
   describe('content', () => {
-
-    beforeEach(async () => await browser.get(
-      testStoryUrl(testStories.modalHeaderCutsOffWithLargeContent),
-    ));
+    beforeEach(
+      async () =>
+        await browser.get(
+          testStoryUrl(testStories.modalHeaderCutsOffWithLargeContent),
+        ),
+    );
 
     const scrollableModalButton = element(
       by.css(`[data-hook="${testPageDataHooks.scrollableModalButton}"]`),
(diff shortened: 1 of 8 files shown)


TalZa1810:
@shlomitc 
When trying to scroll a non-scrollable div, it scrolls it a bit- less than 1 px difference of before & after scrolling. Each time I get a different Y coordinate:
1. 432.43963623046875 Vs. 431.7679748535156 ( 0.671661... difference)
2. 433.25592041015625 Vs. 432.4389953613281 ( 0.81692... difference ) 

Therefore, I think that it's better to check the fix correctness using the applitools (for both the first and second test). 

WDYT?

src/Modal/Modal.e2e.js
```diff
@@ -1,29 +1,87 @@
...
+} from './test/storySettings';
 import { createTestStoryUrl } from '../../test/utils/storybook-helpers';
-import { waitForVisibilityOf } from 'wix-ui-test-utils/protractor';
+import {
+  waitForVisibilityOf,
+  scrollToElement,
+} from 'wix-ui-test-utils/protractor';
+
+//we want to take snapshots both at the beginning and end of the tests
+const eyes = eyesItInstance({ enableSnapshotAtBrowserGet: true });
```
NitayRabi: I Would put this inside `describe` block

src/Modal/Modal.e2e.js
```diff
@@ -1,29 +1,87 @@
...
 
-const eyes = eyesItInstance();
 const { category, storyName } = storySettings;
 
+const getElementByDataHook = (dataHook) => element(by.css(`[data-hook="${dataHook}"]`));
```
NitayRabi: We don't have something to perform this already? (maybe in `'wix-ui-test-utils/protractor'`).
If not - please added to protractor utils so it would be available to other tests

src/Modal/Modal.e2e.js
```diff
@@ -1,29 +1,87 @@
...
+    const modalContentDiv =
+      getElementByDataHook(testPageDataHooks.modalContentDiv);
+
+    const footerDiv =
+      getElementByDataHook(testPageDataHooks.footerDiv);
+
+    const getDivBoundingClientRect = dataHook =>
+      `return document.querySelector("[data-hook='${dataHook}']").getBoundingClientRect();`;
+
+    const getElementCoordinatesByDataHook = dataHook =>
```
NitayRabi: What does this name mean? what's the purpose of this function? 
also - this should also be available for other tests

src/Modal/Modal.e2e.js
TalZa1810: Done

src/Modal/Modal.e2e.js
TalZa1810: This relevant only to the second test so I've added this params just for the second test:
https://github.com/wix/eyes.it

src/Modal/Modal.e2e.js
TalZa1810: "getBoundingClientRect" of a div with a specific dataHook.
I think this is relevant only for this specific test.

PR #3164, opened 2019-03-25
associate table row details with rowData.id

liorur:
associate table row details with rowData.id
this is useful when data is dynamic and rows are added after the initial table render

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x ] Component
  - [ ] Visual test


shlomitc:
Hey, thanks!
Can you please add an example to the storybook's `Table` page with the expanded rows?
It should look something like in this specs https://zpl.io/2vyOpZ5
This means using the `TableActionCell` component for the actions column

shlomitc:
Closing the PR for now as the solution is not required for the original use case and I'm not sure how well it will work with the virtualization. I'll add it as a task in the backlog and link this PR for whoever will implement it.

PR #3166, opened 2019-03-25
add props maxLength, autofocus to EditableTitle

adirou:
### 🔦 Summary
added autoFocus prop 
added maxLength prop

...

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


src/EditableTitle/EditableTitle.js
```diff
@@ -31,7 +37,7 @@ class EditableTitle extends React.Component {
     });
   };
   render() {
-    const { dataHook, className } = this.props;
+    const { dataHook, className, maxLength } = this.props;
```
shlomitc: add to proptypes and add missing test for both custom and default values

adirou: there are not tests in the input wsr for length , besides how do you want us to test it , the inputDriver doesn't have getProps method

adirou: i can test the behaviour of WSR input but it's doesn't make sense

shlomitc: so maybe it's a good time to do it. you can add to the input test and driver `getMaxLength: () => input.maxLength` and use it also in the EditableTitle.

shlomitc: also, the autoFocus should be tested as well

adirou: updated

PR #3172, opened 2019-03-25
Enable Tooltip testkit access appended content

splintor:
### 🔦 Summary
When using `<Tooltip>`'s `appendTo`, `appendToParent` or `appendByPredicate`, the Tooltip's content is not appended to `document.body` but to the specified parent.

When using *Enzyme*, the tested component that contains the `<Tooltip>` might be mounted using Enzyme's `mount()`, which doesn't append the component to `docuemnt.body` by default.
In such cases, `getContentRoot` will not find the Tooltip's content under `document.body` and it needs to look for it under the testkit's `wrapper`.

### DIFF
--- a/src/Tooltip/Tooltip.driver.js
+++ b/src/Tooltip/Tooltip.driver.js
@@ -1,6 +1,7 @@
 import ReactTestUtils from 'react-dom/test-utils';
 import eventually from 'wix-eventually';
 import last from 'lodash/last';
+import * as ReactDom from 'react-dom';
 
 const arrowDirectionToPlacement = {
   top: 'bottom',
@@ -9,7 +10,7 @@ const arrowDirectionToPlacement = {
   right: 'left',
 };
 
-const tooltipDriverFactory = ({ element }) => {
+const tooltipDriverFactory = ({ element, wrapper }) => {
   const getContentRoot = () => {
     const contentRootHook = element.getAttribute('data-content-hook');
     if (!contentRootHook) {
@@ -17,7 +18,22 @@ const tooltipDriverFactory = ({ element }) => {
         `Tooltip.driver: contentRootHook attribute must exist on the Toolrip's root element`,
       );
     }
-    return document.body.querySelector(`[data-hook="${contentRootHook}"]`);
+    const root = document.body.querySelector(
+      `[data-hook="${contentRootHook}"]`,
+    );
+    if (root) {
+      return root;
+    }
+
+    try {
+      const domInstance = ReactDom.findDOMNode(wrapper);
+      return (
+        domInstance &&
+        domInstance.querySelector(`[data-hook="${contentRootHook}"]`)
+      );
+    } catch (ignore) {
+      return null;
+    }
   };
 
   const getTooltipContent = () => getContentRoot().querySelector('.tooltip');
--- a/src/Tooltip/Tooltip.driver.js
+++ b/src/Tooltip/Tooltip.driver.js
@@ -18,18 +18,11 @@ const tooltipDriverFactory = ({ element, wrapper }) => {
         `Tooltip.driver: contentRootHook attribute must exist on the Tooltip's root element`,
       );
     }
-    const root = document.body.querySelector(
-      `[data-hook="${contentRootHook}"]`,
-    );
-    if (root) {
-      return root;
-    }
 
     try {
-      const domInstance = ReactDom.findDOMNode(wrapper);
       return (
-        domInstance &&
-        domInstance.querySelector(`[data-hook="${contentRootHook}"]`)
+        document.body.querySelector(`[data-hook="${contentRootHook}"]`) ||
+        wrapper.getDOMNode().querySelector(`[data-hook="${contentRootHook}"]`)
       );
     } catch (ignore) {
       return null;
--- a/src/Tooltip/Tooltip.driver.js
+++ b/src/Tooltip/Tooltip.driver.js
@@ -15,7 +15,7 @@ const tooltipDriverFactory = ({ element, wrapper }) => {
     const contentRootHook = element.getAttribute('data-content-hook');
     if (!contentRootHook) {
       throw new Error(
-        `Tooltip.driver: contentRootHook attribute must exist on the Toolrip's root element`,
+        `Tooltip.driver: contentRootHook attribute must exist on the Tooltip's root element`,
       );
     }
     const root = document.body.querySelector(


hepiyellow:
I remember a conversation with @shlomitc were we agreed that consumers must render to the body

splintor:
Can you explain why?
I just wasted a few hours trying to make the test in [this PR](https://github.com/wix-private/sites-list/pull/130) until I realized that the tooltip content cannot be found because the component is not attached to `document.body`.
With this PR, this can be avoided, as the testkit will find the content even if it is not attached to `document.body`.

src/Tooltip/Tooltip.driver.js
```diff
@@ -9,15 +10,30 @@ const arrowDirectionToPlacement = {
   right: 'left',
 };
 
-const tooltipDriverFactory = ({ element }) => {
+const tooltipDriverFactory = ({ element, wrapper }) => {
```
shlomitc: Did you try this locally in your project?
I think that `wrapper` is an internal property which isn't available for consumers of the testkit, so I don't think it will not work.

shlomitc:
@splintor - since the Tooltip component is appending itself to the body, it doesn't make sense to mount your component without attaching it to the document dom tree.

Adding the rendered node to the body is a best practice. see also `react-testing-library` doing it by default.
in the `enzyme` case, you should use `attachTo` option of mount to attach to the `document.body`

src/Tooltip/Tooltip.driver.js
splintor: I'm changing the testkit code itself, so it should be available. It is provided by [testkitFactoryCreator](https://github.com/wix/wix-ui/blob/master/packages/wix-ui-test-utils/src/vanilla/vanilla.tsx#L35) used to create the testkit.

I just tested it locally, and `wrapper` is available. I did notice that in my test, `ReactDom.findDOMNode` throws and I need to use `getDOMNode()` instead, but other than that, this change worked, and made the test pass even without attaching my component to the DOM.

splintor:
@shlomitc The `<Tooltip>` component doesn't always append itself to the body. When you use any of the `appendTo`, `appendToParent` or `appendByPredicate` props, the Tooltip is appended to the specified element, which can be inside the wrapper component. So there is no real need to attach the tested component to the body.

Of course, the code will still work if you attach the component to `documnet.body`, as it first tries to look for the tooltip under `document.body`, but if this fails, why not try looking for it under `wrapper`, to handle cases where it was appended to an element within the `wrapper` and the `wrapper` wasn't attached to `document.body`?

shlomitc:
if `Tooltip` is not appended to the body, then the testkit should just work. if it doesn't - can you please provide a reproducible example?

splintor:
@shlomitc 
https://codesandbox.io/s/o948ym4wwy
Currently, `npx yoshi test` fails for this code.
If you remove `appendToParent` in the `<Tooltip>` element in  `src/App/index.js`, the test passes.

It is actually very simple. When `appendTo`, `appendToParent` or `appendByPredicate` is set, the Tooltip is added in a node relative to the component, and not necessarily to `document.body`, but the testkit always looks for the content under `document.body`.

My fix is also simple - if it can't find it under `document.body`, try looking for it under the testkit wrapper.

mykas:
Hey @splintor , we might have solved this issue already in our next generation Tooltip. Look at this query for the content in our new Tooltips testkit:
Screenshot 2019-05-08 at 09 17 07

The method first searches near the element node itself and if not found - searches in the document.body. Can you try in your project to upgrade to the newest Tooltip and see if that works?

shlomitc:
@splintor - can you please give a try to the new tooltip and let us know?

splintor:
Not sure what you mean by "newest Tooltip". I couldn't find this code in github - can you put a link?
I try to upgrade `wix-style-react` to the latest version (`6.18.1`) and the problem persists.

However, when I looked in the history of [Tooltip.driver.js](https://github.com/wix/wix-style-react/blob/master/src/Tooltip/Tooltip.driver.js), I realized that it actually worked as I wanted until @hepiyellow's last PR (https://github.com/wix/wix-style-react/pull/2786). As he wrote in the PR description:
> If we assume that consumers ALWAYS render to the body. Then we can remove the `wrapper. If not, then we need to keep it.

As I wrote above, I don't really understand why you assume that tooltip content will be rendered to the body. In our initial code, it wasn't appended to the document body, and we appended it only because of this problem.

splintor:
I just realized I need to use `upgrade` to get the new `<Tooltip>`, but when I do, I get this errors, so I can't check:
```
component.spec.ts <SiteItem /> indication badge should display a disabled "contributors list" badge when the user has a restricted role [78 ms]

	requestAnimationFrame is not defined
		at Driver.<anonymous> src/client/pages/app/site-list/components/grid-items/site-item/component.driver.tsx:59:7
		at Object.contributorsBadgeTooltipIsShown src/client/pages/app/site-list/components/grid-items/site-item/component.driver.tsx:57
		at Object.<anonymous> src/client/pages/app/site-list/components/grid-items/site-item/component.spec.ts:45:7
		at Context.<anonymous> src/client/pages/app/site-list/components/grid-items/site-item/component.spec.ts:38:5

	From console:
		The above error occurred in the <Popover> component:
		    in Popover (created by Tooltip)
		    in Tooltip (created by Tooltip)
		    in Tooltip (created by Tooltip)
		    in Tooltip (created by class_1)
		    in div (created by class_1)
		    in div (created by class_1)
		    in class_1
		    in Unknown (created by WrapperComponent)
		    in WrapperComponent
		
		Consider adding an error boundary to your tree to customize error handling behavior.
		Visit https://fb.me/react-error-boundaries to learn more about error boundaries. 


		The above error occurred in the <InnerPopper> component:
		    in InnerPopper (created by Context.Consumer)
		    in Popper (created by Popover)
		    in Transition (created by CSSTransition)
		    in CSSTransition (created by Popover)
		    in div (created by Popover)
		    in class_1 (created by OnClickOutside(class_1))
		    in OnClickOutside(class_1) (created by Popover)
		    in Manager (created by Popover)
		    in Popover (created by Tooltip)
		    in Tooltip (created by Tooltip)
		    in Tooltip (created by Tooltip)
		    in Tooltip (created by class_1)
		    in div (created by class_1)
		    in div (created by class_1)
		    in class_1
		    in Unknown (created by WrapperComponent)
		    in WrapperComponent
		
		Consider adding an error boundary to your tree to customize error handling behavior.
		Visit https://fb.me/react-error-boundaries to learn more about error boundaries. 
```

mykas:
@splintor  requestAnimationFrame is not part of Tooltip. Can you check in your spec if you are importing this polyfill in your code?

mykas:
@splintor Also you have to import different TooltipTestkit. Please check the documentation part.

splintor:
I changed to use `TooltipTestkit` constructor, and now I get `Unable to find node on an unmounted component.` when I call it.
I don't have more time to check this.
If you think that the new version of Tooltip handles the problem I'll close the PR.

PR #3176, opened 2019-03-26
<EditableTitle/> minor style fix

Jefoss:
very small fix for the case where the value is long and theres an unwanted horizontal scrollbar


hepiyellow:
@Jefoss So what happens when the content is too long? (ellipsis + tooltip)

Are we talking about the text content (label when it is not in edit mode)?
Or on the content of the `<input/>` when editting?

PR #3184, opened 2019-03-26
add virtualization to <DataTable>

liorur:
Add Virtualizton to DataTable
This feature should allow displaying a scrollable table with a very large number of lines without overloading the browser

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [ ] Visual test


hepiyellow:
As discussed with @liorur 

### Why Table ? Why not a List?
The Use-case of wix-code (which drives this PR) is for a list of log items. Each log item shows a row with only the title, and can be expanded.
 This seems like a List use-case not a Table use-case.

> The solution in this PR uses `react-window` which is not originally meant for `<table/>`. And needs some hacks in order to implement it for a table (changing cell `display` to `block`.)

### UX Scroll
- WSR-UX guidelines suggest that there will be only 1 scrollable container in a Page.
- The Page already implements that Scrollable Container.
  - Maybe the large scroll optimization should be implemented in Page...?
  - If we decide that Page can include an inner ScrollableContainer, then we need to make sure the Page's ScrollableContainer is not scrollable (Stretch Content), and delegate wheel events to the consumer's inner ScrollableContainer.

shlomitc:
1. React-window Because it’s a straightforward solution with minimal changes to the existing component structure and in overall it’s not a bad solution.
2. Not a list because it needs the column structure. Maybe it’s possible to get the same affect without a table, but if it already works - So let’s not over complicate it.
3. Ux scroll - you can try to think of a better solution but for now that’s the quick win

src/DataTable/DataTable.js
```diff
@@ -267,13 +281,55 @@ class DataTable extends React.Component {
...
+      return (
+        <tr style={style} className={this.style.headerSpaceHolderRow}>
+          <td />
+        </tr>
+      );
+    }
+    return this.renderRow(data[index - 1], index - 1, style)[0];
+  };
+
+  getVirtualRowHeight = index => (index ? 60 : 45);
```
shlomitc: extract to variables and name them as `headerRowHeight` and `rowHeigh` to make condition clear. also worth checking if index === 0, just to be on the safe side

src/DataTable/DataTable.js
```diff
@@ -267,13 +281,55 @@ class DataTable extends React.Component {
...
+    );
+  };
+
+  renderVirtualizedTable = () => {
+    const { dataHook, data } = this.props;
+    return (
+      <div data-hook={dataHook}>
+        <List
+          className={classNames(this.style.table, this.style.virtualized)}
+          height={300}
```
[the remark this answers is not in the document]
liorur: or expose a prop

src/DataTable/DataTable.js
liorur: indeed, this was for the POC

liorur:
need to fix problem with space after the last row

shlomitc:
image

liorur:
@shlomitc fixed everything except for AutoSizer, for now i exposed virtualizedTableHeight prop instead

ohana54:
@liorur @shlomitc thanks for all the work invested in this!
Can you please update on the status of the PR? What's missing to get it merged? We'd like to start using this :)

shlomitc:
@ohana54 @liorur indeed great work. will be released in the upcoming version

ohana54:
@shlomitc sounds great! Where can I follow on the release timelines?

shlomitc:
@ohana54 - release is on its way https://github.com/wix/wix-style-react/pull/3202
+-1 hour after merging

ohana54:
@shlomitc awesome, thanks!!

PR #3185, opened 2019-03-26
chore(Accordion): refactor

argshook:
finishing up Accordion PR

still WIP


boriskor:
Good reference for a11y behaviors & aria roles: https://www.w3.org/TR/wai-aria-practices-1.1/#accordion

PR #3189, opened 2019-03-26
Unidrivers - Refactor : Add enhance-unidriver, ReactAdapter with isFocused.

hepiyellow:
### Why ReactBase approach is bad

Well, ReactBase was a hack that takes into account only the React-JSDOM test platform. 
Some methods were implemented in ReactBase that should be in all adapters (as WSR team see it, but Unidriver team disagree).

Once we'll want to implement those methods for all platforms, we'll need to add more platform dependent imports to ReactBase (And rename it to EnhancedBase). This will create a mix-up were tests running in one platform, would import stuff that belongs to another platform).

This refactor changes the strategy of "solving" Unidriver issues:
- from : using ReactBase as a react-only solution inside a component's Unidriver implementation
- to: Using a proper React Unidriver Adapter which enhances the original UniDriver.

This solution is good for methods like `isFocused` which CAN be implemented in all platforms, but UniDriver team is currently refusing to do so.

In this PR I am not implementing all platform adapters.

### Unidriver Migration Scope

We decided that the first-step migration would deal only with React-JSDOM platform.
But we are also writing new component with unidriver, and we need (want) methods that don't exist in Unidriver (and won't be added there).

So we have no choice but to extend/enhance the Unidriver adapters, and create an abstraction over Unidriver that extends the Unidriver API.

@shlomitc your opinion?
@GabiGrin FYI

### DIFF
--- a/scripts/generate-testkit-exports/templates/enzyme.js
+++ b/scripts/generate-testkit-exports/templates/enzyme.js
@@ -2,4 +2,4 @@
 import {
   enzymeTestkitFactoryCreator,
   enzymeUniTestkitFactoryCreator,
-} from 'wix-ui-test-utils/enzyme';
+} from './creators/enzyme';
--- a/scripts/generate-testkit-exports/templates/vanilla.js
+++ b/scripts/generate-testkit-exports/templates/vanilla.js
@@ -2,4 +2,4 @@
 import {
   testkitFactoryCreator,
   uniTestkitFactoryCreator,
-} from 'wix-ui-test-utils/vanilla';
+} from './creators/react-jsdom';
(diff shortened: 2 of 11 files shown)


GabiGrin:
@hepiyellow I think isFocused can match the library.  I think we said to leave it because of jsdom limitations but from a quick check the "activeElement" works well there.

It's the "focus()" action I don't believe fits.

@borisd9  wdyt?

mykas:
@hepiyellow If what @GabiGrin says is true that they can support `isFocused()` then the only missing feature is `focus()` because of JSDOM quirks, right?

borisd9:
@mykas  is right, the discussion is about `focus` method.
you can focus by clicking on something, just like a real user. 

If you have this, and have `isFocused` then you should be set IMO

Issue #3158, opened 2019-03-24
FloatingHelper Content testkit is not exposed


eyaleizenberg:
# 🐛 Bug Report
Missing testkit for FloatingHelper.Content

### 🏗 Relevant Components
FloatingHelper.Content

...

### 😯 Current Behavior
According to the docs, the testkit for the FloatingHelper.Content is accessed by `floatingHelper.ContentTestkitFactory`, however, that constant is not exposed.

...

### 🤔 Expected Behavior
Properly expose the FloatingHelper.Content testkit.

...

### 👣 Steps to Reproduce
The following code will fail:
import {floatingHelper.ContentTestkitFactory as enzymeFloatingHelper.ContentTestkitFactory} from 'wix-style-react/dist/testkit/enzyme';

  - **WSR version**: `6.10.2`
  - **Yoshi version**: `...`

### 👀 Severity

- Critical / Major / Low

gilmeir:
Hey @eyaleizenberg,
Will floatingHelperDriverFactory.getHelperContentDriver() work for you? [see here](https://github.com/wix/wix-ui-backoffice/blob/master/src/components/FloatingHelper/FloatingHelper.driver.ts#L32).

eyaleizenberg:
Hi @gilmeir 
I am not using this after all...

Thanks anyway

PR #3194, opened 2019-03-27
docs(TEST_DRIVERS & TEST_DRIVERS_GUIDELINES): cleanup, update, redact

argshook:
1. redact to lists, less words where possible
1. mentioning how drivers ideally should not be nested
1. other small fixes, some spaces in code examples


docs/contribution/TEST_DRIVERS_GUIDELINES.md
```diff
@@ -1,48 +1,53 @@
...
 
-### Example 1
+## ProTips
 
-An `<Search/>` input might have clear button (little `x`).
+1. Focus driver methods on intent, not granular actions (see examples below)
+1.  **Test the DOM**  
+    Every React prop set by consumer affects the DOM, or behavior of the component (which in turn affects the DOM). Don't assert on React props or inner state, instead assert on the DOM.
+1. Keep drivers flat. The returned object should be flat and simple.
+   Avoid nesting objects or worse, other drivers (it is easy to do but creates implicit dependency which is very hard to maintain)
```
hepiyellow: @argshook 
Can you give an example? (of an implicit dependency which is hard to maintain)

@shlomitc FYI

argshook: FloatingHelper.driver has `getHelperContentDriver`, which has
`getActionButtonDriver`.

FloatingHelperContent driver is not exposed like other drivers (it's non
importable), but instead is part of other driver.

There is a bug with action button in FloatingHelperContent.

Bug fix is to use ButtonNext from wix-ui-core instead of deprecated Button in wix-ui-backoffice.

ButtonNext is unidriver which is async, which would be breaking change
in driver.

From here a plethora of questions open up of how to solve, how not to
break and such. Without the root cause (nested drivers) none of the
questions would arise.

What happened IRL was that i:
* fixed bug by using ButtonNext
* changed FloatingHelperContent driver from `getActionButtonDriver().textContent()` to `getActionButtonText`
* did this under assumption that FloatingHelperContent is not exposed to users (there's even issue open about it, from a confused user of ours https://github.com/wix/wix-style-react/issues/3158)
* oh wait no, it's actually exposed through `FloatingHelper.getHelperContentDriver` - a nested driver.
* i got lucky, because only 3 users use this method: one asked about it, one fixed
my change themselves, one i fixed myself.

let's say there was no bug with button in FloatingHelperContent and we
simply wanted to migrate deprecated `wix-ui-backoffice/Button` to
`wix-ui-core/ButtonNext`. We change the button and its driver. All is
good until we realize that changed driver impacts FloatingHelperContent
and FloatingHelper and in-turn end users.

---

this wall of text here illustrates how nesting drivers is not a good
idea. and it's just one case.

'tis a great moment to share python mantra again https://www.python.org/dev/peps/pep-0020/#id3

docs/contribution/TEST_DRIVERS.md
```diff
@@ -1,49 +1,54 @@
...
+- start by implementing only required functions (and not the entire driver).
+- open issue stating that the new `ConsumedComponent.uni.driver.js` is only partially implemented.
 
 ## Public And Private Drivers
 
-1. The **Public** drivers (`Component.uni.driver.js`) are the ones that exposed to the consumers of the components. They should be simple abstractions over common actions (for example, selecting the third element in the dropdown).
-2. The **Private** drivers (`Component.private.uni.driver.js`) are used for actions on a component that should not be exposed to the user. For example, asserting a class name existance on some component.
-3. The Private drivers are extending the public ones and should be used internally when testing the components.
+1. **Public** drivers (`Component.uni.driver.js`) are exposed to user of components. They should be simple abstractions over common actions (for example, selecting the third element in the dropdown).
+2. **Private** drivers (`Component.private.uni.driver.js`) are used for actions that are not be exposed to user. For example, asserting a class name exists on some component.
```
argshook:
```suggestion
2. **Private** drivers (`Component.private.uni.driver.js`) are used for actions that are not exposed to user. For example, asserting a class name exists on some component.
```

docs/contribution/TEST_DRIVERS.md
```diff
@@ -1,49 +1,54 @@
...
+- Make tests independent from implementation.
+- Abstract complex interactions.
 
 ## Driver Types
 
-### UniDriver (for every new component)
+### Unidriver
 
-All component drivers in this library should be built on top of [unidriver](https://github.com/wix-incubator/unidriver). `unidriver` is a tool to write universal component drivers that can be reused in all test platforms (e.g. ReactDOM/JSDOM, Protractor, Puppeteer, etc...).
+All component drivers in this library should be built with [unidriver](https://github.com/wix-incubator/unidriver). This tools allows to reuse same driver on multiple testing platforms (e.g. ReactDOM/JSDOM, Protractor, Puppeteer, etc...).
```
argshook:
```suggestion
All component drivers in this library should be built with [unidriver](https://github.com/wix-incubator/unidriver). This tool allows to reuse same driver on multiple testing platforms (e.g. ReactDOM/JSDOM, Protractor, Puppeteer, etc...).
```

docs/contribution/TEST_DRIVERS.md
```diff
@@ -1,49 +1,54 @@
...
 
-Public drivers are also published code that needs to be tested.
+Public drivers are published code that needs to be tested.
 We test them simply by using them in our component tests.
 
-### Just a small but very common Gatchas
+### Small But Common Gotchas
 
-If you write a boolean driver method (e.g. `isDisabled()`), remember to test it's related feature both ways!
+If you write a boolean driver method (e.g. `isDisabled()`), remember to test its both ways (when it's `true` and also `false`).
```
argshook:
```suggestion
If you write a boolean driver method (e.g. `isDisabled()`), remember to test it both ways (when it's `true` and also `false`).
```

docs/contribution/TEST_DRIVERS_GUIDELINES.md
```diff
@@ -1,48 +1,53 @@
...
+`<Dropdown/>` allows selecting from a list of options.
 
-## Test the DOM
+User needs to:
+- click on Input (to focus it)
+- press key down (to show list of options)
+- click on certain option
 
-Every React prop set by the consumer has an affect on the DOM, or the behavior of the component (which affects the DOM).
+Instead of providing granular methods for each action, instead abstract them to reflect the intent:
```
argshook:
```suggestion
Instead of providing granular methods for each action abstract them to reflect the intent:
```

docs/contribution/TEST_DRIVERS_GUIDELINES.md
```diff
@@ -52,61 +57,50 @@ For example:
...
-
-```js
-<div>
-  <button>Confirm</button>
-</div>
-```
+This allows testing applications in production.
 
-In this simple case you can locate the button with a `tagName` selector.
+> The name `data-hook` is a Wix naming-convension.
```
argshook:
```suggestion
> The name `data-hook` is a Wix naming convention.
```

PR #3215, opened 2019-04-01
Eyes storybook improvements

shlomitc:


.storybook-test/stories.scss
```diff
@@ -4,7 +4,3 @@ html {
   padding: 0;
   margin: 0;
 }
-
-:global {
-  * { caret-color: transparent; }
```
argshook: i think this can stay, if we ever get to have a test with input
component that has autofocus or something, we'll be safe

otherwise this line it's harmless

shlomitc: since eyes-storybook doesn't take any screenshots - it doesn't have any effect

argshook: eventually it does and i wonder if focus is included in snapshot data. either way, it doesn't really matter. if we'll have an issue, we'll fix the issue :)

PR #3222, opened 2019-04-01
<DataTable/> - style attribute gets to render array

mykas:
This PR fixes an issue introduced in DataTable, where style attribute gets map() method third argument.


NitayRabi:
I think @liorur Added the style for a reason - we might want to pass it in case of virtualisation

mykas:
True, thats why I'll show him this PR.

mykas:
@liorur any comments on this?

mykas:
Found a place where he needs to use it: https://github.com/wix/wix-style-react/commit/8a821c5bf2cbf6eb979efe4edfc51de1773206e1#diff-71ae30a4d5d2e610065bc2cb5ad56b68R293

mykas:
I'll add a proper map usage then :)

PR #3227, opened 2019-04-02
onSubmit is called twice when pressing enter in EditableTitle

aryelu:
### 🔦 Summary
...

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


src/EditableTitle/EditableTitle.js
```diff
@@ -95,13 +95,17 @@ class EditableTitle extends React.Component {
...
-            onEnterPressed={this.onValueSubmission}
+            onEnterPressed={this.onEnterPressed}
           />
         </div>
       </div>
     );
   }
 
+  onEnterPressed = () => {
+    this.wsrInput.blur();
```
mykas: So this make input blurred but the state of the component is still gonna be `focus: true`. Is this the use case?

shlomitc:
@aryelu - I think it's time to add a test for the `enter` key press behavior, to prevent this from recurring

src/EditableTitle/EditableTitle.js
aryelu: the blur will be handled by onBlur

src/Input/Input.uni.driver.js
```diff
@@ -75,7 +75,7 @@ export const testkit = base => {
     focus: async () => await reactBaseInput.focus(),
     blur: async () => await reactBaseInput.blur(),
     keyUp: async () => await reactBaseInput.keyup(),
-    keyDown: async () => await reactBaseInput.keydown(),
+    keyDown: async key => await reactBaseInput.keydown(key),
```
shlomitc: please name `key` as `eventData`

aryelu: 👍

Issue #3209, opened 2019-03-31
`<Page/>` - Add Animations


hepiyellow:
### Mini Header Animation spec

- [x] grey background appears instantly, that's ok now.
- [ ]Foreground (text and such) as above (with overflow hidden on the background container)
  - 0.2s transition when appearing...
    - opacity: 0->100
    - top: -X -> 0 (Goes up)

shlomitc:
won't be fixed soon.closing

PR #3230, opened 2019-04-02
animate the entrance of the pageHeader

mosheKerbel:
fix issue - https://github.com/wix/wix-style-react/issues/3209


mosheKerbel:
@wuwa 
this PR handles this issue https://github.com/wix/wix-style-react/issues/3209
now when you scroll and the page header gets "sticky", it enters with an effect of
opacity + slide in from bottom.
i put the animation to take 0.2s.
could you take a look here and say if it looks good enough?
https://wix-wix-style-react-pr-3230.surge.sh/?selectedKind=2.%20Layout&selectedStory=2.1%20Page%20Layout&full=0&addons=0&stories=1&panelRight=0

wuwa:
can you please make y movement smaller distance - from y+20 --> y

mosheKerbel:
@wuwa the animation was on the x axis, so i changed it from 50 to 20...
did i got you right?

wuwa:
yes, looks good. it is y axis (top property)

mosheKerbel:
we are good to go from Ben's side.
@hepiyellow

hepiyellow:
@wuwa 
BTW: I had an idea, that we can make the `actionsBar` sticky, so it won't be so "jumpy", it will simply stick.

### actionsBar needs to be fixed.

Today we pass `minimized` argument to the `actuiosBar`, so consumers can render it differently when minimized.

I we remove the `minimized` argument, then the actionsBar will be the same in both cases, and then sticky will work just fine.

PR #3235, opened 2019-04-03
<Page/> - add deprecation log when not using "upgrade" flag

mosheKerbel:
@hepiyellow


hepiyellow:
Don't merge yet.
There is a new issue with Page and Safari.
The Overlay header is getting pushed by some content when scrolling.

PR #3237, opened 2019-04-04
`<Tooltip/>` - Add appendTo default predicate for Modal

### LINKED ISSUE #3226: Tooltip in Modal is not shown
# 🐛 Bug Report

### 🏗 Relevant Components

Tooltip, Modal

### 😯 Current Behavior

When rendering `<Tooltip/>` in `<Modal/>`, it is not shown.
The reason is that the Tooltip has a lower zIndex than the Modal, and the Tooltip is rendered onto the body by default.

### 🤔 Expected Behavior

Should be visible.

#### Set attachTo 
Use `<Tooltip appendToParent/>`
OR
Use `appendTo` and give it the DOM element of the Modal.

#### Set zIndex

```
<Tooltip zIndex=5010/> // Modal is 5000
```

### Solution 1 (quick fix)

Add a default predicate, that will attach the Tooltip to the Modal.
https://github.com/wix/wix-style-react/blob/e438dcc51626c6e39dd1811aa529db2f24033fda/src/Tooltip/TooltipContainerStrategy.js#L3

### Solution 2 (Cool Infrastructure)

Add an AppendToProvider (React Context Provider), then provides the "base" DOM element that will be used as a default for the `appendTo` prop of both `<Tooltip/>` (legacy), `<Popover/>` (the new guy)

### 👀 Severity

- Major

hepiyellow:
Resolves #3226 

THis is a quick fix.

I think the predicate solution of Tooltip is not ideal.?

### Solution Using Context Provider

- I suggest a solution where we'll have a `<AppendToProvider/>` 
- The Modal will be such a provider and will provide a `defaultAppendToElement`
- `<Tooltip/>` will be a consumer, and will use the `defaultAppendToElement` instead of the predicate mechanism.

#### Pros

- This method is better since it puts the responsibility of setting the appendTo "base" NOT on the tooltip, but on the context in which the Tooltip is rendered in. So that we don't need to add predicates on ALL Tooltips under a Modal (In this case).

### DIFF
--- a/src/Tooltip/TooltipContainerStrategy.js
+++ b/src/Tooltip/TooltipContainerStrategy.js
@@ -1,6 +1,7 @@
 export class TooltipContainerStrategy {
   constructor(appendTo, appendToParent, appendByPredicate) {
     this._predicates = [
+      element => element.classList.contains('ReactModal__Overlay'),
       element =>
         element.getAttribute('data-class') === 'page-scrollable-content',
       element => element === document.body,
--- a/src/Tooltip/TooltipContainerStrategy.spec.js
+++ b/src/Tooltip/TooltipContainerStrategy.spec.js
@@ -2,6 +2,7 @@ import React from 'react';
 import { TooltipContainerStrategy } from './TooltipContainerStrategy';
 import { mount } from 'enzyme';
 import Page from '../Page';
+import Modal from '../Modal';
 
 describe('TooltipContainerStrategy', () => {
   it('should return body when element is null', () => {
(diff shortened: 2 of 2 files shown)


shlomitc:
Isn’t it a breaking change?

mykas:
I don't think this is problem anymore. The new Tooltip will have `appendTo = "parent"` by default.  This is just a matter of understanding for users and adding some kind of extra provider just makes things even more confusing.

hepiyellow:
Talking with Shlomi, 
we decided to take the assumption that there could not be 2 Tooltips/Popovers opened at the same time (including popover's from BM's header).

And thus, we can simply make the zIndex of the Tooltip higher than the Modal's.

PR #3248, opened 2019-04-08
add disable prop to contact-item builder

malachico:
### 🔦 Summary
...

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test

### DIFF
--- a/src/ContactItemBuilder/ContactItemBuilder.js
+++ b/src/ContactItemBuilder/ContactItemBuilder.js
@@ -46,10 +46,18 @@ ContactItem.propTypes = {
   title: PropTypes.string.isRequired,
   subtitle: PropTypes.string,
   imageUrl: PropTypes.string,
+  disabled: PropTypes.bool,
 };
 
-export const contactItemBuilder = ({ id, title, subtitle, imageUrl }) => ({
+export const contactItemBuilder = ({
   id,
+  title,
+  subtitle,
+  imageUrl,
+  disabled,
+}) => ({
+  id,
+  disabled,
   value: ({ selected }) => (
     <ContactItem
       title={title}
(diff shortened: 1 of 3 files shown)


shlomitc:
please add also an example in the docs so it can be reviewed visually

src/ContactItemBuilder/ContactItemBuilder.spec.js
```diff
@@ -26,5 +26,17 @@ describe('item picker option builder', () => {
     expect(driver.getSubtitle()).toEqual(subtitle);
   });
 
+  it('should return item with disabled prop', () => {
+    const component = (
+      <ContactItem title={title} subtitle={subtitle} disabled />
+    );
+    expect(component.props.disabled).toBeTruthy();
```
hepiyellow: This test is only checking that ContactItem is a JSX item. And prop you pass that way will be available on `component.prop`.

You can calling the method `contactItemBuilder()`?
and expect the result to be an object with the `disabled` property.

src/ContactItemBuilder/ContactItemBuilder.spec.js
```diff
@@ -26,5 +26,17 @@ describe('item picker option builder', () => {
...
+  });
+
+  it('should return item without disabled prop when prop is not passed', () => {
+    const contactItem = <ContactItem title={title} subtitle={subtitle} />;
```
hepiyellow: Same here. You should call `contactItemBuilder()`

src/ContactItemBuilder/ContactItemBuilder.js
```diff
@@ -46,10 +46,18 @@ ContactItem.propTypes = {
   title: PropTypes.string.isRequired,
   subtitle: PropTypes.string,
   imageUrl: PropTypes.string,
+  disabled: PropTypes.bool,
```
hepiyellow: You added `disabled` prop to `<ContactItem/>` but you didn't do anything with it.
Please remove it.

Issue #3122, opened 2019-03-18
<Search/>: error when clearing search input 


ipanasenko:
Reproducible in Storybook: https://wix-wix-style-react.surge.sh/?selectedKind=3.%20Inputs&selectedStory=3.9%20Search&full=0&addons=0&stories=1&panelRight=0&activeTab=Usage

1. Enter text in search input
2. Click X icon
3. See an error:
![image](https://user-images.githubusercontent.com/576152/54548354-63444100-49b0-11e9-86f2-79ca01e69068.png)

I did small investigation and found this https://github.com/wix/wix-style-react/blob/master/src/Input/Input.js#L354 and this https://github.com/wix/wix-style-react/blob/master/src/Input/Input.js#L364

Can someone have a look and try to fix it? We see a lot of these errors in our Sentry: https://sentry.io/organizations/wix_o/issues/936631846/?project=1207945&query=is%3Aunresolved&statsPeriod=14d&utc=false

ipanasenko:
BTW, I only see this error in production react build, somehow it's not thrown in development build

gilmeir:
@ipanasenko can you please try the newest wix-style-react version?

Since the last version you can provide the prop `updateControlledOnClear`.
It's a flag that fixes this error.
It's a breaking change, so this behaviour will become the default only on next major version.

The breaking part is the clear button no longer clear the interval value on controlled input/search, it's up to you to do that in onClear.

Please update if it works for you.

ipanasenko:
Works! Thanks!

PR #3251, opened 2019-04-08
don't reset input value on uncontrolled input (#3122)

gilmeir:
### 🔦 Summary

Fixing: https://github.com/wix/wix-style-react/issues/3122

This removes a hack in the code, that causes an error when a controlled input clear button is clicked.

As part of getting rid of the hack, we aim to separate functionality between controlled / uncontrolled input:
1. No longer reset a controlled Input value when clear button is clicked. User should manage/reset value in the onClear.
2. onChange will no longer be triggered when clicking the clear button, only onClear will be called providing the clear event.

The older functionality is enabled by default for backward compatibility.
To opt in set the prop breakingChangeControlledInputOnClear={true}

### ✅ Checklist - will be updated later
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


src/Input/Input.spec.js
```diff
@@ -583,32 +578,26 @@ describe('Input', () => {
...
         expect(await driver.hasClearButton()).toBe(true);
       });
 
       it('should not be displayed when input text is empty', async () => {
+        // controlled
         const { driver } = render(<Input value="" clearButton />);
         expect(await driver.hasClearButton()).toBe(false);
       });
 
-      it('should clear input and focus it', async () => {
```
shlomitc: I don't think you should delete the previous behavior tests.
they should be here so we know the old behavior works.
only new tests should be applied using the new behavior

src/Input/Input.spec.js
```diff
@@ -583,32 +578,26 @@ describe('Input', () => {
...
-      });
-
-      it('should trigger onChange on clearing as if input just emptied', async () => {
-        const onChange = jest.fn();
+      it('should trigger onClear when clicking the clear button', async () => {
+        // controlled
+        const onClear = jest.fn();
         const { driver } = render(
-          <Input onChange={onChange} value="some value" clearButton />,
+          <Input onClear={onClear} value="some value" />,
```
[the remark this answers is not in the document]
gilmeir: Both new and old logics call onClear.
We're no longer expecting `onChange` to be called only on the new flow.
I'm adding back the old flow tests``

src/Input/Input.spec.js
gilmeir: Yep

src/MultiSelect/InputWithTags.spec.js
```diff
@@ -19,20 +19,48 @@ describe('InputWithTags', () => {
...
 
-    expect(onChange).toHaveBeenCalledTimes(1);
-    expect(onChange.mock.calls[0][0].target.value).toBe('foo');
+      driver.inputDriver().enterText('foo');
+      expect(driver.inputDriver().getValue()).toEqual('foo');
 
-    component.clear();
+      component.clear();
+      expect(driver.inputDriver().getValue()).toEqual('');
+    });
```
gilmeir: take back old code

src/MultiSelect/InputWithTags.spec.js
```diff
@@ -19,20 +19,48 @@ describe('InputWithTags', () => {
...
+      );
+
+      driver.inputDriver().enterText('foo');
+      expect(onClear).toHaveBeenCalledTimes(0);
+
+      component.clear();
+      expect(onClear).toHaveBeenCalledTimes(1);
+    });
+
+    it('should call onChange', () => {
```
gilmeir: irrelevant

src/ModalSelectorLayout/ModalSelectorLayout.js
```diff
@@ -293,13 +294,16 @@ export default class ModalSelectorLayout extends WixComponent {
...
-      searchValue: e.target.value,
+      searchValue,
       isSearching: true,
       items: [],
     });
-  }
+
+  _onSearchChange = e => this._updateSearchValue(e.target.value);
+
+  _onClear = () => this._updateSearchValue('');
```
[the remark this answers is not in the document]
gilmeir: Yes, it's equivalent to when the user triggers `onChange`: we are stating with isSearching that a new search was triggered, for rendering the loader/content

PR #3277, opened 2019-04-21
stop using autoSizeInput

mosheKerbel:
MultiSelect is using InputWithTags
Because of InputWithTags Flexi behavior of gorwing the items, we can simply use width and min-width of 1px on an input field, and it would grow and shrink depending on it's content, without breaking a the line.


hepiyellow:
Does this work on all browsers?

mosheKerbel:
Tested in Chrome, FF and Safari and works well in all.

PR #3281, opened 2019-04-21
<RichTextInputArea/> - support error design

nitayneeman:
This PR adds new props, `status` and `statusMessage`, for rendering `<RichTextInputArea/>` with error design and indicator:
image

Note: Changing toolbar's border to blue when the editor is focused in error mode, is not covered in this PR.

Related to issue:
wix-private/wsr-issues#35


src/Exclamation/Exclamation.uni.driver.js
```diff
@@ -0,0 +1,7 @@
+import { baseUniDriverFactory } from 'wix-ui-test-utils/base-driver';
+
+export const exclamationDriverFactory = base => {
```
nitayneeman: The CI failed because there was no driver file for this component:
```
    Cannot find module '../src/Exclamation/Exclamation.driver' from 'index.js'

      78 | export const editableTitleTestkitFactory = uniTestkitFactoryCreator(load(require('../src/EditableTitle/EditableTitle.uni.driver')));
      79 | export const emptyStateTestkitFactory = testkitFactoryCreator(load(require('../src/EmptyState/EmptyState.driver')));
    > 80 | export const exclamationTestkitFactory = testkitFactoryCreator(load(require('../src/Exclamation/Exclamation.driver')));
         |                                                                     ^
      81 | export const filePickerTestkitFactory = testkitFactoryCreator(load(require('../src/FilePicker/FilePicker.driver')));
      82 | export const floatingHelperTestkitFactory = testkitFactoryCreator(load(require('../src/FloatingHelper/FloatingHelper.driver')));
      83 | export const floatingNotificationTestkitFactory = uniTestkitFactoryCreator(load(require('../src/FloatingNotification/FloatingNotification.uni.driver')));

      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:229:17)
      at Object.require (testkit/index.js:80:69)
```

That's why an empty driver file is created.

src/RichTextInputArea/RichTextInputArea.spec.js
```diff
@@ -76,6 +76,20 @@ describe('RichTextInputArea', () => {
...
+      expect(await driver.hasError()).toBe(true);
+    });
+
+    it('should not render the error indicator when disabled', async () => {
+      const driver = createDriver(<RichTextInputArea disabled error />);
+
+      expect(await driver.hasError()).toBe(false);
+    });
+  });
+
```
nitayneeman: I tried to add a test which checks that `errorMessage` is actually rendered in the tooltip, but I've not managed to do that yet.

Not sure how to import and use here the `tooltipTestkit` properly.

src/RichTextInputArea/RichTextInputArea.uni.driver.js
```diff
@@ -4,14 +4,21 @@ import ReactTestUtils from 'react-dom/test-utils';
...
 
 export default base => {
   return {
     ...baseUniDriverFactory(base),
     isDisabled: async () =>
       Boolean(await getContent(base).attr('contenteditable')),
+    hasError: async () => await getErrorIndicator(base).exists(),
     getContent: () => getContent(base).text(),
     getPlaceholder: () => getPlaceholder(base).text(),
+    getErrorMessage: () => getErrorMessage(base).text(),
```
nitayneeman: I think these methods (`getErrorMessage`, `hoverErrorIndicator`) might be relevant for the consumers.

src/RichTextInputArea/RichTextInputArea.js
```diff
@@ -39,6 +40,8 @@ class RichTextInputArea extends React.PureComponent {
     initialValue: PropTypes.string,
     placeholder: PropTypes.string,
     disabled: PropTypes.bool,
+    error: PropTypes.bool,
+    errorMessage: PropTypes.string,
```
nitayneeman: Something to think about - we don't really need two different props... A single prop, for example `error`, which gets the message might be enough, so that:
1. `undefined` means no error mode
2. An empty string means error mode without a message

argshook: this has already been thought about in different components, and the conclusion was:

`status="error|loading|success|info|..."` and `statusMessage=""`

although, i'm not sure how well this idea lived for, unfortunately each component has it's own different implementation of the same thing

nitayneeman: Talked to @shlomitc about this and we decided to stick with the `error` & `errorMessage` version - mainly to be aligned with `<Input/>` and `<InputArea/>` APIs.

argshook: `Input` considers `error` prop deprecated since 2018-07-31

https://github.com/wix/wix-style-react/pull/1784#discussion_r192712418

everything that uses Input internally also use `status` and `statusMessage`:

ColorInput
NumberInput
NoBorderInput
MultiSelect
Loader

i see this as a mistake in the making. but if you insist, who am i to complain

nitayneeman: Got your point, will be implemented

PR #3282, opened 2019-04-21
`<DatePicker/>`- isOpen should response to props change

TalZa1810:
### 🔦 Summary
Resolves #369 

https://github.com/wix-private/wsr-issues/issues/369

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


src/DatePicker/docs/testStory.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
```
shlomitc: Please add this test with the new visual testing solution - `.visual.js`.

src/DatePicker/DatePicker.spec.js
```diff
@@ -86,6 +86,16 @@ describe('DatePicker', () => {
...
 
+    it('should be shown when getting isOpen prop', () => {
+      const { calendarDriver } = createDriver(
+        <DatePicker isOpen onChange={noop} />,
+      );
+      expect(calendarDriver.isVisible()).toBe(true);
+    });
+
+    //TODO: WIP
+    it('should display a warning message when both isOpen & shouldCloseOnSelect set to true', () => {});
```
shlomitc: I don't think they interfere with each other to require a warning message.
It's ok that when `shouldCloseOnSelect` is turned on - the `onClose` callback will be called.

src/DatePicker/DatePicker.js
```diff
@@ -63,6 +62,25 @@ export default class DatePicker extends WixComponent {
...
+  componentDidUpdate(prevProps, prevState, snapshot) {
+    super.componentDidUpdate(prevProps);
+    const { isOpen } = this.state;
+    const isControlled = this.props.hasOwnProperty('isOpen');
+    if (isControlled && isOpen !== this.props.isOpen) {
+      this.setState({ isOpen: this.props.isOpen });
+    }
+  }
+
+  /*
```
shlomitc: remove comment when done

src/DatePicker/DatePicker.spec.js
TalZa1810: The onClose should be called even though the isOpen prop is set to true?

shlomitc: Yes, these are two different props
`isOpen` - control whether the calendar should be displayed
`shouldCloseOnSelect` - control whether the calendar needs to be closed when selecting a date
`onClose` -  called once the calendar is closed.

Since `isOpen` is acting as a controlling prop, it should not stop the closing behavior from being applied.

hepiyellow:
@talza see my comment on the line when you update the state. 
I don’t think a semi-controlled comp is suitable here.

hepiyellow:
Oh, and please read this:
https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html

PR #3296, opened 2019-04-26
<Tooltip/> - change Tooltip testkit exports

mykas:
This PR adds new Testkit export path convention. The reason for this is because it is impossible to do proper upgrade feature toggle on tooltiptestkifactory without forcing user to initialise the testkit without async await.


src/Tooltip/TooltipNext/MIGRATION-API.md
```diff
@@ -2,11 +2,12 @@
 
 ## TL;DR
 
-- react/enzyme testkit is now *async* !
+- react/enzyme testkit is now _async_ !
```
hepiyellow: Not only enzyme. Also reactDom.

src/Tooltip/TooltipNext/Tooltip.js
```diff
@@ -26,7 +26,7 @@ class Tooltip extends React.PureComponent {
     /**  time in milliseconds to wait before hiding the tooltip. Defaults to 0. */
     exitDelay: PropTypes.number,
     /** moves tooltip content relative to the parent */
-    moveTo: PropTypes.shape({ x: PropTypes.number, y: PropTypes.number }),
+    moveBy: PropTypes.shape({ x: PropTypes.number, y: PropTypes.number }),
```
hepiyellow: I guess this is an unrelated fix

mykas: it is but this is just a proptype change. Mistype in proptype.

src/Tooltip/TooltipNext/MIGRATION-API.md
mykas: you mean that to have TooltipTestkit export just for protractor & puppeteer?

src/Tooltip/TooltipNext/docs/index.story.js
```diff
@@ -27,6 +27,7 @@ import { Layout, Cell } from '../../../Layout';
 import { baseScope } from '../../../../stories/utils/LiveCodeExample';
 import usage from './Usage.md';
 import * as examples from './examples';
+import Readme from '!raw-loader!./README.TESTKIT.md';
```
argshook: raw-loader for markdown is not needed
```suggestion
import Readme from './README.TESTKIT.md';
```

PR #3300, opened 2019-04-28
`Button Family` - Proxy driver from ui-core

hepiyellow:
This fixes broken IconButton driver's `isButtonDisabled`
AND
We will now expose the driver from `wix-ui-core`, instead of reimplementing it (which was also not tested)


src/IconButton/IconButton.uni.driver.js
```diff
@@ -1,13 +1,3 @@
...
-export const iconButtonDriverFactory = base => {
-  const stylableUtil = new StylableUnidriverUtil(stylesheet);
-
-  return {
-    ...baseUniDriverFactory(base),
-    isButtonDisabled: async () =>
-      (await stylableUtil.getStyeState(base, 'disabled')) === 'true',
-  };
-};
+export const iconButtonDriverFactory = buttonNextDriverFactory;
```
mykas: Here it might be bad to expose the wix-ui-core full testkit because IconButton does not contain any textual content. Only the icons. It might confuse people https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/button-next/button-next.uni.driver.ts#L20

src/CloseButton/CloseButton.uni.driver.js
```diff
@@ -1,12 +1,3 @@
...
 
-export const closeButtonDriverFactory = base => {
-  const stylableUtil = new StylableUnidriverUtil(stylesheet);
-  return {
-    ...baseUniDriverFactory(base),
-    isButtonDisabled: async () =>
-      (await stylableUtil.getStyleState(base, 'disabled')) === 'true',
-  };
-};
+export const closeButtonDriverFactory = buttonNextDriverFactory;
```
mykas: The same is for the closebutton with the getButtonTextContent: https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/button-next/button-next.uni.driver.ts#L20

hepiyellow: Fixed

src/IconButton/IconButton.uni.driver.js
hepiyellow: Fixed

PR #3308, opened 2019-04-28
<Page /> Add UniDriver

ritabazak:
Please note that 'Prop Validation' tests (3 tests) are failing on the second deriver when the tests are ran on both drivers on the same time. It does run with success when only one of the drivers is being executed.


mosheKerbel:
i have fixed the conflicts
waiting for the tests to pass, then merge

PR #3310, opened 2019-04-29
WIP: New Carousel Component

ychek:
Guild Week Task:
Improving the Carousel component

### 🔦 Summary
We are improving the Carousel component of the library to support image slideshow, support custom content which is not images, support swipe behavior for mobile and several more.
We use react-slick library under the hood.

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


argshook:
description says
> Improving the Carousel component

but code says that it's not an improvement but rather a new component
`CarouselNew`.

what's the reason for new component? Why can't existing one be improved?

i would like to avoid any of the `ComponentNew`, `new-icons`, `upgrade`
methodologies as they tend to be problematic.

what is the main reason that `CarouselNew` was born?

@shlomitc perhaps you're in the loop already

shlomitc:
should be the same original `Carousel` component, @ychek I assume it's just a WIP commit

ychek:
See new pull request :

#3328

Issue #2971, opened 2019-02-18
Page - Conditional inline rendering


varzager:
There is an issue i keep on facing:
u  enforce allowed children in some of the components.

**for example: `<Page>` can have only one of the `Page.Content`, `Page.Header` etc.**

but they don't support Conditional inline rendering - 
like this:`shouldHaveThis && <Page.FixedContent>`.
 so i get warnings 😞 and add redundant code to overcome that.

would be great if there would be a cross solution for this

thanks 🙏

shlomitc:
@hepiyellow WDYT about it?

hepiyellow:
I’m good with it. If a child is false we can ignore the validation. 

PR welcome.

shlomitc:
done

PR #3311, opened 2019-04-29
<Page/> - don't throw an error on falsy child

gilmeir:


gilmeir:
#2971

src/Page/Page.js
```diff
@@ -516,26 +523,16 @@ Page.propTypes = {
...
 
-    if (
-      children[key].type.displayName !== Page.Header.displayName &&
-      children[key].type.displayName !== Page.Content.displayName &&
-      children[key].type.displayName !== Page.FixedContent.displayName &&
-      children[key].type.displayName !== Page.Tail.displayName
-    ) {
+    const allowedDisplayNames = allowedChildren.map(c => c.displayName);
+    const childDisplayName = child.type.displayName;
+    if (!allowedDisplayNames.includes(childDisplayName)) {
```
shlomitc: why is this condition needed? why not just to allow any node?

gilmeir: 1. I refactored the existing validation to a cleaner code, it already existed
2. Page is a kind of a lay-outing component, with a specific look and feel, I think it's correct to enforce only specific children. The Page children (header/content..) can contain inside them any node the user provides.
3. Starting with this fix the children will no longer be mandatory.

Issue #3304, opened 2019-04-28
eyes-storybook fails due to animation on component mount


gilmeir:
# 🐛 Bug Report

Example:
`Accordion: multiple [1024x768] - Failed 1 of 1`
Fails due to an animation: when the accordion mounts the sections open with animation.

Failing build:
http://pullrequest-tc.dev.wixpress.com/viewLog.html?tab=buildLog&buildTypeId=FedInfra_WixStyleReact&buildId=2621198#_state=1507,2318

Issue on Applitools: https://github.com/applitools/eyes-storybook/issues/13

This issue is to keep track of skipped tests due to this issue.

### 👀 Severity
Major

shlomitc:
The missing feature in `eyes-storybook` to support timeouts is now available after this PR got merged https://github.com/applitools/eyes-storybook/pull/14.

This means that in order to fix this test, we should wrap the Accordion visual test and in `componentDidMount` wait for `x`ms according to `Animator`'s timeouts
This is where we use Animator in the Accordion:
https://github.com/wix/wix-style-react/blob/7a5745e55f4bd95299c62893dac4bac0b2936e07/src/Accordion/AccordionItem/AccordionItem.js#L141

and this is the CSS animations timeouts:
https://github.com/wix/wix-animations/blob/master/src/components/Animator/Animator.scss
https://github.com/wix/wix-animations/blob/master/src/components/Animator/constants/constants.js#L18

shlomitc:
using `Snapper` now makes it easy to solve

PR #3314, opened 2019-04-29
test(*): disable animations and transitions for visual eyes tests

argshook:
visual tests sometimes flake due to animations. for example, accordion
component has expanding animation which takes some miliseconds to
complete

it could be that snapshot is taken during the animation, resulting in
considerable visual diff and a broken build.

an easy remedy is to simply disable any css transition and animation delays and durations

### DIFF
--- a/.storybook-test/stories.scss
+++ b/.storybook-test/stories.scss
@@ -4,3 +4,8 @@ html {
   padding: 0;
   margin: 0;
 }
+
+* {
+  transition: none !important;
+  animation: none !important;
+}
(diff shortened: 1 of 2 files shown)


shlomitc:
see my comment here:
https://github.com/wix/wix-style-react/issues/3304#issuecomment-487385931

argshook:
yep, it should be done with what you said in https://github.com/wix/wix-style-react/issues/3304#issuecomment-487385931

closing, this didn't work anyway

PR #3315, opened 2019-04-29
<Breadcrumbs /> Add UniDriver

ritabazak:


mosheKerbel:
@ritabazak let me know when you finish with what's left here and in the page driver, so i can review.

src/Breadcrumbs/Breadcrumbs.uni.driver.js
```diff
@@ -0,0 +1,63 @@
...
+    clickBreadcrumbAt: async position =>
+      (await optionAt(position))
+        .$('[data-hook="breadcrumb-clickable"]')
+        .click(),
+
+    /** return the active breadcrumb item position or return null if no active item exists */
+    getActiveItemId: async () => {
+      const itemsArray = await base
+        .$$('.itemContainer')
+        .map(item => (item.exists() ? item.getNative() : null)); // eslint-disable-line no-restricted-properties
```
mosheKerbel: use ReactBase, it protect us with the other driver types - https://github.com/wix/wix-style-react/blob/0ce68d570964d838eca65865b52c4f0608d89b28/test/utils/unidriver/ReactBase.js#L11

ritabazak: We decided to use getNative after we discussed this with Erez and this is what he told us to do.

PR #3319, opened 2019-04-30
<NestableList /> drag and drop tree

tzvikaf:
### 🔦 Summary
a new component of list/tree that able to drag and drop
with hierarchy.

### ✅ Checklist
- 📚 Change is documented
  - [x] Story
  - [x] API description

- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


tzvikaf:
most of the component code was copied from https://github.com/echenley/react-dnd-nestable
need to add tests

tzvikaf:
not able to see the details of the security/snyk issue

src/DragAndDrop/docs/SortableList/NestableListExample/NestableListExample.js
```diff
@@ -0,0 +1,93 @@
...
+      },
+    ],
+  };
+
+  handleDrop = items => {
+    // console.log(args);
+    // const nextItems = [...this.state.items];
+    // nextItems.splice(addedIndex, 0, ...nextItems.splice(removedIndex, 1));
+    this.setState({
+      items: items,
```
Faradey27: are you sure that ```items: [...items]``` - not required?

also, what do you think about low level api? for example, someone want to know which item on which position in the tree he is droping(or at list on which parent he droped an item)

src/NestableList/Container.js
```diff
@@ -0,0 +1,67 @@
...
+import pure from 'recompose/pure';
+
+import Item from './Item';
+
+function getDepth(item, childrenProperty) {
+  // returns depth of item and children
+  let depth = 0;
+
+  if (item[childrenProperty]) {
+    item[childrenProperty].forEach(d => {
```
Faradey27: maybe while loop is better here?

Faradey27: also, i think it make sense to test such functions separetly

src/NestableList/DragLayer.js
```diff
@@ -0,0 +1,110 @@
...
+    WebkitTransform: transform,
+    width,
+    height,
+  };
+}
+
+const noopConnectDragSource = el => el;
+
+class CustomDragLayer extends Component {
+  componentWillReceiveProps(nextProps) {
```
Faradey27: can we avoid componentWillReceiveProps? as it deprecated in React@16

src/NestableList/DragLayer.js
```diff
@@ -0,0 +1,110 @@
...
+    if (this.props.isPlaceholder !== nextProps.isPlaceholder) {
+      document.body.classList.toggle('dnd-dragging');
```
Faradey27: do we really need this? why on body and not on first parent?

src/NestableList/Item.js
```diff
@@ -0,0 +1,311 @@
...
+import { DragSource, DropTarget } from 'react-dnd';
+import { getEmptyImage } from 'react-dnd-html5-backend';
+import itemTypes from './itemTypes';
+
+// keep track of horizontal mouse movement
+const mouse = {
+  lastX: 0,
+};
+
+function isSamePosition(prevPosition, nextPosition) {
```
Faradey27: looks like shallow equal

src/NestableList/Item.js
```diff
@@ -0,0 +1,311 @@
...
+    // use empty image as a drag preview so browsers don't draw it
+    // and we can draw whatever we want on the custom drag layer instead.
+    this.props.connectDragPreview(getEmptyImage(), {
+      // IE fallback: specify that we'd rather screenshot the node
+      // when it already knows it's being dragged so we can hide it with CSS.
+      captureDraggingState: true,
+    });
+  }
+
+  shouldComponentUpdate(nextProps, nextState, nextContext) {
```
Faradey27: you can use ```class Item extends React.PureComponent``` instead or even ```class Item extends WixComponent```

src/NestableList/NestableList.js
```diff
@@ -0,0 +1,164 @@
...
+
+  return command;
+}
+
+function replaceNegativeIndex(items, nextPosition, childrenProperty) {
+  let currItems = items;
+
+  return nextPosition.map(nextIndex => {
+    if (nextIndex !== -1) {
+      currItems = currItems[nextIndex][childrenProperty] || [];
```
Faradey27: looks not good, maybe with `for` loop it will be more simple)

src/NestableList/NestableList.js
```diff
@@ -0,0 +1,164 @@
...
+      return nextIndex;
+    }
+
+    return currItems.length;
+  });
+}
+
+function getRealNextPosition(prev, next) {
```
Faradey27: some comments to code required, for me it not clear how we can have not real next position

src/NestableList/NestableList.js
```diff
@@ -0,0 +1,164 @@
...
+      itemsToInsert: [dragItem],
+      childrenProperty,
+    });
+
+    newItems = update(newItems, removeItem);
+    newItems = update(newItems, insertItem);
+
+    this.setState({ items: newItems });
+
+    return Promise.resolve(realNextPosition);
```
Faradey27: why promise?

src/NestableList/NestableList.js
```diff
@@ -0,0 +1,164 @@
...
+
+  componentWillReceiveProps(newProps) {
+    if (newProps.items !== this.state.items) {
+      this.setState({ items: newProps.items });
+    }
+  }
+
+  moveItem = ({ dragItem, prevPosition, nextPosition }) => {
+    const { childrenProperty } = this.props;
+    let newItems = this.state.items;
```
Faradey27: new link required, like .slice()

src/NestableList/NestableList.js
```diff
@@ -0,0 +1,164 @@
...
+    newItems = update(newItems, removeItem);
+    newItems = update(newItems, insertItem);
+
+    this.setState({ items: newItems });
+
+    return Promise.resolve(realNextPosition);
+  };
+
+  dropItem = () => {
+    this.props.onUpdate(this.state.items);
```
Faradey27: not sure about api

src/DragAndDrop/docs/SortableList/NestableListExample/NestableListExample.js
tzvikaf: 1. i'm pretty sure it's not required

2. maybe,
i did a rushed job and expect it to fit the Menu Panel feature i'm working on in ADI,
i expect that i'll need to add more props/callbacks to this component

i would prefer to work on advanced props after i'm done with the Menu Panel feature

src/NestableList/Container.js
tzvikaf: 1. what is the added value of a while,
i feel that it's very easy to understand this function as is.

2. yeah, it belongs in a utils service, i'll see where i can put it

src/NestableList/DragLayer.js
tzvikaf: ok, i used componentDidUpdate, looks good

src/NestableList/DragLayer.js
tzvikaf: i'll remove it

src/NestableList/Item.js
tzvikaf: i'll use it

src/NestableList/Item.js
tzvikaf: ok

src/NestableList/NestableList.js
tzvikaf:
```
function replaceNegativeIndex(items, nextPosition, childrenProperty) {
  let currItems = items;
  const fixedNextPosition = [];

  nextPosition.forEach(nextIndex => {
    if (nextIndex !== -1) {
      currItems = currItems[nextIndex][childrenProperty] || [];
      fixedNextPosition.push(nextIndex);
    } else {
      fixedNextPosition.push(currItems.length);
    }
  });

  return fixedNextPosition;
}
```
don't think it looks better,
`currItems = currItems[nextIndex][childrenProperty] || [];`
is the part that makes it complex but there is no real logical shortcut to skip it,
i think that mostly what it tries is to add the dropped item as the last if there are children

src/NestableList/NestableList.js
tzvikaf: works ok without it, i'll change it
my guess is that the original writer needed to delay it to give the event loop chance
to prepare some things

src/NestableList/NestableList.js
tzvikaf: ok,
i guess the author relied on the update calls,
because they return new array

tzvikaf:
> In general looks good, left several comments
> 
> Also demo looks a little strange(with some css issues) + when I have a lot of nested items and d&d their parent - expeariance feels strange, maybe we need UX help here?

you mean this?
image

src/NestableList/NestableList.js
tzvikaf: i understand, but i prefer it over onDrop,
onDrop might give the illusion that it's controlled component and the dev needs to update the
items,
in truth the items object was already updated

src/DragAndDrop/docs/SortableList/NestableListExample/NestableListExample.js
```diff
@@ -0,0 +1,93 @@
...
+      {
+        id: 4,
+        text: 'Item #4',
+        children: [{ id: 5, text: 'Item #5', children: [] }],
+      },
+    ],
+  };
+
+  handleDrop = items => {
+    // console.log(args);
```
Faradey27: can you remove dead code?

src/DragAndDrop/docs/SortableList/NestableListExample/NestableListExample.js
Faradey27: ok, can you create/update issue about this? with what was done in this PR and what still need to be done?

package.json
```diff
@@ -186,6 +186,8 @@
     "popper.js": "^1.14.5",
     "prop-types": "^15.5.8",
     "rc-slider": "^8.0.2",
+    "react-addons-shallow-compare": "^15.3.1",
+    "react-addons-update": "^15.3.2",
```
argshook: is this necessary? if not, please remove

src/DragAndDrop/docs/SortableList/NestableListExample/NestableListExample.js
argshook: remove commented code if it's not useful

src/NestableList/Container.js
```diff
@@ -0,0 +1,54 @@
...
+              ) : null}
+            </Item>
+          );
+        })}
+      </div>
+    );
+  }
+}
+
+const WrappedContainer = pure(Container);
```
argshook: https://github.com/acdlite/recompose#a-note-from-the-author-acdlite-oct-25-2018

please remove usage of deprecated libraries. it seems like `React.PureComponent` would work just as well here

src/NestableList/DragLayer.js
```diff
@@ -0,0 +1,105 @@
+import React, { Component } from 'react';
+import pure from 'recompose/pure';
+import compose from 'recompose/compose';
+import { DragLayer } from 'react-dnd';
+import itemTypes from './itemTypes';
+
+const layerStyles = {
+  position: 'fixed',
+  pointerEvents: 'none',
+  zIndex: 100,
```
argshook: why `100`? why needed at all?

src/NestableList/DragLayer.js
```diff
@@ -0,0 +1,105 @@
...
+
+  return {
+    transform: transform,
+    WebkitTransform: transform,
+    width,
+    height,
+  };
+}
+
+const noopConnectDragSource = el => el;
```
argshook: this is identity function, not noop, name can be misleading

src/DragAndDrop/docs/SortableList/NestableListExample/NestableListExample.js
```diff
@@ -0,0 +1,93 @@
...
+        style={{ opacity: isPlaceholder ? 0 : 1 }}
+        data-hook={`card-${id}-handle`}
+      >
+        Drag Handle
+      </div>
+    );
+  }
+
+  renderItem = ({
+    isPlaceholder,
```
argshook: wouldn't it be easier to manage code if there were separate `renderItem`, `renderPreview` & `renderPlaceholder`?

src/NestableList/DragLayer.js
```diff
@@ -0,0 +1,105 @@
...
+
+class CustomDragLayer extends Component {
+  getChildren = (items, depth) => {
```
argshook: this is not `getChildren`, it's `renderChildren`

src/NestableList/Item.js
argshook: WixComponent is already PureComponent

src/NestableList/Item.js
```diff
@@ -0,0 +1,317 @@
...
+
+      return connectDropTarget(
+        <div data-hook="nestable-item" ref={this._setRootNode}>
+          {renderItem(renderParams)}
+          {renderChildren && children}
+        </div>,
+      );
+    }
+
+    return compose(
```
argshook: do regular function calls to avoid additional dependency

src/NestableList/Item.js
```diff
@@ -0,0 +1,317 @@
...
+
+export const DropItemTarget = DropTarget(
+  itemTypes.nestedItem,
+  cardTarget,
+  connect => ({
+    connectDropTarget: connect.dropTarget(),
+  }),
+)(DragItemSource);
+
+export const ItemContext = getContext({
```
argshook: please avoid usage of `getContext`. we reeeeally don't want to add sparsely used dependencies

src/NestableList/NestableList.js
```diff
@@ -0,0 +1,149 @@
...
+
+class NestableList extends WixComponent {
+  static defaultProps = {
+    items: [],
+    isRenderDraggingChildren: false,
+    childrenProperty: 'children',
+    childrenStyle: {},
+    onUpdate: () => {},
+    renderItem: () => {
+      throw new Error('NestableList: You must supply a renderItem prop.');
```
argshook: why error if `renderItem` has `isRequired`?

src/NestableList/NestableList.js
```diff
@@ -0,0 +1,149 @@
...
+    renderItem: PropTypes.func.isRequired,
+    moveItem: PropTypes.func.isRequired,
+    dropItem: PropTypes.func.isRequired,
+  };
+
+  state = {
+    items: this.props.items,
+  };
+
+  getChildContext() {
```
argshook: throughout this PR multiple deprecated libraries and deprecated APIs are used. Sorry, i cannot accept such code because i do not want to maintain deprecated things.

for react context, if it really is necessary and there's no other way, please use `create-react-context`. you can inspect Input component https://github.com/wix/wix-style-react/blob/master/src/Input/InputContext.js

src/NestableList/NestableList.js
```diff
@@ -0,0 +1,149 @@
...
+      useDragHandle,
+      maxDepth,
+      threshold,
+      renderItem,
+      moveItem: this.moveItem,
+      dropItem: this.dropItem,
+    };
+  }
+
+  componentWillReceiveProps(newProps) {
```
argshook: please use `getDerivedStateFromProps` together with `react-lifecycles-compat` polyfill, so that we're covered for future

you can see here how to use it https://github.com/wix/wix-style-react/blob/master/src/ColorInput/ColorInput.js#L4

src/NestableList/NestableList.js
```diff
@@ -0,0 +1,149 @@
...
+    if (nextIndex !== -1) {
+      currItems = currItems[nextIndex][childrenProperty] || [];
+      return nextIndex;
+    }
+
+    return currItems.length;
+  });
+}
+
+// fixing a bug where the new path is increased in nesting
```
argshook: where is this bug coming from?

src/NestableList/NestableList.driver.js
```diff
@@ -0,0 +1,49 @@
+import ReactTestUtils from 'react-dom/test-utils';
+import {
+  getInstanceOfDraggableProvider,
+  getInstanceOfDraggableSource,
+  getInstanceOfDraggableTarget,
+} from './driverHelpers';
+
+const nestableListFactory = ({ element, wrapper }) => {
+  // in case if wrapper is coming from enzyme, we want to get it instance
```
argshook: driver should be written as unidriver

https://github.com/wix/wix-style-react/blob/master/docs/contribution/TEST_DRIVERS_GUIDELINES.md
https://github.com/wix/wix-style-react/blob/master/docs/contribution/TEST_DRIVERS.md

src/NestableList/NestableList.js
```diff
@@ -0,0 +1,149 @@
...
+// and the first position jumps too far by 2, and that's why it decreased by 1
+// need to check if the jump can become more severe
+function getRealNextPosition(prev, next) {
+  // moving up a level
+  if (prev.length < next.length) {
+    return next.map((nextIndex, i) => {
+      if (typeof prev[i] !== 'number') {
+        return nextIndex;
+      }
+      return nextIndex > prev[i] ? nextIndex - 1 : nextIndex;
```
argshook: this feels very fragile. could you explain in short what is the main problem you try to solve?

src/NestableList/Item.js
```diff
@@ -0,0 +1,317 @@
...
+
+const cardTarget = {
+  drop(props, monitor) {
+    // clear mouse position
+    mouse.lastX = 0;
+    if (!monitor.didDrop()) {
+      props.dropItem();
+    }
+  },
+  hover(props, monitor, component) {
```
argshook: body of `hover` function should be split into smaller manageable chunks of code

src/DragAndDrop/docs/SortableList/NestableListExample/NestableListExample.js
tzvikaf: might be,
i just don't think that the difference between the states is so big that it requires 3 different functions to handle,
most of the time the dev need just to adjust the rendered item a bit,

if the renderItem start containing switch cases then maybe then it's better to just do 3 different functions.

anyway, i can add those 2 and use renderItem as fallback if the other 2 don't exist,
seems good?

tzvikaf:
@argshook i need the component for a feature in ADI,
i'll continue to work on it until it fits the wix-style-react standard,

i'm not accustomed to develop for wix-style-react so i'll need quite a bit of guidance,
thanks for the helping me

package.json
```diff
@@ -195,6 +197,7 @@
     "react-portal": "^4.1.5",
     "react-transition-group": "^2.2.1",
     "react-window": "^1.7.1",
+    "recompose": "^0.18.0",
```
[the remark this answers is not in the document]
tzvikaf: i'll try to remove it but it's more difficult compared to the others,
i'll need to find an alternative for getContext

src/NestableList/NestableList.js
tzvikaf: it happens when dragging sub-item to become a sub-item of the item below it
it is very brittle but it happens, it goes into situation of jumping 2 indexes,
so it needs to be lowered by one,

because drag and drop is only with one item generally the offset is by one index and the max of the error can be 2

tzvikaf:
i got a deadline by the end of next week,
so i want to focus on the feature and get back here right after that,
is that ok?

argshook:
@tzvikaf thanks for the update!

@shlomitc fyi, perhaps someone from the team could continue so that @tzvikaf is not burdened with work?

shlomitc:
> @tzvikaf thanks for the update!
> 
> @shlomitc fyi, perhaps someone from the team could continue so that @tzvikaf is not burdened with work?

@Faradey27 - can the next guild week keep working on it?

PR #3320, opened 2019-04-30
<FormField /> Add UniDriver

ritabazak:


src/FormField/FormField.uni.driver.js
```diff
@@ -0,0 +1,46 @@
...
+
+export const formFieldUniDriverFactory = base => {
+  const charactersCounter = () => base.$('[data-hook*="formfield-counter"]');
+
+  return {
+    ...baseUniDriverFactory(base),
+    getChildren: async () => {
+      const baseUniDriverElement = base.$('[data-hook*="formfield-children"]');
+      return (await baseUniDriverElement.exists())
+        ? baseUniDriverElement.getNative() // eslint-disable-line no-restricted-properties
```
mosheKerbel: we don't want to access getNative, this is way we added this restirciton inside eslint
when accessing native stuff we want to do it through ReactBase
plz take a look at this marked line for an example:
  https://github.com/wix/wix-style-react/blob/a66e90758276cbb11cc8ea09232430e307ca9853/src/Input/Input.uni.driver.js#L9

src/FormField/FormField.uni.driver.js
```diff
@@ -0,0 +1,46 @@
...
+        : null;
+    },
+    getLabel: async () => {
+      const baseUniDriverElement = base.$('[data-hook*="formfield-label"]');
+      return (await baseUniDriverElement.exists())
+        ? baseUniDriverElement.getNative() // eslint-disable-line no-restricted-properties
```
mosheKerbel: same as above

src/FormField/FormField.spec.js
```diff
@@ -1,220 +1,235 @@
...
-      const driver = createDriver(renderFormField());
-      expect(driver.isRequired()).toEqual(false);
-    });
-  });
+      describe('given only `required` prop', () => {
+        it('should render it inline', async () => {
+          const { driver } = render(renderFormField({ required }));
+          expect(await driver.isRequired()).toEqual(true);
+          expect(
+            !!(await driver.element()).querySelector(
```
mosheKerbel: if element does not exist, unidriver throws an exception
i would add an assert before to make sure the element exist, if not then test will fail with normal description
`await driver.exists()`

src/FormField/FormField.spec.js
```diff
@@ -1,220 +1,235 @@
...
-              label,
-              children: setCharactersLeftOnMount(charactersLeft),
-            }),
-          );
-          expect(driver.isLengthExceeded()).toBe(true);
-        });
+      it('should apply minHeight on label wrapper when there is no children', async () => {
+        const { driver } = render(<FormField label="Text" />);
+        expect(
+          (await driver.element()).querySelector(`.${styles.minLabelHeight}`),
```
mosheKerbel: if element does not exist, unidriver throws an exception
i would add an assert before to make sure the element exist, if not then test will fail with normal description
`expect(await driver.exists()).toEqual(true)`

src/FormField/FormField.uni.driver.js
ritabazak: We decided to use getNative after we discussed this with Erez and this is what he told us to do.

src/FormField/FormField.uni.driver.js
ritabazak: We decided to use getNative after we discussed this with Erez and this is what he told us to do.

PR #3325, opened 2019-04-30
<RadioButton> - Add unidriver

liorcode:


src/RadioGroup/RadioButton/RadioButton.uni.driver.js
```diff
@@ -0,0 +1,19 @@
...
+export const radioButtonUniDriverFactory = base => {
+  const reactBase = ReactBase(base);
+
+  const getRadioButton = async () => (await reactBase.children())[0];
+  const getLabel = async () => (await reactBase.children())[1];
+  const getContent = () => base.$('[data-hook="radio-button-content"]');
+
+  return {
+    ...baseUniDriverFactory(base),
+    check: async () => ReactBase(await getRadioButton()).change(),
```
mosheKerbel: it means that consumers won't be able to use "check function in platforms other than React/jsdom.
Let's try to remove it and to use some combination of click and press the enter key and see if it works.

src/RadioGroup/RadioButton/RadioButton.uni.driver.js
```diff
@@ -0,0 +1,19 @@
+import { baseUniDriverFactory, ReactBase } from '../../../test/utils/unidriver';
+
+export const radioButtonUniDriverFactory = base => {
+  const reactBase = ReactBase(base);
+
+  const getRadioButton = async () => (await reactBase.children())[0];
```
mosheKerbel: try to replace it with the data-hook to radioButton

src/RadioGroup/RadioButton/RadioButton.uni.driver.js
```diff
@@ -0,0 +1,19 @@
...
+  const getLabel = async () => (await reactBase.children())[1];
```
mosheKerbel: try to use the label data-hook

mosheKerbel:
@liorcode plz resolve conflicts and then we can merge

mosheKerbel:
@liorcode 
one of the async tests is  failing

image

liorcode:
> @liorcode
> one of the async tests is failing
> 
> image

Weird, I tried running it several times and couldn't reproduce on my machine...is there anything special about the env it is running on that you can think of?

mosheKerbel:
indeed wierd
did not happen on my machine also
will try to rebuild it on teamcity

mosheKerbel:
note to self - now reproduces on local machine
seems like click() of UniDriver is not triggering the onChange event of a radio button
need to figure it out.

PR #3326, opened 2019-04-30
`<AddItem/>` - migrate Tooltip & migrate to stylable & add unidriver

mykas:
This PR will migrate Tooltip component and update testkit methods to use new Tooltip driver.

- Testkit method `getTooltipDriver()`.Bad decision last year to expose it to users as I am unable to reassign it to new Tooltip driver without breaking users. Instead of this I am adding deprecationLog for our users to use other methods instead.

- Migrated component to use `.st.css aka stylable`. This component is an example on how to extend tooltips stylesheet and use it properly.

- This PR is also refactoring some unit tests and makes sure that `uni.driver` is available.

- Migrate and update documentations.


src/AddItem/AddItem.js
```diff
@@ -67,13 +68,26 @@ class AddItem extends Component {
 
   _renderText = () => {
     const { children, disabled, theme, size } = this.props;
+
+    const textSize = size === 'tiny' ? 'small' : 'medium';
+    const textColor = disabled ? '#CBD3DC' : '#3899EC';
```
shlomitc: Why hard coded color and not from palette? Doesn’t make sense

src/AddItem/AddItem.js
```diff
@@ -60,17 +47,11 @@ class AddItem extends Component {
     /** used for testing */
     dataHook: PropTypes.string,
 
-    /** Tooltip props, leave undefined for no tooltip */
+    /** @deprecated do not use this prop. Make sure to use tooltipcontent */
     tooltipProps: PropTypes.shape(Tooltip.propTypes),
```
shlomitc: Can you check why this was required? I do find sense in passing tooltip configuration such as position related props and appendTo

src/AddItem/AddItem.js
mykas: True, I thought of this as a temporal hack until I can extend `Text.st.css` properly and apply it through  stylable. Though I think I'll just create a new classname here and switch it there.

src/AddItem/AddItem.js
mykas: Yes, I agree. One of my plans on this PR is to expose mostly used popover props `appendTo`, `flip`, `fixed` and maybe `placement`. Other props are not necessary.

src/AddItem/AddItem.js
mykas: O wait actually I can't pass any classname to Text component so it means I can't just pass classname. Will come back to this after `Text.st.css` is fixed.

mykas:
@ZilvinasMaciulis Hey, lets look at this Storybook Page and try to improve descriptions and examples.

PR #3329, opened 2019-04-30
<EditableRow> - add uniDriver

liorcode:


mosheKerbel:
https://github.com/wix/wix-style-react/blob/a3ef3992428ab5faf3d6dc9d0956cfc2045ead32/scripts/generate-testkit-exports/list-all-components.js
This script generate list of components to be generated. The smoke test is using the components for the test. The script does not get EditableRow in it's scan because it is not sitting under a root with the same name. This is an issue to be fixed first in the script.

mosheKerbel:
opened an issue https://github.com/wix/wix-style-react/issues/3333

PR #3342, opened 2019-05-02
<RichTextInputArea/> - add scrollBehavior polyfill 

nitayneeman:
This PR creates adds a new polyfill, called "scrollBehavior", in order to supply the `scrollTo` function for tests that are run in an emulated browser environment, such as `jsdom`. 

Related to issue:
wix-private/wsr-issues#35


testkit/polyfills/scroll-behavior.js
```diff
@@ -0,0 +1,14 @@
+const originalScrollTo = window.scrollTo;
+
+const install = async () => {
+  window.scrollTo = () => {};
```
nitayneeman: Currently, it's not a polyfill but rather a stub

nitayneeman:
Failed because of timeout

PR #3343, opened 2019-05-02
added debounce functionality to the ModalSelectorLayout component

bt4R9:
### 🔦 Summary
Just added a debounce functionality to the Search component inside a ModalSelectorLayout.

### ✅ Checklist
- 📚 Change is documented
  - [x] Story
  - [x] API description
- 🔬 Change is tested
  - [x] Component
  - [x] Visual test

### DIFF
--- a/src/ModalSelectorLayout/docs/index.story.js
+++ b/src/ModalSelectorLayout/docs/index.story.js
@@ -32,7 +32,7 @@ export default {
     itemsPerPage: 4,
     imageSize: 'large',
     withSearch: true,
-    searchDebounceMS: 100,
+    searchDebounceMs: 100,
 
     dataSource: (searchQuery, offset, limit) =>
       new Promise(resolve =>
--- a/src/ModalSelectorLayout/docs/index.story.js
+++ b/src/ModalSelectorLayout/docs/index.story.js
@@ -32,6 +32,7 @@ export default {
     itemsPerPage: 4,
     imageSize: 'large',
     withSearch: true,
+    searchDebounceMS: 100,
 
     dataSource: (searchQuery, offset, limit) =>
       new Promise(resolve =>
--- a/src/ModalSelectorLayout/docs/index.story.js
+++ b/src/ModalSelectorLayout/docs/index.story.js
@@ -32,7 +32,7 @@ export default {
     itemsPerPage: 4,
     imageSize: 'large',
     withSearch: true,
-    searchDebounceMs: 100,
+    searchDebounceMs: 150,
 
     dataSource: (searchQuery, offset, limit) =>
       new Promise(resolve =>
--- a/src/Search/Search.js
+++ b/src/Search/Search.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import { polyfill } from 'react-lifecycles-compat';
 import isEmpty from 'lodash/isEmpty';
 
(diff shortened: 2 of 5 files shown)


src/ModalSelectorLayout/docs/index.story.js
```diff
@@ -32,6 +32,7 @@ export default {
     itemsPerPage: 4,
     imageSize: 'large',
     withSearch: true,
+    searchDebounceMS: 100,
```
argshook:
```suggestion
    searchDebounceMs: 100,
```

bt4R9:
Hi! I'm not sure we should change the behaviour of the `Search` component since it's controlled by `ModalSelectorLayout`. I mean we should provide the value immediately to the `Search` component in order to update the input value, but at the same time we want to get debounced onChange callback. Providing `searchDebounceMs` to the `Search` literally means that the component is no longer controlled by the parent component but we also want an onChange callback that makes current API of the `Search` inconsistent. What do you thing?

bt4R9:
@argshook So in the last commit I moved debounce from the `ModalSelectorLayout` to the `Search` but the API I came up with is kinda inconsistent now because `_isControlled` is in conflict with `_isDebounced` which makes the API ugly. May be you have an idea how to save the old behaviour and smoothly implement debounce functionality? thx

src/Search/Search.js
```diff
@@ -57,19 +79,25 @@ export default class Search extends WixComponent {
...
       this.props.onChange(e);
     } else {
       this.setState({
         inputValue: e.target.value,
       });
     }
+
+    if (this._isDebounced) {
+      // need a copy of the event because of the dobunce
+      this._onChangeDebounced({ ...e });
```
bt4R9: could be `e.persist()`, have no idea what approach is better :)

argshook:
@bt4R9 sorry for taking a while. I submitted a PR to your branch with some changes. please see if it covers your cases and lets merge the functionality 
https://github.com/bt4R9/wix-style-react/pull/1

bt4R9:
@argshook Hi, thx! The only thing I'd like change is the name of `_makeDebounced` method since internally it can both debounced and not debounced. How about `this._onChange = this._makeOnChange();`?

argshook:
> @argshook Hi, thx! The only thing I'd like to change is the name of `_makeDebounced` method since internally it can be both debounced and not debounced. How about `this._onChange = this._makeOnChange();`?

sure thing, you can push changes if you have more in mind :)

let me know when it's ready to merge

bt4R9:
@argshook Let's merge it :)

argshook:
@bt4R9 oh, i just noticed that only ModalSelectorLayout has tests for debounce functionality, but not Search component. That's no good, because ModalSelectorLayout could change or get removed and so tests together with it.

would you be willing to add tests for Search too?
i can do that also but am busy as usual, it'll take time till i reach it

bt4R9:
Sure.

bt4R9:
check it plz
may be we need more cases to check?

argshook:
> may be we need more cases to check?

unless you can think of more. i mostly want tests for regression, so that 6 months from now they would cover us

if build is green, i'll merge, thanks for your work!

bt4R9:
I'm not sure why the build is red. It seems like we have flaking e2e tests =/

argshook:
yeah, we do have visual tests that like to timeout occasionally.
It's ok now, thanks for your work!

PR #3351, opened 2019-05-05
Allow user to set max height

neta-hemyan:
Currently,  `<DropdownBase />` uses `<DropdownLayout />`, which have a default height of 260px. A user of `<DropdownLayout /> can change this property, but it isn't accessible  to a user of `<DropdownBase />`. 

I've added a `maxHeight` property to `<DropdownBase />`, which is then passed to `<DropdownLayout />`

### DIFF
--- a/src/DropdownBase/DropdownBase.js
+++ b/src/DropdownBase/DropdownBase.js
@@ -286,6 +286,7 @@ class DropdownBase extends React.PureComponent {
       options,
       minWidth,
       maxWidth,
+      maxHeight,
     } = this.props;
 
     const { open, selectedId } = this.state;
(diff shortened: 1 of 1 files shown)


src/DropdownBase/DropdownBase.js
```diff
@@ -318,6 +320,7 @@ class DropdownBase extends React.PureComponent {
             }}
           >
             <DropdownLayout
+              maxHeightPixals={maxHeight}
```
nitayneeman: Looks like a typo, probably should be `maxHeightPixels`

nitayneeman: Typically the first prop we pass is `dataHook`, so I would place the new prop after it

nitayneeman:
Please add a component test the covers this behavior

shlomitc:
This PR is stale for a while now so I'm closing it. please re-open when you can fix according to the given comments. thanks

PR #3354, opened 2019-05-05
<RichTextArea/> - deprecate the component

nitayneeman:
This PR deprecates the old `<RichTextArea/>` component.

Related to issue:
wix-private/wsr-issues#35


src/Accordion/docs/index.story.js
```diff
@@ -16,7 +16,7 @@ import {
 
 import { storySettings } from './storySettings';
 import Accordion from '..';
-import RichTextArea from '../../RichTextArea';
+import RichTextArea from '../../Deprecated/RichTextArea';
```
shlomitc: use the new component instead, it's here only for the example anyway

stories/Deprecated/RichTextArea/storySettings.js
```diff
@@ -1,9 +1,8 @@
-import { Category } from '../../../stories/storiesHierarchy';
+import { Category } from '../../storiesHierarchy';
 
 export const storySettings = {
   storyName: 'Rich Text Area',
-  indexCategory: Category.INPUTS,
-  apiCategory: Category.COMPONENTS,
+  category: Category.DEPRECATED,
```
shlomitc: please notice that if the component is used also in the `e2e`, then this means a new baseline, since you change the name of the test

nitayneeman:
Failed because of timeout

PR #3358, opened 2019-05-06
NestableList updates + removing deprecated libraries and including NestableList in main index.js

tzvikaf:
updates and changes in reaction to the comments in PR https://github.com/wix/wix-style-react/pull/3319
also including the NestableList in index.js, looks like it's not showing in dist without it
### 🔦 Summary
  - [x] Code Updates
  - [x] Bug fix


tzvikaf:
the main addition is the 2 last commits,
@argshook i'll do the the tests and driver refactor later,

it looks like i forgot to include NestableList in index.js so i can't see it the deployed WSR,
maybe i just need to wait for version 6.17.2, i don't know how it works

src/NestableList/Item.js
```diff
@@ -275,14 +268,13 @@ class Item extends WixComponent {
...
-    )(
-      <div data-hook="nestable-item">
-        {renderItem(renderParams)}
-        {renderChildren && children}
-      </div>,
+    return connectDropTarget(
+      connectDragSource(
+        <div data-hook="nestable-item">
+          {renderItem(renderParams)}
+          {renderChildren && children}
```
Faradey27: i do not like the name - `renderChildren`, for me it sounds like method, expecially because we usually give `render*` name to methods that return jsx

src/NestableList/NestableList.js
```diff
@@ -60,6 +61,15 @@ class NestableList extends WixComponent {
     dropItem: PropTypes.func.isRequired,
   };
 
+  static getDerivedStateFromProps(props, state) {
```
Faradey27: do we support react 16 latest features?
@argshook

argshook: we still have to be backwards compatible with v15, so `getDerivedStateFromProps` should be polyfilled with `react-lifecycles-compat`, like in ColorInput: 

https://github.com/wix/wix-style-react/blob/master/src/ColorInput/ColorInput.js

argshook: ah, which is done in this PR, so all is good :)

src/NestableList/Item.js
tzvikaf: is **shouldRenderChildren** a better name?

tzvikaf:
i'm not sure what are those tests that fail in the CI build,
they have no connection to my changes, feels flaky

tzvikaf:
how can i fix the travis build? 
i don't understand what's the issue

tzvikaf:
@Faradey27 can you please help me with merging this PR,
there is a bug i left from the previous PR (accessing PropTypes on React, it's undefined on the react version of wix-ui-adi)

Faradey27:
@tzvikaf i think travis will be red and it is ok, about teamcity - looks like flaky test, let just re-trigger it several times, if it will become green - let merge

tzvikaf:
it's green, but i'm not authorized to merge

PR #3360, opened 2019-05-06
fix floating notification type for text

brumeregan:
### 🔦 Summary

changed type for FloatingNotification text prop from PropTypes.string to PropTypes.node


argshook:
you're the best, thanks! :)

Issue #3361, opened 2019-05-06
DataTableDriver.exists() not working as expected


yaarams:
# 🐛 Bug Report

1. testkit.exists() always returns true due to this: https://github.com/wix/wix-style-react/blob/master/src/DataTable/DataTable.driver.js#L14
2. The testkit is eager. It expects the element to be rendered when constructing the testkit.

### 🏗 Relevant Components

...

### 😯 Current Behavior

...

### 🤔 Expected Behavior

...

### 👣 Steps to Reproduce

  - **WSR version**: `...`
  - **Yoshi version**: `...`

### 👀 Severity

- Critical / Major / Low

mosheKerbel:
fixed and release.
https://github.com/wix/wix-style-react/pull/3367

PR #3367, opened 2019-05-07
fix exists() in dataTable driver

mosheKerbel:
Fix issue:
https://github.com/wix/wix-style-react/issues/3361#event-2323865546

When an empty react comment element is created, we would like exists() to return false.

@shlomitc


src/DataTable/DataTable.driver.js
```diff
@@ -11,31 +11,37 @@ const dataTableDriverFactory = ({ element }) => {
...
+    protectedElement.querySelector(`th [data-hook="${index}_info_tooltip"]`);
   const getSortableTitleArrowDesc = index =>
     element.querySelector(
       `th [data-hook="${index}_title"]  [data-hook="sort_arrow_dec"]`,
     );
 
   return {
-    exists: () => !!element,
+    exists: () => {
+      return !!element && element.nodeType !== Node.COMMENT_NODE;
```
shlomitc: why not returning `!isDisplayingNothing`?

PR #3328, opened 2019-04-30
WIP: Carousel Component improvement

ychek:
Guild Week Task by @ychek & @tombenezra :
Improving the Carousel component

### 🔦 Summary

We are improving the Carousel component of the library to support image slideshow, support custom content which is not images, support swipe behavior for mobile and several more.
We use react-slick library under the hood.

### ✅ Checklist
- [ ] 👨‍💻 API change is approved by the UI Infra Developers 
- [ ] 👨‍🎨 UX change is approved by the Design System UX 
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test

PR #3368, opened 2019-05-07
<Carousel/> - improve carousel component

nitayneeman:
This PR adds the ability to render nodes in addition to images. Also, it fixes the broken component tests.

Note: It's a feature branch of https://github.com/wix/wix-style-react/pull/3328 in order to see the diffs in a simple way. After we merge it we'll merge the original as well.


src/Carousel/Carousel.spec.js
```diff
@@ -16,10 +16,26 @@ describe('Carousel', () => {
   const createDriver = createDriverFactory(carouselDriverFactory);
 
   it('should be rendered', () => {
-    const driver = createDriver(<Carousel images={[]} />);
+    const driver = createDriver(<Carousel />);
```
nitayneeman: Unnecessary, because `images` gets an empty array by default

src/Carousel/Carousel.js
```diff
@@ -37,9 +37,11 @@ class Carousel extends React.Component {
 
   _renderImages = images => {
     return images.map((image, index) => (
-      <Proportion aspectRatio={Proportion.PREDEFINED_RATIOS.landscape}>
+      <Proportion
+        key={`${index}${image.src}`}
```
nitayneeman: The `key` prop should be here (and not on the inner div). Most of the tests failed without it.

src/Carousel/Carousel.js
```diff
@@ -92,7 +94,7 @@ class Carousel extends React.Component {
         <WrappedSliderArrow
           arrowSize={arrowSize}
           arrowSkin={arrowSkin}
-          dataHook={dataHooks.prevButton}
+          dataHook={dataHooks.nextButton}
```
nitayneeman: A few tests failed because that's an incorrect dataHook

src/Carousel/Carousel.driver.js
```diff
@@ -5,10 +5,15 @@ export const carouselDriverFactory = ({ element }) => {
       const loader = element.querySelector('[data-hook="loader"]');
       return !!loader;
     },
+    getChildren: () => {
+      // react-slick duplicates the children, so we ditch the cloned nodes
+      return element.querySelectorAll('.slick-slide:not(.slick-cloned)');
```
nitayneeman: I don't think it's a good practice to use these classes, but the library duplicates the supplied nodes in order to support the infinite mode:
image

Not sure if there's a better way to filter them out.

src/Carousel/Carousel.driver.js
```diff
@@ -5,10 +5,15 @@ export const carouselDriverFactory = ({ element }) => {
...
+    },
     getImages: () => {
-      return element
-        .querySelectorAll('[data-hook="carousel-img"]')
-        .map(img => img.src);
+      // Converting the result from NodeList to a real array
+      return [...element.querySelectorAll('[data-hook="carousel-img"]')].map(
```
nitayneeman: Actually we've here the same problem (for example, in case of 3 images - it would return 7)

PR #3372, opened 2019-05-08
`<Table.Loader>` - added component

dorduch:
### 🔦 Summary
...

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test

### DIFF
--- a/src/Table/Table.js
+++ b/src/Table/Table.js
@@ -14,6 +14,7 @@ import {
   TableTitleBar,
   TableContent,
   TableEmptyState,
+  TableLoader,
 } from './components';
 
 export function createColumns({ tableProps, bulkSelectionContext }) {
@@ -81,6 +82,7 @@ export class Table extends React.Component {
   static Titlebar = TableTitleBar;
   static Content = TableContent;
   static EmptyState = TableEmptyState;
+  static Loader = TableLoader;
 
   shouldComponentUpdate() {
     // Table is not really a PureComponent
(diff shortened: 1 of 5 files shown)


shlomitc:
@wuwa - can you please review?
https://wix-wix-style-react-pr-3372.surge.sh/?selectedKind=Components&selectedStory=Table&full=0&addons=0&stories=1&panelRight=0
see the "Table Loader" example

shlomitc:
@wuwa - AFAIK, the table height is defined by its content, meaning if there are no items - it doesn't get any height.
Maybe in real layout, the table should be rendered in a stretched card that takes the entire screen height, but then the loader will not be centered, just at the top

wuwa:
@shlomitc i don't think we can count on full height card stretch. is there a way to give it min height according to expected items height?

shlomitc:
@wuwa - this might still cause a "jump" between the loader and the actual size. I guess we can set something up if this is the required behavior

shlomitc:
@dorduch @wuwa - what do you guys suggest?

wuwa:
can card (containing the table) be set to have min-height in px?

shlomitc:
@wuwa - I don't think it's built in. I guess this can be added with this feature as well if defined well

shlomitc:
It's been hanging for a while now. I'm closing and we'll reopen once we work on it

Issue #3244, opened 2019-04-04
<Text/> - applying a stylable className on ellipsis Text causes errors


shlomitc:
# 🐛 Bug Report

### 🏗 Relevant Components
`<Text/>`

### 😯 Current Behavior
The `Text` component support the ellipsis behavior, adding a `Tooltip` to the truncated text.
In some cases, we want to extend the `Text` component with stylable extended classNames, but get these kind of errors:

![image](https://user-images.githubusercontent.com/6093192/55568089-b2f06f80-5707-11e9-9433-f2e48b8a7395.png)

The reason for it is because internally to the text, the `EllipsedText` component, is propagating the className both for the component and the tooltip, causing an incorrect extending (using `withEllipsedTooltip` HOC).

### 🤔 Expected Behavior
Apply the class name and data-* attributes only to the wrapped component (it's tricky beacuse we override both the `Text` and the `Tooltip` internally.

### 👣 Steps to Reproduce
1. Create the extending text
```css
.extendedText {
  -st-extends: Text;
}

.extendedText:size(medium) {
  color: green;
}
```

2. Apply to component:
```jsx
<Text ellipsis className={styles.extendedText}>hello</Text>
```

3. Render the component in storybook

  - **WSR version**: `^6`
  - **Yoshi version**: `^4`

### Possible Solution
split the css of the Text from Tooltip.

If the fix is too complicated with HOC, I think we should get rid of the HOC solution and move to a render prop children pattern like this:

```jsx
<EllipsedTooltip className={styles.extendedTooltip}>
{
  ({stylesForEllipsis}) => (
     <Text className={classNames(stylesForEllipsis, styles.extendedText)}>hello</Text>
   )
}
</EllipsedTooltip>
```

### 👀 Severity

- Major

shlomitc:
This PR was merged https://github.com/wix/wix-style-react/pull/3378 so it should be ok

PR #3378, opened 2019-05-08
<Text/> - allow passing classname, fix issue when css class is overri…

shlomitc:
still WIP. trying to solve this issue https://github.com/wix/wix-style-react/issues/3244


src/Text/ProxyText.js
```diff
@@ -1,30 +1,36 @@
...
+
+const renderText = props => {
+  return <OriginalText {...props} />;
+};
+
+const renderTextWithEllipsis = props => {
+  const EllipsedText = EllipsedTextWrapper(enhancedProps =>
+    renderText({
+      ...enhancedProps,
+      className: classNames(enhancedProps.className, props.className), //apperantly the HOC is overriding the classname, so we must maintain it manualy
```
mykas: Hey, that is great that we will support passing additional classnames, but as I see from the comments already it does have some issues while letting users pass and override our styles. Wouldn't it be great to teach our users how to extend the Text components stylable file and then add their style?

src/Text/ProxyText.js
```diff
@@ -1,30 +1,36 @@
 import React from 'react';
 import { bool } from 'prop-types';
-import { createHOC } from 'wix-ui-core/dist/src/createHOC';
-import Text from './Text';
+import OriginalText from './Text';
 import { withEllipsedTooltip } from 'wix-ui-core/dist/src/hocs/EllipsedTooltip';
 import ellipsedStyle from '../common/EllipsedTooltip/EllipsedTooltip.st.css';
```
mykas: This comes from wix-ui-backoffice which extends wix-ui-cores Tooltip. I think we should extend our wix-style-react tooltip which has proper stylings applied. what do you think?

shlomitc: There are the current styles, I didn't change them and maybe better not to change in this PR.

In general, please notice that the `withEllipsedTooltip` is targeting the core Tooltip, meaning that if our `wix-style-react` current tooltip has some additions (using the `Text` component for example), then we need to change it to work only with CSS in order to share the same styling logic

src/Text/ProxyText.js
shlomitc: It's not related.

The bug is this - adding any `className` to the `Text` will override the tooltip styles (hover the Text and see):
```jsx
<div style={{width: '50px'}}>
  <Text ellipsis className="some-class">hello world</Text>
</div>
```
Even if you do the class styles extending with stylable, the bug will remain, because you will extend the stylesheet and pass a `className`

src/Text/ProxyText.js
shlomitc: image

mykas: Actually there is nothing special about the wsr Tooltip.st.css. What it does is extends that core Tooltip and adds additional styles. From the example above I see that extending the styles for Tooltip is not working properly yet which means something is breaking it. Maybe that additional `some-class`?

shlomitc: passing `some-class` is intended for the Text, not the Tooltip, but causes the Tooltip to break.

mykas: Could it be that the classsName that is being passed to `EllipsedText` is actually applying on Tooltip and not the Text component?

mykas: O I see now. Its not

src/Text/ProxyText.js
mykas: What I am not sure is this part. Are we sure that echancedProps.classname is not the Tooltips style coming from wix-ui-backoffice? I will check in wix-ui-core.

mykas: Because Tooltips extended style tag should land on Tooltip itself and not the Text. I suspect that this might be the reason for clash

src/Text/ProxyText.js
mykas: Screenshot 2019-05-13 at 10 47 19
EllipsedTooltips classname is landing on Text component. Instead it should stay on Tooltips elements container.

mykas: I think here: https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/hocs/EllipsedTooltip/withEllipsedTooltip.tsx#L92 we get all the classnames that are passed to the EllipsedTooltip and then its a clash.

src/Text/ProxyText.js
```diff
@@ -6,7 +6,7 @@ import tooltip from '../Tooltip/TooltipNext/Tooltip.st.css';
 
 const EllipsedText = withEllipsedTooltip({
   showTooltip: true,
-  tooltipProps: tooltip('root'),
+  tooltipProps: { ...tooltip('root') },
```
argshook: @mykas 	what happened?

mykas: Does not work on protractor. Fails on `tooltip.default is not a function`

PR #3387, opened 2019-05-12
Add component type definitions (#2)

giladsegal:
### 🔦 Summary
...

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test

### DIFF
--- a/src/Calendar/Calendar.uni.driver.d.ts
+++ b/src/Calendar/Calendar.uni.driver.d.ts
@@ -1,0 +1,43 @@
+import {BaseUniDriver} from 'wix-ui-test-utils/unidriver';
+import {SyntheticEventData} from 'react-dom/test-utils';
+import {DropdownLayoutDriver} from '../DropdownLayout/DropdownLayout.driver';
+
+export interface CalendarDriver extends BaseUniDriver {
+  close: () => Promise<void>;
+  isVisible: () => Promise<boolean>;
+  getCurrentMonthWithYear: () => Promise<string>;
+  getNthWeekDayName: (n?: number) => Promise<string>;
+  clickOnNthDay: (n?: number) => Promise<void>;
+  clickDay: (date: Date) => Promise<void>;
+  clickOnNthDayOfTheMonth: (n?: number) => Promise<void>;
+  clickOnSelectedDay: () => Promise<void>;
+  clickOnYearDropdown: () => Promise<void>;
+  clickOnMonthDropdown: () => Promise<void>;
+  clickOnNthYear: (n?: number) => Promise<void>;
+  clickOnPrevMonthButton: () => Promise<void>;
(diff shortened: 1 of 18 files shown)


src/Calendar/Calendar.uni.driver.d.ts
```diff
@@ -0,0 +1,43 @@
...
+  getFocusedDayElement: () => Promise<HTMLElement | null>;
+  pressLeftArrow: () => Promise<void>;
+  pressRightArrow: () => Promise<void>;
+  getSelectedDay: () => Promise<string>;
+  getWidth: () => Promise<CSSStyleDeclaration['width']>;
+  triggerKeyDown: (params: SyntheticEventData) => Promise<void>;
+  isFocusedDayVisuallyUnfocused: () => Promise<boolean>;
+  containsVisuallyUnfocusedDay: () => Promise<boolean>;
+  isTwoMonthsLayout: () => Promise<boolean>;
+  getMonthDropdownDriver: () => Promise<DropdownLayoutDriver>;
```
mosheKerbel: should be DropdownLayoutUniDriver

src/Calendar/Calendar.uni.driver.d.ts
```diff
@@ -0,0 +1,43 @@
...
+  getYearDropdownDriver: () => Promise<DropdownLayoutDriver>;
```
mosheKerbel: should be DropdownLayoutUniDriver

src/Calendar/index.d.ts
```diff
@@ -0,0 +1,51 @@
...
+  | 'pl'
+  | 'it'
+  | 'ru'
+  | 'ja'
+  | 'ko'
+  | 'tr'
+  | 'sv'
+  | 'no'
+  | 'nl'
+  | 'da';
```
mosheKerbel: missing these:
      'zh',
      'th',
      'cs',
https://github.com/wix/wix-style-react/blob/8d37e4d0dc991a90a784a874a375b08dfa2b24ab/src/Calendar/Calendar.js#L419

mosheKerbel:
We are currently not exporting any of the uniDrivers, but the types here are only for the *.uni and not for specific platforms drivers. Is this intended? @shlomitc

shlomitc:
@mosheKerbel - I think I talked with Gilad about creating only for the unidrivers, but actually, we should have for both if we want maximum value from it. @giladsegal - do you remember if this was the conclusion we had?

src/Calendar/Calendar.uni.driver.d.ts
giladsegal: I will check all of these tomorrow, I was OOO for a few days,
After I will finish, we can merge it

PR #3392, opened 2019-05-13
Nested sortable list - revert getDerivedStateFromProps - not works on wix-ui-adi

tzvikaf:
using componentWillReceiveProps instead of getDerivedStateFromProps
because it's not supported in wix-ui-adi


tzvikaf:
hi @argshook,
i know you wanted me to use the getDerivedStateFromProps,
but i guess it's not supported in the react version that used by wix-ui-adi,
the NestableList component doesn't work there with this function.

can you let it pass or is there an easy fix to allow it to work on wix-ui-adi?

tzvikaf:
never mind, i understand what happened,
in wix-style-react the props and state have the same items object,

in wix-ui-adi it's different and so the old props override the new state every time,
i don't like how fragile it is, i'll see what i can do on my side

argshook:
> never mind, i understand what happened,

so can this PR be closed then? urgent to bcc? :)

tzvikaf:
no... it's a difficult bug that i don't think worth the investigation,
how can i know why the props that the getDerivedStateFromProps gets in wix-style-react contains the same items object that the state has
but in wix-ui-adi it's not, 
looks like some code doing assigning from the state into the props
and that code doesn't work or exist in wix-ui-adi.

i don't think the fix is easy,
can you please merge the PR or at least give me a better alternative that isn't that doesn't include using getDerivedStateFromProps?

tzvikaf:
image
 i think this is what i experience,
need to make NestableList controlled

argshook:
i see, well that's unfortunate.
let's go with `componentWillReceiveProps` but please then remove the polyfill references and add a comment so that next developer know about the issue.

argshook:
thanks for the update!

tzvikaf:
thanks!

PR #3394, opened 2019-05-13
Add a puppeteer testkit bundle

ohana54:
### 🔦 Summary
TL;DR - This will allow Sled test writers to use the WSR puppeteer drivers in their tests.

When writing Sled tests, you can only use packages that exist in the lambda environment.
We want to create a bundle for the puppeteer drivers which we will upload to the lambda environment so it will be available during test execution.
We could do nothing, and upload the entire WSR module, but it's too big (around 400MB) and we are limited in space.

cc @tompere


testkit/webpack.config.js
```diff
@@ -0,0 +1,32 @@
...
+        use: {
+          loader: 'babel-loader',
+        },
+      },
+    ],
+  },
+  plugins: [
+    // Puppeteer drivers don't use styles, so we want to ignore them
+    // by replacing them with empty objects
+    new webpack.NormalModuleReplacementPlugin(
```
shlomitc: why is this needed? are there any drivers that currently do import scss files?

package.json
```diff
@@ -39,7 +39,8 @@
     "test:types": "tsc --noEmit -p tsconfig.declarations.json",
     "copy-public-styles": "cp src/*.scss .",
     "prebuild": "npm run generate-testkit-exports",
-    "build": "npm run lint && npm run proxy-wix-ui-icons && npm run transpile && npm run storybook:build",
+    "build": "npm run transpile && npm run storybook:build && npm run build-puppeteer-testkit",
```
shlomitc: maybe it can run in parallel to other tasks instead of sequential?

testkit/webpack.config.js
ohana54: Unfortunately yes - but not in the way you think (as you showed me, no driver is actually directly importing styles).
The problem is something like this:
1. puppeteer.js (the entry point) imports a **uni** button driver
2. The uni driver imports scss files somewhere in its dependency graph
So if I have this kind of import:
`import { buttonNextDriverFactory } from 'wix-ui-core/drivers/unidriver';`
Webpack will try to import **everything** exported by `wix-ui-core/drivers/unidriver`

Assuming we don't need styles, like we discussed, I thought the safest easiest thing is to just ignore these imports. I'm open to more suggestions :)

package.json
ohana54: Done

ohana54:
@yavorsky please let us know when we can have your input, everything else is done :)
Thanks!

package.json
```diff
@@ -39,7 +39,8 @@
     "test:types": "tsc --noEmit -p tsconfig.declarations.json",
     "copy-public-styles": "cp src/*.scss .",
     "prebuild": "npm run generate-testkit-exports",
-    "build": "npm run lint && npm run proxy-wix-ui-icons && npm run transpile && npm run storybook:build",
```
yavorsky: @shlomitc are we ok with removing `lint` from build script?

testkit/webpack.config.js
yavorsky:
```
So if I have this kind of import:
import { buttonNextDriverFactory } from 'wix-ui-core/drivers/unidriver';
Webpack will try to import everything exported by wix-ui-core/drivers/unidriver
```

It shouldn't with `sideEffects: false` in module's package.json. And if this module won't be transpiled with babel to commonjs modules.

testkit/webpack.config.js
```diff
@@ -0,0 +1,32 @@
...
+      /\.(s?css)$/,
+      path.join(__dirname, './fake-style-module.js'),
+    ),
+  ],
+  mode: 'development',
```
yavorsky: should it be `development` for e2e testing ?

testkit/webpack.config.js
```diff
@@ -0,0 +1,32 @@
...
+    path: path.resolve('./dist/testkit'),
+    filename: 'puppeteer-testkit-bundle.js',
+  },
+  module: {
+    rules: [
+      {
+        test: /\.js$/,
+        exclude: /node_modules/,
+        use: {
+          loader: 'babel-loader',
```
yavorsky: maybe we can use options here and pass `babel-preset-env` with `last 2 chrome versions` so it will be compiled faster just for latest chrome?

package.json
ohana54: Looks like a bad merge? I never intended to remove anything, I'll fix

testkit/webpack.config.js
ohana54: I'm using the same babel config as the rest of the app, I don't know if I can diverge. @shlomitc ?

testkit/webpack.config.js
ohana54: But it does :)

testkit/webpack.config.js
ohana54: It's the driver, so I personally prefer it won't be minified in case I need to step into.

testkit/webpack.config.js
shlomitc: sounds like an optimization, not really sure it is a must, so depends on you guys to decide

testkit/webpack.config.js
shlomitc: makes sense

testkit/webpack.config.js
shlomitc: @yavorsky - maybe the right thing will not be taking the drivers from the cherry-picked `wix-ui-core/drivers` entry? just take them from `dist/src`?

yavorsky: Yeah. It's not working b/c unidriver.js has commonjs modules.
We can:
- move it to src, so we can use both `dist/src/drivers` and `dist/es/src/drivers`.
-  update the script, to create `es` files in own directory: `import { buttonNextDriverFactory } from 'wix-ui-core/drivers/es/unidriver'`

testkit/webpack.config.js
ohana54: I prefer not to add the complication of a separate build configuration, the build output is exactly the same as other drivers we supply to consumers.

ohana54:
@yavorsky and @shlomitc thank you for the review and the insightful comments!

Issue #3401, opened 2019-05-14
<DropdownLayout /> Enable default marking of option


morsh:
# ✨ Feature Request

### 🔦 Scope

<DropdownLayout />
I want to expose a new property that would enable controlled marking an options

...

### 💁 Explanation

This feature is required to enable <InputWithOptions /> to mark the first selectable option when supplying a new set of options.

Part of the problem is that today, when clicking on `Enter` without hovering on any option simply closes the dropdown dialog without selecting any option.

...

### 🐾 Possible solution 

Expose a `defaultHoveredId` that would be used whenever options are changed (instead of state)

...

### 👀 Severity

- Major

argshook:
closed by https://github.com/wix/wix-style-react/pull/3409

PR #3409, opened 2019-05-15
DropdownLayout - enable mark option by default

morsh:
### 🔦 Summary
Enable providing an option `markedOption` with the following specifications:
- false [default]: no change
- true: If no option is marked, mark first selectable option (mark no select)
- number/string: If no option is marked, mark option with given id

Issue: https://github.com/wix/wix-style-react/issues/3401
...

### ✅ Checklist
- 📚 Change is documented
  - [X] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [X] Component
  - [ ] Visual test


morsh:
Approved by: Neil Osman

argshook:
@morsh hey, sorry im taking so long, was attending a conference and am sitting through workshop now :)

myself i'd probably go with a slightly different approach (instead of new prop, i would add new property to `option` object.

But i don't want to block, so this is good to merge imo.

Let me know if you have any thoughts, if not, i'll merge :)

morsh:
@argshook - I thought about a prop in option, but this would require synchronisation with marking functionality + potentially marking two options as marked.
This way, you can decide to mark the first option (our requirement) or any other option you decide (if you provide a number instead of `true`.

I'd be happy if you could merge :)
Thanks!

argshook:
> but this would require synchronisation with marking functionality + potentially marking two options as marked.

yes, and that's ok in my mind, because DropdownLayout could be used in other components as building block, where multiple makred options make sense (tags input, for example)

anyway, i don't want to block and we don't currently need additional functionality, so it's good to be merged. Thanks for contribution!

PR #3411, opened 2019-05-15
<CardGalleryItem/> - adjust titles to not be mandatory

mosheKerbel:
[According to spec]
(https://app.zeplin.io/project/5c8f6f1f9f489505cac756dc/screen/5cc97116201cf7684b654ffd) when titles and subtitles are not supplied, card should be without footer.

Without footer:
image
image

With footer:
image
image

@NitayRabi


src/CardGalleryItem/test/CardGalleryItem.visual.js
```diff
@@ -13,7 +13,7 @@ const tests = [
     describe: 'basic',
     its: [
       {
-        it: 'badge - none',
+        it: 'without badge',
```
shlomitc: renaming the name of it will create a new baseline, so I wouldn't recommend doing it in the same PR where you change some behavior.
such refactors should be done separately

mosheKerbel: yes i know, i generally follow this practice, but in this case i think it's "forgivable" because it's a new feature. this part was only merged yesterday and the behaviour is super expected.
WDYT?

shlomitc: ok

src/CardGalleryItem/CardGalleryItem.js
```diff
@@ -86,8 +86,15 @@ class CardGalleryItem extends React.Component {
       dataHook,
     } = this.props;
 
+    const footerExists = title || subtitle;
```
NitayRabi: I Would prefer using `!isEmpty(title) || !isEmpty(title)` from lodash to handle `' '`

src/CardGalleryItem/CardGalleryItem.uni.driver.js
```diff
@@ -2,16 +2,19 @@ import { baseUniDriverFactory } from 'wix-ui-test-utils/base-driver';
 
 const getTitle = base => base.$('[data-hook="title"]');
 const getBadge = base => base.$('[data-hook="badge"]');
+const getSubtitle = base => base.$('[data-hook="subtitle"]');
 const getPrimaryAction = base => base.$('[data-hook="primary-action"]');
 const getSecondaryAction = base => base.$('[data-hook="secondary-action"]');
 const getHoverComponent = base => base.$('[data-hook="hover-component"]');
```
NitayRabi: I don't know if this is the PR for this, but I would create a `DataHooks` file to contain all "magic strings" used within testing and in file. 
You can also add method `byDataHook` to not repeat `'[data-hook="${DataHook}"]'`

src/CardGalleryItem/test/CardGalleryItem.spec.js
```diff
@@ -18,12 +18,24 @@ describe('CardGalleryItem', () => {
...
     const driver = createDriver(<CardGalleryItem title="Card" />);
 
     expect(await driver.getTitle()).toBe('Card');
   });
 
+  it('should not render subtitle by default', async () => {
+    const driver = createDriver(<CardGalleryItem />);
+
+    expect(await driver.getSubtitle()).toBeNull();
+  });
```
NitayRabi: Missing some testing for passing falsey values to title + subtitle?

src/CardGalleryItem/CardGalleryItem.uni.driver.js
mosheKerbel: done

PR #3414, opened 2019-05-15
NestableList - add onDragStart and onDragEnd callbacks

tzvikaf:
added callback for when drag starts and ends,
pass dragged item component props as argument


tzvikaf:
it's a small useful update,
the main gripe is the test,
i couldn't separate simulation of starting and ending drag

PR #3415, opened 2019-05-15
<Carousel/> - refactor the component to use react-slick

nitayneeman:
This PR refactors the `<Carousel/>` component to use react-slick instead of implementing the behaviour internally.

Most of the code is directly taken from https://github.com/wix/wix-style-react/pull/3328.


src/Carousel/Carousel.global.scss
```diff
@@ -0,0 +1,27 @@
+@import "~slick-carousel/slick/slick.scss";
```
shlomitc: I think that this file is important in order to load the global styles, but not for overrides later (.slick-list and others). let's say there's another carousel with different styles, this will override all styles which is not a good thing.

Is it possible to create another file `Carousel.scss` and inside of it make the overrides?

```scss
.root {
  :global {
    .slick-list {
          ...
     }
  }
}
```

src/Carousel/Pagination/Pagination.global.scss
```diff
@@ -0,0 +1,10 @@
+@import "../../colors.scss";
```
shlomitc: maybe this file can also be not `.global`, as described in the previous comment

src/Carousel/Pagination/Pagination.scss
```diff
@@ -1,18 +1,22 @@
...
+    border-radius: 6px;
+    margin: 0 6px;
+
+     &:hover {
+      background-color: $B27;
+    }
 
-  &[data-active="true"] {
-    background-color: $B10;
+     & button {
```
shlomitc: what is this button?

src/Carousel/docs/index.story.js
```diff
@@ -17,21 +28,56 @@ const imagesExamples = [
...
+  sections: [
+    tab({
+      title: 'Usage',
+      sections: [
+        importExample({
+          source: "import Carousel from 'wix-style-react/Carousel';",
+        }),
+      ],
+    }),
+
```
shlomitc: you can add one simple example, just so the `Usage` won't be empty

PR #3420, opened 2019-05-16
fix forgotten static method

bt4R9:
@argshook, we forgot to turn getDerivedStateFromProps to the static method. So I changed it a bit, all tests are passed, check it please.


argshook:
indeed, it was forgotten :) Thanks!

PR #3424, opened 2019-05-16
<CardGalleryItem/> - change badge to be a node

mosheKerbel:
@shlomitc


src/CardGalleryItem/CardGalleryItem.uni.driver.js
```diff
@@ -9,8 +10,10 @@ const getHoverComponent = base => base.$('[data-hook="hover-component"]');
 const cardGalleryItemDriverFactory = base => ({
   ...baseUniDriverFactory(base),
   getTitle: () => getTitle(base).text(),
-  getBadgeText: async () =>
-    (await getBadge(base).exists()) ? getBadge(base).text() : null,
+  getBadge: async () =>
+    (await getBadge(base).exists())
+      ? ReactBase(getBadge(base)).prop('firstChild')
```
mosheKerbel: temporary
we already have solution in unidriver(https://github.com/wix-incubator/unidriver/pull/80)
just waiting for Gabi to release.

mosheKerbel:
@NitayRabi

NitayRabi:
Do we want to allow any kind of node rendering under badge prop? do we want to validate it's a badge? may cause some weird behaviour - @shlomitc - Opinions?

mosheKerbel:
> Do we want to allow any kind of node rendering under badge prop? do we want to validate it's a badge? may cause some weird behaviour - @shlomitc - Opinions?

I had some thoughts about it but i think it's ok to give the consumer the responsibility for it.
I though of adding a warning of the react props validation but ran into HOC issue becuase badge is a focusuble comopnent composite.
@shlomitc WDYT?

mosheKerbel:
@NitayRabi updating
discussed the issue with @shlomitc 
in order to "protect" the client we would ask for the badge props and then construct it ourself, this is the alternative.
In this case we are fine with giving this responsibility to the consumer to act right and pass a Badge node.

PR #3425, opened 2019-05-16
<Carousel/> - migrate e2e to visual grid

nitayneeman:
This PR creates a visual spec for `<Carousel/>`. This file contains a single test that just checks a basic rendering of images (instead of checking that in e2e).


nitayneeman:
Failed because of timeout of another component

PR #3430, opened 2019-05-19
<Carousel/> - support `children` prop

nitayneeman:
This PR adds the ability to render custom nodes instead of images (images are still supported):
image

Related to #3328.


nitayneeman:
Failed because of a timeout

PR #3437, opened 2019-05-20
<CardGalleryMenuItem/> - Add popover menu

mosheKerbel:


src/CardGalleryItem/CardGalleryItem.js
```diff
@@ -69,27 +75,45 @@ class CardGalleryItem extends React.Component {
...
-      badge,
-      backgroundImageUrl,
-      primaryActionProps,
-      secondaryActionProps,
-      className,
-      dataHook,
-    } = this.props;
-
-    const footerExists = title || subtitle;
+  _renderFooter(title, subtitle) {
```
mosheKerbel: this is just a refactor of extracting it out of the main render function

mosheKerbel:
@shlomitc @NitayRabi

src/CardGalleryItem/docs/ExampleBasic.js
```diff
@@ -1,23 +1,61 @@
 import React from 'react';
 
 import CardGalleryItem from 'wix-style-react/CardGalleryItem';
-import { Container, Row, Col } from 'wix-style-react/Grid';
+import { Col, Container, Row } from 'wix-style-react/Grid';
```
mosheKerbel: plz do not review the storybook examples yet, this is just a quick & dirty one so UX could review the component changes.

mosheKerbel:
I rebased to master and updated the icons, so now the badge looks fine in storybook.

mosheKerbel:
@mykas

src/CardGalleryItem/CardGalleryItem.js
```diff
@@ -69,27 +72,44 @@ class CardGalleryItem extends React.Component {
...
           [styles.hoveredContentWithFooter]: footerExists,
         })}
         data-hook={DataHooks.hoverContent}
       >
+        {settingsMenu && (
+          <div
+            className={styles.settingsMenuContainer}
+            data-hook={DataHooks.settingsMenu}
+          >
+            <SettingsMenu />
```
mykas: Hey, I think you can just do: `{settingsMenu}` here

src/CardGalleryItem/CardGalleryItem.uni.driver.js
```diff
@@ -13,30 +15,36 @@ const getSecondaryAction = base =>
...
+      return style.match(/url\("?([^"]*)"?\)/)[1];
+    },
+    click: () => getHoverComponent(base).click(),
+    getPrimaryActionLabel: () => getPrimaryAction(base).text(),
+    clickOnPrimaryAction: () => getPrimaryAction(base).click(),
+    getSecondaryActionLabel: () => getSecondaryAction(base).text(),
+    clickOnSecondaryAction: () => getSecondaryAction(base).click(),
+    getHoverComponent: () => getHoverComponent(base),
+    getHoveredContent: () => base.$(`[data-hook=${DataHooks.hoverContent}]`),
+    getSettingsMenu: async () =>
```
mykas: What does this method return?

src/CardGalleryItem/CardGalleryItem.js
```diff
@@ -69,27 +72,44 @@ class CardGalleryItem extends React.Component {
...
-    const hoveredContent = (
+    const SettingsMenu = settingsMenu;
+    return (
       <div
         className={classNames(styles.hoveredContent, {
           [styles.hoveredContentWithFooter]: footerExists,
         })}
         data-hook={DataHooks.hoverContent}
       >
+        {settingsMenu && (
```
mykas: I would suggest moving this to separate `renderSettingsMenu()` function too.

mosheKerbel: done

src/CardGalleryItem/CardGalleryItem.uni.driver.js
mosheKerbel: updated it
this is for the user to validate that his menu prop was passed succesfully.

src/CardGalleryItem/CardGalleryItem.js
mosheKerbel: done

src/CardGalleryItem/CardGalleryItem.uni.driver.js
```diff
@@ -13,30 +15,38 @@ const getSecondaryAction = base =>
...
+      return style.match(/url\("?([^"]*)"?\)/)[1];
+    },
+    click: () => getHoverComponent(base).click(),
+    getPrimaryActionLabel: () => getPrimaryAction(base).text(),
+    clickOnPrimaryAction: () => getPrimaryAction(base).click(),
+    getSecondaryActionLabel: () => getSecondaryAction(base).text(),
+    clickOnSecondaryAction: () => getSecondaryAction(base).click(),
+    getHoverComponent: () => getHoverComponent(base),
+    getHoveredContent: () => base.$(`[data-hook=${DataHooks.hoverContent}]`),
+    getSettingsMenu: async () => {
```
mykas: Regarding this. I don't think that user needs to test that he actually passed his PopoverMenu through our component. This part is only need to be tested on our side so it shouldn't be exposed to our users.

shlomitc: in case of render slots, I think it is a good option to return the element itself.
https://github.com/wix/wix-style-react/blob/master/docs/contribution/TEST_DRIVERS_GUIDELINES.md#render-slots

it also goes with the philosophy that whatever prop you give, you have a way to know the connection was done well

mykas: So instead of user passing custom data-hook on his component and searching for it with the (in this case PopoverMenu) testkit - we should expose the getter that does that for him?

shlomitc: user should pass dataHook anyway, as he's the one responsible for the rendering. that's also how he instantiates the testkit.

mykas: True, but I am just thinking whats the point in this case.
This:
```
const component = ...
const testkit = cardGalleryItemTestkitFactory({ wrapper:  component, dataHook: 'random' })
const popoverMenuTestkit = popoverMenuTestkitFactory(testkit.getSettingsMenu())
...
```
And this:
```
const component = ...
const popoverMenuTestkit = popoverMenuTestkitFactory({ wrapper: component, dataHook:'random-popovermenu'})
...
```

argshook: since we're dealing with composition, there's no reason to have dedicated testkit method for getting composed components, as that's already possible by using testkit in regular way (what Mykolas showed)

also, what we consider now `getSettingsMenu` might very well not make sense in future, when "settings menu" needs to support a checkbox or something. Then the name will be misleading, potentially requiring us to implement additional method and deprecate `getSettingsMenu`.

so i would try to avoid all that jazz and simply use separate testkits

mosheKerbel:
i think we are fine with merging this
guys? @mykas @shlomitc

src/CardGalleryItem/docs/index.story.js
```diff
@@ -62,15 +95,97 @@ export default {
...
+        title: 'Description',
+        sections: [
+          importExample(
+            "import CardGalleryItem from 'wix-style-react/CardGalleryItem';",
+          ),
+          divider(),
+          title('Examples'),
+          ...[
+            {
+              title: 'Menu on hover',
```
shlomitc: should show examples for both the IconButton and TextButton

shlomitc: add a description for the composition

mykas: Hey Shlomi, do you think that we should do the documentation on the same PR ? Because Moshe needs to align all the descriptions with UX. So maybe this can be done after he comes to Vilnius?

src/CardGalleryItem/CardGalleryItem.uni.driver.js
shlomitc: the render slot driver pattern was already well discussed, I don't think now is the time to reopen it. we can have a separate discussion about it. now let's keep following the guidelines.

regarding the `settingsMenu` prop naming - this widget is more opinionated than other containers such as `Card`, so in this case, I think this it is good to have a more semantic usage than "top-right" for example.

src/CardGalleryItem/docs/index.story.js
mosheKerbel: i added the descriptions and changed the appearance of the examples like the new UX (e.g ColorInput).
For more ux story concerns we will do it in a separate effort.

shlomitc: UX stories are on the lowest priority at the moment for this PR.

As we discussed before - we create a story per component, which focuses on APIs and common usages.
UX story is an additional story, which its product is the UX, so they should be done together with them

mosheKerbel:
All issues disuccssed were handled
Teamcity fails because of an unrelated test timeout

PR #3440, opened 2019-05-20
Storybook - fix broken components

nitayneeman:
It appears that about five hours ago, webpack released a new version (v4.32.0) which bundles and minifies our storybook static files in a way that makes React to trigger the following error:
image


nitayneeman:
It has already passed the checks, so I merge it

nitayneeman:
Hey @yavorsky, maybe do you have any clue why the latest version of webpack (https://github.com/webpack/webpack/releases/tag/v4.32.0) leads to the issue above?

yavorsky:
Interesting. It could be related to `Stack Trace for Errors in rejected Promises from import() now includes the stack from where the import() was called` since Tooltip is using dynamic imports.

I'll check the webpack codebase update a bit later. 

Thanks!

PR #3446, opened 2019-05-21
Nested sortable list - adding classes props

tzvikaf:
enabling passing of react classes to control the styling of inner nestable-list components
top level container
container
item


src/DragAndDrop/docs/SortableList/NestableListExample/NestableListExample.js
```diff
@@ -67,6 +81,9 @@ export default class NestableListExample extends React.Component {
         <div className={styles.root}>
           <h3 className={styles.title}>Draggable Area</h3>
           <NestableList
+            itemClass={styles.item}
```
Faradey27: what do you think about having `theme: {class1, class2}` prop?

Faradey27: also do we really need such flexibility? maybe we can just have several predefined skins, like in other components?

tzvikaf: i'm not a ux or presume to know the future, so i'm not sure about predefined skins.
but i do think that maybe settings also constant classes is good.

theme looks better, i'll update

PR #3449, opened 2019-05-21
<Input/> - added `warning` status

malachico:
### 🔦 Summary
...

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test

added warning state to input component. 
image


NitayRabi:
- Regarding code - is AmaterialTheme still needed - @shlomitc ? beyond that seems good, in general (not for this scope) we should use `Input.Affix` compositions to achieve this behaviours (both for warning, and error).
- Regarding UX - you based the design on [this](https://zpl.io/2pBr6Lj) spec? was the colors for hover, focus approved by design system? @LittleTal @milkyfruit

shlomitc:
Amaterial theme not needed
looks like it's not matching the UX specs, the icon should definitely be different
https://zpl.io/2pBr6Lj
image

src/Input/Input.js
```diff
@@ -469,7 +470,11 @@ Input.propTypes = {
   disabled: PropTypes.bool,
 
   /** Input status - use to display an status indication for the user. for example: 'error' or 'loading' */
-  status: PropTypes.oneOf([Input.StatusError, Input.StatusLoading]),
+  status: PropTypes.oneOf([
```
NitayRabi: Please update the comment to reflect the `warning` option in the docs.

src/Input/ThemedInputWarningSuffix.js
```diff
@@ -0,0 +1,89 @@
...
+import styles from './Input.scss';
+
+const placementToMoveBy = {
+  right: { x: 2, y: -10 },
+  left: { x: -2, y: -10 },
+  top: { x: 0, y: -10 },
+  bottom: { x: 0, y: -8 },
+};
+
+const AmaterialWarningSuffix = ({
```
NitayRabi: `AmaterialWarningSuffix` is no longer relevant

LittleTal:
As @shlomitc mentioned, the warning icon is not aligned:

image

Should be:

image

malachico:
hi @LittleTal and @shlomitc thanks a lot for the review. 
I couldn't find a triangle alert icon,
as I understood from Jasmin Adam, it should be added to WSR, and then we will switch the circle to triangle.
can we use for now with the circle one until the triangle will be added and then replace it? I don't want this to delay our development. 
thanks.

LittleTal:
Here is the icon: 
https://wix-wix-ui-icons-common.surge.sh/?selectedKind=Icons%2FSystem&selectedStory=Normal&full=0&addons=0&stories=1&panelRight=0

malachico:
Hi, thanks a lot for the reviews, I've fixed your comments :)
image

PR #3452, opened 2019-05-22
<InputWithOptions/> - prevent page refresh on Enter press

gilmeir:
https://github.com/wix/wix-style-react/issues/2657


src/InputWithOptions/docs/tests/TestInsideForm.js
```diff
@@ -0,0 +1,59 @@
...
+    };
+
+    const onSelect = option => {
+      const value = option.value;
+      this.setState({
+        value,
+        selectedId: option.id,
+      });
+
+      console.log(
```
nitayneeman: Remove this line

src/InputWithOptions/docs/tests/TestInsideForm.js
```diff
@@ -0,0 +1,59 @@
...
+        `Selected option id=${JSON.stringify(option)}, value=${value}`,
+      );
+    };
+
+    const onManuallyInput = value => {
+      this.setState({
+        selectedId: -1,
+      });
+      console.log(`Manually selected ${value}`);
```
nitayneeman: Remove this line

src/InputWithOptions/InputWithOptions.private.protractor.driver.js
```diff
@@ -0,0 +1,23 @@
+import { dropdownSelector } from './InputWithOptions.protractor.driver';
+import dropdownLayoutDriverFactory from '../DropdownLayout/DropdownLayout.protractor.driver';
+//import inputDriverFactory from '../Input/Input.protractor.driver';
+
+export default component => {
+  const dropdown = component.$(dropdownSelector);
+  const dropdownLayoutDriver = dropdownLayoutDriverFactory(dropdown);
+
+  const input = component.$('input');
+  //const inputDriver = inputDriverFactory({ element: input });
```
nitayneeman: Good practice, it should work :)

src/InputWithOptions/InputWithOptions.e2e.js
```diff
@@ -2,10 +2,17 @@ import { inputWithOptionsTestkitFactory } from '../../testkit/protractor';
...
+  storySettings,
+  insideFormStorySettings,
+  testStories,
+} from './docs/storySettings';
+import privateDriverFactory from './InputWithOptions.private.protractor.driver';
+import { eventually } from '../../test/utils/unit/eventually';
 
 describe('InputWithOptions', () => {
   let driver;
+  let privateDriver;
```
nitayneeman: Actually the private driver is enough, I'd change `driver` to be the private one

src/MultiSelect/docs/tests/TestTabsSwitches.js
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable no-console */
 import React from 'react';
-import MultiSelect from 'wix-style-react/MultiSelect';
+import MultiSelect from '../../MultiSelect';
```
shlomitc: revert

src/MultiSelect/MultiSelect.e2e.js
```diff
@@ -192,6 +192,7 @@ describe('MultiSelect - Focus behaviour', () => {
 
   async function focusOnMultiSelect() {
     const firstElement = $(`[data-hook="input-for-focus-1"]`);
+
```
shlomitc: revert

src/MultiSelect/MultiSelect.e2e.js
```diff
@@ -204,6 +205,7 @@ describe('MultiSelect - Focus behaviour', () => {
 
     await driver.click();
     await driver.hoverItemById('AL');
+
```
shlomitc: revert

src/InputWithOptions/docs/tests/TestInsideWrapperForm.js
```diff
@@ -0,0 +1,54 @@
+/* eslint-disable no-console */
```
shlomitc: I would suggest for this test, not to wait for browser refresh (which can give us false-negatives), but rather check the dom.
for example - when onSubmit is called - do some change the dom.
in the test itself - wait that the dom indicator was not changed after the right amount time

src/InputWithOptions/InputWithOptions.protractor.driver.js
```diff
@@ -1,21 +1,24 @@
 import dropdownLayoutDriverFactory from '../DropdownLayout/DropdownLayout.protractor.driver';
```
shlomitc: 1. spreading the input driver is not the best practice, so I wouldn't encourage it.
2. changing the driver to use the `InputDriver` is correct, but just make sure it doesn't break other tests

src/InputWithOptions/InputWithOptions.protractor.driver.js
```diff
@@ -1,21 +1,24 @@
...
+import inputDriverFactory from '../Input/Input.protractor.driver';
 
 const driverFactory = component => {
   const dropdownLayoutDriver = dropdownLayoutDriverFactory(
-    component.$('[data-hook="dropdown-layout-wrapper"]'),
+    component.$(`[data-hook="dropdown-layout-wrapper"]`),
+  );
+
+  const inputDriver = inputDriverFactory(
+    component.$(`[data-hook="input-wrapper"]`),
```
shlomitc: not sure this is needed at all. since `InputWithOptions` renders `<Input/>` on it's own (like 371 in InputWithOptions), it might make sense that it will render its own data-hook. this might not be trivial, because `Input` kind of overrides it `wsr-custom-input`, but in that case, maybe it's better to look for this hook instead

src/InputWithOptions/InputWithOptions.js
```diff
@@ -171,7 +171,11 @@ class InputWithOptions extends WixComponent {
     return (
       <div>
         {dropDirectionUp ? this._renderDropdownLayout() : null}
-        <div data-input-parent className={this.inputClasses()}>
+        <div
+          data-input-parent
+          className={this.inputClasses()}
+          data-hook="input-wrapper"
```
shlomitc: looks redundant

src/InputWithOptions/InputWithOptions.private.protractor.driver.js
```diff
@@ -0,0 +1,16 @@
...
+import inputDriverFactory from '../Input/Input.protractor.driver';
+
+export default component => {
+  const dropdownLayoutDriver = dropdownLayoutDriverFactory(component);
+  const inputDriver = inputDriverFactory(component);
+
+  const pressEnter = () => inputDriver.element().sendKeys(protractor.Key.ENTER);
+
+  return {
+    ...dropdownLayoutDriver,
```
shlomitc: Again, I wouldn't expose these drivers and spread them, expose only the function that you need

src/InputWithOptions/InputWithOptions.e2e.js
```diff
@@ -1,13 +1,21 @@
-import { inputWithOptionsTestkitFactory } from '../../testkit/protractor';
```
shlomitc: try to simplify the changes in the file. we can do the refactors later

src/Input/Input.js
```diff
@@ -405,7 +405,7 @@ class Input extends Component {
     const { customInput: CustomInputComponent, dataRef, ...rest } = props;
     const inputProps = {
       ...(CustomInputComponent
-        ? { 'data-ref': dataRef, ...rest, 'data-hook': 'wsr-custom-input' }
+        ? { 'data-ref': dataRef, ...rest, 'data-hook': 'wsr-input-custom' }
```
shlomitc: this might break other tests

src/Input/Input.js
```diff
@@ -231,7 +231,7 @@ class Input extends Component {
     });
 
     return (
-      <div className={styles.inputWrapper}>
+      <div className={styles.inputWrapper} data-hook="input-wrapper">
```
shlomitc: this is not needed. we usually put `data-hook`s on the top wrapper as it comes from `dataHook` prop.
in this case, it's a long time mistake, where the data-hook is propagated in the `renderInput()` function and set directly to the native `<input/>`.
so in that case, I think it's better no to try to wrap it now.

src/InputWithOptions/InputWithOptions.private.protractor.driver.js
gilmeir: Sure.
I've changed it so the private driver spreads the public driver (so we won't need to init both) + add the private functions to private driver

PR #3453, opened 2019-05-22
test(storybook): disable font smoothing

argshook:
temporary measure to check if it affects applitools screenshots


argshook:
no longer relevant, closing

PR #3454, opened 2019-05-22
omit getDerivedStateFromProps for the Search component

bt4R9:
@argshook hi! Let's discuss this change.
It seems like we don't need it because we reset the `value` in `componentDidUpdate`. What do you think?


argshook:
@bt4R9 thanks for taking care of this! i update a little to remove polyfill entirely.

PR #3455, opened 2019-05-22
Nested sortable list - pass droppedItem to the onUpdate callback

tzvikaf:
adding droppedItem parameter to the onUpdate function,

it's an object that has relevant data of the droppedItem,
main intent was for bi but can help for other stuff,

the object contains the data of the item, it's new position in the tree and depth,
there is a bit more data that might be relevant in the future


src/NestableList/NestableList.js
```diff
@@ -93,8 +93,8 @@ class NestableList extends WixComponent {
     return realNextPosition;
   };
 
-  dropItem = () => {
-    this.props.onUpdate && this.props.onUpdate(this.state.items);
+  dropItem = item => {
+    this.props.onUpdate && this.props.onUpdate(this.state.items, item);
```
argshook: does `onUpdate` really need all items to be passed?
i'd like the api to be better thought out, cause we are not able to do breaking changes easily.

if it is necesary to return `this.state.items` and also `item` in `onUpdate`, it should be an object imo:
`this.props.onUpdate({ items: this.state.items, item })`
this way we are free to add more stuff in future and not rely on
positional arguments

tzvikaf: true

shlomitc:
@argshook - what's the status of this PR?

argshook:
@shlomitc most of the code was moved to wix-ui-adi and is not really relevant for wix-style-react anymore.

ideally, as we spoke, this should be part of wix-ui-core.

closing.

tzvikaf:
it was moved to adi-web, 
i'll need to to put in a common project because the component may be used also in the responsive editor project

Issue #2368, opened 2018-10-24
Feature Request - Disabled State


LittleTal:
## The request
Disabled state in all components should be changed according to a new design.

Motivation: currently disabled state colors are not accessible- the contrast ratio is very low.

New disabled colors mapping:
https://docs.google.com/spreadsheets/d/1_Q2-zui1EwPqQ3JzE0zgevmIfTbEIbojgsbbz8h5aF8/edit?usp=sharing

Note that new colors are not Hex colors, it's `D10` and `D80` with an alpha channel.
It works well on a colored background. 

Example of components with new disables colors in Zeplin:
https://zpl.io/VQ994An

@shlomitc What do you think?

shlomitc:
mostly done

PR #3461, opened 2019-05-23
Foundation - add stylable stylesheet files for colors, opacity, spacing, shadows etc.

mykas:
This PR is two initiatives:
- Move all wix-ui-backoffice consumer stylable files to use styles from wix-style-react
- Update the foundation according to specs.


mykas:
Maybe we can map this issue with this pr: https://github.com/wix/wix-style-react/issues/2368 @shlomitc @gilmeir what do you think?

mykas:
I have added a javascript based stylable formatter(https://stylable.io/docs/references/formatters) in order enable us to move our components disable state with color + alpha channel. The function: `calc_rgba` accepts two arguments hex and alpha channel value to achieve desire color with alpha channel.

PR #3466, opened 2019-05-24
chore(Page.Deprecated): remove e2e and examples

argshook:
This component is deprecated since v6.3.0 (more than 3 months)

component code is not removed, only documentation and e2e
e2e is often timing out and takes ~2 minutes, removing this is a low hanging fruit in improving e2e time


shlomitc:
Although 3 months past, we don’t know if our consumers did migrate.

Ask, I think that the safest thing to do is get ready for the next major version and only then remove old tests.

Another option - migrate the new page to the visual grid and save the time there

argshook:
> Although 3 months past, we don’t know if our consumers did migrate.

removing tests has nothing do to with users though, if the test is problematic i don't see why we should hang on to it, especially when component is deprecated.

Page and Page.Deprecated are ones taking longest e2e time, it very often spans over 5 minutes (twice). 45 mins is our CI limit, we're reaching it quite often already

shlomitc:
We write tests in order to have more confidence. The fact that it takes a lot of time is an issue we must solve, not remove tests.

In my opinion, rewriting the updated Page to the visual testing will do the same effect of reducing the build time, only we don't compromise on removing the tests.

argshook:
do you suggest to rewrite e2e test for deprecated component?

shlomitc:
For the non deprecated component

argshook:
closing, maybe we'll get to it next time

Issue #3365, opened 2019-05-07
`<FormField/>` - label text overlaps bounding box when text is too long


mykas:
# 🐛 Bug Report

### 🏗 Relevant Components

`<FormField/>`

### 😯 Current Behavior

When `FormField`s label is too long and this happens on different languages then it will overlap the FormFIelds boundaries.
![image](https://user-images.githubusercontent.com/8919190/57283570-20e9d700-70b8-11e9-85ec-ff12990f16c8.png)

### 🤔 Expected Behavior

FormField label should show ellipsis when space is not enough for label to show the full label:\

![image](https://user-images.githubusercontent.com/8919190/57283766-8dfd6c80-70b8-11e9-8f7a-1a4fda7b9b87.png)

### 👣 Steps to Reproduce

Use FormField label with label position from the left side and add very long label text.

### 👀 Severity

Major

PR #3467, opened 2019-05-26
<FormField/> - Fix ellipsis

gilmeir:
#3365 

<Text/> adds the required features:
* ellipsis
* tooltip

### DIFF
--- a/src/FormField/FormField.js
+++ b/src/FormField/FormField.js
@@ -194,7 +194,7 @@ class FormField extends React.Component {
           >
             <Text
               size={labelSize}
-              for={id}
+              htmlFor={id}
               tagName={'label'}
               ellipsis
               data-hook="formfield-label"
--- a/src/FormField/FormField.js
+++ b/src/FormField/FormField.js
@@ -192,13 +192,15 @@ class FormField extends React.Component {
               [styles.minLabelHeight]: !children,
             })}
           >
-            <Label
-              appearance="T1"
+            <Text
               size={labelSize}
-              children={label}
               for={id}
+              tagName={'label'}
+              ellipsis
               data-hook="formfield-label"
-            />
+            >
+              {label}
+            </Text>
 
             {required && asterisk}
             {this._renderInfoIcon()}
--- a/src/FormField/FormField.js
+++ b/src/FormField/FormField.js
@@ -14,6 +14,11 @@ const labelPlacements = {
   left: 'left',
 };
 
+const sizeToPixels = {
+  small: 18,
+  medium: 24,
+};
+
 const asterisk = (
   <div
     data-hook="formfield-asterisk"
@@ -116,10 +121,13 @@ class FormField extends React.Component {
   _hasCharCounter = () => typeof this.state.lengthLeft === 'number';
 
   _renderInfoIcon = () => {
-    const { infoContent, infoTooltipProps } = this.props;
+    const { infoContent, infoTooltipProps, labelSize } = this.props;
+    const size = sizeToPixels[labelSize];
+
     return (
       infoContent && (
         <InfoIcon
+          size={size}
           dataHook="formfield-infotooltip"
           className={styles.infoIcon}
           tooltipProps={{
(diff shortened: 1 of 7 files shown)


src/FormField/FormField.scss
```diff
@@ -46,6 +46,7 @@ $childrenMargin: 6px;
     display: flex;
     align-items: center;
     flex: 1 0 auto;
+    width: 100%;
```
nitayneeman: I think that `max-width` would be better here

src/FormField/FormField.visual.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import FormField from './FormField';
+import Input from '../Input';
+
+storiesOf(`FormField`, module).add('Label ellipsis', () => (
```
nitayneeman: I would change the structure of this file according to our convention, like:
```javascript
const tests = [
  {
    describe: 'Something',
    its: [
      {
        it: 'Something',
        props: {
        },
      },
    ],
  },
]

  its.forEach(({ it, props }) => {
    storiesOf(`FormField/${describe}`, module).add(it, () => (
      <FormField {...props} />
    ));
  });
});
```

src/FormField/FormField.js
```diff
@@ -159,6 +165,22 @@ class FormField extends React.Component {
...
+  _renderLabel = props => {
+    const { label, labelSize, id } = this.props;
+
+    return (
+      <Text
+        size={labelSize}
+        htmlFor={id}
+        tagName={'label'}
+        data-hook="formfield-label"
+        {...props}
```
shlomitc: why?

src/FormField/FormField.scss
```diff
@@ -69,6 +127,16 @@ $childrenMargin: 6px;
   }
 }
 
+.root.small .label * {
```
shlomitc: we never style with `*`. if we must do it, we need to target the direct child.
I'm not a CSS expert, but maybe it's better to apply the styles to the specific elements using a className

src/FormField/FormField.scss
```diff
@@ -31,25 +53,61 @@ $childrenMargin: 6px;
...
+  align-self: center;
+}
+
+.label {
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-start;
+  overflow: hidden;
+
+  & * {
```
shlomitc: same comment for `*`

src/FormField/FormField.scss
```diff
@@ -31,25 +53,61 @@ $childrenMargin: 6px;
...
+.labelWrapper {
+  display: flex;
+  align-self: center;
+}
+
+.label {
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-start;
+  overflow: hidden;
```
shlomitc: why was this needed?

src/FormField/FormField.scss
```diff
@@ -31,25 +53,61 @@ $childrenMargin: 6px;
...
+
+.medium .minLabelHeight {
+  line-height: $mediumSize;
+  min-height: $mediumSize;
+}
+
 .minLabelHeight {
-  min-height: $inlineLabelMinHeight;
-  line-height: $inlineLabelMinHeight;
+  overflow: hidden;
```
shlomitc: why is this needed? it should be solved with `Text` applying ellipsis

src/FormField/FormField.js
```diff
@@ -159,6 +165,22 @@ class FormField extends React.Component {
     (labelPlacement === labelPlacements.left ||
       labelPlacement === labelPlacements.right);
 
+  _renderLabel = props => {
```
shlomitc: why all props are needed?

src/FormField/FormField.js
```diff
@@ -173,46 +195,42 @@ class FormField extends React.Component {
...
-            <Label
-              appearance="T1"
-              size={labelSize}
-              children={label}
-              for={id}
-              data-hook="formfield-label"
-            />
-
+            {this._renderLabel({
+              ellipsis: true,
```
shlomitc: 1. what is the case for not applying ellipsis?
2. better to pronounce it as a business case. for example, if it's an inline label concern, then it should pass this as a flag and internally in the function, apply the ellipsis

src/FormField/FormField.visual.js
```diff
@@ -0,0 +1,39 @@
...
+  );
+};
+
+const test = (it, props) => ({ it, props });
+
+const tests = [
+  {
+    describe: 'Label',
+    its: [
+      test('Ellipsis', {
```
shlomitc: test names should affect what you see.
this means - it should show ellipsis in case of long text

src/FormField/FormField.visual.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import FormFieldComponent from './FormField';
```
shlomitc: this should be named FormField

src/FormField/FormField.visual.js
```diff
@@ -0,0 +1,39 @@
...
+import Input from '../Input';
+
+const FormField = props => {
```
shlomitc: this is a `FormFieldWrapper`

src/FormField/FormField.visual.js
```diff
@@ -0,0 +1,39 @@
...
+  return (
+    <div style={{ width: 285, padding: 50 }}>
+      <FormFieldComponent {...props}>
+        <Input />
```
shlomitc: this is only one use case. so if the wrapper is general, it should not contain the children.
`<Input/>` is a test configuration, meaning it should come from the props of a give test

yaarams:
@gilmeir please update https://jira.wixpress.com/browse/WCD-8479
when this is ready.

shlomitc:
fixed in this PR https://github.com/wix/wix-style-react/pull/3506

PR #3476, opened 2019-05-27
use wix-ui-framework to generate enzyme & vanilla testkit exports

argshook:
this is internal change that should not affect users

---

migrate `scripts/generate-testkit-exports` to `wuf export-testkits`

1. remove `scripts/generate-testkit-exports`
2. use `wuf update`
3. use `wuf export-testkits`

---

new `.wuf` folder appeared:

```sh
.wuf
├── required-component-files.json # folder shape of component 
├── components.json # list of all components that adhere to the shape (created with `wuf update`)
├── generator # previously folder was at root, now under .wuf 
│   ├── codemods # same as before
│   └── templates # same as before
└── testkits # collection of testkit exporting related files 
    ├── definitions.js # former testkits/testkit-definitions.js
    ├── enzyme.template.js # former scripts/generate-testkit-exports/templates/enzyme.js
    └── vanilla.template.js # former scripts/generate-testkit-exports/templates/vanilla.js
```

---

`npm run generate-testkit-exports` used to run `node scripts/generate-testkit-exports`. This script was tailored for wix-style-react only. We're moving now to generic solution, `wuf export-teskits`

`npm run generate-testkit-exports` now runs:
  1. `npm run generate-components-list` (`wuf update` with flags)
  2. `npm run generate-vanilla-testkit-exports` (`wuf export-testkits` with flags for vanilla)
  3. `npm run generate-enzyme-testkit-exports` (`wuf exports-testkits` with flags for enzyme)

`npm run generate-components-list` runs `wuf update`.
This creates `.wuf/components.json` with all components that are similar to the shape defined in `.wuf/required-component-files.json`

once `.wuf/components.json` is created, it is possible to generate testkit exports for them.

`npm run generate-vanilla-testkit-exports` and `npm run generate-enzyme-testkit-exports` both run `wuf export-testkits` but with different flags (for enzyme and vanilla testkits respectively)

this is a step towards having only `wuf update` do all the work


scripts/generate-testkit-exports/generate.js
```diff
@@ -2,7 +2,7 @@ const path = require('path');
 const fs = require('fs');
 
 const listAllComponents = require('./list-all-components');
-const TESTKIT_DEFINITIONS = require('../../testkit/testkit-definitions');
+const TESTKIT_DEFINITIONS = require('../../.wuf/testkits/definitions.js');
```
rokija: `const TESTKIT_DEFINITIONS = require('../../.wuf/testkits/definitions');` 😄

argshook: you're absolutely right, thanks! fixed

argshook:
fyi, this is postponed until Friday

PR #3477, opened 2019-05-28
Add no color tooltip to <Swatches />

MSiliunas:
### 🔦 Summary
Add tooltip to <Swatches /> as in spec.

### ✅ Checklist
- 📚 Change is documented
  - [x] Story
  - [x] API description


src/Swatches/Swatches.js
```diff
@@ -53,12 +62,16 @@ Swatches.propTypes = {
 
   /** If true shows no color option */
   showClear: bool,
+
+  /** If showClear is true than this message appears on hover */
```
argshook:
```suggestion
  /** optional message to display in tooltip when showClear is true */
```

src/Swatches/test/Swatches.spec.js
```diff
@@ -73,5 +73,20 @@ describe('Swatches', () => {
 
       expect(await driver.isSwatchSelectedAt(0)).toBe(true);
     });
+
+    it('should add tooltip', async () => {
+      const { driver } = render(<Swatches showClear colors={['#000000']} />);
+
+      expect(await driver.hasTooltip()).toBe(true);
```
argshook: @MSiliunas  what is the intent of this test? it says `should add tooltip` but when there is no `showClearMessage`, tooltip shouldn't be added?

MSiliunas: @argshook It checks if the tooltip is added when `showClear` is true. It has default message.

argshook: oh, ok. probably would be a good idea to assert default message too

MSiliunas: done 👍

src/Swatches/Swatch.js
```diff
@@ -1,10 +1,25 @@
 import React from 'react';
 import { withFocusable } from 'wix-ui-core/dist/src/hocs/Focusable/FocusableHOC';
 import styles from './Swatches.st.css';
+import Tooltip from '../Tooltip';
 
 class Swatch extends React.PureComponent {
   static propTypes = {};
 
+  addPopover = children => {
```
mykas: I guess you change this to addTooltip too.

src/Swatches/Swatch.js
```diff
@@ -29,6 +44,7 @@ class Swatch extends React.PureComponent {
         style={{ backgroundColor: color }}
       />
     );
+    return color ? button : this.addPopover(button);
```
mykas: Same is here: addTooltip.

src/Swatches/Swatches.js
```diff
@@ -53,12 +62,16 @@ Swatches.propTypes = {
 
   /** If true shows no color option */
   showClear: bool,
+
+  /** optional message to display in tooltip when showClear is true */
+  showClearMessage: string,
```
mykas: The default value for this should be empty string and the proptype should be node because user might want to provide translation node here.

src/Swatches/Swatches.js
```diff
@@ -53,12 +62,16 @@ Swatches.propTypes = {
...
 };
 
 Swatches.defaultProps = {
   colors: [],
   size: 'small',
   selected: '',
+  showClearMessage: 'No Color',
```
mykas: Empty string. No business logic for our components.

PR #3480, opened 2019-05-28
<FormField/> - Added a method to driver that checks if info tooltip exists

zzgab:
### 🔦 Summary
Added a `hasTooltip()` method in FormField testkit, because when the `infoContent` prop is empty there is no (i) icon and tooltip, and `getInfoContent()` failed.

### ✅ Checklist
- 🔬 Change is tested
  - [X] Component


src/FormField/FormField.spec.js
```diff
@@ -110,6 +110,11 @@ describe('FormField', () => {
           const { driver } = render(renderFormField(props));
           expect(await driver.getInfoContent()).toBe(infoContent);
         });
+
```
mosheKerbel: Missing a positive scenario - "should have tooltip..."

zzgab: Pushed.

PR #3489, opened 2019-05-30
`<ListItemAction/>` - add missing paddingSize prop

mykas:
This PR is adding missing passingSize prop that changes based on text size.


src/ListItemAction/ListItemAction.js
```diff
@@ -4,6 +4,11 @@ import styles from './ListItemAction.st.css';
 import Text from '../Text';
 import Box from '../Box';
 
+const textSizetoPaddingMap = {
```
mosheKerbel: textSizeToPaddingMap

PR #3493, opened 2019-05-30
<InputWithOptions/> - prevent submit event default behaviour on Enter key

gilmeir:
#2657


src/Layout/docs/examples/Form.js
```diff
@@ -9,9 +9,49 @@ import Checkbox from 'wix-style-react/Checkbox';
 import Text from 'wix-style-react/Text';
 import Button from 'wix-style-react/Button';
 import RadioGroup from 'wix-style-react/RadioGroup';
-
+import MultiSelect from 'wix-style-react/MultiSelect';
 import styles from '../styles.scss';
 
+class MultiSelectForTest extends React.Component {
```
shlomitc: why did you need to add this?

src/Input/Input.js
```diff
@@ -405,7 +405,7 @@ class Input extends Component {
     const { customInput: CustomInputComponent, dataRef, ...rest } = props;
     const inputProps = {
       ...(CustomInputComponent
-        ? { 'data-ref': dataRef, ...rest, 'data-hook': 'wsr-custom-input' }
+        ? { 'data-ref': dataRef, ...rest, 'data-hook': 'wsr-input-custom' }
```
shlomitc: why is this change necessary?

src/Layout/docs/examples/Form.js
gilmeir: I wanted this use case (input with options inside form) to be available in a storybook example, so one can easily try out hitting Enter easily.

Are there any constraints about changing this example?

gilmeir: I wanted to add this use case to the Form example in storybook, so if anyone wants to manually test the behaviour of InputWithOptions inside a form they can open this example.

src/Input/Input.js
gilmeir: This allows us to find both custom and not custom Input elements in the protractor inputDriver:
https://github.com/wix/wix-style-react/blob/inputwoptions-enter-inside-form/src/Input/Input.protractor.driver.js#L4

Both data-hooks are defined here in Input:
1. https://github.com/wix/wix-style-react/blob/inputwoptions-enter-inside-form/src/Input/Input.js#L200
2. https://github.com/wix/wix-style-react/blob/inputwoptions-enter-inside-form/src/Input/Input.js#L408

src/InputWithOptions/InputWithOptions.protractor.driver.js
```diff
@@ -1,21 +1,24 @@
...
-    isFocused: () => isFocused(input),
+    click: () => inputDriver.click(),
+    getInput: () => inputDriver.element(),
+    getInputText: () => inputDriver.getText(),
+    isFocused: () => inputDriver.isFocused(),
     element: () => component,
     /** Check wether the options dropdown is open */
     isOptionsShown: () => dropdownLayoutDriver.getDropdown().isDisplayed(),
-    enterText: text => input.clear().sendKeys(text),
+    enterText: text => inputDriver.enterText(text),
```
shlomitc: might be a breaking change, let's revery this line

src/InputWithOptions/InputWithOptions.js
```diff
@@ -171,7 +171,11 @@ class InputWithOptions extends WixComponent {
     return (
       <div>
         {dropDirectionUp ? this._renderDropdownLayout() : null}
-        <div data-input-parent className={this.inputClasses()}>
+        <div
+          data-input-parent
+          className={this.inputClasses()}
+          data-hook="input-with-options--input"
```
shlomitc: this doesn't looks like the right place to put the data-hook.
if you want to use the `inputDriver`, then you should mark the `<Input/>` itself with this data hook

src/Layout/docs/examples/Form.js
shlomitc: it adds complexity that isn't needed for this example, let's remove it

src/Input/Input.protractor.driver.js
```diff
@@ -1,8 +1,9 @@
 import { isFocused } from 'wix-ui-test-utils/protractor';
 
 const inputDriverFactory = component => {
-  const input = component.$('[data-hook="wsr-input"]'),
-    clearButton = component.$('[data-hook="input-clear-button"]');
+  // selector doesn't use data-hook since Input consumer can provide his own customInput
+  const input = component.$('input');
```
shlomitc: see the explanation regarding `input-with-options--input`

src/Input/Input.js
gilmeir: reverting this change

PR #3498, opened 2019-05-31
<DropdownBase/> - expose flip&fixed params and pass it down to popover

mosheKerbel:
@mykas


mosheKerbel:
Merged
PR was red because of an unrelated async timeout test

PR #3503, opened 2019-05-31
Dependency: update `husky` && husky hook config

sidoruk-sv:
### 🔦 Summary
Update husky to v2.3.0 and migrate husky hook config to latest format


shlomitc:
Many thanks!

PR #3502, opened 2019-05-31
Dependency: update `jscodeshift` (now doesn't use deprecated `nomnom` package)

sidoruk-sv:
Due to `nomnom` deprecation, I get such a warning on my project install:

> npm WARN deprecated nomnom@1.8.1: Package no longer supported. Contact support@npmjs.com for more info.

`nomnom` is coming from `jscodeshift` which is a dependency of `wix-style-react` 

Since version `jscodeshift@0.6.2` it doesn't use `nomnom` anymore: https://github.com/facebook/jscodeshift/commit/3fc6ca1c1e4b95c13123cab59270057e33a0418a

PR #3511, opened 2019-06-04
Revert "Dependency: update `jscodeshift` (now doesn't use deprecated `nomnom` package)"

argshook:
Reverts wix/wix-style-react#3502

`jscodeshift` uses babel7 since `0.6.0`, it interferes with other tooling and breaks wsr users builds.

Downgrading this dep is a workaround, but the dep is not critical anyway.
it should go away from wsr once `wix-ui-framework` supports all the "code plumbing".


sidoruk-sv:
@argshook can you point to some issue/PR relates to `wix-ui-framework to have the ability to track on that:
https://github.com/wix/wix-ui/tree/master/packages/wix-ui-framework

PR #3526, opened 2019-06-05
<Table/> - add `deselectRowsByDefault` prop

Oripi:
### 🔦 Summary
This feature allows deselecting rows when clicking on bulk selection when some rows are selecting. This behaviour is the current one we're using in old projects and we'd like to preserve it.

### ✅ Checklist
- 📚 Change is documented
  - [x] Story
  - [x] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [ ] Visual test


nitayneeman:
The PR looks good.

What do you think about this functionality?
When the new `deselectRowsByDefault` prop is true, it behaves like this:
demo

LittleTal:
@nitayneeman Looks good

Issue #3426, opened 2019-05-17
ImageViewer submits form on add image


jurasm:
# 🐛 Bug Report

### 🏗 Relevant Components

ImageViewer

### 😯 Current Behavior

ImageViewer's add button is now IconButton without `type="button"` which on press submits form. In earlier versions `6.14.0` it was div, so there wasn't an issue.

### 🤔 Expected Behavior

On pressing add image it should not submit form.

### 👣 Steps to Reproduce

Try pressing on add image.

```
<form onsubmit={() => alert('Form was submitted!')}>
    <ImageViewer/>
</form>
```
  - **WSR version**: `6.19.0`
  - **Yoshi version**: `^3.0.0`

### 👀 Severity

- Major

PR #3527, opened 2019-06-05
`<AddItem/>` - set type 'button'

Neophen:
### 🔦 Summary
Adding a button type to AddItem component and adding type button from ImageViewer
Fixes: https://github.com/wix/wix-style-react/issues/3426


src/AddItem/AddItem.js
```diff
@@ -138,6 +142,7 @@ class AddItem extends Component {
         {...style('root', { theme }, this.props)}
         data-hook={dataHook}
         disabled={disabled}
+        type={type}
```
mykas: I think it is enough to set it to be static: `button`. I don't see a case where it can be used with `form`.

Neophen: ok

PR #3530, opened 2019-06-05
<EmptyState /> add bottom padding to empty state

morsh:
### 🔦 Summary
...

### ✅ Checklist
- [X] 👨‍💻 API change is approved by the UI Infra Developers 
- [X] 👨‍🎨 UX change is approved by the Design System UX 
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [X] Component
  - [ ] Visual test


shlomitc:
@wuwa - Mor suggests adding a custom margin to the `EmptyState` component. was that approved? Don't we want to have it as a different theme in addition to the ones we have?

wuwa:
yes was approved. new theme is overkill (also old themes actually) - border would be better as separate prop.

src/EmptyState/EmptyState.js
```diff
@@ -81,6 +92,9 @@ EmptyState.propTypes = {
   /** The Empty State image, can be either a string representing the image URL, or a node to render instead */
   image: PropTypes.node,
 
+  /** The Empty State image bottom margin. If not specified, use the default padding defined in the css */
+  imageBottomMargin: PropTypes.number,
```
shlomitc: I think that instead of limiting ourselves with just the marginButton, let's pass `classNames` object props, targeting named by the same CSS API. In this case:
```js
classNames: PropTypes.shape({
  imageContainer: PropTypes.string,
})
```

morsh: @shlomitc - done

PR #3539, opened 2019-06-06
`<ListItemAction/>` - add autoFocos prop

mykas:
This PR will add autoFocus prop which can be enabled to focus automatically on page load.


mykas:
Neil quote:
`Keyboard behavior
tab to focus on the triggering element. 
space or enter on the triggering element toggle content element (remember to toggle the aria-expanded bool on the triggering element and the aria-hidden bool on the content element). When expanding the content element shifts the focus to the first or selected menu-item.
Only the keyboard up/down arrow keys should be used to navigate the menu items in a loop.
Esc and Tab keys select the current-menu-item and set the focus back to the triggering element.
`

This part: `When expanding the content element shifts the focus to the first or selected menu-item.`

And also an example of menu item js in w3c: https://www.w3.org/TR/wai-aria-practices-1.1/examples/menu-button/js/MenuItemAction.js

mykas:
Dropdownlayout supports only selection based mechanism ant not the focus. PopoverMenu needs to have focus based navigation. This prop is the easy way to achieve first focused element behaviour. Much better then attaching refs and focusing when component is mounted.

PR #3544, opened 2019-06-10
Feat/add color in color picker

MSiliunas:
### 🔦 Summary
- Added color preview in `<ColorPicker />` converter;
- Color preview has ability to act as 'Add' button which invokes callback;
- `<ColorPicker />` now support `empty` state which returns / accepts empty string. i.e. if you remove hex or other code in the converter;

### ✅ Checklist
- [x] 👨‍💻 API change is approved by the UI Infra Developers @mykas 
- 📚 Change is documented
  - [x] Story
  - [x] API description
- 🔬 Change is tested
  - [x] Component
  - [ ] Visual test


src/ColorPicker/ColorPickerConverterRGB.js
```diff
@@ -1,50 +1,64 @@
...
   };
 
   constructor(props) {
     super(props);
-    this.state = mapValue(props.current.rgb().object(), Math.round);
+    this.state = getRgbOrTransparent(props.current);
+  }
+
+  isInputsEmpty() {
+    return this.state.r === '' && this.state.g === '' && this.state.b === '';
```
argshook:
```
isInputsEmpty() {
  const { r,g,b } = this.state;
  return [r,g,b].every(value => value === '');
}
```

leaving just for fun, you don't have to do this :)

src/ColorPicker/utils.js
```diff
@@ -1,4 +1,40 @@
...
+  } catch (error) {
+    return null;
+  }
+}
+
+export function getHexOrTransparent(_color) {
+  return isTransparent(_color) ? '' : _color.hex();
+}
+
+export function getRgbOrTransparent(_color) {
```
argshook: for me this name indicates, that return value could be `transparent` but it actually can be `{r: '', g: '', b: ''}`. Also, one branch is about strings whilst other is about numbers (`Math.round`). is this ok?

perhaps it would be easier to operate with rgba values and avoid switching type?

src/ColorPicker/ColorPickerConverterViewer.js
```diff
@@ -0,0 +1,84 @@
...
+
+import Color from 'color';
+import styles from './ColorPickerConverter.st.css';
+import { string, func, oneOfType, node, object } from 'prop-types';
+import { withFocusable } from 'wix-ui-core/dist/src/hocs/Focusable/FocusableHOC';
+import Add from '../new-icons/Add';
+import Tooltip from '../Tooltip';
+
+function getContrastColor(bg, light = '#ffffff', dark = '#162d3d') {
+  try {
```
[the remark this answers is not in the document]
MSiliunas: `Color` lib fails if it can't resolve any color. In this case it might throw an exception if `bg` is corrupted color.

src/ColorPicker/ColorPicker.js
```diff
@@ -100,21 +107,22 @@ export default class ColorPicker extends WixComponent {
...
+        {children && (
           <div className={css.children}>
-            <Swatches
-              colors={preset}
-              showClear={showClear}
-              onClick={this.onSwatchClick}
-              selected={selectedSwatch}
-              dataHook="color-picker-swatches"
-            />
+            {React.cloneElement(children, {
```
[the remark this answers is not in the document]
MSiliunas: What I wanted to achieve here is the color picker intercepting `onClick` callback on child so it could invoke color picker's inner change function.

src/ColorPicker/utils.js
MSiliunas: Maybe it should be named `getRgbOrEmpty` as the intention of this helper is to get empty strings if color object has `alpha === 0`.
The empty strings represents empty state in the `ColorPickerConverterRgb` (same in Hsb).

src/ColorPicker/ColorPickerConverterViewer.js
```diff
@@ -0,0 +1,84 @@
+import React, { PureComponent } from 'react';
+
+import Color from 'color';
+import styles from './ColorPickerConverter.st.css';
+import { string, func, oneOfType, node, object } from 'prop-types';
+import { withFocusable } from 'wix-ui-core/dist/src/hocs/Focusable/FocusableHOC';
+import Add from '../new-icons/Add';
```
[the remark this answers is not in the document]
MSiliunas: 👍

src/ColorPicker/ColorPicker.js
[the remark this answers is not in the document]
MSiliunas: 👍 changed and tests added

MSiliunas:
> * If color is already added to the swatch, `AddSmall` icon should be removed and the callback turned off.

This is controlled by user implementation currently. Color Picker is not aware of available swatches.

src/ColorInput/ColorInput.js
```diff
@@ -120,8 +123,9 @@ class ColorInput extends React.Component {
         onCancel={this.cancel}
         onConfirm={this.confirm}
         onClickOutside={this.confirm}
-        preset={preset}
-        showClear={showClear}
+        colorPickerChildren={colorPickerChildren}
```
argshook:
```suggestion
        children={colorPickerChildren}
```
this would make more sense imo

src/ColorPicker/ColorPicker.js
```diff
@@ -46,17 +46,27 @@ export default class ColorPicker extends WixComponent {
...
 
+    /** Handle color add button click. If not passed plus icon is not visible. */
+    onAdd: func,
+
     /** Children would be rendered above action buttons */
-    children: node,
+    children: oneOfType([node, func]),
+
+    /** Content to show in add button tooltip. Doesn't appear if `onAdd` is not passed. */
+    addTooltipContent: oneOfType([string, node]),
```
argshook:
```suggestion
    addTooltipContent: node,
```

src/ColorPicker/ColorPickerConverterHsb.js
```diff
@@ -61,18 +72,15 @@ export default class ColorPickerConverterHsb extends WixComponent {
     },
   ) {
     this.setState({ [part]: value }, () => {
-      const _color = safeColor(this.state);
-      if (_color) {
+      const isMissingData =
+        this.state.h === '' || this.state.s === '' || this.state.l === '';
```
argshook: again, just for fun, you don't have to do this. i just like the symmetry of `some` and `every` in such cases

```js
const { h, s, l } = this.state;
const isMissingData = [h,s,l].some(value => value === '')
```

src/ColorPicker/ColorPickerConverterViewer.js
```diff
@@ -0,0 +1,87 @@
...
+      return light;
+    }
+  } catch (err) {}
+}
+
+class ColorPickerConverterViewer extends PureComponent {
+  static propTypes = {
+    color: object,
+    onAdd: func,
+    addTooltipContent: oneOfType([string, node]),
```
argshook:
```suggestion
    addTooltipContent: node,
```

src/ColorPicker/docs/index.story.js
```diff
@@ -9,18 +10,25 @@ export default {
...
+    children: [
       {
-        value: ['red', 'green', 'yellow', '#fff', 'rgb(0, 0, 0)', 'magenta'],
-        label: "['red', 'green', 'yellow', '#fff', 'rgb(0, 0, 0)', 'magenta']",
+        label: 'node',
+        value: <div>a node here</div>,
+      },
+      {
+        label: 'function',
+        value: ({ change }) => <div>node from function</div>,
```
argshook: it is not `change`, it is `changeColor` :)
also, would be nice to show how it can be called, maybe with swatches component, why not

MSiliunas:
> * [ ]  Add default value to `addTooltipContent` : "Add Color"
> * [ ]  `emptyPlaceholder` default value should be : "#"

@milkyfruit I was told that there should not be any built in placeholders by @mykas

src/ColorInput/ColorInput.js
MSiliunas: we talked about this with @mykas and agreed that `children` wouldn't be clear enough on where the elements will be added

src/ColorPicker/docs/index.story.js
MSiliunas: It appears that storybook doesn't recognise function passed as children. But adding and example anyways.

src/ColorInput/ColorInput.js
argshook: i'm talking about a prop of `ColorViewer`, not `ColorPicker`

MSiliunas: my bad, sorry 🎳

src/ColorPicker/docs/index.story.js
argshook: hmm, it used to work, something must've changed somewhere. ignore it for now

PR #3551, opened 2019-06-10
<Stepper/> - generate new  component with props validation

mosheKerbel:


src/beta/Stepper/test/Stepper.spec.js
```diff
@@ -1,22 +1,58 @@
...
+      render(<Stepper active={1} steps={[{}]} />);
+      expect(consoleErrorSpy).toHaveBeenCalledWith(
+        expect.stringContaining(
+          'Warning: Failed prop type: The prop `steps[0].text` is marked as required in `Stepper`, but its value is `undefined`',
+        ),
+      );
+    });
+
+    it('should warn when steps length is less than 2', async () => {
+      expect(true).toEqual(true); // todo: impl
```
mosheKerbel: i am considering validating the steps size.
1 < stepSize < 8
can do it as a part of the props-validation by using a HOC as in this example
https://www.ian-thomas.net/custom-proptype-validation-with-react/

no other component is using it. will need to add it as a HOC and add tests.
if i do it i prefer adding it as a task and get back to it later.

WDYT?
@shlomitc

mosheKerbel:
> @mosheKerbel are you sure this shouldn't be a compound component? i imagine this shape will grow in future, require tooltips and stuff

It shouldn't. You can take a look here, the component is very restricted.
https://app.zeplin.io/project/5c8f6f1f9f489505cac756dc/screen/5cc8015405f4f16514851df4

argshook:
> > @mosheKerbel are you sure this shouldn't be a compound component? i imagine this shape will grow in future, require tooltips and stuff
> 
> It shouldn't. You can take a look here, the component is very restricted.
> [app.zeplin.io/project/5c8f6f1f9f489505cac756dc/screen/5cc8015405f4f16514851df4](https://app.zeplin.io/project/5c8f6f1f9f489505cac756dc/screen/5cc8015405f4f16514851df4)

if only i get a nickel for each "shouldn't" that later turns into "should" :)

i would implement with compound idea, but ignore me

mosheKerbel:
> if only i get a nickel for each "shouldn't" that later turns into "should" :)
> 
> i would implement with compound idea, but ignore me

i kind of disagree because it contradicts with the idea of this component spec.

mosheKerbel:
@nitayneeman @shlomitc

src/Stepper/Stepper.js
```diff
@@ -0,0 +1,72 @@
...
+
+    this._validateNumberOfSteps(props.steps);
+  }
+
+  componentDidUpdate() {
+    this._validateNumberOfSteps(this.props.steps);
+  }
+
+  _validateNumberOfSteps(steps) {
+    if (!steps || !steps.length || !steps.length < 2 || steps.length > 7) {
```
nitayneeman: I would export these magic numbers to variables and reuse them

src/Stepper/Stepper.js
```diff
@@ -0,0 +1,72 @@
...
+      console.error(
+        'Warning: Failed prop: The prop `steps` in `Stepper` has to be an array in the size of 2 to 7.',
+      );
+    }
+  }
+
+  _renderStep = (step, idx) => {
```
nitayneeman: As a convention - I usually place the private methods at the bottom (after `render`)

src/Stepper/docs/index.story.js
```diff
@@ -0,0 +1,106 @@
...
+
+          divider(),
+
+          title('Examples'),
+
+          code({
+            title: 'Stepper',
+            compact: true,
+            source:
+              '<div style={{padding: "20px 20px", backgroundColor: "white"}}>' +
```
nitayneeman: Why not template strings?

src/Stepper/test/Stepper.spec.js
```diff
@@ -0,0 +1,79 @@
...
+    consoleErrorSpy = jest
+      .spyOn(global.console, 'error')
+      .mockImplementation(jest.fn());
+  });
+
+  afterEach(() => {
+    consoleErrorSpy.mockRestore();
+  });
+
+  describe('Stepper', () => {
```
nitayneeman: Maybe it's worth to add a test which verifies that when you pass X number of steps it actually renders the correct number of steps

src/Stepper/Stepper.js
mosheKerbel: actually i'm used to another convention where the main render is in the bottom of the file and the other private renders above it :)
do we have any guideline for wsr?

src/Stepper/test/Stepper.spec.js
mosheKerbel: all this kind of tests will be in the pending next PR

PR #3557, opened 2019-06-11
add full width to section helper

adimorwix:
### 🔦 Summary

add fullWidth prop to section helper

### ✅ Checklist
- 📚 Change is documented
  - [ Y] Story


src/SectionHelper/index.js
```diff
@@ -27,6 +27,9 @@ export const HELPER_APPEARANCE = {
 class SectionHelper extends WixComponent {
   render() {
     const { showCloseButton, onClose } = this.props;
+    const contentStyle = {
+      maxWidth: this.props.fullWidth ? '100%' : null,
```
shlomitc: That's not what you're trying to achieve...
what you want is to remove the `max-width` constrains from here:
https://github.com/wix/wix-style-react/blob/master/src/SectionHelper/styles.scss#L18

so instead of inlining the style tag, make a change in the css, applying

```
.fullWidth.content {
  max-width: 0
}
```

and in your code
```js
className={classnames(styles.content, {[styles.fullWidth]: this.props.fullWidth})};
```

PR #3563, opened 2019-06-12
chore(GeneratedTestComponent): remove

argshook:
this component was used to test component generator. It is no longer
used as test basis and not regenerated anymore.

component generator and its tests were extracted to separate
wix-ui-framework package

thus, GeneratedTestComponent is no longer relevant in this package


shlomitc:
If I remember correctly, generating the test component happened on each CI run, to tests that the generated component works.

So the test was on the output of the template, and not the generator, something that is unique per project

argshook:
it used to happen before `wuf` yes, but the test was to check for git status, which was almost always falsy and thus not really useful.

test can still be done, but i don't want generated test component to be part of all the testkit exports, type definitions & storybook.

i'll bring the test back but not in a way that does not interfere with aforementioned  files

PR #3571, opened 2019-06-16
<Stepper/> - most basic stepper functionality

mosheKerbel:
@nitayneeman @shlomitc 
I initialy tried to break into smaller PRs but then they were too dependent
Anyways this PR is the initial one for the Stepper component.
Divided small "features" into commits. Let me know if it is comfortable enough reviewing it this way.


src/Stepper/test/Stepper.spec.js
```diff
@@ -0,0 +1,209 @@
...
+        createDriver(<Stepper activeStep={1} steps={steps} />);
+        expect(consoleErrorSpy).toHaveBeenCalledWith(
+          expect.stringContaining(
+            `Warning: Failed prop: The prop 'steps' at [1] is invalid. Active step can not be disabled. Falling back to 'normal' type for step[1]`,
+          ),
+        );
+      });
+    });
+    describe('Clickable', () => {
+      it('Disabled step should not be clickable', async () => {
```
nitayneeman: Why not wrapping by an internal `describe` instead of attaching it to the test description? 
For example:
```
describe('Disabled step', () => {
  it('should not be clickable', async () => {
})
```

src/Stepper/docs/index.story.js
```diff
@@ -0,0 +1,119 @@
...
+          code({
+            title: 'Stepper',
+            compact: true,
+            source: `<div style={{padding: "20px 20px", backgroundColor: "white"}}>
+              <Stepper activeStep={2} steps={steps} onClick={onClick}/>
+              </div>`,
+          }),
+
+          code({
+            title: 'Stepper',
```
nitayneeman: Probably not part of this PR but should be explained better

src/Stepper/dataHooks.js
```diff
@@ -0,0 +1,3 @@
+export default {
+  step: 'STEP',
```
nitayneeman: I think it'd be better to add a prefix of `stepper` to the dataHook, for example: "stepper-step" (relevant for next dataHooks as well)

That's the convention I'm familiar with

src/Stepper/Stepper.js
```diff
@@ -0,0 +1,108 @@
...
+  };
+
+  _validateStepType = (steps, activeStep) => {
+    if (
+      activeStep &&
+      typeof activeStep === 'number' &&
+      steps[activeStep].type === STEP_TYPES.DISABLED
+    ) {
+      console.error(
+        `Warning: Failed prop: The prop 'steps' at [${activeStep}] is invalid. Active step can not be disabled. Falling back to 'normal' type for step[${activeStep}]`,
```
nitayneeman: Not sure if it's a warning or error

src/Stepper/Step/StepCircle/StepCircle.st.css
```diff
@@ -0,0 +1,102 @@
...
+
+.text:disabled > a {
+  color: value(D80);
+}
+
+.text:selected > a {
+  color: value(D80);
+}
+
+.text:disabled > a:hover {
```
nitayneeman: Redundant?

src/Stepper/Stepper.js
```diff
@@ -0,0 +1,108 @@
...
+
+  _validateNumberOfSteps(steps) {
+    if (
+      !steps ||
+      !steps.length ||
+      steps.length < MIN_STEPS ||
+      steps.length > MAX_STEPS
+    ) {
+      console.error(
+        `Warning: Failed prop: The prop 'steps' in 'Stepper' has to be an array in the size of ${MIN_STEPS} to ${MAX_STEPS}.`,
```
nitayneeman: Same here - warning / error?

src/Stepper/Step/StepCircle/StepCircle.st.css
mosheKerbel: actually i must use this state, because i am overriding Text component which has a defenition for `.text > a:hover`

src/Stepper/Stepper.js
mosheKerbel: you are right it should be an error

src/Stepper/Stepper.js
mosheKerbel: you are right it should be an error

PR #3589, opened 2019-06-18
Feat/wuf beta components

argshook:
this is an update to `npm scripts` to use `wix-ui-framework@2.4.1`:

1. remove a bunch of `generate-*` scripts from `package.json`
1. use `scripts/update-components.sh` with `wuf` commands inside
1. use `scripts/update-beta-components.sh` with `wuf` commands inside for beta components
1. add `.wuf/beta` folder for beta components. It's mainly used with `wuf`, not manually
1. reuse regular component generator codemods for beta components
1. wuf all the way, one more step closer to just `wuf update`
1. add `npm run generate:beta` for generating beta components


.wuf/beta/components.json
```diff
@@ -0,0 +1,5 @@
+{
+  "PopoverMenu": {
```
argshook: same as `.wuf/components` but for beta components. this file is generated

.wuf/beta/generator/codemods/index.js
```diff
@@ -0,0 +1,28 @@
+// Here we reuse codemods of non beta components
+// only configure to output files in beta folder
+
+module.exports = [
```
argshook: this file is almost the same as `.wuf/generator/codemods/index.js`. It is used during `npm run generate:beta`

.wuf/beta/testkits/definitions.js
```diff
@@ -0,0 +1,43 @@
...
+ *   unidriver?: false;
+ *
+ *   // set path to testkit if path is not following convention
+ *   // this file will be imported and wrapped with appropriate testkit factory
+ *   // (one of testkitFactoryCreator, uniTestkitFactoryCreator, enzymeTestkitFactoryCreator or enzymeUniTestkitFactoryCreator
+ *   testkitPath?: function;
+ * }
+ */
+
+module.exports = {
```
argshook: this file is same as `.wuf/testkits/definitions.js`, but for beta components

scripts/update-components-beta.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
+# create beta components list
+wuf update \
```
argshook: bash scripts are scary, i know, but it's just an easy way out of horrific `script` tags in `package.json` without any additional overhead

PR #3593, opened 2019-06-18
<Stepper/> - test stepper adjust itself on window resize

mosheKerbel:
@nitayneeman @argshook


src/Stepper/test/Stepper.e2e.js
```diff
@@ -0,0 +1,44 @@
...
+  waitForVisibilityOf,
+  scrollToElement,
+} from 'wix-ui-test-utils/protractor';
+import { createStoryUrl } from '../../../test/utils/storybook-helpers';
+
+import { stepperTestkitFactory } from '../../../testkit/protractor';
+import { storySettings } from './storySettings';
+
+const { category, storyName } = storySettings;
+const autoExampleUrl = createStoryUrl({
```
nitayneeman: AFAIK, we try not to test on top of the playground but rather prefer creating a dedicated test page. For example:
https://github.com/wix/wix-style-react/blob/master/.wuf%2Fgenerator%2Ftemplates%2Fsrc%2FComponent%2Ftest%2FComponent.e2e.js#L15

Issue #3497, opened 2019-05-30
<RichTextInputArea /> links are not target=_blank anymore, bad interaction with Business Manager


dovilel:
# 🐛 Bug Report

### 🏗 Relevant Components

`<RichTextInputArea />`

### 😯 Current Behavior

If there is a link added to the text area, user clicking on it causes Business Manager to fire an internal navigation event (businessManager.onNavigation).
With the old deprecated component (RichTextArea), links had target=_blank on them, which are not considered by BM to be internal navigation.

The constructed value does have target=_blank in both components.

### 🤔 Expected Behavior

Expect links to have no effect in Business Manager and to be consistent between input area and output value.

### 👣 Steps to Reproduce

Not sure how to give example with BM, since I can't find new RichTextInputAreas there.
Easy way to see behavior that happens:
1. Install Events App.
2. Go to BM -> Events.
3. Click Add Event
4. Scroll to bottom for the old RichTextArea
5. Add a link.
6. Inspect it and remove target="_blank"
7. Click on the link - see popup appearing.

  - **WSR version**: 16.5
  - **Yoshi version**: `...`

### 👀 Severity

- Major

argshook:
closed by #3602, release pending on #3611

PR #3602, opened 2019-06-19
<RichTextInputArea/> - add `target` attribute to rendered links

nitayneeman:
This PR fixes the following issue:
https://github.com/wix/wix-style-react/issues/3497

It attaches a `target` attribute to the DOM elements of the links which are rendered within the content after adding links. The default value is `_blank`, much like the behavior in the legacy `<RichTextArea/>`.

Notice it doesn't change the output value, which remains the same.


nitayneeman:
Failed because of timeout

PR #3605, opened 2019-06-20
`<Text/>`- visual testing, removing the relevant e2e tests and updating docs

TalZa1810:
### 🔦 Summary
visual testing and removing the relevant e2e tests. Moreover, updating documentation.

Related PR: https://github.com/wix/wix-style-react/pull/3377 
(for docs update , styling & anchor visual test). 

Note:
Still got some issues with the ellipsis visual test.
looks like it doesn't trigger on the initial rendering. in the protractor test, it worked only after scrolling.
The best option will be trying to solve it as part of the ellipsis HOC

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [x] Visual test


src/Text/test/Text.visual.js
```diff
@@ -0,0 +1,65 @@
...
+};
+
+const sizes = Object.keys(SIZES);
+
+const skins = Object.keys(SKINS);
+
+const weight = Object.keys(WEIGHTS);
+
+const tests = [
+  {
```
nitayneeman: It would be better if we stick with our conventions:
```
{
  it: 'Should...',
  props: {...}
}
```

src/Text/test/Text.visual.js
```diff
@@ -0,0 +1,65 @@
...
+              Text Component
+            </Text>
+          </Cell>
+        );
+      })}
+    </Layout>
+  ));
+});
+
+storiesOf(`Text/Light prop`, module).add('light prop', () => (
```
nitayneeman: Maybe we should attach this spec to the `tests` array

nitayneeman:
Related to wix-private/wsr-issues#451

PR #3606, opened 2019-06-20
<Calendar/> - FilterDate should not affect calendar if date is modified

aarnoldaas:
### 🔦 Summary
FilterDate provides date that is used in calendar modifying date can break calendar. 

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [ ] Visual test


src/Calendar/Calendar.js
```diff
@@ -266,7 +266,7 @@ export default class Calendar extends WixComponent {
     return {
       disabledDays: excludePastDates
         ? { before: new Date() }
-        : date => !filterDate(date),
+        : date => !filterDate(new Date(date)),
```
mosheKerbel: i am assuming that creating a new instance of Date solves some bug you noticed?
i don't really understand what the bug you are experiencing.

src/DatePicker/DatePicker.spec.js
```diff
@@ -542,6 +542,23 @@ describe('DatePicker', () => {
     });
   });
 
+  describe('`filterDate` prop', () => {
```
mosheKerbel: you are changing the Calendar component, 
why testing DatePicker and not Calendar itself?

src/DatePicker/DatePicker.spec.js
```diff
@@ -542,6 +542,23 @@ describe('DatePicker', () => {
...
+    it('should not affect calendar if date is modified', () => {
+      const date = new Date(2017, 9, 2);
+      const { inputDriver } = createDriver(
+        <DatePicker
+          onChange={noop}
+          filterDate={_date => _date.setUTCDate(3)}
```
mosheKerbel: this function should return true/false, why do you return a date?

aarnoldaas: I wanted to show that modifying date in callback breaks Calendar. I will add return false in next line.

src/DatePicker/DatePicker.spec.js
aarnoldaas: Good point, in my project I use DatePicker so I started from test in DatePicker

src/Calendar/Calendar.js
aarnoldaas: We had a bug because we added timeZoneOffset to value provided in filter date callback. This in some edge cases shifted all calendar days by one day. Example Jul 15 Monday, become Tuesday.

aarnoldaas: We need to compare dates not according to browser timezone so we were modifying date. Date picker return value with computer timezone. 
It's bad that we modified date in callback, but result was unexpected so I wanted to fix it in library so other wont face this issue.

mosheKerbel: now i get it
you want to protect the date from the external filterDate func, so nobody could accidentally override it?

aarnoldaas: Yes

mosheKerbel: good fix 👍 
just took me a while to figure out what's going on here.

mosheKerbel:
ping @aarnoldaas 
we are one step from closing that one...let's push it.

aarnoldaas:
@mosheKerbel pushed changes

PR #3608, opened 2019-06-20
<Tag/> - Replace tag thumbnail varations e2e->visual

mosheKerbel:
@nitayneeman @TalZa1810 
just the beginning of the work on tag
need to render it a bit more pretty before merge


nitayneeman:
Related to wix-private/wsr-issues#451

PR #3616, opened 2019-06-23
<Stepper/> - fix border

mosheKerbel:
These states should not have a border according to deign spec.


mosheKerbel:
All good with tests - only 1 failed due to flaky timeout
Merging.

PR #3618, opened 2019-06-23
visual testing- creating a util for all test scenarios when all possible values are given

TalZa1810:
### 🔦 Summary
In many of our visual tests, we have a repetitive pattern - we have an object with multiple values and we wish to create an object for all visual tests. For example:
`<Text/>` component has multiple skins:
```
const SKINS = {    
    standard: 'standard',    
    error: 'error',       
    ....
};
```

We would like to have a util that by giving `propName` and `propValues` it will return an object with all tests scenarios as follows:

```
const skinsTestScenarios= {
    its: [ 
                { it: 'standard' , props: { skin:  'standard' } }, 
                { it: 'error', props: { skin:  'error' } }, 
                .... 
          ]
};
```


test/utils/visualTest/visualTestingUtils.spec.js
```diff
@@ -0,0 +1,65 @@
...
+        },
+        {
+          it: propValue2,
+          props: { [propName]: propValue2 },
+        },
+      ],
+    };
+
+    describe('return value', () => {
+      it('should be similar to "describeObject" object', () => {
```
TalZa1810: Do you have a better suggestion for the test description?

test/utils/visualTest/visualTestingUtils.js
```diff
@@ -0,0 +1,34 @@
...
+  Object.keys(propValues).map(propValue => {
+    const testScenario = { it: propValue, props: { [propName]: propValue } };
+    describeToReturn.its.push(testScenario);
+  });
+
+  return describeToReturn;
+};
+
+const variablesValidation = ({ propName, propValues, describeName }) => {
+  const propNameError = 'in valid prop name';
```
mosheKerbel: Nitpicking but invalid not in valid

test/utils/visualTest/visualTestingUtils.js
```diff
@@ -0,0 +1,34 @@
+export const createDescribeObject = ({
+  propName = '',
```
mosheKerbel: pattern is a bit odd, you give it a default value which turns out to be invalid when you continue.
I would skip settings default values and in the validation just check for the type and if exists look at content:

```
if (!propName) {
  // err
}

if (!propValues || Object.keys(propValue).length === 0) {
  // err
}

```

mosheKerbel:
nice! i like the result
just 2 things and we can merge:
1. `visualTestingUtils` - let's make all function es6 arrow functions.
2. is it ok or we have some un-intended repetition in the use cases?
 
skins
image

sizes
image

TalZa1810:
@mosheKerbel 
As for the arrow function- fixed. As for the sizes of the components, they are all [small ](https://github.com/wix/wix-style-react/blob/cf6398364cdd75d7b7a8c5272f21be969932cf6d/src/CloseButton/CloseButton.visual.js#L13)([default prop](https://wix-wix-style-react.surge.sh/?activeTab=API&addons=0&full=0&panelRight=0&selectedKind=5.%20Buttons&selectedStory=5.4%20CloseButton&stories=1))



The light skin should be wrapped in a coloured `<Box/>`, should consider adding this as a different test scenario or leaving it as is.

test/utils/visualTest/visualTestingUtils.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  if (!propValues || !Object.keys(propValues).length) {
+    throw new Error(propValuesError);
+  }
+};
+
+export const renderChildrenNodes = ({ its }, componentToRender) =>
+  its.map(({ props }, i) => (
+    <VisualTestComponent key={i} child={componentToRender} {...props} />
```
jonathanadler: Super picky, and not really important, but this can be written:
```jsx
<VisualTestComponent key={i}>
    {React.cloneElement(componentToRender, props)}
</VisualTestComponent>
```
And then `VisualTestComponent` will be:
```jsx
render() {
    return (
        <div style={...}>
            {this.props.children}
        </div>
    );
}
```

test/utils/visualTest/visualTestingUtils.js
```diff
@@ -0,0 +1,52 @@
...
+  });
+
+  return its;
+};
+
+const createTestScenario = ({ propName, propValue }) => {
+  return { it: propValue, props: { [propName]: propValue } };
+};
+
+const variablesValidation = ({ propName, propValues }) => {
```
jonathanadler: Nice 👍

test/utils/visualTest/visualTestingUtils.js
```diff
@@ -0,0 +1,52 @@
+import React, { PureComponent } from 'react';
+import { node } from 'prop-types';
+
+export const createVisualTests = ({ propName, propValues }) => {
+  variablesValidation({ propName, propValues });
+
+  const its = { its: [] };
```
jonathanadler: Super super picky:
maybe change the name to `describe` or `description`:
```js
const description = { its: [] };
```

test/utils/visualTest/visualTestingUtils.js
```diff
@@ -0,0 +1,52 @@
+import React, { PureComponent } from 'react';
+import { node } from 'prop-types';
+
+export const createVisualTests = ({ propName, propValues }) => {
```
jonathanadler: Another naming comment:
Maybe we should call it something like `createVisualTestsByProp` or something like that.
Just so it will be understood that this creates a specific test according to the prop enum, and not a general visual test.
wdyt?

TalZa1810: Sounds good @jonathanadler. Done.

test/utils/visualTest/visualTestingUtils.js
TalZa1810: Done :)

TalZa1810:
I've fixed it. Let me know if you have more comments or we can merge and start using it :)
@jonathanadler  @mosheKerbel @nitayneeman

test/utils/visualTest/visualTestingUtils.js
```diff
@@ -0,0 +1,48 @@
...
+  if (!propName) {
+    throw new Error(propNameError);
+  }
+
+  if (!propValues || !Object.keys(propValues).length) {
+    throw new Error(propValuesError);
+  }
+};
+
+export const renderChildrenNodes = ({ its }, componentToRender) =>
```
nitayneeman: We should consider renaming to something like `renderTestComponents` or `renderEyeTests` IMO.
It's going to be used apparently a lot, so it should have a proper semantic name.

mosheKerbel:
approved 1 applitools diff
failed also due to 1 unrelated async test
merging.

PR #3622, opened 2019-06-24
Components Size - script to generate components bundles and analyse components sizes

mykas:


package.json
```diff
@@ -105,13 +107,13 @@
...
     "@babel/core": "^7.2.2",
     "@babel/parser": "^7.3.1",
-    "@emotion/core": "0.13.1",
     "@babel/plugin-proposal-class-properties": "^7.3.0",
     "@babel/plugin-proposal-export-namespace-from": "^7.2.0",
     "@babel/plugin-transform-modules-commonjs": "^7.2.0",
     "@babel/plugin-transform-runtime": "7.2.0",
     "@babel/preset-env": "^7.3.1",
     "@babel/preset-react": "^7.0.0",
+    "@emotion/core": "0.13.1",
```
shlomitc: if you already touched this - is it still required?

mykas: nope, gonna revert

mykas: not sure why this have changed here. Gonna revert

mykas: But if you ask if this is still required - yes. This was added last time because of storybook bug where applitools storybook dependency is overriding our dependency. Until we upgrade to storybook 5 - this will be needed.

PR #3374, opened 2019-05-08
Ellipsis and tooltip on long badges

matan04:
### 🔦 Summary
...
Add ellipsis and tooltip to long badge text

### ✅ Checklist
- [ ] 👨‍🎨 UX change is approved by the Design System UX

PR #3628, opened 2019-06-25
<Badge/> - support ellipsis in case of long text

nitayneeman:
This PR is based on https://github.com/wix/wix-style-react/pull/3374/.

It adds the ability to render `<Badge/>` with an ellipsis, in case the parent's width is limited. Also, the E2Es are replaced with visual tests.


shlomitc:
I wouldn't change the E2E before I know the new behavior works.
for example - in the previous PR, adding the ellipsis broke the E2E. so if we did this, it means that we wouldn't catch the issue

nitayneeman:
> I wouldn't change the E2E before I know the new behavior works.
> for example - in the previous PR, adding the ellipsis broke the E2E. so if we did this, it means that we wouldn't catch the issue

I'm still checking it, but it seems like a specific issue with the Protractor driver (the right element is caught, but the click isn't triggered anymore).

Anyway, the new behavior works - I verified it manually and with visual tests (including tests for BadgeSelect).

nitayneeman:
> > I wouldn't change the E2E before I know the new behavior works.
> > for example - in the previous PR, adding the ellipsis broke the E2E. so if we did this, it means that we wouldn't catch the issue
> 
> I'm still checking it, but it seems like a specific issue with the Protractor driver (the right element is caught, but the click isn't triggered anymore).
> 
> Anyway, the new behavior works - I verified it manually and with visual tests (including tests for BadgeSelect).

Well, it seems to be related to the position of the click. Once you click on the suffix icon, it always works. But when you click on the text, which wrapped by the ellipsis HOC, you need to click twice - the first is to focus the element and the second is the actual click (and then you can continue clicking without any problem). Notice that if you focus the element using the keyboard and then click, it works.

I've managed to pass the test by simulating a few clicks instead of one - but indeed, it's a real issue.

demo

nitayneeman:
> > > I wouldn't change the E2E before I know the new behavior works.
> > > for example - in the previous PR, adding the ellipsis broke the E2E. so if we did this, it means that we wouldn't catch the issue
> > 
> > 
> > I'm still checking it, but it seems like a specific issue with the Protractor driver (the right element is caught, but the click isn't triggered anymore).
> > Anyway, the new behavior works - I verified it manually and with visual tests (including tests for BadgeSelect).
> 
> Well, it seems to be related to the position of the click. Once you click on the suffix icon, it always works. But when you click on the text, which wrapped by the ellipsis HOC, you need to click twice - the first is to focus the element and the second is the actual click (and then you can continue clicking without any problem). Notice that if you focus the element using the keyboard and then click, it works.
> 
> I've managed to pass the test by simulating a few clicks instead of one - but indeed, it's a real issue.
> 
> demo
> 
> Edit: I've tried to disable the `withFocusable` HOC and now it works as expected (and all E2Es pass successfully). So apparently the combination of those makes this issue for a reason.

I've tried to create a simple HOC and render it instead of the `withEllipsedTooltip` HOC and actually got the same result. It means that's not really related to `withEllipsedTooltip`, but rather an issue of any HOC that's wrapped by `FocusableHOC` 🤔

nitayneeman:
> > > > I wouldn't change the E2E before I know the new behavior works.
> > > > for example - in the previous PR, adding the ellipsis broke the E2E. so if we did this, it means that we wouldn't catch the issue
> > > 
> > > 
> > > I'm still checking it, but it seems like a specific issue with the Protractor driver (the right element is caught, but the click isn't triggered anymore).
> > > Anyway, the new behavior works - I verified it manually and with visual tests (including tests for BadgeSelect).
> > 
> > 
> > Well, it seems to be related to the position of the click. Once you click on the suffix icon, it always works. But when you click on the text, which wrapped by the ellipsis HOC, you need to click twice - the first is to focus the element and the second is the actual click (and then you can continue clicking without any problem). Notice that if you focus the element using the keyboard and then click, it works.
> > I've managed to pass the test by simulating a few clicks instead of one - but indeed, it's a real issue.
> > demo
> > Edit: I've tried to disable the `withFocusable` HOC and now it works as expected (and all E2Es pass successfully). So apparently the combination of those makes this issue for a reason.
> 
> I've tried to create a simple HOC and render it instead of the `withEllipsedTooltip` HOC and actually got the same result. It means that's not really related to `withEllipsedTooltip`, but rather an issue of any HOC that's wrapped by `FocusableHOC` 🤔

So, after figuring out that it's related to `FocusableHOC` somehow, I tried the following:

1. Created another empty HOC, just to be sure, that replaces `FocusableHOC` which wraps `<Badge/>` and contains the second empty HOC. The thing is it worked, what guaranteed me it's probably about the `FocusableHOC`.

2. Started to disable almost everything, and saw that when `focusableOnFocus: () => this.setState({ // Whatever field, because it also reproduced for an empty object })` is there, it doesn't work for a reason.

3. Created a project that claimed to simulate the issue, but couldn't reproduce that (even though I used the same scenario and React version) - https://stackblitz.com/edit/react-hocs-example

4. Instead of an empty Ellipsis HOC, I tried to evaluate the content as a reference for a render function:
```
render() {
 const Comp = () => (
   <div>
     <span>My HOC</span>
   </div>
 );

 return <div>{<Comp />}</div>;
}
```
But the issue wasn't solved.

5. When I evaluated the function like `return <div>{Comp()}</div>;` it suddenly worked. So I started to think about bundling stuff, and tried to analyze the diff between the outputs:

Screen Shot 2019-06-30 at 15 58 16

On the right is the render function of step 4 whereas the left is after this step, and it makes sense. 

I'm not sure I understand yet why just invoking the render function works, whereas creating an element that wraps <Comp/> (or any HOC) doesn't - but it gets me closer to the solution.

BTW, I'm documenting this mainly for myself (but possibly for the future). 😎

nitayneeman:
Accepting the visual diff and merging the PR.

I verified that all E2Es pass successfully.

PR #3634, opened 2019-06-25
<Stepper/> -  resize steps

mosheKerbel:
@jonathanadler


src/Stepper/Step/Step.js
```diff
@@ -70,18 +69,9 @@ class Step extends React.PureComponent {
...
-            'stepTextWrapper',
-            {
-              disabled: type === STEP_TYPES.DISABLED,
-              selected: active,
-            },
-            this.props,
-          )}
-          style={{ maxWidth: `${stepSize}px` }}
-        >
+        <div {...styles('stepTextWrapper', {}, this.props)}>
```
jonathanadler: Since you removed the `-st-states` from `stepTextWrapper`, I think that you can have:
```
<div className={styles.stepTextWrapper}>
```
instead of:
```
<div {...styles('stepTextWrapper', {}, this.props)}>
```

mosheKerbel:
12 applitools Stepper tests failed due to the changes - this is good!
merging and will approve changes after.

PR #3643, opened 2019-06-26
Change Input components disabled colors

gilmeir:


src/new-colors.scss
```diff
@@ -0,0 +1,11 @@
+@import './colors.scss';
```
nitayneeman: I'm wondering why do we need another file for the new colors, because we've a dedicated file for that purpose (`colors.scss`).

I'm afraid that in the future we'll have "new-new-colors", "new-new-new-colors" and so on

gilmeir: removing the whole css variables & new-colors, getting back to sass rgba function, created a different issue for migrating to css variables https://github.com/wix/wix-style-react/issues/3704

gilmeir:
Irrelevant, implemented in #3718,

PR #3645, opened 2019-06-26
fix: moving common types from index.d.ts

BrooklynKing:
### 🔦 Summary
**Problem:**
Assume my project's tsconfig.json has
`skipLibCheck: false` and `noImplicitAny:true`
if I import component, that has `.d.ts`, and it has inside its import of your `index.d.ts`, then my typescript build would fail, since almost everything inside your `index.d.ts` right now is implicitly any.

**Solution**
I just moved all common types, that could be reused, to separated files. Not sure about the name "common.d.ts".


shlomitc:
That's great! thanks!

PR #3650, opened 2019-06-30
split the test script into multiple test scripts

netanelgilad:
@shlomitc This time, I split the commands, and made the `test` command only run on master. So we will not have a case where we have the e2e tests (with eyes) running twice on master or PR.


shlomitc:
still not merging - the test command runs all tests in PR although not supposed to

netanelgilad:
@shlomitc When the build is green this can be merged, `test` script fixed.

PR #3653, opened 2019-07-01
<MessageBoxMarketerialLayout/> - Add unidriver

justenor:
Migrate MessageBox-MarketerialLayout to unidriver


src/MessageBox/MarketerialLayout/MessageBoxMarketerialLayout.driver.js
```diff
@@ -22,6 +22,11 @@ const messageBoxMarketerialLayoutDriverFactory = ({ element }) => {
       element.querySelector('[data-hook="message-box-title"]').textContent,
     getContentBySelector: selector => element.querySelector(selector),
     getImageSrc: () => element.querySelector('[data-hook="header-image"]').src,
+    isClassPresentInPrimaryButton: className =>
```
mosheKerbel: `isClassPresentInPrimaryButton
isClassPresentInHeaderCloseButton`

I did not talk about it in our session - so this is a good time @justenor @Lukermelesh 

Theses methods are very specific to our test, we don't want to expose them to the consumer.
Even though many methods here exist that should not be exposed to the user to begin with (bad testing), i don't want to break users who used it, so we are not changing removing them.
But new methods like these, let's move to private driver.

Please look at the following example here to see how we work with private drivers:
https://github.com/wix/wix-style-react/blob/f439429aec56de31df5e227861f58c2b08daa154/src/Accordion/Accordion.spec.js#L5

src/MessageBox/MarketerialLayout/MessageBoxMarketerialLayout.uni.driver.js
```diff
@@ -0,0 +1,37 @@
...
+    getTitle: () => getMessageBoxTitleBase().text(),
+    getContentBySelector: async selector =>
+      (await base.$(selector).exists()) ? base.$(selector) : null,
+    getImageSrc: () => getHeaderImageBase().attr('src'),
+    isClassPresentInPrimaryButton: className =>
+      getPrimaryButtonBase().hasClass(className),
+    isPrimaryButtonDisabled: async () =>
+      await getPrimaryButtonBase()._prop('disabled'),
+    isClassPresentInHeaderCloseButton: async className =>
+      (await getCloseButtonBase().attr('class')).includes(className),
```
mosheKerbel: you can simply use hasClass like in line 31, no?

src/MessageBox/MarketerialLayout/MessageBoxMarketerialLayout.uni.driver.js
```diff
@@ -0,0 +1,37 @@
...
+    clickOnPrimaryButton: () => getPrimaryButtonBase().click(),
+    clickOnSecondaryButton: () => getSecondaryButtonBase().click(),
+    closeMessageBox: () => getCloseButtonBase().click(),
+    getTitle: () => getMessageBoxTitleBase().text(),
+    getContentBySelector: async selector =>
+      (await base.$(selector).exists()) ? base.$(selector) : null,
+    getImageSrc: () => getHeaderImageBase().attr('src'),
+    isClassPresentInPrimaryButton: className =>
+      getPrimaryButtonBase().hasClass(className),
+    isPrimaryButtonDisabled: async () =>
```
mosheKerbel: i see that the spec was missing the opposite tests for boolean methods (before your migration)
plz add them.

src/MessageBox/MarketerialLayout/MessageBoxMarketerialLayout.uni.driver.js
justenor: No, the problem is that classes in header close button are like 'ad5678sdf-dark'. I couldn't think of the way how to get this full class as they were hardcoded till now in tests: https://github.com/wix/wix-style-react/pull/3653/files#diff-ddb72bfee42d0560253ceca2b5a1bfbbL107 . Do you have any idea how to solve this?

src/MessageBox/MarketerialLayout/MessageBoxMarketerialLayout.driver.js
justenor: Fixed

src/MessageBox/MarketerialLayout/MessageBoxMarketerialLayout.uni.driver.js
justenor: Fixed

src/MessageBox/MarketerialLayout/MessageBoxMarketerialLayout.uni.driver.js
mosheKerbel: ok, so it all begins from a test that in my opinion should have been a snapshot test and not unitest, but in this scope we are not going to turn everything around.
so, i suggest - look for the real class, and not just a portion of it.
this long id that you mentioned is the hashing of css modules that is taken from the button stylesheet.
so you can import the style
```import styles from '../../Button/Button.st.css';```
and instead of looking for ```.dark```, look for ```styles.dark```

justenor: Thanks! Fixed

mosheKerbel:
great job
merging. failed only due to 1 falky unrelated async timeout test.
@justenor notice our convention, i changed it to it in the PR name
```<MessageBoxMarketerialLayout/> - Add unidriver```

PR #3658, opened 2019-07-02
<InputArea /> - Add UniDriver

Lukermelesh:


src/InputArea/InputArea.spec.js
```diff
@@ -1,365 +1,386 @@
...
+      it('passes the name and value attribute', async () => {
+        const onChangeMock = jest.fn();
+        const props = {
+          name: 'gal',
+          onChange: onChangeMock,
+        };
+        const driver = createDriver(<InputAreaForTesting {...props} />);
+        await driver.enterText('some text');
+        const eventTarget = onChangeMock.mock.calls[0][0].target;
+        expect(eventTarget.name).toEqual('gal');
```
Lukermelesh: @mosheKerbel I had to change the test, because here `eventTarget` is an HTML element now instead of an object, because I used the new UniDriver `enterValue` method

mosheKerbel:
approved.
all tests passed beside 4 unrelated flaky timeout tests
merged.

PR #3662, opened 2019-07-02
<PopoverMenu> adjust popover menu appendTo prop type

Jefoss:
the prop type for popover menu is very specific and does not correlate to the <Tooltip> component it uses. (it is spamming us with prop type warnings since we use a DOM element)

### DIFF
--- a/src/PopoverMenu/PopoverMenu.js
+++ b/src/PopoverMenu/PopoverMenu.js
@@ -7,7 +7,6 @@ import More from '../new-icons/More';
 import PopoverMenuItem from '../PopoverMenuItem';
 import classnames from 'classnames';
 import {
-  any,
   oneOf,
   oneOfType,
   bool,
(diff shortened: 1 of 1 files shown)


Jefoss:
@yavorsky technically there are more things the tooltip prop can accept thats why i went with the same type as there

nitayneeman:
Let's wait until the build finishes successfully before merging

src/PopoverMenu/PopoverMenu.js
```diff
@@ -11,6 +11,8 @@ import {
   oneOfType,
   bool,
   element,
+  func,
```
nitayneeman: Your build fails because you have a linting issue - `func` is imported twice

Jefoss: removed the import

PR #3666, opened 2019-07-02
Infra - use ui-autotools for ssr sanity tests

mosheKerbel:
@shlomitc 
@jonathanadler


src/Stepper/Stepper.meta.js
```diff
@@ -2,7 +2,26 @@ import Stepper from './Stepper';
 import Registry from '@ui-autotools/registry';
 
 const metadata = Registry.getComponentMetadata(Stepper);
+metadata.nonEventListenerTestCompliant = true;
```
jonathanadler: Why is this needed?

mosheKerbel: this test is currently failing for react 15 becuase it uses Hydrate
need to modify the autotools code first.

mosheKerbel: they have a good fallback for this in the SSR but in this test they don't

jonathanadler: But isn't it `nonReactStrictModeCompliant` that fixes that?
I thought `nonEventListenerTestCompliant` is different, but I might be totally wrong.

mosheKerbel:
merged.
failed due to 1 flaky unrelated timeout test.

PR #3675, opened 2019-07-03
<Modal /> - add unidriver

justenor:


src/Modal/Modal.spec.js
```diff
@@ -3,214 +3,249 @@ import eventually from 'wix-eventually';
...
         );
-        expect(driver.getContentStyle().maxHeight).toBe('100vh');
+        expect(await driver.getContentLabel()).toEqual(props.contentLabel);
       });
 
-      it('content position should be relative', () => {
-        const { driver } = render(<Modal {...props} />);
-        expect(driver.getContentStyle().position).toBe('relative');
+      describe('maxHeight', () => {
+        it('should render maxHeight passed in props', async () => {
```
justenor: @mosheKerbel I've split this test in two separate tests because it was buggy: driver always gets first modals' content, so this test was only checking if first modal was correctly rendered and was passing just because props were identical to both modals.

src/Modal/Modal.uni.driver.js
```diff
@@ -0,0 +1,37 @@
...
+        : null,
+    isScrollable: async () =>
+      !(await getPortal().hasClass('portalNonScrollable')),
+    closeButtonExists: () => getCloseButton().exists(),
+    clickOnOverlay: () => getOverlay().click(),
+    clickOnCloseButton: () => getCloseButton().click(),
+    getContentStyle: async () => await ReactBase(getContent()).getStyle(),
+    getContentLabel: () => getContent().attr('aria-label'),
+    getZIndex: async () => {
+      const style = await ReactBase(getOverlay()).getStyle();
```
mosheKerbel: avoid using ReactBase whenever possible
```
const style = await getOverlay()._prop('style');
```

mosheKerbel: also in function `getContentStyle`

justenor: Fixed

PR #3678, opened 2019-07-03
Fix typo: `checbox` -> `checkbox`

sidoruk-sv:
### 🔦 Summary
**NOTE: clickRowChecbox -> clickRowCheckbox can break somebodies tests**


sidoruk-sv:
@shlomitc, please review)

sidoruk-sv:
> Looks good.
> 
> However, because it's a breaking change and since it gonna be a while until we release a major release - we should apparently keep the old function as well for both drivers, but trigger a deprecation log when using it.
> 
> Basically it means to just use `deprecationLog` as described here:
> [/docs/internal/DEPRECATION_GUIDE.md@`master`](https://github.com/wix/wix-style-react/blob/master/docs/internal/DEPRECATION_GUIDE.md) [](/wix/wix-style-react/blob/HEAD@{2019-07-07T04:38:31Z}/docs/internal/DEPRECATION_GUIDE.md)

Cool, thanks! 
Did not know about this ability.

Deprecation warning added ✅

PR #3695, opened 2019-07-05
Add missing puppeteer and protractor driver exports for components mi…

justenor:
…grated to unidriver


mosheKerbel:
👍 
i just realised we also need it in the equivalent d.ts files

Issue #3664, opened 2019-07-02
Add more control over the Text's tooltip in case of ellipsis


Jefoss:
# ✨ Feature Request

### 🔦 Scope
Text component with ellipsis

### 💁 Explanation
currently the  developer has no control over the ellipsis tooltip - (the props are hardcoded - https://github.com/wix/wix-style-react/blob/master/src/Text/ProxyText.js#L7-L10) and in some cases i believe there are things you would want to customize like hide/show delays etc..

### 🐾 Possible solution 

the solution could be pretty straightforward, propagate the tooltip props from the Text component to the tooltip component instead of having it hard coded.

### 👀 Severity
Low

mykas:
Feature developed.

shlomitc:
@mykas - is this issue still relevant?

mykas:
Hi, this feature is already developed and merged. Closing it.

PR #3701, opened 2019-07-05
`<Text/>` - expose tooltip props

mykas:
This PR will expose tooltip props for our Text component usage that wants to control ellipsis floating behaviour. https://github.com/wix/wix-style-react/issues/3664


ipanasenko:
Is it possible to also add `zIndex` prop? Do you need an issue for this?

src/Text/ProxyText.js
```diff
@@ -1,21 +1,72 @@
...
+  ]),
+  /** `ellipsis` prop. Whether to enable the flip behaviour. This behaviour is used to flip the Tooltips placement when it starts to overlap the target element. */
+  flip: bool,
+  /** `ellipsis` prop. Whether to enable the fixed behaviour. This behaviour is used to keep the Tooltip at it's original placement even when it's being positioned outside the boundary. */
+  fixed: bool,
+  /** `ellipsis` prop. Tooltip content placement in relation to target element */
+  placement: string,
+  /** `ellipsis` prop. Tooltip timeout value. */
+  timeout: number,
+  /** `ellipsis` prop. Tooltip content max width value. */
```
ipanasenko: why does it say "`ellipsis` prop" everywhere?

mykas: This is to distinguish that this effects only the ellipsis. We are currently developing a feature in our storybook to split exposed tooltip props table from the main api table.

mykas:
It would be better if you could open a PR for this feature.

src/Text/ProxyText.js
ipanasenko: Oh, okay. Makes sense

ipanasenko:
https://github.com/wix/wix-style-react/pull/3725

PR #3702, opened 2019-07-05
`<Text/>` - migrate text component to sections with new examples

mykas:
This PR is improving Text components documentation part.


mykas:
@milkyfruit Please check Text component examples. Be aware that this is not a UX story but rather AP API story as I understand we don't have index number for Text component?

PR #3712, opened 2019-07-08
<Stepper> - changes to docs

mosheKerbel:


mosheKerbel:
test/applitools - ran all 522 - the hook here is to a wrong applitools link
test3 - failed due to 1 unrelated flaky test.

PR #3723, opened 2019-07-09
Disabled state selections (round 3)

gilmeir:


src/RadioGroup/RadioGroup.scss
```diff
@@ -78,12 +78,13 @@ $radioButtonDisabledShadow: $B50 0 0 0 1px;
...
-    border-color: $D60;
-    background-color: $D70;
+    color: $D10-30;
+    border-color: $D10-30;
+    background-color: $D10-05;
   }
 
   &.checked .radio:after {
-    background-color: $D55;
+    background-color: $D10-30;
```
gilmeir: check if it should be 05

src/Slider/Slider.js
```diff
@@ -131,6 +131,9 @@ Slider.propTypes = {
```
gilmeir: 1. check with Lital about the knob cursor
2. visual test for disabled

gilmeir:
Eyes is failing due to one wrong diff. I rejected it, and releasing.

PR #3725, opened 2019-07-09
feat(Text): pass zIndex, hideDelay and showDelay prop to Tooltip

ipanasenko:


mykas:
Hey, thanks for the PR! Can you please add two more props: `hideDelay` & `showDelay`? https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/tooltip/Tooltip.tsx#L38

This would close one another issue: https://github.com/wix/wix-style-react/issues/3664

ipanasenko:
@mykas done

PR #3735, opened 2019-07-11
<InputWithOptions/> - Fixing dropdown textOverflow propagate to the input

adicicurel:
Adding e2e test

### 🔦 Summary
Fixing Dropdown textOverflow issue, but setting it to the input it contains.

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [x] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [ ] Visual test


src/InputWithOptions/InputWithOptions.e2e.js
```diff
@@ -75,6 +76,25 @@ describe('InputWithOptions', () => {
     });
   });
 
+  describe('textOverflow clip', () => {
```
nitayneeman: Actually we don't really benefit from these tests a lot since we don't use `eyes` at all. Moreover, we're trying to not add new E2Es (because these are pretty slow tests) as long as we've other options.

Could you move these tests to InputWithOptions.spec.js?

src/InputWithOptions/docs/tests/TestTextOverflow.js
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
```
nitayneeman: In case we move the tests to InputWithOptions.spec.js, this file is redundant

mosheKerbel: > Overall looks good.
> 
> Notice this issue which happens when you hover the input and scroll horizontally:
> issue
> 
> It should be fixed before we merge.

@adicicurel @nitayneeman  no need to handle, this is a known issue in chrome
https://bugs.chromium.org/p/chromium/issues/detail?id=596804

package.json
```diff
@@ -134,7 +134,7 @@
     "commander": "^2.19.0",
     "concurrently": "^3.4.0",
     "copy": "^0.3.2",
-    "enzyme": "^3.3.0",
+    "enzyme": "^3.10.0",
```
mosheKerbel: plz revert

PR #3739, opened 2019-07-11
<NumberInput/> - reduce bundle size

NitayRabi:
- [x] Consume lodash function directly

image


src/NumberInput/NumberInput.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import _ from 'lodash';
+import isNil from 'lodash/isNil';
```
argshook: @NitayRabi 	
is it really needed at all?
`isNil` is just `item == null`
https://github.com/lodash/lodash/blob/master/isNil.js

PR #3742, opened 2019-07-11
Docs: Simplify Page component docs

shlomitc:
Simple is better


src/Page/docs/index.story.js
```diff
@@ -77,69 +92,266 @@ export default {
...
+                    Page Content
+                  </Page.Content>
+                </Page>
+                `,
+              compact: false,
+            },
+            {
+              title: 'Stretch to full size',
+              description:
+                'A Page stretches according to its container height. use `calc(100vh)` for a standalone page or `calc(100vh - 48px)` if top bar exists',
```
nitayneeman: Using `calc` for standalone is redundant - it could be just `100vh`

src/Page/docs/index.story.js
```diff
@@ -77,69 +92,266 @@ export default {
...
+              source: `
+                <div style={{height: 'calc(40vh)'}}>
```
nitayneeman: Also here and in all examples below - it could be just `height: '40vh'`

src/Page/docs/index.story.js
shlomitc: done

src/Page/docs/index.story.js
shlomitc: done

PR #3747, opened 2019-07-12
Docs: Improve the Card Layout story and add a Card doc

shlomitc:
- Make the 2.2 Card Layout more user-friendly
- Add a standalone Card documentation page


stories/CardLayout/index.story.js
```diff
@@ -0,0 +1,75 @@
...
+        ],
+      }),
+    ]),
+
+    importExample(`
+import Card from 'wix-style-react/Card';
+const { Header, Subheader, Content, Divider} = Card;
+    `),
+
+    title('Layout'),
```
nitayneeman: Maybe the title should be "Examples", as we do in all other places (or "Layout Examples")? Because I don't think this title is so informative

PR #3754, opened 2019-07-15
<Checkbox/> - UniDriver changes to support more than jsdom

mosheKerbel:
* Use data attr instead of css selectors to avoid wrong hashing that not match between diff. platforms.
* Use `label` driver without the `vanilla` testkit wrapper.
* Tooltip internal driver for checkbox is still using the `vanilla` testkit wrapper because it is less important for the crash course. This will be done in a separate effort, because this case of drivers of drivers that use `body` param, need to be solved using a shared util that will create `body` param based on the test platform (more time consuming then what we have at this scope).

@shlomitc


src/Checkbox/Checkbox.uni.driver.js
```diff
@@ -46,14 +48,18 @@ export const checkboxUniDriverFactory = base => {
      */
     hasFocusState: () => base.attr('data-focus'),
     isChecked,
-    isDisabled: () => base.hasClass('disabled'),
-    isIndeterminate: () => base.hasClass('indeterminate'),
-    hasError: () => base.hasClass('hasError'),
+    isDisabled: async () =>
```
shlomitc: let's extract the magic strings of `data-disabled` and similars.
also, consider not having `true` and `false` as statics

PR #3756, opened 2019-07-15
use wuf to generate protractor testkits exports

argshook:
This PR uses `wuf export-testkits` to generate `testkits/protractor.js` file.

that file has a LOT of inconcistencies and unfortunately requires a lot
of configurations.

this is an unfortunate but necesary step towards full automation.

the configs will be removed as each component clears out it's
snowflakynes, like for example:

* some testkits are exported named, some default
* some have unidriver protractor but legacy enzyme/vanilla or vice-versa
* some testkits exported with different name than component
* some testkits have weird paths
* etc...

almost all of the snowflakynes can be mitigated and don't affect
wix-style-react user at all. However, this is a job for another PR.


.wuf/testkits/protractor.template.ejs
```diff
@@ -0,0 +1,152 @@
...
+export const tooltipTestkitFactory = protractorTestkitFactoryCreator(
+  tooltipDriverFactory,
+);
+
+export const TooltipTestkit = protractorUniTestkitFactoryCreator(
+  tooltipNextDriverFactory,
+);
+<%
+const order = [
+  'Input',
```
argshook: this is temporary, so that `testkit/protractor.js` diff is smaller and easier to review

.wuf/testkits/definitions.js
```diff
@@ -85,7 +91,11 @@ module.exports = {
   Header: {
     testkitPath: "../src/Card/Header/Header.driver",
     // TODO: this is actually  Card.Header, but is exported just as header
-    skipSanityTest: true
+    skipSanityTest: true,
+    namedTestkitExport: true,
```
argshook: all of the `namedTestkitExport` should be gone eventually. the reason is that some testkits use `export default`, some have names and those are not consistent also.

.wuf/testkits/definitions.js
```diff
@@ -192,119 +231,196 @@ module.exports = {
...
+  },
+
+  TimeInput: {
+    namedTestkitExport: true,
+    testkitImportName: "timeInputUniDriverFactory",
+    protractorTestkitPath: "../src/TimeInput/TimeInput.uni.driver"
+  },
+
+  Notification: {
+    protractorUnidriver: true,
```
argshook: there are cases when protractor exports unidriver, but enzyme/vanilla is still legacy. Thus, more config

testkit/protractor.js
```diff
@@ -135,10 +162,10 @@ export const fieldWithSelectionCompositeTestkitFactory = protractorTestkitFactor
   fieldWithSelectionCompositeDriverFactory,
 );
 
-import googleAddressWithInputDriverFactory from '../src/GoogleAddressInputWithLabel/GoogleAddressInputWithLabel.protractor.driver';
+import googleAddressInputWithLabelDriverFactory from '../src/GoogleAddressInputWithLabel/GoogleAddressInputWithLabel.protractor.driver';
 
 export const googleAddressWithInputTestkitFactory = protractorTestkitFactoryCreator(
```
argshook: notice the typo in the name? it can't be changed now, cause that's breaking change :(

argshook:
build is passing, despite the unrelated timeout.

This means the temporary `order` array in template can be removed.
It will make bigger diff, but it's because automated export has
different order, not because the exports are different

testkit/protractor.js
argshook: this is just one example to showcase why those configs are needed for automation. without config there would be breaking changes

PR #3760, opened 2019-07-16
Add new feature props to Carousel component

mokhovyk:
Add dots, variableWidth and className props to Carousel component


src/Carousel/Carousel.spec.js
```diff
@@ -307,6 +314,21 @@ describe('Carousel', () => {
...
+            dots={false}
+            images={[
+              { src: 'image1.jpg' },
+              { src: 'image2.jpg' },
+              { src: 'image3.jpg' },
+            ]}
+          />,
+        );
+
+        expect(await driver.isPageNavigationDotExists()).toBeFalsy();
```
NitayRabi: `toBe(false)` (more strict)

mokhovyk: > `toBe(false)` (more strict)

resolved

mokhovyk:
> Hey,
> 
> What's the use case for exposing these props (`dots`, `variableWidth`)?
> I'm wondering because this behavior is not defined in the design system at the moment

The use case is the implementation of carousel in discovery feed (here is a link to the design https://zpl.io/2v6Ym7P)

nitayneeman:
> > Hey,
> > What's the use case for exposing these props (`dots`, `variableWidth`)?
> > I'm wondering because this behavior is not defined in the design system at the moment
> 
> The use case is the implementation of carousel in discovery feed (here is a link to the design https://zpl.io/2v6Ym7P)

Hey @milkyfruit / @LittleTal,

The following Carousel specification is a bit different from what's specified in our design system:
image

https://zpl.io/ankKD8n

Are you familiar with that?

milkyfruit:
@nitayneeman it seems more than dots differ from existing component:

1. Previous and Next buttons overlay the content
2. Part of the future/past content is already visible
3. Button's style isn't part of Wix Style library

@mokhovyk would `variableWidth` solve # 1 and # 2 issues?

For a full context, current component specs are given below

Zeplin:
https://zpl.io/2EBGeGe

React:
https://wix-wix-style-react.surge.sh/?activeTab=Usage&addons=0&full=0&panelRight=0&selectedKind=Components&selectedStory=Carousel&stories=1

mokhovyk:
> @nitayneeman it seems more than dots differ from existing component:
> 
> 1. Previous and Next buttons overlay the content
> 2. Part of the future/past content is already visible
> 3. Button's style isn't part of Wix Style library
> 
> @mokhovyk would `variableWidth` solve # 1 and # 2 issues?

@milkyfruit `variableWidth` solves #2. With custom `className` there are options to solve #1

mokhovyk:
> 1. I think we should test the `variableWidth` prop somehow either - because it's part of the component API.
> 2. We should add informative examples in the story page that demonstrate how to use the `dots` &  `variableWidth` props (a dedicated example for each prop).

1.  `variableWidth` visual test was added
2. two more examples `dots` & `variableWidth` props  were added

nitayneeman:
> > 1. I think we should test the `variableWidth` prop somehow either - because it's part of the component API.
> > 2. We should add informative examples in the story page that demonstrate how to use the `dots` &  `variableWidth` props (a dedicated example for each prop).
> 
> 1. `variableWidth` visual test was added
> 2. two more examples `dots` & `variableWidth` props  were added

Great!
Let's wait until the build finishes and then I'll merge.

Issue #3681, opened 2019-07-03
InputWithOptions text caret moves to text start when pressing the up key


naamagr:
# 🐛 Bug Report

### 🏗 Relevant Components

InputWithOptions

### 😯 Current Behavior

When the user enters some text in the `InputWithOptions`, some options may appear in the drop-down dialog and the user can navigate between them by using the up and down keys. 
pressing the down key behaves as expected, but after pressing the up key, the text caret moves to text start, so the user cannot easily continue his search.

### 🤔 Expected Behavior

The caret stays where it was (like it does when pressing the down key )

### 👣 Steps to Reproduce

It can be reproduced easily in the storybook:
https://wix-wix-style-react.surge.sh/?selectedKind=Components&selectedStory=InputWithOptions&full=0&addons=0&stories=1&panelRight=0

1. Click on the search box 
2. Insert some text 
3. Move down and then up over the options with the up and down keys
4. The caret moves to text start

### 👀 Severity

Major
It affects search feature in business manager, and probably all users of InputWithOptions

gilmeir:
Hitting the `ArrowUp` key on `Input` component brings the cursor to the beginning of the text.
`InputWithOptions` composes `Input`, so in this case we will event.preventDefault() on `ArrowUp` key.

PR #3762, opened 2019-07-16
<InputWithOptions/> - fix Input cursor jumping to input start on ArrowUp

gilmeir:
#3681


src/InputWithOptions/InputWithOptions.driver.js
```diff
@@ -48,6 +48,8 @@ const inputWithOptionsDriverFactory = ({ element }) => {
       const { element: optionElm } = dropdownLayoutDriver.optionById(optionId);
       return !!optionElm().querySelector(`[data-hook=highlighter-${optionId}]`);
     },
+    enterText: text => inputDriver.enterText(text),
```
nitayneeman: Because we're already export the whole inputDriver (which is not so good practice) - these methods (`enterText `, `getCursorLocation`) are duplicated and actually redundant... 
Instead of wrapping them you can use directly in your test out of `inputDriver`.

In the future we're going to ditch the `inputDriver` from being exported and just export the relevant methods as you did - but at the moment,  having both cases, is a bit confusing...

What do you think?

src/InputWithOptions/InputWithOptions.uni.driver.js
```diff
@@ -53,6 +53,8 @@ export const inputWithOptionsUniDriverFactory = base => {
           .exists())
       );
     },
+    enterText: text => inputDriver.enterText(text),
```
nitayneeman: Also here

src/InputWithOptions/InputWithOptions.driver.js
gilmeir: Agree.
Following the talk in the weekly, since the driver already exists and is exported to our clients, we'll stick to being consistent rather then right ;)

PR #3763, opened 2019-07-16
move all wix-ui-core usages to standalone version

ranyitz:
### 🔦 Summary

This PR should move `wsr` to work without need to transpile stylable on the consumer side. In fact, it should just work with common webpack configs like `create-react-app`'s.

Full explanation in this repo - https://github.com/wix-a/wsr-output-standalone
_______________________

This is a wip, currently it replaces all usages of `wix-ui-core/dist` with `wix-ui-core/dist/standalone`.

All `--st-from` calls have been redirected to `src` rather than dist, as @barak007 instructed.

This version would be availble for testing under:

```
npm i wix-style-react@branch-3763-merge
```
_____________

The next step would be to output a standalone version like [has been done](https://github.com/wix/wix-ui/pull/1348) on `wix-ui-core`, but instead the existing one.


src/Tooltip/index.d.ts
```diff
@@ -3,7 +3,7 @@ import { WixComponentProps } from '../BaseComponents/WixComponent';
 import {
   AppendTo,
   Placement,
-} from 'wix-ui-core/dist/src/components/popover/Popover.d';
+} from 'wix-ui-core/src/components/popover/Popover.d';
```
shlomitc: not sure if this is correct

shlomitc:
@ranyitz 
1. let's just make sure that this transpilation script works good also for the `standalone` verson
https://github.com/wix/wix-style-react/blob/master/scripts/transpileSrc.js#L77
2. we still have the `wix-ui-backoffice` references, which also provides `.st.css` files. maybe we should have a standalone version there as well?

src/Tooltip/index.d.ts
ranyitz: That's correct, thanks!

ranyitz:
@shlomitc Yeah, we'll need to transform `wix-ui-backoffice` as well.

For now, when the current stage is done (only `wix-ui-core`) we'll be able to see if `wsr` tests are passing and we could start checking it on different projects.

In case everything goes smoothly we will add `wix-ui-backoffice` and then `wix-style-react` itself.

.storybook/webpack.config.js
```diff
@@ -1,12 +1,30 @@
 const merge = require("lodash/merge");
 const path = require("path");
 const wixStorybookConfig = require("yoshi/config/webpack.config.storybook");
+const { StylableWebpackPlugin } = require("@Stylable/webpack-plugin");
```
argshook:
```suggestion
const { StylableWebpackPlugin } = require("@stylable/webpack-plugin");
```

i believe this is a typo (capital S), noticed on clean install

shlomitc:
closing until relevant again

PR #3766, opened 2019-07-16
add batches to .ci_config for steps parallelisation

gilmeir:
I had a meeting with @dmgor about stabilizing our CI and about gaps I collected from the team.

##  Meeting summary
We want to split our build in CI into parallel parts.
We already have that today, but our goal is to:
1. Being able to define and change the different steps by ourselves: the name of the step (instead of "testX") and the command that is being run in this step.
2. Get direct support from CI, better understand the interface.

### Suggestion brought up by Dmitriy:
CI has a new feature in beta version, for splitting build into steps.

It's supported by CI, has been recently released, and is in beta phase, with a few clients using it.
For example, [guinea-pig](http://tc.dev.wixpress.com/project.html?projectId=Guineapig2_Guineapig2_Parallel) (the experiments web app frontend) project is using it.

All we need to do:
1. Create a new file named `.ci_config` in the root of the project (already exists in wsr). The file is a JSON file. Example:
```
{
  "batches": [
    "lint",
    "test:e2e",
    "test:unit"
  ]
}
```
2. Add there a list of steps to run in parallel.
3. The steps are names of npm scripts from our package.json, each step will be executed WITH `npm run ${stepName}`
4. Every step name in TC will be defined by the package.json script name you put in `.ci_config`
5. This solution will override the existing parallelisation we have.
6. This solution applies for both dev & pull-request team-cities.
7. It has better support from CI team, without needing the infra Dev-Ex team to bridge between us. 
8. Dmitriy suggested we move to their new solution, as it was developed to be used by all teams, instead of our custom solution done inside our infra group.

The steps are the same as today.


netanelgilad:
@shlomitc @dmgor 
Let's try to get this merged, and move to the automatic solution.
I'm going to change the PR to run the current script as in `test:1`, `test:2`, etc. And then after we see everything works well we can make the scripts more semantic with this PR https://github.com/wix/wix-style-react/pull/3979

PR #3769, opened 2019-07-17
<Input/> - replace old puppeteer driver with unidriver for testkit export

mosheKerbel:
@shlomitc


src/Input/Input.puppeteer.driver.js
```diff
@@ -1,3 +1,9 @@
+/*
```
shlomitc: If you have backward compatibility - just delete it

src/Input/Input.uni.driver.js
```diff
@@ -47,13 +47,14 @@ export const testkit = base => {
     hasClearButton: async () => await clearButtonNode.exists(),
     clickClear: async () => await ReactBase(clearButtonNode).click(),
     getValue: async () => await input.value(),
```
shlomitc: Worth adding a test for it (in the same expectation of getValue)

src/Input/Input.puppeteer.driver.js
mosheKerbel: i wanted to, but it is still used in another puppeteer driver so i leave it for now, until we replace the other puppeteer driver with unidriver

src/Input/Input.uni.driver.js
mosheKerbel: yep good point

mosheKerbel:
failed due to 3 flaky unrelated tests
merged.

PR #3773, opened 2019-07-17
feat(<ModalPreviewLayout />): POC | working example

david-pe:
### 🔦 Summary
...

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


src/ModalPreviewLayout/docs/index.story.js
```diff
@@ -0,0 +1,150 @@
...
+        >
+          Open Preview Modal
+        </Button>
+        <Modal
+          isOpen={this.state.isOpenFullScreenModal}
+          onRequestClose={closePreviewModal}
+          shouldCloseOnOverlayClick
+          contentLabel="Preview modal example"
+          scrollable={false}
+          scrollableContent={false}
```
nitayneeman: It seems like `shouldCloseOnOverlayClick`, `scrollable` and `scrollableContent` are redundant here because you apply this behavior as part of `<ModalPreviewLayout/>` either way

src/ModalPreviewLayout/docs/index.story.js
```diff
@@ -0,0 +1,149 @@
...
+          Open Preview Modal
+        </Button>
+        <Modal
+          isOpen={this.state.isOpenFullScreenModal}
+          onRequestClose={closePreviewModal}
+          contentLabel="Preview modal example"
+          scrollable={false}
+          scrollableContent={false}
+        >
+          <ModalPreviewLayout
```
nitayneeman: Can you improve this example so it will be close to the spec?

One/two buttons as actions and card inside the content, nothing more:

image

src/ModalPreviewLayout/ModalPreviewLayout.st.css
```diff
@@ -0,0 +1,38 @@
...
+.title {
+  max-width: 198px;
+}
+
+.innerOverlay {
+  display: flex;
+  justify-content: center;
+  top: 100px;
+  position: relative;
+  margin-bottom: 100px;
```
nitayneeman: According to sepc the margin on the top & bottom should be 24px

src/ModalPreviewLayout/test/ModalPreviewLayout.spec.js
```diff
@@ -0,0 +1,94 @@
...
+
+  it('should render the preview content', async () => {
+    const contentDataHook = 'preview-content';
+    const props = {
+      ...requiredProps,
+      children: <div data-hook={`${contentDataHook}`} />,
+    };
+    const { driver } = render(<ModalPreviewLayout {...props} />);
+
+    expect(await driver.contentByDataHook(contentDataHook)).not.toBeNull();
```
nitayneeman: I think that in order to make this test more reliable, we should verify the content instead of ensuring it's not null.

For example:
```javascript
 const props = {
      ...requiredProps,
      children: <div data-hook={`${contentDataHook}`}>Some content</div>,
    };
```

And then:

```suggestion
    expect(await driver.contentByDataHook(contentDataHook).text()).toBe('Some content');
```

src/ModalPreviewLayout/test/ModalPreviewLayout.spec.js
```diff
@@ -0,0 +1,94 @@
...
+
+  it('should render the title', async () => {
+    const contentDataHook = 'preview-title';
+    const props = {
+      ...requiredProps,
+      title: <div data-hook={`${contentDataHook}`} />,
+    };
+    const { driver } = render(<ModalPreviewLayout {...props} />);
+
+    expect(await driver.contentByDataHook(contentDataHook)).not.toBeNull();
```
nitayneeman: Same here - more reliable expectation

src/ModalPreviewLayout/test/ModalPreviewLayout.spec.js
```diff
@@ -0,0 +1,94 @@
...
+
+  it('should render the actions', async () => {
+    const contentDataHook = 'preview-actions';
+    const props = {
+      ...requiredProps,
+      actions: <div data-hook={`${contentDataHook}`} />,
+    };
+    const { driver } = render(<ModalPreviewLayout {...props} />);
+
+    expect(await driver.contentByDataHook(contentDataHook)).not.toBeNull();
```
nitayneeman: Same here - more reliable expectation

src/ModalPreviewLayout/test/ModalPreviewLayout.visual.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import ModalPreviewLayout from '../ModalPreviewLayout';
+
+const commonProps = {
+  //use for repeated props across the tests (e.g. {buttonText: 'example'})
```
nitayneeman: Please pass here title and children

src/ModalPreviewLayout/test/ModalPreviewLayout.visual.js
```diff
@@ -0,0 +1,29 @@
...
+};
+
+const tests = [
+  {
+    describe: 'sanity', // prop name (e.g. size)
+    its: [
```
nitayneeman: I think we should test the following cases in visual:

1. Modal with title and simple content (inner content that doesn't take the full width)
2. Modal with title and actions
3. Modal with title and content that takes the full width

src/ModalPreviewLayout/test/ModalPreviewLayout.spec.js
david-pe: I copied from the messagebox specs, so it's a trade-off with consistency

nitayneeman: We've a lot of practices in the project and apparently will not achieve absolute consistency in the near future...

In this case I'd go with the better solution

david-pe: sure, np

src/ModalPreviewLayout/test/ModalPreviewLayout.visual.js
david-pe: I'll import the scrollable example
but we need to click on the button
do you know of an easy way for this in visual tests?

src/ModalPreviewLayout/docs/index.story.js
```diff
@@ -0,0 +1,149 @@
...
+
+          columns([
+            importExample(
+              "import ModalPreviewLayout from 'wix-style-react/ModalPreviewLayout';",
+            ),
+          ]),
+
+          divider(),
+
+          title('Examples'),
```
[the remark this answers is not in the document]
david-pe: removed examples since we can't render the fullscreen within the example area

src/ModalPreviewLayout/docs/index.story.js
```diff
@@ -0,0 +1,149 @@
...
+            shouldCloseOnOverlayClick
+            title={<div>Preview modal</div>}
+            actions={
+              <div
+                style={{ height: 20, width: 20, backgroundColor: 'white' }}
+              ></div>
+            }
+          >
+            <div
+              style={{ background: 'white', width: '600px', height: '2000px' }}
```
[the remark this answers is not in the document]
david-pe: our content is always fixed width, good enough for now

src/ModalPreviewLayout/test/ModalPreviewLayout.visual.js
david-pe: only #1 for now

src/ModalPreviewLayout/ModalPreviewLayout.uni.driver.js
```diff
@@ -0,0 +1,24 @@
...
+      base.$('[data-hook="preview-modal-inner-overlay"]').click(),
+
+    /** Get the title node */
+    previewTitle: () => base.$('[data-hook="preview-modal-title"]'),
+
+    /** Get the actions node */
+    previewActions: () => base.$('[data-hook="preview-modal-actions"]'),
+
+    /** Get the content node */
+    previewContent: () => base.$('[data-hook="preview-modal-content"]'),
```
nitayneeman: It's a bit petty, but we've sort of convention of "get" prefix for getters, as you can see in any other driver:
https://wix-wix-style-react.surge.sh/?activeTab=TestKits&addons=0&full=0&panelRight=0&selectedKind=9.%20Modals&selectedStory=9.3%20Custom%20Modal&stories=1

For example: `getTitle`, `getActions`, `getContent`

src/ModalPreviewLayout/test/ModalPreviewLayout.visual.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import ScrollableContentPreviewModal from '../docs/examples/ScrollableContent/ScrollableContent';
```
nitayneeman: Not sure it's a good idea to reuse the docs example. It means that every time we'll change the example, the test will fail.

So instead of coupling, I'd just provide the props for basic visual sanity example

src/ModalPreviewLayout/test/ModalPreviewLayout.visual.js
nitayneeman: You can use the driver to open the modal.

See how we do interactive visual tests here:
https://github.com/wix/wix-style-react/blob/master/src%2FRichTextInputArea%2Ftest%2FRichTextInputArea.visual.js#L139

src/ModalPreviewLayout/ModalPreviewLayout.uni.driver.js
david-pe: what a waste of chars but ok

src/ModalPreviewLayout/test/ModalPreviewLayout.visual.js
david-pe: to test this you need more than the component, you need <Modal />
I see the <MessageBox /> visual tests used this method

nitayneeman: What do we need besides the component? We're only talking about rendering it with basic props (even not within `<Modal/>`, but just as is).

This test basically checks the API and also that everything is rendered in the right place.

src/ModalPreviewLayout/docs/index.story.js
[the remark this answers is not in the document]
milkyfruit: Even it's impossible to render in the example area, I think we must display examples in some format. It could show static snippet with a small description and a CTA button to open a preview.

ZilvinasMaciulis:
UX here. 

Despite a couple of comments, the thing looks nice. 

1. Animation - using standard modal fade/scale animation looks weird for this type of modal. We need to simplify it to fade in only.

2. Can't see the example with navigation left/right.

3. Scroll bar should not hide behind the header. Instead, the scroll bar should be squeezed between the header and bottom of the browser. 

4. Bg Color Overlay - should be  D10, Alpha 96% for better content exposure.

5. Spacings between the buttons should be 12px.

6. Action area left to right:
* The secondary button is a Text Button, Small, Light, Underline: None, Affix: Prefix.
* Primary action button is a Button, Small, Light, Secondary.
* More action button is an Icon Only Button, Small, Light Secondary.

7. Also, this wasn't mentioned before but there might be a need to display a loader for bigger files. We could use 11.6 Loader of size Medium without a label.

nitayneeman:
> UX here.
> 
> Despite a couple of comments, the thing looks nice.
> 
> 1. Animation - using standard modal fade/scale animation looks weird for this type of modal. We need to simplify it to fade in only.
> 2. Can't see the example with navigation left/right.
> 3. Scroll bar should not hide behind the header. Instead, the scroll bar should be squeezed between the header and bottom of the browser.
> 4. Bg Color Overlay - should be  D10, Alpha 96% for better content exposure.
> 5. Spacings between the buttons should be 12px.
> 6. Action area left to right:
> 
> * The secondary button is a Text Button, Small, Light, Underline: None, Affix: Prefix.
> * Primary action button is a Button, Small, Light, Secondary.
> * More action button is an Icon Only Button, Small, Light Secondary.
> 
> 7. Also, this wasn't mentioned before but there might be a need to display a loader for bigger files. We could use 11.6 Loader of size Medium without a label.

Thanks for the review!

Regarding 1 - how's it different from the standard modal and what's special in this case?

Regarding 2 - the navigation buttons aren't part of the scope of this task 

Regarding 3-6 - @david-pe can you adjust according to these instructions?

Regarding 7 - OK, I'll add it to our backlog

src/ModalPreviewLayout/docs/index.story.js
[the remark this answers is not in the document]
nitayneeman: I agree with @milkyfruit.

@david-pe I'll take this as a task and improve the examples

ZilvinasMaciulis:
1 - the main difference is the header.  Previous modals don't have a  header.  They are smaller centered objects positioned in the center of the page, even a very wide and tall modal of the media manager is centered. Having them appear from the center feels natural. 

In the preview modal, the width of the header occupies viewport from coast to coast. The moment of "docking" the header into its position feels weird. Probably because of the gap between the viewport and the header. Also, this animation is not common for headers. Headers do not scale up from the page center.

nitayneeman:
> 1 - the main difference is the header. Previous modals don't have a header. They are smaller centered objects positioned in the center of the page, even a very wide and tall modal of the media manager is centered. Having them appear from the center feels natural.
> 
> In the preview modal, the width of the header occupies viewport from coast to coast. The moment of "docking" the header into its position feels weird. Probably because of the gap between the viewport and the header. Also, this animation is not common for headers. Headers do not scale up from the page center.

Okay, got your point. My only concern here is the fact that the animation changes from the various modals and actually not consistent. Are we fine with that?

Maybe we should apply only fade-in for all modal types?

ZilvinasMaciulis:
Although being a modal "Preview Modal" has a different appearance, so it's ok to have sort of different animation. I don's see a need to change the animation of other modals. Scale/fade works fine for all of them.

PR #3775, opened 2019-07-18
Feat/wuf generate puppeteer testkits exports

argshook:
This is similar to PR that automates protractor testkits exports
https://github.com/wix/wix-style-react/pull/3756 but this time for puppeteer

the goal is to generate exactly the same looking file as the current
manual one.

because of that, `.wuf/testkits/definitions.js` has grown really big,
but that's to accomodate all the different variations each component
has and to keep generated output the same.

many of these definitions are easily removable, but that's the task for another PR.


argshook:
all tests pass which means `order` can be removed from puppeteer.template.ejs

it will make diff bigger but not because there are changes but because order of testkits exports in generated file is different

PR #3777, opened 2019-07-19
<Breadcrumbs/> - fix UniDriver

mosheKerbel:
use data-attributes instead of ReactBase and class lookups
to support more tests platforms and not only jsdom.

@shlomitc


mosheKerbel:
failed due to 1 related flaky test, meging.

PR #3785, opened 2019-07-21
Split e2e tests

yanivefraim:
This is a POC for splitting e2e tests into two: tests starting with the letters: `a-H` and `I-Z`.
It should wait for this [PR](https://github.com/wix/yoshi/pull/1235) to be merged first (and a new Yoshi version to be released)


protractor1.conf.js
```diff
@@ -0,0 +1,13 @@
+module.exports.config = {
+  specs: [
+    'test/{A..H}**/**/*.e2e.js',
+    'src/{A..H}**/**/*.e2e.js',
+    'test/{a..z}**/**/*.e2e.js',
```
gilmeir: Here are the combinations as I understand we want to cover. 
Here you catch all a-z

protractor1.conf.js:
1. a-h
2. A-H

protractor2.conf.js:
3. i-z
4. I-Z 

Also, I suggest renaming the files protractor-a-to-i.conf.js, protractor-h-to-z.conf.js
so it's a bit more meaningful

mosheKerbel:
To make sure all tests are executed i suggest after the next normal build, currently it's building one, we count down the tests executed in applitools and compare it to another PR without the split.

mosheKerbel:
@yanivefraim notice that i pushed force to resolve the conflict

mosheKerbel:
ok applitools executed all tests - 528, as it should.
i'm just checking now about the failing tests - in test1
i suspect the latest Yoshi brought us other updated react dependencies that casuse there warnings which fail the tests

mosheKerbel:
> ok applitools executed all tests - 528, as it should.
> i'm just checking now about the failing tests - in test1
> i suspect the latest Yoshi brought us other updated react dependencies that casuse there warnings which fail the tests

fixed in other PRs and test1 now pass
merging.

Issue #3555, opened 2019-06-11
<SectionHelper/> / <FloatingNotification/> - change design styles


shlomitc:
Follow the docs to change the styling of the components
https://docs.google.com/spreadsheets/d/1CIf6II23mi0_A8O9GSNsP1F0yrrPl3ljeRJe9ciT2r8/edit?usp=sharing

shlomitc:
done

PR #3786, opened 2019-07-21
<SectionHelper/>, <FloatingNotification/> - change themes colors

gilmeir:
#3555


src/Tag/Tag.scss
```diff
@@ -21,6 +21,10 @@
   }
 }
 
+.text {
+  color: $D20 !important;
```
nitayneeman: Check if it's possible to override in a better way

gilmeir: I ended up using `<Text/>` API to set that color, on css.
Using `secondary` gives text color (D20),
and using `light` gives text color only for the dark-theme (D80 === white)

So we're good. I'm merging soon

Issue #3691, opened 2019-07-04
type definitions file of exported testkits should be automated


argshook:
# ✨ Feature Request

Currently these files are manual:

- `testkit/index.d.ts`
- `testkit/enzyme.d.ts`
- `testkit/protractor.d.ts`
- `testkit/puppeteer.d.ts`

### 🔦 Scope

related to users using typescript and wix-style-react testkits

### 💁 Explanation

these files should be automated because writing them manually is error prone, forgetful and just boring

### 🐾 Possible solution 

`wuf export-testkits --template .wuf/testkits/types.template.ejs` for each platform

### 👀 Severity

- Low

PR #3796, opened 2019-07-22
use wuf to generate testkits type definitions

argshook:
This PR uses `wuf` to automate:

1. `testkit/*.d.ts`
1. `testkit/beta/*.js`
1. `testkit/beta/*.d.ts`

closes https://github.com/wix/wix-style-react/issues/3691

this is the last PR which automates everything `testkit/*`! :tada:


.wuf/testkits/enzyme-typescript.template.ejs
```diff
@@ -0,0 +1,58 @@
...
+<%
+  const typed = {
+    Avatar: 'EnzymeUniTestkitFactory<AvatarDriver>',
+    Badge: 'EnzymeTestkitFactory<BadgeDriver>',
+    BadgeSelect: 'EnzymeTestkitFactory<BadgeSelectDriver>',
+    Box: 'EnzymeUniTestkitFactory<BoxDriver>'
+  };
+
+  const typeless = [
+    'ListItemAction',
```
argshook: this and other `typeless` arrays in other templates should be temporary.

current goal is to generate same file looking that we had before automation.

each component has it's own history and it's not a matter of this PR

PR #3798, opened 2019-07-22
POC(Perfer) - add perfer to run through our bundles

mykas:
Adding a perfer to threshold our component sizes.


package.json
```diff
@@ -43,6 +43,7 @@
     "test:e2e-runtime": "cd ./test-runtime/imports && NODE_ENV=production npm run test",
     "test:eyes-storybook": "eyes-storybook -c ./.storybook-test",
     "test:types": "tsc --noEmit -p tsconfig.declarations.json",
+    "test:size": "npm run generate-components-bundles && npx @wix/perfer --verbose",
```
Schniz: this should be `test:performance`, and maybe you need to do `npm run build` before that?

PR #3803, opened 2019-07-22
Updated width to support string.

Gongreg:
This allows to set width to 100% instead of just using the number.


nitayneeman:
Flaky tests, merging.

PR #3804, opened 2019-07-22
Cheatsheet- Tooltip family

TalZa1810:
### 🔦 Summary
`Tooltip` Family


stories/Introduction/Cheatsheet/ComponentsFamilies/TooltipFamily.js
```diff
@@ -13,74 +13,134 @@ import FloatingHelper from 'wix-style-react/FloatingHelper';
...
+          <PopoverMenu.MenuItem text="Edit" prefixIcon={<Edit />} />
+          <PopoverMenu.MenuItem
+            text="Delete"
+            prefixIcon={<Delete />}
+            skin="destructive"
+          />
+        </PopoverMenu>
+      </Box>
+    </Layout>
+    <Box height="150px" />
```
TalZa1810: I had to add this since otherwise, the `FloatingHelper` example was overlapping the`Popover` example. Let me know if you have a better suggestion :)

TalZa1810:
@ayeletr fixed :)

stories/Introduction/Cheatsheet/ComponentsFamilies/TooltipFamily.js
```diff
@@ -13,74 +13,156 @@ import FloatingHelper from 'wix-style-react/FloatingHelper';
...
           </Box>
-        }
-        initiallyOpened
-        placement="right"
-        target={<Box height="150px" />}
-      />
-    </SingleComponentSideBySide>
+          {showFloatingHelper && (
+            <FloatingHelper
+              target={<Box height="200px" />}
```
nitayneeman: Not sure what's the target here... I guess it should be the `<TextButton/>`?

stories/Introduction/Cheatsheet/ComponentsFamilies/TooltipFamily.js
```diff
@@ -13,74 +13,156 @@ import FloatingHelper from 'wix-style-react/FloatingHelper';
...
+              initiallyOpened
+              placement="right"
+              onClose={this.onFloatingHelperClose}
+              content={
+                <FloatingHelper.Content
+                  title="Don’t forget to setup payments"
+                  body="In order to sell your music you need to choose a payment method."
+                  actionText="Ok, Take Me There"
+                  onActionClick={() => null}
```
nitayneeman: Redundant?

stories/Introduction/Cheatsheet/ComponentsFamilies/TooltipFamily.js
```diff
@@ -13,74 +13,156 @@ import FloatingHelper from 'wix-style-react/FloatingHelper';
...
+          <PopoverMenu.MenuItem text="Add" prefixIcon={<AddSmall />} />
+          <PopoverMenu.MenuItem text="Edit" prefixIcon={<EditSmall />} />
+          <PopoverMenu.MenuItem
+            text="Delete"
+            prefixIcon={<DeleteSmall />}
+            skin="destructive"
+          />
+        </PopoverMenu>
+      </Box>
+      <Box>
```
nitayneeman: Not sure why we need these `<Box/>`s (also in line 79) as wrappers

stories/Introduction/Cheatsheet/ComponentsFamilies/TooltipFamily.js
nitayneeman: You can open the tooltip on the right, and define `minHeight` for the layout as well (instead of empty div)

stories/Introduction/Cheatsheet/ComponentsFamilies/TooltipFamily.js
```diff
@@ -13,74 +13,156 @@ import FloatingHelper from 'wix-style-react/FloatingHelper';
...
+);
+
+const PopoverExample = () => (
+  <SingleComponentSideBySide
+    name="7.2 Popover"
+    componentsNames={['<Popover/>', '<Popover.Content/>', '<Popover.Element/>']}
+  >
+    <Popover showArrow shown placement="right">
+      <Popover.Element>
+        <div />
```
nitayneeman: In general, popovers are aimed to be opened by an element, but here we have an empty one... It's okay that the popover is shown by default, but I think we should simulate a correct use case

Some text like "Popover on the bottom/right" is pretty enough

stories/Introduction/Cheatsheet/ComponentsFamilies/TooltipFamily.js
TalZa1810: The `<Box/>` was used to help with the positioning of the Popover. When using Both `<Layout/>` and `<PopoverMenu/>`, the popover is not placed at the bottom, even though I set the `placement` to `bottom`. It was either not using a `Layout` at all or using it with Boxes. The Layout is used to display the two menus aligned.

stories/Introduction/Cheatsheet/ComponentsFamilies/TooltipFamily.js
TalZa1810: I wanted to make sure that the `<PopoverMenu/>` items do not overlap the elements display below. Eventually, I wrapped them in a `<Box/>` (which as explained below was used also to make sure that the popover is displayed at the bottom of the button) and set each one 'minHeigh' prop. I think it's a more elegant way to make sure they don't overlap the `FloatingHelper` example.

stories/Introduction/Cheatsheet/ComponentsFamilies/TooltipFamily.js
TalZa1810: I added a "Hover Me" button and used the `onMouseLeave` and `onMouseEnter` events. WDYT?

TalZa1810:
As for the screenshot above (The `Popover` overlapping the names of the elements), this happens only when the Dev Mode is open (responsiveness).
I'll check how we can resolve this and fix this in a different PR.

stories/Introduction/Cheatsheet/ComponentsFamilies/TooltipFamily.js
nitayneeman: Sounds fine by me

PR #3817, opened 2019-07-24
<Sidebar/> - improve docs with index page

nitayneeman:
This PR demonstrates a BusinessManager-like sidebar using the sidebar visual components. Also, it adds a new category for "Navigation" stuff.

There are a few things we should be aware:

1. When the divider is in full mode, it has no margins (discussed with @wuwa)
2. We don't provide at the moment visual components such as the back button, timer or edit site button
3. A11y hasn't been implemented yet
4. We should think about a simpler way to implement the error suffix icon
5. @ayeletr suggested reducing the right margin to `24px` instead of `30px`, so the text will have more space 
6. The sidebar item has a transition for showing the chevron on hover but the fade-out currently doesn't work well

Link to the story page:
https://wix-wix-style-react-pr-3817.surge.sh/?selectedKind=6.%20Navigation&selectedStory=6.1%20Sidebar&full=0&addons=0&stories=1&panelRight=0

The example looks like:
Screen Shot 2019-07-24 at 12 01 15

Related to wix-private/wsr-issues#454


nitayneeman:
I'm attaching the feedback after reviewing with @ayeletr.

`<Sidebar/>`:

- [ ] The width should be 222px (and not 220px)

`<SidebarHeader/>`, `<SidebarDivider/>` & `<SidebarSectionTitle/>`:

- [ ] Change left and right margins to 24px

`<SidebarSectionItem/>`:

- [ ] The width of the text should be 154px
- [ ] The height of the text should be 36px (reduce the padding accordingly)
- [ ] Change right margin to 18px when there's a suffix
- [ ] Change right margin to 42px when there's no suffix
- [ ] Add prop that removes the chevron on hover
- [ ] Remove the ellipsis from the text (should be multiple lines in case of long text)
- [ ] Hide the chevron on hover in case of a suffix 
- [ ] The chevron colors should be always the same as the text

Issue #3821, opened 2019-07-24
There is no way to specify data-hook for PopoverMenu.MenuItem


splintor:
# 🐛 Bug Report

### 🏗 Relevant Components

`PopoverMenu` (beta)

### 😯 Current Behavior

`PopoverMenu` automatically sets the data-hooks of its child `MenuItem`s as `popover-menu-{id}` (https://github.com/wix/wix-style-react/blob/master/src/beta/PopoverMenu/PopoverMenu.js#L236)

### 👀 Severity

Major - this is currently breaking our automation tests - https://wix.slack.com/archives/CEV9B2VGD/p1563869779001000

mykas:
Hi, thanks for reporting this. After discussion we will surely add `dataHook`to next release.

splintor:
Sounds super logical.

dmitriweissmann:
This is also a concern for QA automaton. We use a data hooks or other attributes to identify the element and if the element in it's place.
For example in case of popover-menu-{id} we have no way to know what it is unless we start interaction with the element. for example it's a menu with 3 entries, currently there is no way to tell what is what and if there are all required and only the required.

mykas:
Make sense

mykas:
fixed.

PR #3822, opened 2019-07-24
`<PopoverMenu.MenuItem/>`, `<PopoverMenu.Divider/>` - beta - allow overwriting internal dataHook

mykas:
This PR will expose dataHook prop for PopoverMenu beta compound component Menuitem. Fixing: https://github.com/wix/wix-style-react/issues/3821


src/beta/PopoverMenu/test/PopoverMenu.private.uni.driver.js
```diff
@@ -3,12 +3,19 @@ import { dropdownBasePrivateDriverFactory } from '../../../DropdownBase/Dropdown
...
+        : `[data-hook="popover-menu-${option}"]`,
+    );
 
   return {
     ...publicDriverFactory(base),
     keyDownTrigger: key => keyDown(key),
     keyDownOption: async (option, key) => getOption(option).pressKey(key),
-    getMenuItem: async option => await getOption(option).getNative(),
+    getDivider: async dataHook => getOption('', dataHook).getNative(),
+    getMenuItem: async (option, dataHook) =>
```
splintor: Why do we need to pass `option` as menuItem? `dataHook` should be enough.
You should check if `option` is a number and then decide if you want to use it as index (and search for `popover-menu-${option}`) or as `dataHook`.
Another option is to make it explicit and have two separate methods: `getMenuItemByIndex` and `getMenuItemByLabel`.

But having an additional `option` parameter that is ignored if `dataHook` is passed looks like a strange API.

Also, you need to update `PopoverMenu.uni.driver.js` and `PopoverMenu.uni.driver.d.ts` to make it possible to call `clickAtChild()` with `dataHook` and not just with `index`.

mykas: Actually this is `private.uni.driver` which is not exposed. The method `clickAtChild()` is using DropdownBase infrastructure to click on the option by specifying the index number.

splintor: Any plans to add `getChildByDataHook()` or `clickAtChildByDataHook()`?

mykas: No probs, you can open a PR for this. I suggest reusing the `clickAtchild()` method to accept dataHook as second parameter. The first parameter should stay as `option`. There are users that uses in this way.

PR #3825, opened 2019-07-24
<Stepper/> - add key to div

mosheKerbel:
missing key gives warning in react16, adding this key solves it


mosheKerbel:
failed due to 3 unrelated flaky tests

PR #3827, opened 2019-07-24
<DataTable/> - forward ref

mosheKerbel:
In react 16 we are getting this warning:
`Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()`

wrapping the inner component with forwardRef solves it.
@shlomitc @nitayneeman


mykas:
But isn't React.forwardRef a feature of React 16? Are we stoping the support of React 15?

mosheKerbel:
> But isn't React.forwardRef a feature of React 16? Are we stoping the support of React 15?

good point! i missed that part
i will see if there is a pollyfill i'll add it and if not i will check what version is used of react we use

mosheKerbel:
@mykas fixed

src/DataTable/DataTable.js
```diff
@@ -303,6 +304,11 @@ class DataTable extends React.Component {
 
   getVirtualRowHeight = () => this.props.virtualizedLineHeight;
 
+  renderVirtualizedTableElementWithRefForward = () =>
+    React.forwardRef((props, ref) =>
```
mykas: I wonder if this can still throw an error if we use `autootols sanity`. can you run it through with react 15 installed?

mosheKerbel: > I wonder if this can still throw an error if we use `autootols sanity`. can you run it through with react 15 installed?

I ran `autotools sanity` it ran fine in react 15
but currently the only comp. that actually add a simulation test is the `Stepper`.
(using the sanity test for all other existing components is another effort we need to make)

mykas: Anyway, this looks to me like a hack ,but if it works it works.

mosheKerbel: i agree, the reason i'm fine with it is that the majority of our consumers are react16
and the component is actually works fine with both react15 & react16, as we did not get issues about it.

PR #3839, opened 2019-07-25
<DatePicker/> - fixes for react 16

mosheKerbel:
will serve us when we move to react16
@mykas @shlomitc


mosheKerbel:
3 failing tests that were already in master
merging.

PR #3845, opened 2019-07-28
Use support_ref_forward from a single place

mosheKerbel:
Using it from one place will make it easier to search for it and remove it once we move to next major version.


mosheKerbel:
failed due to 2 unrelated flaky tests
merging

PR #3850, opened 2019-07-29
<ColorInput/> - fix error & errorMessage

mosheKerbel:
Currently error message is never displayed, it is not tested so we did not notice it.
The reason?
error was not disturcted from props and was passed to `input` as part of `...rest` which somehow fucked it up.
@mykas @shlomitc


mykas:
I have no idea why do we have to pass: `error` and `status="error"` in order for `errorMessage` to work. It is quite confusing. And also I did not run test on this because we do not that features are are already tested on the used component.

mosheKerbel:
> I have no idea why do we have to pass: `error` and `status="error"` in order for `errorMessage` to work. It is quite confusing. And also I did not run test on this because we do not that features are are already tested on the used component.

well this is a solid proof that we do have to test it because it does not work.
regarding the other thing i agree, but it requires a change in the api which is a bit more time consuming whereas this change i did was instant and solves it.

mykas:
I don't think this change solves it. Now you miss `error` prop on Input. In storybook the errorMessage works only if you pass both: `status="error"` and `error`

mykas:
Honestly I think this should be solved on `Inputs` side. The `errorMessage` should work if status=`error` is passed.

argshook:
> Honestly I think this should be solved on `Inputs` side. The `errorMessage` should work if status=`error` is passed.

There was a discussion, multiple times in the past, that instead of
`error` and `errorMessage` we should do `status` and `statusMessage`.


`error` and `errorMessage` are marked as deprecated in Input for a long
time now.

However, due to double implementations and forever lasting deprecation,
we have this confusion.

we should move towards `status` and `statusMessage` wherever possible,
including this case.

so in this case, `status="error"` should be the fix

mykas:
I agree with but by only passsing `status="error"` is not enough here. The input will not show `errorMessage` if `error` prop is not passed too.

mosheKerbel:
> I don't think this change solves it. Now you miss `error` prop on Input. In storybook the errorMessage works only if you pass both: `status="error"` and `error`

true but this is what the current API says, so it does work as you would expect it to.

mosheKerbel:
> > Honestly I think this should be solved on `Inputs` side. The `errorMessage` should work if status=`error` is passed.
> 
> There was a discussion, multiple times in the past, that instead of
> `error` and `errorMessage` we should do `status` and `statusMessage`.
> 
> `error` and `errorMessage` are marked as deprecated in Input for a long
> time now.
> 
> However, due to double implementations and forever lasting deprecation,
> we have this confusion.
> 
> we should move towards `status` and `statusMessage` wherever possible,
> including this case.
> 
> so in this case, `status="error"` should be the fix

cool, changes like this should be tested as well and it takes more time.
can be done in a separate effort. 
for now we are getting a quick win with this PR to solve the bug and act like API declares.

argshook:
> for now we are getting a quick win with this PR to solve the bug and act like API declares.

sure thing, i just wanted to shed some history :)

PR #3858, opened 2019-07-30
Feat: add popoverProps prop to <ColorInput/>

kranux:
### 🔦 Summary

We want to have an ability to configure popover from `<ColorInput/>`. 

To achieve that we will pass prop `popoverProps ={{}}` to <ColorInput/>, for example:

```jsx

<ColorInput
  popoverProps={{
    zIndex: 2019
  }}
>
```
### ✅ Checklist
- [ x] 👨‍💻 API change is approved by the UI Infra Developers @mykas 

- 📚 Change is documented
  - [ x] API description


kranux:
@mykas we already have some `popover` prefixed props in `<ColorInput />` but I assume that it's not in the scope of current PR?

mykas:
We will need to leave those and later deprecated. The deprecation part I will handle later.

mykas:
@shlomitc Please review this part. This is the first component that will use common props pattern for passing popover props. 

**Pros**
- This pattern is great if only one popover or tooltip

**Cons**
- Not so great if multiple components popovers or tooltips inside.

I suggest using different pattern for components that are structured out of multiple system components: `use composition and expose render slots`.

src/ColorInput/components/ColorViewer.js
```diff
@@ -27,9 +27,11 @@ export class ColorViewer extends React.Component {
       onAddColor,
       addTooltipContent,
       placeholder,
+      popoverProps,
     } = this.props;
     return (
       <Popover
+        {...(popoverProps || {})}
```
mykas: I don't think you need this `|| {}` if you have defined the default props value. Just do `{...popoverProps}`.

kranux: Added defaultProps;

mykas:
Looks good. Lets merge it and release it so you could play with it.

Issue #3859, opened 2019-07-30
Add light <Text /> variation


LittleTal:
# 🐛 Bug Report

### 🏗 Relevant Components
`<Text/>`
...

### 😯 Current Behavior
Light disabled text variation is missing which cause text on dark bg to disappear (it uses the dark variation)

![image](https://user-images.githubusercontent.com/23282301/62131085-246db280-b2e3-11e9-947b-c22119a85a35.png)

...

### 🤔 Expected Behavior
Need to add disabled state variation for light text 
Text color should be D80-70

...

### 👣 Steps to Reproduce

  - **WSR version**: `...`
  - **Yoshi version**: `...`

### 👀 Severity
Critical

PR #3861, opened 2019-07-31
<Text/> - change disabled+light color

gilmeir:
#3859


gilmeir:
@LittleTal 
There are a few states missing in the Zeplin spec, so I'm listing them down here.
It refers only to `disabled` skin:
regular - D10
secondary - D20
light - D80-70
light + secondary - D80-70

LittleTal:
@gilmeir There are only 2 types of disabled, dark and light (primary and secondary on both cases should be the same) , I've updated [here](https://zpl.io/aNmZW5Q) the disabled for light

src/Text/Text.st.css
```diff
@@ -73,11 +73,15 @@
 }
 
 .root:skin(disabled),
-.root:skin(disabled):light,
 .root:skin(disabled):secondary {
   color: value(D10-30);
 }
 
+.root:skin(disabled):light,
+.root:skin(disabled):light:secondary {
```
gilmeir: Do I really need to add the :secondary selector? it's for both true/false

src/Text/test/Text.spec.js
```diff
@@ -85,17 +85,6 @@ describe('Text', () => {
         const { driver } = render(<Text light>Hello</Text>);
         expect(await driver.isLight()).toBe(true);
       });
-
```
gilmeir: Currently the component doesn't propagate `light` as is to the st.css.
I've changed that, `light` currently propagated as is,
and whether/how to style it - it's in the css

gilmeir: Also, we don't want to maintain visual testing as part as unit tests,
and we already have a `text variations` visual test to test all the permutations.

src/Text/test/Text.visual.js
```diff
@@ -74,7 +74,13 @@ storiesOf('Text/Styles', module).add('styles variations', () => (
             {Object.keys(SKINS).map(skin =>
               [false, true].map(light =>
                 [false, true].map(secondary => (
-                  <span style={{ margin: '12px' }}>
+                  <span
+                    style={{
+                      margin: '12px',
+                      backgroundColor:
+                        light && skin === 'disabled' ? 'black' : 'transparent',
```
gilmeir: dark background for light colors

src/Text/Text.st.css
```diff
@@ -77,8 +77,7 @@
   color: value(D10-30);
 }
 
-.root:skin(disabled):light,
-.root:skin(disabled):light:secondary {
+.root:skin(disabled):light {
```
gilmeir: @mosheKerbel 
it was unnecessary. 
merging.

PR #3865, opened 2019-07-31
make notification close button optional

ariel-wine:
### 🔦 Summary
...

### ✅ Checklist
- [✅ ] 👨‍🎨 UX change is approved by the Design System UX benb@wix.com
- 📚 Change is documented
  - [✅] Story
  - [✅] API description
- 🔬 Change is tested
  - [✅] Component
  - [✅] Visual test


src/Notification/docs/index.story.js
```diff
@@ -21,14 +21,27 @@ import typesReadme from './types.readme.md';
 const exampleChildren = [
   {
     label: 'Just text',
+    value: [<Notification.TextLabel>Notification text</Notification.TextLabel>],
```
argshook:
```suggestion
    value: <Notification.TextLabel>Notification text</Notification.TextLabel>,
```

there's a bug hidden somewhere in `wix-storybook-utils` which that `[object Object]` initially instead of `label`. As a quick workaround, which in this case works, passing value without array shows correct label

PR #3868, opened 2019-08-01
chore(package.json): remove `jscodeshift` dependency

### LINKED ISSUE #3864: Dependencies: `jscodeshift` produces deprecation warning for project that depends on wix-style-react
# ✨ Feature Request

### 🔦 Scope

For now, when installing the project with the dependency of `wix-style-react` user will get deprecation warning about `babel-preset-es2015`:

> npm WARN deprecated babel-preset-es2015@6.24.1: :raised_hands:  Thanks for using Babel: we recommend using babel-preset-env now: please read https://babeljs.io/env to update!

### 💁 Explanation

Deprecation warning about `babel-preset-es2015` comes from `jscodeshift` (that is a `dependency` of `wix-style-react`). 
As @argshook described, `jscodeshift` is used in `wix-style-react` for **internal** stuff only.

### 🐾 Possible solution

To eliminate that deprecation warning, @argshook suggest:
- ditch that dependency;
- or at least move it to devDependencies.

### 👀 Severity

- Low

argshook:
it is only needed for internal use and even then it's run with `npx`,
rendering `jscodeshift` dependency unnecessary

closes #3864


shlomitc:
Where’s the documentation change? Where is the devDependency?

argshook:
> Where’s the documentation change?

`migrate-to-icons-v2` script which uses `jscodeshift` is relevant only for people migrating from v4 to v5. That version has `jscodeshift` as `dependency` so no need to additionally install it

moreover, `migrate-to-icons-v2` script can be removed from codebase, cause, again, it's
only relevant for people migrating from v4 to v5. If we're going to remove that script,
there's no point for documenting it further.

> Where is the devDependency?

for `migrate-to-icons-v2` `devDependency` is not needed nor would it be helpful

for `npm run make-async-spec` `package.json` has this:

```
npx jscodeshift -t scripts/unidriver-migration/src/make-async/make-async.js
```

so, `npx` ensures `jscodeshift` will work even without entry in `devDependencies`

so, i take that neither docs change nor `devDed` is needed

PR #3873, opened 2019-08-01
UniDriver migrations

mosheKerbel:
purpose of this PR is to reduce the dependency of ReactBase and use regular UniDriver interface when we can.
@shlomitc


mosheKerbel:
failed due to 2 unrelated flaky tests
merged

PR #3863, opened 2019-07-31
temp test

mosheKerbel:
@argshook reproduction of the colorInput testkit in storybook.

PR #3880, opened 2019-08-04
<ColorInput/> - drivers & test changes

mosheKerbel:
PR main purpose is for `<ColorInput/>` to be supported in React 16.
Currently many tests fail for this component in React 16.
As part of the required changes i have touched other areas.

Following have been changed:
1. Datahooks extraction
2. `Size` prop - support `medium` in `<input/>` instead of converting `medium` to `normal`.
3. UniDriver refactor - get rid of the `...` of nested components. "Breaking changes" are accepted here because only 1 consumer is using this testkit. These changes does not brake him.
4. Currently auto teskit docs generation produces a blank page. https://github.com/wix/wix-style-react/pull/3863 arijus will check it. In the meanwhile i am using a manual testkit readme so i had to modify it.

@mykas @shlomitc


src/ColorInput/ColorInput.js
```diff
@@ -85,9 +85,10 @@ class ColorInput extends React.Component {
   static getDerivedStateFromProps(props, state) {
     if (!state.active && props.value !== state.value) {
       return {
-        ...state,
         value: extractHex(props.value),
       };
+    } else {
+      return null;
```
mykas: No need for else. Just do `return;`

src/Input/Input.driver.js
```diff
@@ -103,6 +103,7 @@ const inputDriverFactory = ({ element }) => {
     isDisabled: () => element.classList.contains(styles.disabled),
     isOfStyle: style => element.classList.contains(styles[`theme-${style}`]),
     isOfSize: size => element.classList.contains(styles[`size-${size}`]),
+    getSize: () => element.getAttribute('data-size'),
```
mykas: I think it is not a great idea to mix two components changes in one pr. In this pr you have added additional size `medium` for Input and also some new testkit method. This should be documented in the changelog later I'd say extract these changes and do another pr just for this.

mosheKerbel: will do

src/ColorInput/ColorInput.js
mosheKerbel: can get rid of the `else`, though i prefer  explicit `return null` over `return`

mykas: return; just returns undefined and in javascript we avoid `null` values. Be aware of these kind of issues: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/null#Difference_between_null_and_undefined

mykas: For example a place like this: https://github.com/wix/wix-style-react/blob/master/src/ColorInput/components/ColorViewer.js#L14 could go very wrong if null values comes in :)

mosheKerbel: `Warning: %s.getDerivedStateFromProps(): A valid state object (or null) must be returned. You have returned undefined.,ColorInput`
so it's either `return null` or `return {}`
`return {}` is fine with me

mykas: if it requires an object then sure.

src/Input/Input.driver.js
mosheKerbel: https://github.com/wix/wix-style-react/pull/3891

mosheKerbel:
failed due to unrelated flaky tests
merged

PR #3884, opened 2019-08-04
Cheatsheet- Picker family

TalZa1810:
### 🔦 Summary
Picker Family


stories/Introduction/Cheatsheet/ComponentsFamilies/PickerFamily.js
```diff
@@ -1,6 +1,467 @@
...
+    } else if (index < selectedIndex) {
+      selectedIndex = 1;
+    }
+    this.setState({
+      options: options.filter((option, i) => index !== i),
+      selectedIndex,
+    });
+  };
+
+  onRadioOptionToggle = index => {
```
mosheKerbel: is it really necessary to have 2 representations of the selectedIndex in the state? 
1. option.isSelcted
2. selectedIndex

stories/Introduction/Cheatsheet/ComponentsFamilies/PickerFamily.js
```diff
@@ -1,6 +1,467 @@
...
+  render() {
+    const { color } = this.state;
+    return (
+      <SingleComponentSideBySide
+        name="10.4 Color Picker"
+        componentsNames={['<ColorPicker/>']}
+      >
+        <Preview>
+          <ColorPicker
+            onCancel={() => 'Cancelled'}
```
mosheKerbel: i would either use an empty `noop` `func () => {}`
or actually use the info that i get and print it to the console `(e) => console.log('Cancelled', e);`
same for the confirm func.

mosheKerbel: and for other places as well

TalZa1810: It doesn't let me commit the `console.log`s. That's why I used these eventually. There's a preset hook that prevents from committing `console.log`s. The first option looked to me a bit redundant, no?

mosheKerbel: yeah this preset guards us, so our production code won't include such things
but this case is valid, because it's just a code example
so you can use
`/* eslint-disable no-console */` on top of the file.

stories/Introduction/Cheatsheet/ComponentsFamilies/PickerFamily.js
TalZa1810: I've removed the `selectedIndex` attribute from the state. Let me know what you think and if we can merge :)

PR #3894, opened 2019-08-06
<InputWithOptions /> testkit, fix when dropDirectionUp is true

audriuss:
### 🔦 Summary
<InputWithOptions /> testkit was failing when component with `dropDirectionUp` was rendered. This prop changes order of elements, but in testkit there was `element && element.childNodes[0]` which depends on the order. So changed testkit, that it would select by `data-hook`

### ✅ Checklist
- 🔬 Change is tested
  - [X] Component


mykas:
Will be fixed on this PR https://github.com/wix/wix-style-react/pull/3876

PR #3898, opened 2019-08-06
fix(<ModalPreviewLayout />): end scrollbar below header strip

david-pe:
### 🔦 Summary
...

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


src/ModalPreviewLayout/ModalPreviewLayout.st.css
```diff
@@ -11,10 +11,10 @@
 }
 
 .overlay {
-  align-self: start;
+  /* align-self: start; */
   width: 100vw;
   height: 100vh;
-  overflow-y: auto;
+  overflow-y: hidden;
   overflow-x: hidden;
```
nitayneeman: If you hide both `overflow-x` & `overflow-y` it could be just:
```suggestion
  overflow: hidden;
```

src/ModalPreviewLayout/ModalPreviewLayout.st.css
```diff
@@ -11,10 +11,10 @@
 }
 
 .overlay {
-  align-self: start;
+  /* align-self: start; */
```
nitayneeman: Redundant?

src/ModalPreviewLayout/docs/examples/ScrollableContent.js
```diff
@@ -56,8 +56,8 @@ export class ScrollableContentPreviewModal extends Component {
               </Box>
             }
           >
-            <Box>
-              <img src={ascendInvoice} />
+            <Box marginBottom="24px">
```
nitayneeman: Why do we need to do it here?

I think it should part of the layout, the same as we have `margin-top`

PR #3913, opened 2019-08-08
`<CardGalleryItem/>` - support background image as node element

danarish4:
### 🔦 Summary
...

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


src/CardGalleryItem/CardGalleryItem.js
```diff
@@ -165,17 +169,32 @@ class CardGalleryItem extends React.Component {
...
+                      {
+                        [styles.contentWithFooter]: footerExists,
+                      },
+                    )}
+                  >
+                    {backgroundImageNode}
+                  </div>
+                )
+              )}
+              {badge && this._renderBadge(badge)}
```
mykas: I am not sure if this is a good idea to move this part from the div to outside. There might be some tests running on this and it could potential break some testkit methods. Please double check it if this is not breaking.

src/CardGalleryItem/CardGalleryItem.js
```diff
@@ -165,17 +169,32 @@ class CardGalleryItem extends React.Component {
...
-                  [styles.contentWithFooter]: footerExists,
-                })}
-                style={{
-                  backgroundImage: `url(${backgroundImageUrl})`,
-                }}
-                data-hook={DataHooks.backgroundImage}
-              >
-                {badge && this._renderBadge(badge)}
-              </div>
+              {backgroundImageUrl ? (
```
mykas: This part became a `feature toggle` thing. For future maintainability make sure to extract this part so separate internal method for example: `renderBackground`.

mykas:
Thanks for your PR. Here are some things to do before this PR can be become ready to be merged:
- [ ] Make sure that background image stuff is being tested with unit tests. 
- [ ] Add an example in storybook that shows how node based item is used.

src/CardGalleryItem/CardGalleryItem.js
mykas: Also the switching part does not look great, because you switch between two different structures while instead you could actually do one div with conditional styles etc.

PR #3915, opened 2019-08-08
<StatisticsWidget /> - add handlers for mouse and keyboard actions

maxvektor:
### 🔦 Summary
`Tab` for navigation (only between active items)
`Hover` for background change
`Click`, `Enter`, `Space` for callback

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


src/StatisticsWidget/StatisticsWidget.js
```diff
@@ -35,10 +37,74 @@ class StatisticsWidget extends React.PureComponent {
...
+
+    return (
+      <div className={styles.root} data-hook={dataHook}>
+        {firstFive.map(this._renderStat)}
+      </div>
+    );
+  }
+}
+
+class StatisticsItem extends React.PureComponent {
```
maxvektor: It is getting big and stateful. Is it ok to make a separated component? 
Maybe we should move it to a different file?

src/StatisticsWidget/test/StatisticsWidget.spec.js
```diff
@@ -201,4 +201,41 @@ describe('StatisticsWidget', () => {
...
+
+    it('should not call onclick, when prop not passed', async () => {
+      data.statistics[0].onClick = undefined;
+      const { driver } = render(<StatisticsWidget {...data} />);
+      await driver.clickStatistics(0);
+
+      expect(onClick).toBeCalledTimes(0);
+    });
+
+    it('should call onClick', async () => {
```
maxvektor: As I understand, we don't need to test enter and space as far as we don't expose keypress in a driver.
Will try to test it in visual tests.

maxvektor:
There is no visual tests for now. Will try to write some.

src/StatisticsWidget/StatisticsWidget.js
```diff
@@ -35,10 +37,74 @@ class StatisticsWidget extends React.PureComponent {
...
+  static displayName = 'StatisticsItem';
+
+  componentDidMount() {
```
mosheKerbel: not sure why u need to use this lifecycle event and not simply add keyDown event to your main div in line 175
`onKeyDown={onClick ? this._getSpaceOrEnterHandler(onClick) : undefined}`

src/StatisticsWidget/test/StatisticsWidget.spec.js
maxvektor: I faced an issue while implementing this one. Tab button doesn't affect focus inside `jsdom`. 
https://github.com/jsdom/jsdom/issues/2102

>Firing events does not change focus, as you can see: https://jsbin.com/qiresuzisi/edit?html,console,output
Please actually test the code in browsers next time.

The common tip is to use `element.focus()` witch is exposed by unit driver only via `getNative`. I don't think it is a good thing to use it because it does not test really anything and also is implemented differently in a protractor.

I made some tests just be pressing enter on items but we need to deal with tabs in a different way.

src/StatisticsWidget/test/StatisticsWidget.private.uni.driver.js
```diff
@@ -14,5 +17,22 @@ export const statisticsWidgetPrivateDriverFactory = (base, body) => {
...
+      const stat = await getStatsItem(index);
+      await stat.pressKey('Enter');
+    },
+
+    pressSpaceKey: async index => {
+      const stat = await getStatsItem(index);
+      await stat.pressKey(' ');
+    },
+
+    hasTabIndex: async index => {
```
maxvektor: workaround to test focus with UT

src/StatisticsWidget/StatisticsItem/StatisticsItem.js
```diff
@@ -0,0 +1,143 @@
...
+import DataHooks from '../dataHooks';
+import DataAttrs from '../dataAttrs';
+
+import styles from './StatisticsItem.st.css';
+
+class StatisticsItem extends React.PureComponent {
+  static displayName = 'StatisticsItem';
+
+  _getFocusableProps = () => {
+    // add focusable hooks only when item is clickable
```
mosheKerbel: no need for comment, can be removed.

src/StatisticsWidget/StatisticsItem/StatisticsItem.js
```diff
@@ -0,0 +1,143 @@
...
+      ? {
+          onFocus: focusableOnFocus,
+          onBlur: focusableOnBlur,
+          tabIndex: 0,
+        }
+      : {};
+  };
+
+  _getSpaceOrEnterHandler = handler => event => {
+    if (event.defaultPrevented) {
```
mosheKerbel: can you explain the purpose? when does it get true?

src/StatisticsWidget/StatisticsItem/StatisticsItem.js
```diff
@@ -0,0 +1,143 @@
...
+      return;
+    }
+
+    const { key, keyCode } = event;
```
mosheKerbel: do we need both? isn't it enough to just use key? 
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key

src/StatisticsWidget/StatisticsWidget.js
```diff
@@ -42,7 +32,7 @@ class StatisticsWidget extends React.PureComponent {
     ),
   };
 
-  _renderStat = (stat, key) => <FocusableItem {...stat} key={key} />;
+  _renderStat = (stat, key) => <StatisticsItem {...stat} key={key} />;
```
mosheKerbel: not critical but can be aligned inside the `map` instead of extracting to this func...

maxvektor: It will create function on every render. If it is not critical, will do it.

src/StatisticsWidget/StatisticsItem/StatisticsItem.js
maxvektor: This is just a recommendation from the same artice: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key#Example

As I understood, is is possible to catch an event after it was prevented. Looks strange for me, maybe the deal is they using `true` as 3rd argument. Don't see any use of it, Will delete.

src/StatisticsWidget/StatisticsItem/StatisticsItem.js
maxvektor: IE dose not support this.
Screen Shot 2019-08-13 at 11 34 13 AM

maxvektor: just realized it is not about enter or spacebar. Will remove.

mosheKerbel:
failed to unrealted flaky tests

PR #3917, opened 2019-08-09
`<Dropdown/>` - improve documentation

mykas:
This PR will improve documentation part.


stories/Dropdown/storySettings.js
```diff
@@ -0,0 +1,6 @@
+import { Category } from '../storiesHierarchy';
+
+export const storySettings = {
+  category: Category.SELECTION,
```
shlomitc: I think this should sit under the `Components` category, with the story name of simply `Dropdown`

mykas: But this is UX story :)

mykas: Aren't we splitting two different stories?

shlomitc: let's drop the UX story for now. component API is much more important at this stage.
we have the Cheatsheet to bridge the gap between UX symbols and component APIs

src/Dropdown/docs/examples.js
```diff
@@ -0,0 +1,159 @@
+export const simple = `
```
shlomitc: Wouldn’t it be better to expose the examples as separate files? That way you can get a better linting

src/Dropdown/docs/index.story.js
```diff
@@ -1,49 +1,40 @@
...
-const options = [
-  { id: 0, value: 'Option 1' },
-  { id: 1, value: 'Option 2' },
-  { id: 2, value: 'Option 3' },
-  { id: 3, value: 'Option 4' },
-];
+const liveCode = config =>
+  code({
+    previewProps: {
+      style: { backgroundColor: '#f0f4f7' },
```
shlomitc: Is this needed? Shouldn’t default styles be enough?

src/Dropdown/docs/examples.js
mykas: I think linting in the examples is good enough. For me it is much simpler to just add an example under the string and leave it. The code looks cluttered when i see multiple files with examples and names not matching with what example is really doing.

src/Dropdown/docs/index.story.js
mykas: just a copy paste stuff.

src/Dropdown/docs/examples.js
mykas: As long as examples are small enough - linting is not an issue.

mykas: Screenshot 2019-08-14 at 10 18 49

shlomitc: worth checking if there's a way to write these examples as a valid js and import them in a way it can be stringified and be feed to the `code()` section

mykas: But how is this not valid? if it fails - it fails on storybook :) Anyway, the only time I would like wrap it with React component and show it for our users is when I need to use state. Other times I want to showcase the usage of the single component. From the interviews we gather with Domas - we got the idea that simple examples with minimal code are GREAT. So honestly I try to code like this, but we can discuss this when you come.

PR #3918, opened 2019-08-09
<StatisticsWidget />- tests for mouse and keyboard actions

maxvektor:
### 🔦 Summary
e2e tests

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


package.json
```diff
@@ -195,6 +195,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.3.1",
+    "babel-register": "^6.26.0",
```
maxvektor: remove

src/StatisticsWidget/test/StatisticsWidgetAccessibilityTest.js
```diff
@@ -0,0 +1,85 @@
...
+
+  render() {
+    return (
+      <div style={{ marginLeft: 100, marginTop: 100 }}>
+        <StatisticsWidget
+          dataHook={storySettings.dataHook}
+          statistics={this.state.statistics}
+        />
+        <ul>
+          {this.state.messages.map(text => (
```
maxvektor: did not found a better solution to test clicks.

mosheKerbel:
> but in our case, we are trying to avoid e2e tests as much as we can, as they are very expensive.
> 
> as an alternative this is how we do it:
> 
> 1. behavioural tests - unit tests
>    in your case - make sure the click callback is executed when it should.
> 2. visual tests
>    in your case - the hover style is shown on hover - you have examples in the stepper visual on how to do it
>    the focus style is shown - can also be done similar to the hover, with interactive tests.
> 
> with these you can cover all tests u did here, but without using e2e

if something is not clear enough on how to impl it, let me know.

PR #3927, opened 2019-08-12
chore: add .prettierrc.js with settings from "eslint-config-yoshi-base"

ipanasenko:
This will allow WebStorm to format files using Prettier rules.


.prettierrc.js
```diff
@@ -0,0 +1,12 @@
+let prettierConfig = {
```
mosheKerbel: i don't really understand what it does
i mean i didn't have something like that configured on my env.
and when i run lint it does change the quotes to single and verify trailing commas
i'm guessing this is doing it?
https://github.com/wix/wix-style-react/blob/b703f080155a62bbdaf20ff248ec53a92b37dc13/package.json#L271

ipanasenko: My PR is for Webstorm. There's hotkey in Webstorm for "Format using Prettier". Webstorm can't find Prettier settings (because they are hidden inside yoshi) and formats using default Prettier settings.
This PR allows Webstorm to use Prettier settings, defined in Yoshi config.

It doesn't change how yoshi, eslint, or whatever works, it's just to allow Webstorm to know about Prettier config

mosheKerbel: yeah i'm on Webstorm, will talk to in Slack

PR #3929, opened 2019-08-12
cheatsheet- adding links to TextInput and TextArea

TalZa1810:
### 🔦 Summary
I've added links to `Text Input` and `Text Area` as an example of implementation for the other `InputFamily`.


stories/Introduction/Cheatsheet/sharedComponents/ComponentsNames.js
```diff
@@ -1,20 +1,38 @@
...
-      >
-        <Text size="small" weight="thin" secondary>
-          {name}
-        </Text>
-      </Box>
-    ))}
+    {componentsNames.map((name, i) => {
+      let componentName;
+
+      if (typeof name === 'string') {
```
TalZa1810: I've added the `if else` statement in order not to break the current implementation. Once I'll finish with all the links, I'll remove it. The same thing for the `ComponentStorybookName` component.

TalZa1810:
I'm not sure about some of the variables and functions namings so I'd love to hear your feedback. Later on, after finishing with all the inputs, I thought about creating the `<SingleComponentSideBySide/>` wrapper dynamically (similar thing to what we've done with the `visual tests` files.

stories/Introduction/Cheatsheet/ComponentsFamilies/InputFamily.js
```diff
@@ -24,29 +31,43 @@ import FormField from 'wix-style-react/FormField';
...
-  >
-    <FormField label="Text Area">
-      <InputArea placeholder="Value" />
-    </FormField>
-  </SingleComponentSideBySide>
-);
+const { familyName, symbols } = inputFamilyMapping;
+
+const TextInputExample = () => {
+  const { UxStory, componentsNames } = symbols[0];
```
mosheKerbel: now u just have 2 but i imagine that later on you will have links to all the examples? could be even 10?
i think it's better to describe them with meaningful names instead of indexes inside an array

stories/Introduction/Cheatsheet/sharedComponents/ComponentStorybookName.js
```diff
@@ -2,18 +2,31 @@ import React from 'react';
...
-      secondary={!isDeveloped}
-      light={!isDeveloped}
-      size="medium"
-      weight="bold"
-    >
-      {name}
-    </Text>
-  </Box>
-);
+const ComponentStorybookName = ({ name, isDeveloped }) => {
```
mosheKerbel: it's a bit confusing becuase `name` is not really a name anymore, but a name with a url.
my suggestion - keep sending the name as a prop, but when needed, send also the url.
your if will then check for a url existence.

stories/Introduction/Cheatsheet/sharedComponents/ComponentsNames.js
mosheKerbel: i addressed this issue above.

stories/Introduction/Cheatsheet/urlsMapping/inputsUrlsMapping.js
```diff
@@ -0,0 +1,51 @@
+//3. Inputs
+import { familyNames } from './shared';
+
+const symbolsNames = {
+  Text_Input: '3.1 Text Input',
```
mosheKerbel: is there any reason to use "_" in these name?
it just grabs your eyes, while the other naming is done with CamelCase, which i think is enough.

stories/Introduction/Cheatsheet/ComponentsFamilies/InputFamily.js
TalZa1810: Later on, I plan to run on the array and add the props dynamically (something similar to the [visual tests](https://github.com/wix/wix-style-react/blob/master/src/IconButton/test/IconButton.visual.js#L7)).

The order of the symbols matters, that's why I used an array instead of an object.

mosheKerbel: i see
as long as we won't have long lists of indexed refs then sababa.

mosheKerbel:
test3-unrleated flaky test failed
merged.

stories/Introduction/Cheatsheet/urlsMapping/inputsUrlsMapping.js
TalZa1810: It's to distinguish between a component name and a UxStory.

mosheKerbel: i know
i'm not saying to remove the collection
i'm talking about the keys

mosheKerbel: `Text_Input` to become `TextInput`

PR #3922, opened 2019-08-11
`<RichTextInputArea>` - keep newlines of empty html elements in initial value

mosheKerbel:
resolves this [issue](https://github.com/wix/wix-style-react/issues/3768)
@nitayneeman 
@mykas

PR #3935, opened 2019-08-13
 <RichTextInputArea> - keep newlines of empty html elements in initial value

mosheKerbel:
@nitayneeman 
it was fixed [here](https://github.com/wix/wix-style-react/pull/3922) but then i [reverted](https://github.com/wix/wix-style-react/pull/3930) it
because the links were not converted, so i've added this code now and it gets the job done.
```
    const content = convertFromHTML({
      htmlToEntity: (nodeName, node, createEntity) => {
        if (nodeName === 'a') {
          return createEntity('LINK', 'MUTABLE', { url: node.href });
        }
      },
    })(value);
```

tests? well your snapshot tests are enough and should have caught the issue earlier, it's just that Applitools couldn't tell the diffs. This is why i opened the [issue](https://wix.slack.com/archives/CM5F7P8K1/p1565685156024300) to them


mosheKerbel:
test2 failed because it is not rebased to `statisticsWidget` , it is ok to merge.

PR #3953, opened 2019-08-15
`<ToggleButton/>` - new component

ipanasenko:
### 🔦 Summary
Two new components: `ToggleButton` and `ToggleButtonGroup`. Should be used together.

https://zpl.io/bz66eGz
image

### ✅ Checklist
- 📚 Change is documented
  - [x] Story
  - [x] API description
- 🔬 Change is tested
  - [x] Component
  - [x] Visual test


src/ToggleButtonGroup/docs/index.story.js
```diff
@@ -0,0 +1,158 @@
...
+      issueUrl: 'https://github.com/wix/wix-style-react/issues/new',
+    }),
+    tabs({
+      tabs: [
+        tab({
+          title: 'Description',
+          sections: [
+            columns([
+              description({
+                text: `Icon Button is used when action is lower priority than a regular action or there’s no space available to place a button with text.`,
```
ipanasenko: Oooops, updating description here...

src/ToggleButtonGroup/docs/index.story.js
```diff
@@ -0,0 +1,158 @@
...
+            }),
+
+            divider(),
+
+            columns([title('Examples')]),
+
+            ...[
+              {
+                title: 'Uncontrolled',
+                text: 'TODO',
```
ipanasenko: Oooops, updating description here...

src/ToggleButton/ToggleButton.js
```diff
@@ -0,0 +1,104 @@
...
+    /** Click event handler  */
+    onClick: func,
+    /** Applies active styles */
+    active: bool,
+    /** Applies disabled styles */
+    disabled: bool,
+    /** String based data hook */
+    dataHook: string,
+    /** Tooltip content that will appear on hover */
+    tooltip: node,
```
mykas: prop `tooltip` doesn't say anything. I suggest naming it like this: `tooltipContent`. I understand that this seems like too much, but we have experience where users are not really understanding such props until they open storybook or have typings.

src/ToggleButton/ToggleButton.js
```diff
@@ -0,0 +1,104 @@
...
+  render() {
+    const {
+      skin,
+      priority,
+      size,
+      className,
+      children,
+      active,
+      value,
+      dataHook = value,
```
mykas: Why equals to value?

src/ToggleButton/ToggleButton.js
```diff
@@ -0,0 +1,104 @@
...
+      tooltip,
+      tooltipPlacement,
+      ...rest
+    } = this.props;
+
+    const buttonClassNames = cx(
```
mykas: This way of switching classnames is a way with sass, but not with stylable. This should be applied on Tooltip and use like this: `{...styles('toggleButton', { skin, active }, this.props }`. As you see the middle parameter is the state of the stylable.

src/ToggleButton/ToggleButton.js
```diff
@@ -0,0 +1,104 @@
...
+    return tooltip ? (
+      <Tooltip
+        className={cx(styles.tooltip, className)}
+        upgrade
+        content={tooltip}
+        placement={tooltipPlacement}
+        moveBy={tooltipPlacement === 'top' ? { x: 0, y: 2 } : { x: 10, y: 0 }}
+        size="small"
+      >
+        <ButtonNext {...buttonProps} className={buttonClassNames}>
```
mykas: Here you should use our Button component instead of `ButtonNext` from wix-ui-core. If you want to change the stylings of the button according to the new specs I can show you how to extend the button stylesheet and override values.

src/ToggleButton/ToggleButton.st.css
```diff
@@ -0,0 +1,108 @@
...
+  background-color: value(D80);
+  color: value(D10);
+}
+
+.toggleButton:hover:not(:disabled) {
+  background-color: value(D80);
+  color: value(B10);
+}
+
+.toggleButton:active:not(:disabled),
```
mykas: toggle:active and toggle.active?

src/ToggleButton/docs/index.story.js
```diff
@@ -0,0 +1,137 @@
...
+          title: 'Description',
+          sections: [
+            columns([
+              description({
+                text: `Control Button is used when action is lower priority than a regular action or there’s no space available to place a button with text.`,
+              }),
+            ]),
+
+            importExample({
+              source:
```
mykas: Just a little change. No need for `source:` you can pass string to `importExample('string')`

src/ToggleButton/test/ToggleButton.e2e.js
```diff
@@ -0,0 +1,21 @@
+import { eyesItInstance } from '../../../test/utils/eyes-it';
+import { createTestStoryUrl } from '../../../test/utils/storybook-helpers';
+import { storySettings, testStories } from './storySettings';
+
+describe('ToggleButton', () => {
```
mykas: For static eyes screenshots we use eyes storybook debug. I saw already that you have defined eyes for .visual so this file can be deleted. Problem with this that it takes a lot of time of our ci builds and .visual stuff does not.

src/ToggleButton/test/testStory.js
```diff
@@ -0,0 +1,80 @@
+/* eslint-disable react/prop-types */
```
mykas: No need for this as .visual stuff already covers it.

src/ToggleButtonGroup/ToggleButtonGroup.js
```diff
@@ -0,0 +1,158 @@
+import React, {
```
mykas: This component should be another PR. Make sure to extract this.

test/jest-setup.js
```diff
@@ -1,7 +1,10 @@
 import Enzyme from 'enzyme';
```
mykas: Should not be in this PR.

.prettierrc.js
```diff
@@ -0,0 +1,12 @@
+let prettierConfig = {
```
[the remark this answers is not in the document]
ipanasenko: It's already in separate PR for 4 days: https://github.com/wix/wix-style-react/pull/3927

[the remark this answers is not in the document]
mykas: Ok, but for now it shouldn't be in this PR.

src/ToggleButton/ToggleButton.js
ipanasenko: It's just default value for `dataHook`, it allows to query by buttons by value. Helped me avoid duplication in tests: `value="adjust"` vs `value="adjust" dataHook="adjust"`

mykas: Ok, but implementation stuff should not be according to our unit tests. We write for your users and test like our users do so please leave it undefined.

src/ToggleButton/ToggleButton.st.css
ipanasenko: Yes, because one is when user interacts with button, and another is when `active` prop provided. If it's too confusing, we can rename `active` prop to `selected`, or whatever

mykas: ok, give a minute I'll check.

mykas: yes, that seems correct. The `:active` stuff is browser when on mouse key press and when user wants to select it it should be `selected`

ipanasenko: Okay, I'll rename this prop

src/ToggleButton/ToggleButton.js
```diff
@@ -0,0 +1,104 @@
...
+    /** Applies active styles */
+    active: bool,
+    /** Applies disabled styles */
+    disabled: bool,
+    /** String based data hook */
+    dataHook: string,
+    /** Tooltip content that will appear on hover */
+    tooltip: node,
+    /** Tooltip placement that will appear on hover */
+    tooltipPlacement: oneOf(['top', 'right']),
```
mykas: Another thing about the Tooltip. The thing with the tooltip is that we have many different cases where users needs tooltip with different functionality. So we decided that we should instead expose all props of the Tooltip so that our users would be able to modify it on demand. So please change this prop to: `tooltipProps: object`.

src/ToggleButton/test/ToggleButton.spec.js
```diff
@@ -0,0 +1,72 @@
...
+
+    it('should apply className when active', async () => {
+      const props = { active: true, children: <More /> };
+      const { driver } = render(<ToggleButton {...props} />);
+
+      expect(await driver.isButtonActive()).toEqual(true);
+    });
+  });
+
+  describe(`'as' prop`, () => {
```
mykas: This part is already covered in buttons implementation. I think its redundant to test again this feature :)

src/ToggleButton/ToggleButton.st.css
```diff
@@ -0,0 +1,116 @@
+:import {
+  -st-from: "wix-ui-core/dist/src/components/button-next/button-next.st.css";
+  -st-default: Button;
+}
+
+:import {
+  -st-from: "wix-ui-core/dist/src/themes/backoffice/colors.st.css";
```
mykas: This should come from: `../Foundation/stylable/colors.st.css`;

src/ToggleButton/ToggleButton.js
```diff
@@ -0,0 +1,110 @@
...
+      });
+
+    const buttonProps = {
+      ...rest,
+      'data-hook': dataHook,
+    };
+
+    const { placement: tooltipPlacement = 'top' } = tooltipProps;
+
+    return tooltipContent ? (
```
mykas: So in this case we have a way for our users to have `ToggleButton` without the tooltip. Problem with this is that our users needs to use this all the times with the tooltip according to our ux. The reason for this that icon representation alone cannot explain user the meaning of this so tooltip should cover this case. So if users is not specifying `tooltipContent` he should still see empty tooltip so he would understand that this is mandatory.

src/ToggleButton/ToggleButton.js
```diff
@@ -0,0 +1,93 @@
...
+
+    const { placement: tooltipPlacement = 'top' } = tooltipProps;
+
+    return (
+      <Tooltip
+        {...styles('tooltip', {}, { ...rest, className })}
+        upgrade
+        size="small"
+        content={tooltipContent}
+        placement={tooltipPlacement}
```
mykas: Lets use placement inside tooltipProps and in default props we define `placement: top`;

src/ToggleButton/ToggleButton.js
```diff
@@ -0,0 +1,93 @@
...
+        moveBy={tooltipPlacement === 'top' ? { x: 0, y: 2 } : { x: 10, y: 0 }}
+        {...tooltipProps}
+      >
+        <ButtonNext
+          {...rest}
+          data-hook={dataHook}
+          data-selected={selected}
+          {...styles('toggleButton', { skin, selected, size }, rest)}
```
mykas: Lets move this up near the `{...rest}`

src/ToggleButton/ToggleButton.js
```diff
@@ -0,0 +1,93 @@
...
+      ...rest
+    } = this.props;
+
+    const childSize = iconChildSize[size];
+
+    const { placement: tooltipPlacement = 'top' } = tooltipProps;
+
+    return (
+      <Tooltip
+        {...styles('tooltip', {}, { ...rest, className })}
```
mykas: Lets delete this spread. The whole spread.

src/ToggleButton/ToggleButton.js
```diff
@@ -0,0 +1,93 @@
...
+      >
+        <ButtonNext
+          {...rest}
+          data-hook={dataHook}
+          data-selected={selected}
+          {...styles('toggleButton', { skin, selected, size }, rest)}
+        >
+          {children &&
+            cloneElement(children, {
+              size: childSize,
```
mykas: Lets remove the `size` property.

src/ToggleButton/ToggleButton.st.css
```diff
@@ -0,0 +1,121 @@
...
+/* icon size   */
+
+.toggleButton.iconMedium {
+  width: 24px;
+  height: 24px;
+}
+
+/* tooltip */
+
+.tooltip {
```
mykas: Not needed anymore.

src/ToggleButton/ToggleButton.st.css
```diff
@@ -0,0 +1,121 @@
...
+  -st-from: "../Tooltip/TooltipNext/Tooltip.st.css";
+  -st-default: Tooltip;
+}
+
+:import {
+  -st-from: "../Foundation/stylable/colors.st.css";
+  -st-named: B10, D55, D80, B20, B50, B40, F00, D10, D20, D30;
+}
+
+.toggleButton {
```
mykas: Because we will use only in one place the stylesheet - change the naming of this to `.root`.

src/ToggleButton/ToggleButton.st.css
```diff
@@ -0,0 +1,121 @@
+:import {
+  -st-from: "wix-ui-core/dist/src/components/button-next/button-next.st.css";
+  -st-default: Button;
+}
+
+:import {
```
mykas: Remove redundant

src/ToggleButton/ToggleButton.uni.driver.js
```diff
@@ -0,0 +1,23 @@
...
+    /** returns true if component exists */
+    exists: buttonDriver.exists,
+    /** returns the component element */
+    element: buttonDriver.element,
+    /** click on the element */
+    click: buttonDriver.click,
+    /** returns true if button is disabled */
+    isButtonDisabled: buttonDriver.isButtonDisabled,
+    /** returns true if button has skin applied */
+    hasSkin: skin => base.hasClass(styles[skin]),
```
mykas: This part is a bit problematic. The import for stylable file will not work on: `proctractor` and `puppeteer`. Lets do the same data attribute stuff here to. Just do in the source code in return method like this:
```
const skins = ['standard', 'dark'].reduce((prev, curr) => ({...prev, [`data-skin-${curr}`]: !!this.props[curr]  }), {})
```
Feel free to improve it.

src/ToggleButton/ToggleButton.uni.driver.js
```diff
@@ -0,0 +1,27 @@
+import { buttonNextDriverFactory } from 'wix-ui-core/drivers/unidriver';
+import { tooltipDriverFactory } from '../Tooltip/TooltipNext/Tooltip.uni.driver';
+
+export const toggleButtonDriverFactory = base => {
+  const tooltipDriver = tooltipDriverFactory(base);
```
mykas: you should pass `body` to second parameter in order for the tooltip to work. You can get it from `toggleButtonDriverFactory` second parameter.

src/ToggleButton/ToggleButton.uni.driver.js
```diff
@@ -0,0 +1,27 @@
...
+  // Not using Omit so that AutoDocs will generate properly
+  return {
+    /** returns true if component exists */
+    exists: buttonDriver.exists,
+    /** returns the component element */
+    element: buttonDriver.element,
+    /** click on the element */
+    click: buttonDriver.click,
+    /** returns true if button is disabled */
+    isButtonDisabled: buttonDriver.isButtonDisabled,
```
mykas: I guess here you should use buttonBaseElement.

src/ToggleButton/docs/index.story.js
```diff
@@ -0,0 +1,137 @@
...
+    as: 'button',
+    children: <CropRotate />,
+    skin: 'standard',
+    priority: 'primary',
+    size: 'medium',
+    tooltip: 'I am a tooltip',
+    selected: false,
+    disabled: false,
+    tooltipContent: 'Crop & Rotate',
+    tooltipPlacement: 'top',
```
mykas: move this to tooltipProps

src/ToggleButton/docs/index.story.js
```diff
@@ -0,0 +1,137 @@
...
+  },
+  exampleProps: {
+    onClick: () => 'Clicked!',
+    children: icons,
+    as: ['button', 'a', 'span', 'div'],
+    tooltipPlacement: ['top', 'right'],
```
mykas: same as above

src/ToggleButton/ToggleButton.uni.driver.js
ipanasenko: Why? I want to reuse `isButtonDisabled` from buttonDriver, like in icon button driver

src/ToggleButton/ToggleButton.uni.driver.js
ipanasenko: Didn't help me, still getting
```
ToggleButton › `tooltipContent` prop › should set tooltip content from `tooltipContent` prop

    TypeError: Cannot read property '$' of undefined

      at Object.<anonymous> (node_modules/wix-ui-core/src/components/popover/Popover.common.uni.driver.ts:9:17)
      at step (node_modules/tslib/tslib.js:136:27)
      at Object.next (node_modules/tslib/tslib.js:117:57)
      at fulfilled (node_modules/tslib/tslib.js:107:62)
```

mykas: The body part must be passed in the private driver too, because in your test your are mounting actually private.driver.

src/ToggleButton/ToggleButton.uni.driver.js
mykas: But you are passing base to the buttonNextDriverFactory. The base includes: Tooltip + Button. You should pass directly the `togglebutton-trigger` element not the root.

src/ToggleButton/ToggleButton.js
```diff
@@ -0,0 +1,88 @@
...
+
+  static propTypes = {
+    /** render as some other component or DOM tag */
+    as: oneOfType([func, object, string]),
+    /** Used for passing any wix-style-react icon. For external icon make sure to follow ux sizing guidelines */
+    children: node,
+    /** Button skins */
+    skin: oneOf(['standard', 'dark']),
+    /** Button priority */
+    priority: oneOf(['primary']),
```
mykas: I don't see in our specs this value. The button only supports primary so no point of adding this.

src/ToggleButton/ToggleButton.js
```diff
@@ -0,0 +1,88 @@
...
+    /** Button size */
+    size: oneOf(['medium']),
```
mykas: Same is here. whats the point of defining the size of it only has one size?

src/ToggleButton/ToggleButton.st.css
```diff
@@ -0,0 +1,110 @@
...
+}
+
+.root:skin(dark):disabled {
+  background-color: value(D80);
+  color: value(D55);
+}
+
+/* button size */
+
+.root:size(medium) {
```
mykas: redundant code. The same size is already defined in the root.

src/ToggleButton/docs/testkit.md
```diff
@@ -0,0 +1,108 @@
...
+
+## ToggleButton TestKit API
+
+| method           | arguments | returned value     | description                        |
+| ---------------- | --------- | ------------------ | ---------------------------------- |
+| exists           | -         | `Promise<boolean>` | returns true if element in the DOM |
+| element          | -         | `Promise<element>` | returns the component element      |
+| click            | -         | `Promise<void>`    | clicks on the button               |
+| isButtonDisabled | -         | `Promise<boolean>` | returns true if button is disabled |
+| isButtonSelected | -         | `Promise<boolean>` | returns true if button is selected |
```
mykas: Don't forget to add getTooltipContent method.

src/ToggleButton/ToggleButton.st.css
```diff
@@ -0,0 +1,96 @@
...
+  -st-named: B10, D55, D80, B20, B50, B40, F00, D10, D20, D30;
+}
+
+.root {
+  -st-states: skin(enum(standard, dark)), selected, size(enum(medium));
+  -st-extends: Button;
+}
+
+.root {
+  min-width: 0;
```
mykas: do we really need this ?

ipanasenko: yes, because there's min-width on button (I guess 84px) and we need smaller than that

src/ToggleButton/ToggleButton.uni.driver.js
ipanasenko: I've provided correct base to buttonNextDriverFactory: https://github.com/wix/wix-style-react/pull/3953/commits/a50f5ff8d66ad194bdfa82ab5990d483d5493ffe

src/ToggleButton/ToggleButton.st.css
mykas: Oh ok

src/ToggleButton/docs/index.story.js
```diff
@@ -0,0 +1,136 @@
...
+      issueUrl: 'https://github.com/wix/wix-style-react/issues/new',
+    }),
+    tabs({
+      tabs: [
+        tab({
+          title: 'Description',
+          sections: [
+            columns([
+              description({
+                text: `Control Button is used when action is lower priority than a regular action or there’s no space available to place a button with text.`,
```
splintor: `Control Button`? Isn't this a `Toggle Button`?

ipanasenko: You are correct, it's a type, thanks!

milkyfruit:
@ipanasenko @mykas the story should be indexed as 5.6 Toggle Button. 5.5 is for Add Item.

PR #3963, opened 2019-08-16
`<MultiSelectCheckbox/>` - improve tests and prepare for inputWithOptions migration to Popover

mykas:
This PR is another preparation for InputWithOptions migration to use Popover mechanism.


src/MultiSelectCheckbox/MultiSelectCheckbox.spec.js
```diff
@@ -100,12 +98,13 @@ describe('multiSelectCheckbox', () => {
     });
 
     it('should not lose Focus or close the list on selection with a mouse click', async () => {
-      const { inputDriver, dropdownLayoutDriver } = createDriver(
+      const { inputDriver, dropdownLayoutDriver, driver } = createDriver(
         <MultiSelectCheckbox options={options} />,
       );
 
       await inputDriver.focus();
```
shlomitc: Is this still needed?

mykas: I am not sure. I just fix this so that it would not fail on my inputWithOptions. Many things needs to be revisited of course :)

shlomitc: My only concern when changing tests is will it require a change for the consumer. If so - it might be considered a breaking change and that’s not a good thing.

If it’s only refactoring for a new and better named function that’s fine.

mykas: It will not break our users. I don't think that our users testing `.focus` stuff on our components. This part was changed because actually in the component logic - focus() does not open the menu. This test was wrong in the beginning because it assumes that menu is always available for testing because of how dropdownlayout is done. It will be always in the DOM, but visible through css.

shlomitc: Worth going through some github projects. I think people are using focus or something similar like keyDown. Otherwise there’s not other option to open the selection menu

PR #3970, opened 2019-08-19
Swatches - Allow add button which shows color picker.

alexgoncharwix:
### 🔦 Summary
...

### ✅ Checklist
- [x ] 👨‍💻 API change is approved by the UI Infra Developers 
- [ x] 👨‍🎨 UX change is approved by the Design System UX 
- 📚 Change is documented
  - [ x] Story
  - [ x] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ x] Component
  - [ ] Visual test

### DIFF
--- a/src/Swatches/ColorPreviewAddButton/ColorPreviewAddButton.js
+++ b/src/Swatches/ColorPreviewAddButton/ColorPreviewAddButton.js
@@ -1,0 +1,70 @@
+import React, { PureComponent } from 'react';
+
+import Color from 'color';
+import styles from './ColorPreviewAddButton.st.css';
+import { func, node, object } from 'prop-types';
+import { withFocusable } from 'wix-ui-core/dist/src/hocs/Focusable/FocusableHOC';
+import ColorPreviewAddIcon, {
+  ColorPreviewAddIconSize,
+} from './ColorPreviewAddIcon';
+
+function getContrastColor(bg, light = '#ffffff', dark = '#162d3d') {
+  try {
+    const color = Color(bg);
+    const luminosity = color.luminosity();
+
+    if (luminosity > 0.5) {
+      return dark;
+    } else {
+      return light;
+    }
+  } catch (err) {}
+}
+
+class ColorPreviewAddButton extends PureComponent {
+  static propTypes = {
+    color: object,
+    onClick: func,
+    addTooltipContent: node,
+    iconSize: ColorPreviewAddIconSize,
+  };
+
+  static defaultProps = {
+    onClick: () => {},
+  };
+
+  onAddClick = () => {
+    const { color, onClick } = this.props;
+
+    onClick(color.hex());
+  };
+
+  render() {
+    const { color, focusableOnFocus, focusableOnBlur, iconSize } = this.props;
+    const noColorSelected = color.alpha() === 0;
+
+    const styleProps = {
+      noColorSelected,
+    };
+
+    return React.createElement(
+      'button',
+      {
+        style: {
+          backgroundColor: noColorSelected ? undefined : color.hex(),
+        },
+        'data-hook': 'color-preview-add-button',
+        onFocus: focusableOnFocus,
+        onBlur: focusableOnBlur,
+        ...styles('preview', styleProps, this.props),
+        onClick: this.onAddClick,
+      },
+      <ColorPreviewAddIcon
+        iconSize={iconSize}
+        color={getContrastColor(color.hex())}
+      />,
+    );
+  }
+}
+
+export default withFocusable(ColorPreviewAddButton);
--- a/src/Swatches/ColorPreviewAddButton/ColorPreviewAddButton.st.css
+++ b/src/Swatches/ColorPreviewAddButton/ColorPreviewAddButton.st.css
@@ -1,0 +1,42 @@
+:import {
+  -st-from: "wix-ui-core/dist/src/hocs/Focusable/Focusable.st.css";
+  -st-default: Focusable;
+}
+
+:import {
+  -st-from: "../../Foundation/stylable/colors.st.css";
+  -st-named:
+    B00,
+    B00-24, B00-42,
+    F00
+}
+
+.preview {
+  -st-extends: Focusable;
+  -st-states: noColorSelected;
+  flex: 0 0 24px;
+  border-radius: 6px;
+  border: solid 1px value(B00-24);
+  width: 24px;
+  height: 24px;
+  cursor: pointer;
+  padding: 0;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+
+  /* Disable native focus */
+  outline: none;
+}
+
+.preview:hover {
+  border-color: value(B00-42);
+}
+
+.preview:noColorSelected {
+  background-color: value(F00);
+}
+
+.preview:not(:disabled):focus-visible {
+  box-shadow: 0 0 0 3px value(F00);
+}
--- a/src/Swatches/Swatches.js
+++ b/src/Swatches/Swatches.js
@@ -1,19 +1,25 @@
 import React from 'react';
-import { array, func, string, oneOf, bool, node } from 'prop-types';
+import { array, bool, func, node, oneOf, string } from 'prop-types';
 import resolveColor from 'color';
 import styles from './Swatches.st.css';
 import Swatch from './Swatch';
+import AddColorButton from './AddColorButton/AddColorButton';
+import { ColorPreviewAddIconSize } from './ColorPreviewAddButton/ColorPreviewAddIcon';
 
 /** Color swatches */
 const Swatches = props => {
   const {
     colors,
     onClick,
+    onAdd,
     selected,
     size,
     dataHook,
     showClear,
     showClearMessage,
+    showAddButton,
+    addButtonTooltip,
+    iconSize,
   } = props;
 
   const hexColors = colors.map(color => resolveColor(color).hex());
@@ -21,6 +27,14 @@ const Swatches = props => {
 
   return (
     <div {...styles('root', { size }, props)} data-hook={dataHook}>
+      {showAddButton && (
+        <AddColorButton
+          tooltip={addButtonTooltip}
+          iconSize={iconSize}
+          onAdd={onAdd}
+        />
+      )}
+
       {showClear && (
         <Swatch
           color=""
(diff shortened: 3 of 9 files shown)


src/Swatches/ColorPreviewAddButton/ColorPreviewAddButton.js
```diff
@@ -0,0 +1,70 @@
+import React, { PureComponent } from 'react';
```
mykas: Also I am not sure this needs to be extracted. There are lots of boilerplate here. You have two places attached with `withFocusable`. Only one place should be done and it should be `AddColor` component.

src/Swatches/ColorPreviewAddButton/ColorPreviewAddButton.st.css
```diff
@@ -0,0 +1,42 @@
...
+    B00-24, B00-42,
+    F00
+}
+
+.preview {
+  -st-extends: Focusable;
+  -st-states: noColorSelected;
+  flex: 0 0 24px;
+  border-radius: 6px;
+  border: solid 1px value(B00-24);
```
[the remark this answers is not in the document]
alexgoncharwix: Double checked - the corder exists in the spec.

src/Swatches/ColorPreviewAddButton/ColorPreviewAddButton.js
```diff
@@ -0,0 +1,70 @@
...
+
+  render() {
+    const { color, focusableOnFocus, focusableOnBlur, iconSize } = this.props;
+    const noColorSelected = color.alpha() === 0;
+
+    const styleProps = {
+      noColorSelected,
+    };
+
+    return React.createElement(
```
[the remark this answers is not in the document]
alexgoncharwix: Done.

src/Swatches/Swatches.js
```diff
@@ -65,6 +82,15 @@ Swatches.propTypes = {
 
   /** optional message to display in tooltip when showClear is true */
   showClearMessage: node,
+
+  /** If true shows add button which triggers colors picker*/
+  showAddButton: bool,
+
+  /** Tooltip of add button */
+  addButtonTooltip: string,
```
mykas: This doesn't say anything. Is this suppose to be the content of the tooltip?

src/Swatches/ColorPreviewAddButton/ColorPreviewAddButton.js
```diff
@@ -0,0 +1,70 @@
...
+        style: {
+          backgroundColor: noColorSelected ? undefined : color.hex(),
+        },
+        'data-hook': 'color-preview-add-button',
+        onFocus: focusableOnFocus,
+        onBlur: focusableOnBlur,
+        ...styles('preview', styleProps, this.props),
+        onClick: this.onAddClick,
+      },
+      <ColorPreviewAddIcon
```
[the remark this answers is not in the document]
alexgoncharwix: This is to follow Single responsibility principle and to minimize `ColorPreviewAddButton.js` file size. Could we keep it as is?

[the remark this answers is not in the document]
mykas: Yes, but in our case this is already small component. I don't think nesting multiple files will help us in the future maintain this. We understand the single responsibility part but it in this case this makes is more complex.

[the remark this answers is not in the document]
mykas: When we debug the components - sometimes it is much easier to have one single file to see the whole logic. This is only an Add button and you are splitting it in to 3 separate files.

src/Swatches/ColorPreviewAddButton/ColorPreviewAddButton.st.css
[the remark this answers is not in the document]
mykas: Please share your design specs.

src/Swatches/ColorPreviewAddButton/ColorPreviewAddButton.js
alexgoncharwix: `ColorPreviewAddButton` - is just a stateless button with no business logic.

alexgoncharwix: Maybe `ColorPreviewAddButton` should be focusable instead of `AddColor `?

mykas: Once again I don't see any benefits separating this part from the `AddColor` button. You just create another class with couple of methods to handle onClick which can be inside `AddColor` too. The focus should appear on the button and you can use the focusableOnBlur and focusableOnFocus props to pass it and control it.

src/Swatches/test/Swatches.spec.js
```diff
@@ -1,8 +1,10 @@
 import React from 'react';
 import { createRendererWithUniDriver, cleanup } from '../../../test/utils/unit';
+import eventually from 'wix-eventually';
 
 import Swatches from '../Swatches';
 import { swatchesPrivateDriverFactory } from './Swatches.private.uni.driver';
+import SideMenu from '../../Deprecated/SideMenu';
```
mykas: Redundant code

src/Swatches/test/Swatches.spec.js
```diff
@@ -88,5 +90,87 @@ describe('Swatches', () => {
...
+        expect(await driver.isColorPickerShown()).toBe(false);
+      });
+
+      it('should show color picker on add button click', async () => {
+        const { driver } = render(
+          <Swatches colors={['#000000']} showAddButton />,
+        );
+
+        await driver.clickAddColorButton();
+        await eventually(async () => {
```
mykas: Is this because of the Popovers timeout?

src/Swatches/ColorPreviewAddButton/ColorPreviewAddButton.js
alexgoncharwix: Ok, I'll merge it. But I think such approach leads to very big components which hard to read.

src/Swatches/test/Swatches.spec.js
alexgoncharwix: Removed.

src/Swatches/Swatches.js
```diff
@@ -65,6 +81,15 @@ Swatches.propTypes = {
...
   showClearMessage: node,
+
+  /** If true shows add button which triggers colors picker*/
+  showAddButton: bool,
+
+  /** Text for add button tooltip*/
+  addButtonMessage: string,
+
+  /** Size of Plus icon inside add button */
+  iconSize: AddColorIconSize,
```
mykas: this prop is not descriptive enough. should be something like: `addButtonIconSize`.

src/Swatches/AddColor/AddColor.js
```diff
@@ -0,0 +1,129 @@
...
+    });
+  };
+
+  confirm = () => {
+    this.hideColorPicker();
+    this.props.onAdd(this.state.color.hex());
+  };
+
+  render() {
+    const { tooltip, iconSize } = this.props;
```
mykas: This `tooltip` prop should be renamed to `tooltipContent`.

src/Swatches/AddColor/AddColor.js
```diff
@@ -0,0 +1,129 @@
...
+    });
+  };
+
+  setColor = color => {
+    this.setState({
+      color,
+    });
+  };
+
+  confirm = () => {
```
mykas: lets rename this: `onColorPickerConfirm`

src/Swatches/AddColor/AddColor.js
```diff
@@ -0,0 +1,129 @@
...
+  };
+
+  hideColorPicker = () => {
+    this.setState({
+      isColorPickerShown: false,
+      color: EMPTY_COLOR,
+    });
+  };
+
+  setColor = color => {
```
mykas: lets rename this: `onColorPickerChange`

src/Swatches/AddColor/AddColor.js
```diff
@@ -0,0 +1,129 @@
...
+  static defaultProps = {
+    onAdd: () => {},
+  };
+
+  state = {
+    color: EMPTY_COLOR,
+    isColorPickerShown: false,
+  };
+
+  toggleColorPicker = () => {
```
mykas: lets rename this to: onButtonClick()

src/Swatches/AddColor/AddColor.js
```diff
@@ -0,0 +1,129 @@
...
+  render() {
+    const { tooltip, iconSize } = this.props;
+    const { isColorPickerShown, color } = this.state;
+    const isEmptyColor = color.alpha() === 0;
+    const noColorSelected = !isColorPickerShown || isEmptyColor;
+    const buttonColor = noColorSelected ? undefined : color;
+
+    return (
+      <Popover
+        appendTo="parent"
```
mykas: The append value should be better be: `window` in this case because this will be used inside the color picker so its not a good idea to append to color picker container.

src/Swatches/AddColor/AddColor.st.css
```diff
@@ -0,0 +1,47 @@
...
+:import {
+  -st-from: "../../Foundation/stylable/mixins/calc_rgba.js";
+  -st-default: calc_rgba;
+}
+
+.root {
+  -st-states: selected;
+  -st-extends: Focusable;
+
+  flex: 0 0 24px;
```
mykas: Why is this needed?

src/Swatches/docs/index.story.js
```diff
@@ -117,6 +117,20 @@ export default {
             }),
           ]),
 
+          columns([
+            description({
+              title: 'Add button',
+              text:
+                'Pass `showAddButton` prop to get the Add Color button. In addition pass `onAdd` to handle newly selected color and `addButtonMessage` to control tooltip of the add button',
```
mykas: Change this to:

Swatches support button that is meant to add new color preset. Enable it by passing `showAddButton` prop and make sure to defined the tooltip message with prop `addButtonMessage` to inform users on the action meaning.

src/Swatches/docs/index.story.js
```diff
@@ -129,9 +143,12 @@ export default {
                   return (
                     <Swatches
                       showClear
```
mykas: Full interactive example should include the whole logic on adding new color. With states etc.

PR #3972, opened 2019-08-19
`<Popover/>` - add a request animation frame polyfill on testing

mykas:
This PR is adding requestAnimationFrame method polyfill  on user unit testing. Jest supports this by default and mocha is not.


shlomitc:
worth mentioning that this was required because of refactoring the `InputWithOptions` components to work with `Popover`. This is potentially a breaking change for consumers testing with mocha (jest env includes it).

I mentioned it in the breaking changes suggestions for next major version

PR #3979, opened 2019-08-20
Semantic test scripts

nitayneeman:
This PR makes the test scripts more semantic. 

I always forget what's the responsibility of each script every time I encounter that during the CI, and then I need to return to package.json - instead of understanding it immediately.

Assuming these scripts have a dedicated responsibility, I don't see any concern in naming them semantically.

What do you think? @shlomitc @netanelgilad


nitayneeman:
FYI @argshook @mykas @mosheKerbel @gilmeir @TalZa1810

mykas:
I think it doesn't work :D I was thinking the same thing lately

shlomitc:
I think this should be done as some kind of a CI configuration.
@netanelgilad - can we make this happen?

nitayneeman:
> I think it doesn't work :D I was thinking the same thing lately

I guess we just need to adjust `wix-style-react-tests-composite` properly?

netanelgilad:
I think this is a good chance to test out the automatic solution by CI, so you'll be able to configure this stuff yourself in `.ci-config` (currently it was done manually)

shlomitc:
awesome. @nitayneeman is for your service if any changes/monitoring is needed

package.json
```diff
@@ -29,12 +29,12 @@
...
-    "test1": "npm run test:unit",
-    "test2": "npm run test:eyes-storybook",
-    "test3": "YOSHI_PROTRACTOR_CONFIG=./protractor1.conf.js npm run test:e2e",
-    "test4": "YOSHI_PROTRACTOR_CONFIG=./protractor2.conf.js npm run test:e2e",
-    "test5": "npm run test:types",
-    "test6": "npm run lint && npm run sanity",
+    "test1:unit": "npm run test:unit",
+    "test2:visual": "npm run test:eyes-storybook",
+    "test3:e2e": "YOSHI_PROTRACTOR_CONFIG=./protractor1.conf.js npm run test:e2e",
+    "test4:e2e": "YOSHI_PROTRACTOR_CONFIG=./protractor2.conf.js npm run test:e2e",
```
mosheKerbel: same alias for both jobs?

PR #3993, opened 2019-08-21
Fix stylable paths for wix-ui-core/themes

yavorsky:
### 🔦 Summary
Since we are using script to transform wix-ui paths to contain `es` files and all paths match, we need to update root themes imports so script will update `wix-ui-core/dist/src/themes/backoffice` -> `wix-ui-core/dist/es/src/themes/backoffice`.
This will fix some style issues with stylable classnames based on paths


shlomitc:
solves issues like these:
image

image

PR #3995, opened 2019-08-21
Extend `no-restricted-imports` pattern to check for non-dist wix-ui-core imports

yavorsky:
### 🔦 Summary
...
Add eslint rule to disallow wsr devs to import modules from `wix-ui-core`'s root.
Since we have build process to replace `wix-ui-core/dist/src` with `wix-ui-core/dist/es/src` imports for all files from `dist/es` it will be helpful to check this during lint. I excluded drivers, since it's not a part of the build process and don't included into wix-ui-core's `dist`.

Also with this rule, I was able to find 2 incorrect paths for ES modules we had before.

I know that it would be good to have message for patterns, but it's not implemented yet on the eslint side: https://github.com/eslint/eslint/issues/11843


shlomitc:
amazing. thanks a lot!

Issue #3946, opened 2019-08-14
 <ModalSelectorLayout> - add max-height


liorur:
simply add maxHeight prop next to height prop so its possible for the modal to to be as small as needed but not grow beyond the screen
```
height="auto"
maxHeight="100%"
```

shlomitc:
sounds legit. want to make a PR?

shlomitc:
done

PR #4011, opened 2019-08-25
<ModalSelectorLayout/> - add maxHeight prop

liorur:
### 🔦 Summary
simply add maxHeight prop next to height prop so its possible for the modal to to be as small as needed but not grow beyond the screen

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


liorur:
fixes #3946

mosheKerbel:
@liorur 
Changing the default `height` from `100%` to `auto` will change the way the modal is displayed inside a container for other users.
consider the following [jsbin](https://jsbin.com/lufosewela/edit?html,output)
image

I think you will be able to achieve what you want even if you keep the default height at `100%`

Plus, plz add a description for the height and maxHeight prop using the `/** comment */` so it will be reflected in the docs api.

mosheKerbel:
Tests passed
Failed on 1 flaky unrelated test.
Merging

PR #4017, opened 2019-08-26
<ColorInput/> - remove manual testkit docs

mosheKerbel:
Calling the functions explicitly in the driver solves an issue with auto-docs, which allows it to generate the testkit docs correctly

@argshook i wonder if it's worth merging this
because i'm getting rid of manual docs which needs to be synced with the code
**but** i'm loosing the `await` and `async` which are mandatory for uniDriver and are currently not generated by autodocs.

current one to compare with:
https://wix-wix-style-react.surge.sh/?activeTab=Testkit&addons=0&full=0&panelRight=0&selectedKind=3.%20Inputs&selectedStory=3.11%20ColorInput&stories=1


argshook:
i'm working on updating testkit docs to be more up to date and reflect
async/await stuff so it will come from autodocs in a week or so.

this can be merged imo, but if you say missing async/await will be
misleading for the time being, then no need

mosheKerbel:
Cool
in that case i see no rush, i will merge it once autodocs issues are resolved.

shlomitc:
@argshook - any update?

argshook:
Unfortunately, the fixes i were talking about [here](https://github.com/wix/wix-style-react/pull/4017#issuecomment-524909149) didn't address issue that fixed in this PR.

In particular, testkit that exports functions referencing other testkits
is problematic for parser.

in code:

```diff
-    cancel: colorPickerTestkit.cancel,
+    cancel: () => colorPickerTestkit.cancel(),
```

it makes sense to pass just reference, instead of creating function
expression and then invoking. However, parser gets stuck on reference
trying to look deeper until it crashes.

---

therefore, for now it's good to merge this PR. The testkit tab of ColorInput will be consistent with other story pages and we'll cleanup some markdown

PR #4019, opened 2019-08-26
indication and presence added to avatar

adirou:
### 🔦 Summary
presence and indication features are added
...

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


src/Avatar/Avatar.js
```diff
@@ -4,24 +4,52 @@ import PropTypes from 'prop-types';
...
+              className={classNames(
+                avatarStyle(size, color && `color${capitalize(color)}`),
+              )}
+            />
+          </div>
+          {presence && <div {...styles('presence', { size, presence })}></div>}
+        </div>
+        {renderIndication && (
+          <div
+            data-hook="indication"
```
nitayneeman: Please change to `avatar-indication` so the dataHook will be unique:
```suggestion
            data-hook="avatar-indication"
```

src/Avatar/Avatar.js
```diff
@@ -4,24 +4,52 @@ import PropTypes from 'prop-types';
...
+    size,
+    presence,
+    indication,
+    color,
+    onClick,
+    dataHook,
+    className,
+    ...rest
+  } = props;
+  const renderIndication = indication && sizeNumber(size) > 36;
```
nitayneeman: I would assign this magic number to a const so that meaning would be clear:
```suggestion
  const minRenderSize = 36;
  const renderIndication = indication && sizeNumber(size) > minRenderSize;
```

src/Avatar/Avatar.js
```diff
@@ -4,24 +4,52 @@ import PropTypes from 'prop-types';
...
-      {...rest}
-      data-hook={dataHook}
-      className={classNames(
-        className,
-        avatarStyle(size, color && `color${capitalize(color)}`),
-      )}
-    />
+    <div className={className}>
+      <div {...styles('avatarContainer', { size, indication, presence })}>
+        <div className={styles.avatarPresenseContainer}>
```
nitayneeman: Maybe redundant? not sure we need this container

src/Avatar/Avatar.js
```diff
@@ -4,24 +4,52 @@ import PropTypes from 'prop-types';
...
+          <div {...styles('coreAvatar', { size, presence })}>
+            <CoreAvatar
+              {...rest}
+              data-hook={dataHook}
+              className={classNames(
+                avatarStyle(size, color && `color${capitalize(color)}`),
+              )}
+            />
+          </div>
+          {presence && <div {...styles('presence', { size, presence })}></div>}
```
nitayneeman: I think that you don't have to use Stylable in this case because you already injected `size` & `presence` as part of `avatarContainer`:
```suggestion
          {presence && <div className={styles.presence}></div>}
```

But maybe it will complicate the syntax in the style file?

src/Avatar/Avatar.js
```diff
@@ -4,24 +4,52 @@ import PropTypes from 'prop-types';
...
+        </div>
+        {renderIndication && (
+          <div
+            data-hook="indication"
+            onClick={onClick}
+            {...styles('indication', { size, presence })}
```
nitayneeman: Same wonder here

src/Avatar/Avatar.js
```diff
@@ -60,13 +88,19 @@ Avatar.propTypes = {
   className: PropTypes.string,
   /** Applied as data-hook HTML attribute that can be used to create an Avatar driver in testing */
   dataHook: PropTypes.string,
+  /** Avatar presence. Options like 'online' mean that the conatct is online */
+  presence: PropTypes.oneOf(['online', 'offline', 'busy']),
+  /** A node to be rendered as Indication.*/
+  indication: PropTypes.node,
+  /** function which triggered on indication click.*/
```
nitayneeman: Worth mentioning the following, IMO:
```suggestion
  /** function which triggered on indication click. Notice it will be clickable just in case `indication` is provided */
```

src/Avatar/Avatar.js
```diff
@@ -77,6 +111,7 @@ function AvatarDefaultPlaceholder() {
 
 Avatar.defaultProps = {
   placeholder: <AvatarDefaultPlaceholder />,
+  size: 'size48',
```
nitayneeman: I know this is the default according to spec, but I afraid that it might break consumers code

src/Avatar/Avatar.uni.driver.js
```diff
@@ -1,3 +1,8 @@
-export {
-  avatarDriverFactory as avatarUniDriverFactory,
-} from 'wix-ui-core/dist/src/components/avatar/avatar.uni.driver';
+import { avatarDriverFactory } from 'wix-ui-core/dist/src/components/avatar/avatar.uni.driver';
+
+export default base => {
+  return {
+    ...avatarDriverFactory(base),
+    clickIndication: () => base.$('[data-hook="indication"]').click(),
```
nitayneeman: I would import this dataHook from a constant file

src/Avatar/Avatar.uni.driver.js
```diff
@@ -1,3 +1,8 @@
-export {
-  avatarDriverFactory as avatarUniDriverFactory,
-} from 'wix-ui-core/dist/src/components/avatar/avatar.uni.driver';
+import { avatarDriverFactory } from 'wix-ui-core/dist/src/components/avatar/avatar.uni.driver';
```
nitayneeman: Because the fact that Avatar.driver.js is also being used, we need to add the following method there as well

src/Avatar/Avatar.spec.js
```diff
@@ -1,7 +1,33 @@
...
+import { createRendererWithUniDriver, cleanup } from '../../test/utils/unit';
+import avatarUniDriverFactory from './Avatar.uni.driver';
+import { PhotoCamera } from 'wix-ui-icons-common';
 
 describe('Avatar', () => {
   it('should have correct displayName', async () => {
     expect(Avatar.displayName).toEqual('Avatar');
   });
+
+  describe('[async]', () => {
```
nitayneeman: We should also add `[sync]` describe that uses Avatar.driver.js

src/Avatar/Avatar.st.css
```diff
@@ -2,12 +2,163 @@
...
+}
+
+.avatarContainer {
+  -st-states:size(enum(size90, size72, size60, size48,
+  size36, size30, size24, size18)), indication, presence;
+  position: relative;
+  display: inline-block;
+}
+
+.avatarPresenseContainer {
```
nitayneeman: I think this one is redundant, worth to check

src/Avatar/docs/index.story.js
```diff
@@ -132,6 +162,53 @@ export default {
...
         />
       </Cell>
+
+      <Cell>
+        <LiveCodeExample
+          compact
+          scope={{ styles }}
+          title="Presence"
+          initialCode={`
+          <div className={styles.container}>
```
nitayneeman: Remove this line

src/Avatar/docs/index.story.js
```diff
@@ -132,6 +162,53 @@ export default {
...
+        />
+      </Cell>
+
+      <Cell>
+        <LiveCodeExample
+          compact
+          scope={{ React, styles, IconButton, PhotoCamera }}
+          title="Indication"
+          initialCode={`
+          <div className={styles.container}>
```
nitayneeman: Remove this line

src/Avatar/Avatar.js
```diff
@@ -1,27 +1,69 @@
...
 import { Avatar as CoreAvatar } from 'wix-ui-core/dist/src/components/avatar';
 import { avatar as avatarStyle } from 'wix-ui-core/dist/src/themes/backoffice';
 
-import style from './Avatar.st.css';
+import styles from './Avatar.st.css';
 import { capitalize } from '../utils/cssClassUtils';
 
+const getSizeNumber = size => Number(size.substring(4));
+const minIndicationRenderSize = 36;
+const minSmallIconButten = 60;
```
nitayneeman: Typo:
```suggestion
const minSmallIconButton = 60;
```

src/Avatar/Avatar.st.css
```diff
@@ -2,12 +2,169 @@
...
 .placeholder {
   fill: value(F00);
   height: inherit;
   width: inherit;
   border-radius: inherit;
-}
\ No newline at end of file
+}
+.externalContainer{
+  line-height: 0;
```
nitayneeman: We should add a comment above this solution

src/Avatar/Avatar.st.css
```diff
@@ -2,12 +2,169 @@
...
+  align-items: center;
+  position: absolute;
+  border-radius: 50%;
+  box-shadow: value(shadow10);
+  box-sizing: border-box;
+  bottom: 0;
+  right: 0;
+}
+.indication:hover{
+  cursor: pointer;
```
nitayneeman: Still relevant?

src/Avatar/Avatar.st.css
```diff
@@ -2,12 +2,169 @@
...
+
+.avatarContainer:size(size90) .presence{
+  width: 12px;
+  height: 12px;
+  top: 1px;
+  right: 4px;
+}
+
+
+
```
nitayneeman: Remove these empty lines

nitayneeman:
Failed because of flaky tests

PR #4022, opened 2019-08-27
`<ToggleButton>` - added support for 'small' size

dorduch:


mykas:
@milkyfruit please review if this is what we want to have in general. We don't have the official specs for this. Is this valid?

milkyfruit:
@dorduch where the small size is planned to be used?

dorduch:
@milkyfruit  we are upgrading wsr version in the product page, and we have info icons with tooltip that now need to be `ToggleButton` instead of just `Button`

milkyfruit:
@dorduch could you share the Zeplin files? I cannot approve this PR until it's not clear if it's a single use solution or this size property could scale to other products as well.

The current commit allows to change not the component size, but the icon's size only. Which isn't exactly component's size.

In this case I think it could be solved without a new prop, instead allowing to pass `IconNameSmall`.

dorduch:
zpl.io/bAG04Ne

milkyfruit:
@mykas talked with @dorduch in person, this PR was to meant to help migrate stores to React and he was looking for a solution how to display info icon in a form field:
image
I suggested using `<FormField infoContent="tooltips text"/>` instead.

Closing this PR for now.

PR #4024, opened 2019-08-27
`<FillPreview />` - Implement new component

alexgoncharwix:
### 🔦 Summary
...

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


src/FillPreview/FillPreview.js
```diff
@@ -0,0 +1,260 @@
...
+
+import Tooltip from '../Tooltip';
+import IconAdd from '../../new-icons/Add';
+import IconAddSmall from '../../new-icons/AddSmall';
+import styles from './FillPreview.st.css';
+import { withFocusable } from 'wix-ui-core/dist/src/hocs/Focusable/FocusableHOC';
+import { func, oneOf, oneOfType, string, node, bool } from 'prop-types';
+import Proportion from '../Proportion';
+import { parseColorGradient } from './GradientParser';
+import isUrl from 'is-url';
```
mykas: Here in wix-style-react we only use external libraries when we really need to. For example this package: `is-url` is just a function with some regex. https://github.com/segmentio/is-url/blob/master/index.js

It would be best if you recreate this in `utils` folder and use it directly without downloading extra packages.

src/FillPreview/FillPreview.js
```diff
@@ -0,0 +1,260 @@
...
+    );
+  }
+
+  return fillInfo.element || null;
+};
+
+class FillPreview extends React.Component {
+  constructor(props) {
+    super(props);
+    this.resolveFillType = memoizeOne(resolveFillType);
```
mykas: can you explain me a bit why do we need to use this here?

src/FillPreview/FillPreview.js
```diff
@@ -0,0 +1,260 @@
...
+  }
+
+  wrapWithTooltip(element) {
+    const { addModeTooltipContent, mode, disabled } = this.props;
+    return addModeTooltipContent && mode === 'add' && !disabled ? (
+      <Tooltip
```
mykas: Actually this is a bit confusing to understand. Instead of rendering different tree on mode change - you should use `disabled` prop and you wouldn't need a special hoc for switching the component render.

src/FillPreview/docs/index.story.js
```diff
@@ -0,0 +1,204 @@
...
+              title: 'Add Icon',
+              text:
+                'Pass `mode="add"` and control icon size via `addIconSize` prop (possible values are `small` and `normal`, default value is `small`). ' +
+                'Icon size becomes light or dark based on background color luminosity.',
+            }),
+
+            code({
+              compact: true,
+              source: `
+                <div style={{ width: '190px', display: 'grid', gridTemplateColumns: 'repeat(4, 1fr)', gridGap: 10 }}>
```
mykas: We usually show examples without custom css. It would be better if you could reuse our <Layout/> component which is already coming with css grid built in.

src/FillPreview/docs/index.story.js
```diff
@@ -0,0 +1,204 @@
...
+                    <FillPreview mode="add" />
+                    <FillPreview mode="add" addIconSize="small" fill="#28c197"/>
```
mykas: addIconSize should be in different example. remember - one feature = one example block.

src/FillPreview/docs/index.story.js
```diff
@@ -0,0 +1,204 @@
...
+              title: 'Selected and disabled states',
+              text:
+                'Pass `selected` or `disabled` props accordingly. If a component in a disabled it cannot be selected. ' +
+                'Also it cannot be selected when `mode==="add".`',
+            }),
+
+            code({
+              compact: true,
+              source: `
+                <div style={{ width: '140px', display: 'grid', gridTemplateColumns: 'repeat(3, 1fr)', gridGap: 10 }}>
```
mykas: same is here. Should be used either with <Layout/> or <Box/> component.

src/FillPreview/test/FillPreview.e2e.js
```diff
@@ -0,0 +1,40 @@
+import {
```
mykas: lets drop this file and also the test story. Currently for static visual tests we use .visual file in order not to make our builds worse.

src/FillPreview/test/FillPreview.private.uni.driver.js
```diff
@@ -0,0 +1,25 @@
...
+  return {
+    ...publicDriverFactory(base),
+
+    addIconExists: async () =>
+      base.$('[data-hook="fill-preview-add-icon"]').exists(),
+
+    hasTooltip: async () =>
+      base.$('[data-hook="fill-preview-tooltip"]').exists(),
+
+    getTooltipText: async () => tooltipTestkit.getTooltipText(),
```
mykas: should be public

src/FillPreview/test/FillPreview.private.uni.driver.js
```diff
@@ -0,0 +1,25 @@
...
+
+    isSelected: async () =>
```
mykas: should be public

src/FillPreview/FillPreview.js
```diff
@@ -0,0 +1,260 @@
...
+        url: fill,
+      };
+    }
+
+    // Simple color (human name or hex)
+    try {
+      return {
+        color: resolveColor(fill),
+      };
+    } catch (e) {}
```
mykas: How is this `catch` will handle your case? I don't think resolveColor will actually throw an error to make this try unresolved.

src/FillPreview/FillPreview.js
```diff
@@ -0,0 +1,260 @@
...
+      colorStops.map(colorStop => {
+        return resolveColor(colorStop);
+      });
+    }
+  }
+
+  return [];
+}
+
+const resolveFillType = fill => {
```
mykas: For me this function feels like pattern matcher.

mykas: The if statements does not feel like understandable.

mykas: I would do something like this:
```
const fillType = [
  {
    when: fill => isString(fill) && isUrl(fill),
    make: fill => ({ url: fill }),
  },
  {
    when: fill => isString(fill) && resolveColor(fill),
    make: fill => ({ color: resolveColor(fill) }),
  },
  {
    when: fill => isString(fill) && parseColorGradient(fill),
    make: fill => ({ gradient: parseColorGradient(fill) }),
  },
  {
    when: fill => React.isValidElement(fill),
    make: fill => ({ element: fill }),
  },
  {
    when: fill => !fill,
    make: () => ({ isEmpty: true }),
  },
];

const resolveFillType = fill =>
  fillType.find(({ when }) => when(fill)).make(fill);
```

src/FillPreview/FillPreview.js
```diff
@@ -0,0 +1,260 @@
...
+import { parseColorGradient } from './GradientParser';
+import isUrl from 'is-url';
+import memoizeOne from 'memoize-one';
+import resolveColor from 'color';
+
+const COLOR_ICON_LIGHT = '#ffffff';
+const COLOR_ICON_DARK = '#162d3d';
+const COLOR_ICON_EMPTY_COLOR = '#3899ec';
+
+export const FillPreviewAddIconSize = oneOf(['small', 'normal']);
```
mykas: redundant extraction. Why do you need to export this?

src/FillPreview/FillPreview.js
```diff
@@ -0,0 +1,260 @@
...
+
+  // Element
+  if (React.isValidElement(fill)) {
+    return {
+      element: fill,
+    };
+  }
+
+  return {
+    isUnknown: true,
```
mykas: This api does not make sense. You create some kind of flags to check the state of the fill then you have another pattern matcher here:https://github.com/wix/wix-style-react/pull/4024/files#diff-cf30b6db8804e39dbc53b7e9139dcdb6R102

Where you return css params based on filltype value. Maybe just return css value   and you wouldn't need this `getStyle` function.

src/FillPreview/FillPreview.js
```diff
@@ -0,0 +1,260 @@
...
+const COLOR_ICON_LIGHT = '#ffffff';
+const COLOR_ICON_DARK = '#162d3d';
+const COLOR_ICON_EMPTY_COLOR = '#3899ec';
+
+export const FillPreviewAddIconSize = oneOf(['small', 'normal']);
+export const FillPreviewMode = oneOf(['fill', 'add']);
+
+function getContrastColor(
+  colorsList,
+  light = COLOR_ICON_LIGHT,
```
mykas: Any point making this a parameter overridable if we use it internally and never pass light or dark? or am I something missing?

src/FillPreview/test/FillPreview.e2e.js
alexgoncharwix: Ok, drop plz.

.wuf/components.json
```diff
@@ -385,5 +382,11 @@
   },
   "StatisticsWidget": {
     "path": "src/StatisticsWidget"
+  },
+  "FillPreview": {
+    "path": "src/FillPreview"
+  },
+  "LabelledElement": {
```
mykas: I think we have some code here that is not suppose to be here.

PR #4028, opened 2019-08-28
Enable making table rows unselectable

splintor:
### 🔦 Summary
Enable making some rows unselectable, so a checkbox does not appear for them in the selection column.

This is needed for the new list view of sites list, because folders should not currently be selected, as the selection is for moving sites, and folders cannot be moved.

### ✅ Checklist
- [X] 👨‍💻 API change is approved by the UI Infra Developers @nitayneeman 
- [X] 👨‍🎨 UX change is approved by the Design System UX @shlomitc 
- 📚 Change is documented
  - [X] Story
- 🔬 Change is tested
  - [X] Component
  - [X] Visual test


src/Table/Table.js
```diff
@@ -111,6 +113,17 @@ export class Table extends React.Component {
...
 
   render() {
+    let hasUnselectables = null;
+    let allIds = this.props.data.map((rowData, rowIndex) =>
+      rowData.unselectable
+        ? (hasUnselectables = hasUnselectablesSymbol)
+        : defaultTo(rowData.id, rowIndex),
+    );
+
+    if (hasUnselectables === hasUnselectablesSymbol) {
```
nitayneeman: I like symbols, but not sure what's the benefit here - why is `hasUnselectables` not just a boolean?

src/Table/docs/index.story.js
```diff
@@ -146,7 +146,8 @@ export default {
             {
               title: 'Table with Toolbar',
               description:
-                'This is a basic table that implements a toolbar using `<Table.ToolbarContainer/>` - which receives the SelectionContext (`showSelection` is true in order to enable the selection). Notice that the whole table is wrapped by `<Card/>` so that the toolbar seems like part of the table. In addition, we render the main toolbar when there are no selected rows, and an actions toolbar when some rows are selected.',
+                'This is a basic table that implements a toolbar using `<Table.ToolbarContainer/>` - which receives the SelectionContext (`showSelection` is true in order to enable the selection). Notice that the whole table is wrapped by `<Card/>` so that the toolbar seems like part of the table. In addition, we render the main toolbar when there are no selected rows, and an actions toolbar when some rows are selected.' +
```
nitayneeman: Because we're trying to simplify the examples as much as possible (and we still have a long way to there), I think it would be better to export this part to a dedicated example - something like "Table with Unselectable Rows" with the minimal stuff that necessary (most of the header logic is redundant for example)

nitayneeman:
Obviously we've conflicts due to https://github.com/wix/wix-style-react/pull/4027

src/Table/Table.js
splintor: I don't know what are the possible values of `row.id` and didn't want to limit the user from using a boolean for the `id`, so I used a value which surely cannot be used as an `id`. But if you think we can assume `row.id` is never `true`, I can change this.

src/Table/docs/index.story.js
splintor: 👍 Done

src/Table/Table.js
```diff
@@ -16,6 +16,8 @@ import {
   TableEmptyState,
```
nitayneeman: We forgot to add the new prop to the columns PropTypes:
https://github.com/wix/wix-style-react/blob/4bdc210a386495a08414abc15fe0fbdb31a79b62/src/Table/Table.js#L212

src/Table/Table.js
nitayneeman: Fair enough

src/Table/Table.js
splintor: What new prop? `unselectable` is a row field, not a column prop...

nitayneeman: You're right. I meant to the `data` prop description so it would be part of the auto-generated API:
https://github.com/wix/wix-style-react/blob/master/src%2FTable%2FTable.js#L198

splintor: Fixed - please review

PR #4031, opened 2019-08-29
<FloatingNotification/> - Moved the max-width property from the Container to Text

yogevitz:
### 🔦 Summary

As advised by @milkyfruit,
Made it possible to change the `width` of a `Floating Notification`, by moving the `max-width` property to the `Text` inside the `Floating Notification`'s container.

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


yogevitz:
@mosheKerbel 
This means that I would set the `width` prop in-line for the div of the container:
image
I just thought it would be less versatile...

mosheKerbel:
> @mosheKerbel
> This means that I would set the `width` prop in-line for the div of the container:
> image
> I just thought it would be less versatile...

Sorry not sure that i understand
But the idea was to expose `width` prop for the `<FloatingNotificatoin/>` comopnent itself.

mosheKerbel:
we're out for today (team activity) so i'll be able to re-review Tomorrow.

yogevitz:
Thanks, I will update the PR with the changes

mosheKerbel:
test1:unit and test4:e2e are failing due to unrelated flaky tests, so no problem here
test3:e2e - is failing over diffs in shapshots of the actual FloatingNotification, i need to figure it out first.

mosheKerbel:
@nitayneeman could you please take this one?

nitayneeman:
Hey @yogevitz,

I'm checking why the E2Es are broken.
Can you add an example with different width in the meantime?
https://github.com/wix/wix-style-react/blob/master/src%2FFloatingNotification%2Fdocs%2Findex.story.js#L84

nitayneeman:
The E2Es failed because we remove `maxWidth` from the parent so that in case the element is pretty wide it actually makes sense...

A test for example (before the change):
image

After:
image

So I'm not sure we should remove `maxWidth` because it's a breaking change. But we can approach in another way. Le'ts bring back `maxWidth` as it was before, but in case `width` is given we'll change both `width` and `maxWidth` to the value.

Like that:
```javascript
<div
        data-hook={dataHook}
        className={classNames(styles.root, styles[type], className)}
        style={{width, maxWidth: width}}
      >
```

It means that when `width` isn't given it behaves as it was before, and otherwise - `width` takes control.

nitayneeman:
> @yogevitz @nitayneeman tested the component, from UX side `width` prop works fine.

Great!

I just wanna make sure the CI passes successfully and then we'll merge it.

src/FloatingNotification/FloatingNotification.js
```diff
@@ -52,27 +52,35 @@ class FloatingNotification extends React.PureComponent {
...
+    /** The width of the content container of the notification */
+    width: PropTypes.string,
   };
 
   static defaultProps = {
     type: NOTIFICATION_TYPES.STANDARD,
     buttonProps: {},
     textButtonProps: {},
     showCloseButton: true,
+    width: 'auto',
```
nitayneeman: I think that this what breaks the E2Es. We're already familiar with a rendering bug related to the `<Text/>` usage in this component, and that default value triggers the bug indirectly.

My suggestion is to remove this line, because when `width` isn't provided it won't be rendered anyway (undefined value) - which is good.

src/FloatingNotification/FloatingNotification.scss
```diff
@@ -12,6 +12,7 @@ $buttonMaxWidth: 25%;
   height: 36px;
   padding-left: 12px;
   padding-right: 6px;
+  width: auto;
```
nitayneeman: This is also redundant

src/FloatingNotification/FloatingNotification.scss
```diff
@@ -26,6 +27,7 @@ $buttonMaxWidth: 25%;
   .text,
   .textButton {
     margin-left: 6px;
+    max-width: 540px;
```
nitayneeman: Because we returned `max-width` to the parent level - this line is also redundant

yogevitz: I think this isn't redundant because we want to set a default `max-width` in case it is not given.
Notice that I change the `max-width` property only if given a `width` in the component render function: `width && (style.maxWidth = width);`
Agree?

nitayneeman: I agree we need a default value for `maxWidth`. The thing is we also have it on the root:
https://github.com/wix/wix-style-react/blob/e12e3c08744efc2e36bbc02a13a5b20cbab6184b/src/FloatingNotification/FloatingNotification.scss#L16

So I think it will still work, exactly as it was before we did the changes.
You can verify it when you do the changes and please let me know if it doesn't work  :)

nitayneeman:
@yogevitz Let me know once you push the fixes

PR #4039, opened 2019-09-02
docs(*): remove redundant properties from various sections

argshook:
* `importExample` supports string as argument, no need for `{ source: 'string' }`
* `importExample` does not need to be within `columns`, it was agreed with ux some time ago
* `columns` supports array as argument, no need for `{ items: [] }`
* removed some unnecessary abstractions

harmless update :)


argshook:
unrelated e2e time outs in builds, otherwise good to merge

PR #4045, opened 2019-09-03
Add <LabelledElement> component - initial implementation

giladsegal:
I am still missing visual tests, will add them ASAP


stories/index.js
```diff
@@ -201,3 +201,9 @@ require('../src/beta/PopoverMenu/docs/index.story');
 
 // TODO: move to correct position
 require('../src/StatisticsWidget/docs/index.story');
+
+// TODO: move to correct position
```
nitayneeman: Move it to the correct position as the TODO says :)

Which is:
https://github.com/wix/wix-style-react/blob/34e74a13f3075e888c4ef183d2598d102eddbb0e/stories/index.js#L143

src/LabelledElement/test/LabelledElement.spec.js
```diff
@@ -0,0 +1,120 @@
...
+    const onBlurSpy = jest.fn();
+    const { driver } = createDriver({
+      children: <input data-hook="my-input" onBlur={onBlurSpy} />,
+    });
+    const childrenDriver = driver.getChildrenDriver();
+    const childInputDriver = await childrenDriver.$('[data-hook="my-input"]');
+
+    await childInputDriver.click();
+    const reactBase = ReactBase(childInputDriver);
+    await reactBase.blur();
```
nitayneeman: I think it would be better to implement `blur` method which uses ReactBase within the private driver, so that the spec won't have to deal with ReactBase at all

src/LabelledElement/LabelledElement.js
```diff
@@ -0,0 +1,102 @@
...
+
+  render() {
+    const { hasFocus, inputId } = this.state;
+    const { dataHook, label, children } = this.props;
+
+    const placeLabelOnTop = hasFocus || !this._isInputEmpty();
+
+    return (
+      <div className={styles.root} data-hook={dataHook}>
+        <label
```
nitayneeman: WSR exports the `<Label/>` component from wix-ui-core:
https://wix-wix-style-react.surge.sh/?selectedKind=Components&selectedStory=Label&full=0&addons=0&stories=1&panelRight=0

So, consider using it. 

I think the main benefit is you get testkit methods such as `getLabelText` and `getLabelForAttribute`, and you don't need to implement on your own (check out the "Testkit" tab).

src/LabelledElement/LabelledElement.uni.driver.js
```diff
@@ -0,0 +1,21 @@
...
+  const getLabel = () => base.$('[data-hook="labelledElement-label"]');
+
+  return {
+    ...baseUniDriverFactory(base),
+
+    /** Gets the content of rendered label */
+    getLabelText: () => getLabel().text(),
+
+    /** Gets the ID of assoicated input element */
+    getLabelForAttribute: () => getLabel().attr('for'),
```
nitayneeman: As said above, `getLabelText` & `getLabelForAttribute` could be reused by Label's driver

src/LabelledElement/LabelledElement.uni.driver.js
```diff
@@ -0,0 +1,21 @@
...
+
+    /** Checks if position the label is positioned at thetop of the input element  */
+    isLabelAtTop: () => getLabel().hasClass('labelTop'),
```
nitayneeman: Also I think we could place `getLabelForAttribute` &  `isLabelAtTop` within the private driver, because these apparently don't really interest the consumer (they deal with internal concerns of the component)

src/LabelledElement/LabelledElement.js
```diff
@@ -0,0 +1,102 @@
...
+          data-hook="labelledElement-label"
```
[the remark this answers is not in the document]
giladsegal: If you like this convention we should probably add it to the generator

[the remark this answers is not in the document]
nitayneeman: Definitely, I'll open an issue for that

src/LabelledElement/LabelledElement.scss
```diff
@@ -0,0 +1,32 @@
+@import '../common';
+
+.root {
+  text-align: center;
```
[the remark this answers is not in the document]
giladsegal: good catch, leftovers from the generator I think

src/LabelledElement/LabelledElement.uni.driver.js
giladsegal: Personally I don't like the concept of "private" drivers - if you want to test it, maybe someone else will want to test it as well, unless you have a really good reason (e.g. using `ReactBase` I prefer to avoid it)

argshook: sometimes we want to ensure behaviour by using reusable function in a consistent manner but without exposing it as public api. that's why the private testkit idea was born.

public testkit are public api, if we publish something internal there, it's gonna get used and we're locked in, unable to fix/update.

also, public testkit should be written in a way that would mimic user
actions and not things like `getClassName` or `getStylableState`. Those
cases should be part of private testkit

giladsegal: Ok - I can move `getLabelForAttirubute` to the private driver, but `isLabelAtTop` should remain public I think

nitayneeman: In case you think the consumer could benefit from `isLabelAtTop` and actually use that - go for it

nitayneeman:
The CI fails because of flaky tests

nitayneeman:
Hey, 

This input is going to be used for Premium (Checkout).
Is there a chance that any of you could do a UX review, please? @LittleTal @ayeletr

LittleTal:
> Hey @LittleTal / @ayeletr,
> 
> This component is going to be used for Premium (Checkout):
> https://wix-wix-style-react-pr-4045.surge.sh/?selectedKind=WIP&selectedStory=LabelledElement&full=0&addons=0&stories=1&panelRight=0
> 
> Is there a chance that any of you could do a UX review, please?

Need to fix:
1. Text cursor and BG color on hover disappear while hovering the label 
2. Need to add margin top for input text, currently:
image

should be:
image
 
(Maybe label position should be different)

3. Label text size is 16px, should be 12px
4. Label with no value(placeholder) is not vertically center aligned:
image

giladsegal:
Hey @LittleTal 
Just a few clarifications about this component:
It is a base component without relevant styles, it just takes care of the label and transition.
We will implement two components that style the input properly (`InputWithLabel` and `AutocompleteWithLabel`)

Regarding your comments:
1. Will be taken care of
2. Will be implemented in `InputWithLabel` and `AutocompleteWithLabel`
3. Label `font-size` is `16px` indeed but it is scaled down by 25% which effectively changes the `font-size` to `12px`
4. Will be implemented in `InputWithLabel` and `AutocompleteWithLabel`
5. Can you please clarify? The focus behavior is the same as `Input` component

nitayneeman:
> Hey @LittleTal,
> Just a few clarifications about this component:
> It is a base component without relevant styles, it just takes care of the label and transition.
> We will implement two components that will style the input properly (`InputWithLabel` and `AutocompleteWithLabel`)
> 
> Regarding your comments:
> 
>     1. Will be taken care of in the scope of this PR
> 
>     2. Will be implemented in `InputWithLabel` and `AutocompleteWithLabel`
> 
>     3. Label `font-size` is `16px` indeed but it is scaled down by 25% which effectively changes the `font-size` to `12px`
> 
>     4. Will be implemented in `InputWithLabel` and `AutocompleteWithLabel`
> 
>     5. Can you please clarify? The focus behavior is the same as `Input` component
> 
> 
> This is the PR for [InputWithLabel](https://github.com/wix/wix-style-react/pull/4051) (still a WIP)

Once you push a fix for the first comment, we'll merge this PR.
The rest will be handled as part of https://github.com/wix/wix-style-react/pull/4051.

LittleTal:
@giladsegal @nitayneeman 
5. looks like it was fixed, maybe it was a bug

PR #4052, opened 2019-09-05
Autocomplete with label

alonshmiel1:
- [ ] - Storybook description tab header + examples should contain options
- [ ] - Clicking on the chevron should trigger input focus
- [ ] - Chevron should be vertically centered
- [ ] - Add testkit documentation
- [ ] - Add e2e tests
- [ ] - Add tests for `InputWithOptions` predicate prop + documentation
- [ ] - Verify that onSelect is working in uncontrolled mode


src/InputWithOptions/InputWithOptions.js
```diff
@@ -109,9 +109,12 @@ class InputWithOptions extends Component {
   }
```
nitayneeman: What is the motivation for these changes?

Let's do them in a separate PR and also make sure to cover them by appropriate tests

src/AutocompleteWithLabel/AutocompleteWithLabel.js
```diff
@@ -0,0 +1,87 @@
...
+      this.setState({ value: event.target.value });
+    };
+
+    const {
+      label,
+      options,
+      value,
+      dataHook,
+      size: _size,
+      ...rest
```
nitayneeman: Let's remove that, so we'll use only the necessary props (once we need some <Input/> props, we would add them explicitly)

nitayneeman:
@yakirn Do you know who is going to complete this PR?

yakirn:
@nitayneeman best consult with @assafswix

nitayneeman:
> @nitayneeman best consult with @assafswix

@assafswix  I'm closing this PR at the moment.
Let's reopen when there is someone who's dedicated to complete the work.

src/AutocompleteWithLabel/docs/index.story.js
```diff
@@ -0,0 +1,104 @@
...
+  exampleProps: {
+    options: [{ label: 'options', value: options }],
+  },
+
+  sections: [
+    header({
+      issueUrl: 'https://github.com/wix/wix-style-react/issues/new',
+      sourceUrl:
+        'https://github.com/wix/wix-style-react/tree/master/src/AutocompleteWithLabel/',
+      component: <AutocompleteWithLabel label="label" />,
```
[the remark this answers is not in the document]
saarkuriel: What part?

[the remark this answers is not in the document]
nitayneeman: `component: <AutocompleteWithLabel label="label" />`

Another option is adding there a better example instead, but we'll have anyway as part of the Story examples below, so I wouldn't invest effort on that

.wuf/testkits/definitions.js
```diff
@@ -727,10 +721,7 @@ module.exports = {
   },
 
   AutoCompleteComposite: {
-    vanillaLegacyTestkit: true,
-    enzymeLegacyTestkit: true,
-    protractorLegacyTestkit: true,
-    noPuppeteerTestkit: true,
+    noPuppeteerTestkit: true, // was not exported before automation, keeping that way for now
```
nitayneeman: This fails the CI. 

It should be:
```suggestion
    vanillaLegacyTestkit: true,
    enzymeLegacyTestkit: true,
    protractorLegacyTestkit: true,
    noPuppeteerTestkit: true,
```

.wuf/components.json
```diff
@@ -400,5 +400,8 @@
   },
   "Divider": {
     "path": "src/Divider"
+  },
+  "LabelledElement": {
```
argshook: Where are all these changes from? Rebasing on latest master should not require any of the generated changes

argshook:
note for whoever is going to work next on this:

i'd prefer component to be:
```diff
-AutocompleteWithLabel
+AutoCompleteWithLabel
```

that's simply because we already have `AutoComplete` and `AutoCompleteComposite`.

yes, it's quite a requirement for a very minor thing but we already have
many tiny inconcistencies like that and when they add up, it causes confusion.

if whoever is going to work next on this has troubles with mass rename,
please contact me, i'm happy to help

src/AutoCompleteWithLabel/test/AutoCompleteWithLabel.visual.js
```diff
@@ -0,0 +1,99 @@
...
+    storiesOf(`AutoCompleteWithLabel/${describe}`, module).add(it, () => (
+      <AutoCompleteWithLabel {...commonProps} {...props} />
+    ));
+  });
+});
+
+interactiveTests.forEach(({ describe, its }) => {
+  its.forEach(({ it, props, componentDidMount }) => {
+    storiesOf(`InputWithLabel/${describe}`, module).add(it, () => (
+      <InteractiveEyeTest {...props} componentDidMount={componentDidMount} />
```
nitayneeman: Minor but we could use there `commonProps` as well:
 ```suggestion
      <InteractiveEyeTest {...commonProps} {...props} componentDidMount={componentDidMount} />
```

File: src/AutoCompleteWithLabel/docs/index.story.js [hunk 7]
```diff
@@ -0,0 +1,114 @@
...
+  component: AutoCompleteWithLabel,
+  componentPath: '..',
+
+  componentProps: {
+    value: '',
+    label: 'my label',
+  },
+
+  exampleProps: {
+    options: [{ label: 'options', value: options }],
```
nitayneeman: Do you have some good example of `suffix` to add there?

Currently when you try to insert some text within the "suffix" field in Storybook it throws an error that makes the page be blank

src/AutoCompleteWithLabel/AutoCompleteWithLabel.js
```diff
@@ -0,0 +1,120 @@
...
+  static propTypes = {
+    dataHook: PropTypes.string,
+    /** label to apear in input */
+    label: PropTypes.string.isRequired,
+    options: PropTypes.arrayOf(optionValidator).isRequired,
+    suffix: PropTypes.arrayOf(PropTypes.element),
+    status: PropTypes.oneOf([
+      Input.StatusError,
+      Input.StatusWarning,
+      Input.StatusLoading,
```
nitayneeman: Maybe it's worth to write the values explicitly hardcoded because our AutoDocs doesn't know to deal with that at the moment

src/AutoCompleteWithLabel/docs/index.story.js
saarkuriel: I think that since `InputWithLabel` provides this options, this component should too cause this is the same UX...

saarkuriel:
> 3. Hover cursor should be pointer
>    image

This is the same as in `InputWithOptions`, cause once you click on the component you can start writing. Are you sure I should change it?

LittleTal:
@saarkuriel Ok, no problem, let's keep it as it is

PR #4053, opened 2019-09-05
Add additionalInfo for <StatisticsWidget /> section

MSiliunas:
### 🔦 Summary
Adds `additionalInfo` prop for statistics items.
We will need to show our insights below the value in each section. `percentage` does not cover our case so it seems like a good idea to have an ability to render whatever is needed.

### ✅ Checklist
- 📚 Change is documented
  - [x] Story
  - [x] API description
- 🔬 Change is tested
  - [ ] Component
  - [x] Visual test


mosheKerbel:
@MSiliunas can you add a mock of how you expect the widget to look?

src/StatisticsWidget/StatisticsWidget.js
```diff
@@ -29,6 +30,7 @@ class StatisticsWidget extends React.PureComponent {
         percentage: PropTypes.number,
         invertedPercentage: PropTypes.bool,
         onClick: PropTypes.func,
+        additionalInfo: PropTypes.node,
```
argshook: i would call this `children`

src/StatisticsWidget/StatisticsItem/StatisticsItem.js
```diff
@@ -134,6 +134,16 @@ class StatisticsItem extends React.PureComponent {
     );
   };
 
+  _renderAdditionalInfo = additionalInfo => {
+    if (!additionalInfo) {
+      return null;
+    }
+    if (typeof additionalInfo === 'string') {
+      return <div>{additionalInfo}</div>;
```
argshook: what is the div for? if `additionalInfo` (or i'd prefer `children`) is for arbitrary content, then i think there shouldn't be any assumptions

MSiliunas: in case `string` is passed I figured that it should be wrapped in something? or should we just render what is passed?

src/StatisticsWidget/StatisticsWidget.js
MSiliunas: renamed 👌

src/StatisticsWidget/StatisticsItem/StatisticsItem.js
MSiliunas: returns `children` as is

mosheKerbel:
hey @MSiliunas thanks for the contribution to the library, we encourage it and appreciate it.
The way we work with new features is a bit different.
You should open us an issue with a feature request and if needed, ping us in `wix-style-react` slack channel to accelerate our response. 
This way our UX are able to review it and see if the requirement makes sense in terms of ux, before the **actual development starts**, so we can save time.

For this use case, i've talked to @LittleTal 
Please add here the Zepplin spec to your requirement for her to review and see if this solution fits.

MSiliunas:
> hey @MSiliunas thanks for the contribution to the library, we encourage it and appreciate it.
> The way we work with new features is a bit different.
> You should open us an issue with a feature request and if needed, ping us in `wix-style-react` slack channel to accelerate our response.
> This way our UX are able to review it and see if the requirement makes sense in terms of ux, before the **actual development starts**, so we can save time.
> 
> For this use case, i've talked to @LittleTal
> Please add here the Zepplin spec to your requirement for her to review and see if this solution fits.

@mosheKerbel @LittleTal 
I talked with Drąsius and he said that today he should have something ready to show. For now we only had wireframes.

argshook:
@MSiliunas 	hey there, maybe you have updates?

MSiliunas:
Drasius had to speak with your UX about this. Im on vacation right now.

kranux:
@argshook we talked with @mykas that it makes sense to pass `...rest` props to <div> as it makes more flexible and helps us achieving our functionality:

image

We need to show / hide tooltip on `children` while mouseEnter / leave statistics card.

argshook:
I added a test for `...rest`.

as we all are agreeing, this is good to merge.
Thanks for the effort!

PR #4054, opened 2019-09-05
Add prop to <Input> to hide error suffix for error status

giladsegal:


src/Input/Input.js
```diff
@@ -486,6 +491,9 @@ Input.propTypes = {
   /** The status (error/loading) message to display when hovering the status icon, if not given or empty there will be no tooltip */
   statusMessage: PropTypes.node,
 
+  /** When set to true, this input won't display error suffix for error status   */
+  hideErrorStatusSuffix: PropTypes.bool,
```
shlomitc: Why not make it  for any status type?

PR #4063, opened 2019-09-08
`<Button/>`- fixing isButtonDisabled testkit function

### LINKED ISSUE #3160: <Button/>/<TextButton> testkit `isButtonDisabled` doesn't work
# 🐛 Bug Report

### 🏗 Relevant Components

`<Button/>`
`<TextButton/>`

### 😯 Current Behavior

Using Testkit's `isButtonDisabled` returns `false` even when buttons are disabled 
for e.g using - 
```
await (await (await buttonDriver.element()).getProperty(
            'disabled'
          )).jsonValue()
```
works while `isButtonDisabled` doesn't

### 🤔 Expected Behavior

testKits method should work according to docs description

### 👣 Steps to Reproduce
Create a `<Button disabled>` component and use the testkit to see if it's disabled.

  - **WSR version**: `6.7.3`
  - **Yoshi version**: `3`

### 👀 Severity

- Critical / Major / Low

TalZa1810:
### 🔦 Summary
Resolves #3160 

`ButtonNext` already uses `aria-disabled` when button is disabled. Changed the `Button` unidriver to use the `ButtonNext` unidriver instead of checking the stylable disable state:
https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/button-next/button-next.tsx#L75

Should be merged after the `wix-ui-core` ButtonNext unidriver PR is merged:
https://github.com/wix/wix-ui/pull/1466

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [ ] Visual test


TalZa1810:
@NitayRabi 
I think there's coverage for this change. Let me know if you think we should add more tests.

NitayRabi:
Just to make sure we stay covered I would create a E2E to navigate to page with disabled button and testing its disabled.

shlomitc:
I prefer not adding new E2Es just to verify that a driver works. We should expect unidriver to work. Unitest should be enough

shlomitc:
Btw, shouldn’t be a driver feature of the core component?

TalZa1810:
I'm creating another PR for the wix-ui-core `ButtonNext` to resolve the issue for the `TextButton`. 
The same change for the `ButtonNext` unidriver. The wsr `Button` unidriver don't use the `ButtonNext` unidriver, it has it's own driver.

TalZa1810:
so can I merge this one if all tests pass? @shlomitc

shlomitc:
Don’t know about the TextButton, but should t the solution be the same for all button types? Solving it in the core and reusing the driver sounds like the right way to go

TalZa1810:
I see what you mean. Yep, I'll fix this in `wix-ui-core` and add only extra function in the driver (`hasSkin`) instead of having a duplication of the functions both here and in `wix-ui-core`.

TalZa1810:
Updated to use `ButtonNext` unidriver. 
Should be merged after the wix-ui-core `ButtonNext` unidriver PR is merged:
wix/wix-ui#1466

mosheKerbel:
approved 
can be merged after it's using the new core in ci tests

PR #4070, opened 2019-09-09
feat(ImageViewer): add a prop for render Remove Button

miroshSa:
### 🔦 Summary
For MMgr new feature related to board, we need to hide RemoveButton, because for board creation user should always have an image for the board.  

### ✅ Checklist
- 🔬 Change is tested
  - [x] Component


src/ImageViewer/tests/ImageViewer.spec.js
```diff
@@ -125,6 +125,13 @@ describe('ImageViewer', () => {
           await driver.clickRemove();
           expect(onRemoveImage).toHaveBeenCalled();
         });
+
+        it('should NOT render remove button when prop "showRemoveButton" equals to "false"', async () => {
```
mosheKerbel: @stanislavpa 
looks great
plz add also the positive case
`should **render** remove button when prop "showRemoveButton" equals to "true"'`

PR #4073, opened 2019-09-10
docs(PageLayout): add example of stretched <AddItem/>

nitayneeman:
This PR improves the PageLayout docs to demonstrate how to stretch vertically `<AddItem/>` when it's contained inside a layout.


mosheKerbel:
👍 
The only thing i wonder about is how useful it is for devlopers when it is inside a UX story.

mosheKerbel:
merging this 
i think we should also make the documentation of `<Proportion/>` comopnent better, and add reference to it from the `<AddItems/>` docs as well.

PR #4085, opened 2019-09-12
docs - testing examples

mosheKerbel:
@argshook


mosheKerbel:
merged
flaky tests are not related.

mosheKerbel:
This is a compromise
When we move to react 16, we will also be able to move to storybook 5 and then add "hidden" stories of readme which we'll be able to navigate to, instead of navigating to github.

PR #4087, opened 2019-09-12
`<FillButton/>` - new component

mykas:
This component should cover the Swatches missing feature for AddButton.


src/FillButton/utils.js
```diff
@@ -0,0 +1,53 @@
+import gradient from 'gradient-parser';
+import { isString } from '../utils/StringUtils';
+import resolveColor from 'color';
+import { COLOR_ICON_LIGHT, COLOR_ICON_DARK } from './constants';
+
+export const parseColor = fill => {
+  if (!isString(fill)) {
```
argshook:
```suggestion
  if (typeof fill === 'string') {
```
rašyk kaip žmogus

PR #4092, opened 2019-09-12
<BarChart> - add component

maxvektor:


src/BarChart/BarChart.js
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * Array of statistic items
+     *  * `value` - value of the statistic. Displayed as big text in the first row.
+     *  * `description` - short label under the bar.
+     *  * `descriptionInfo` - long description.
+     */
+    statistics: PropTypes.arrayOf(
+      PropTypes.shape({
+        value: PropTypes.number.isRequired,
```
maxvektor: + valueInShort

src/BarChart/BarChart.js
```diff
@@ -0,0 +1,74 @@
...
+      </Heading>
+    </div>
+  );
+
+  _renderItem = ({ value, description, descriptionInfo }, key) => (
+    <div
+      className={styles.item}
+      key={key}
+      data-hook={dataHooks.bar}
+      style={{
```
maxvektor: cssVariable? stylable?

src/BarChart/docs/index.story.js
```diff
@@ -0,0 +1,106 @@
...
+
+export default {
+  category: storySettings.category,
+  storyName: 'BarChart',
+
+  component: BarChart,
+  componentPath: '..',
+
+  componentProps: {
+    buttonText: 'Hello World!',
```
maxvektor: text

src/BarChart/test/BarChart.visual.js
```diff
@@ -0,0 +1,29 @@
...
+import { storiesOf } from '@storybook/react';
+import BarChart from '../BarChart';
+
+const commonProps = {
+  //use for repeated props across the tests (e.g. {buttonText: 'example'})
+};
+
+const tests = [
+  {
+    describe: 'sanity', // prop name (e.g. size)
```
maxvektor: tests

src/BarChart/BarChart.js
```diff
@@ -0,0 +1,146 @@
...
+class BarChart extends React.PureComponent {
+  static displayName = 'BarChart';
+
+  static propTypes = {
+    /** Applied as data-hook HTML attribute that can be used to create driver in testing */
+    dataHook: PropTypes.string,
+
+    /**
+     * Array of items
+     *  TODO: we have value and valueInShort in statistics. Maybe rename value to weight?
```
argshook: this todo is no longer relevant imo

src/BarChart/BarChart.js
```diff
@@ -0,0 +1,146 @@
...
+     *  * `value` - value of the statistic. This prop is used for sorting bars. Displayed as big text on a bar, when there is no caption prop.
+     *  * `caption` -  Displayed as big text on a bar.
```
argshook: i would rename `caption` to `label` because in wix-style-react there are
already places with `value` and `label` pairs, so it's mostly for
consistency

src/BarChart/BarChart.js
```diff
@@ -0,0 +1,146 @@
...
+     *  * `captionInShort` - Is shown instead a caption(value) when there is not enough space.
+     *  * `description` - short label under the bar.
+     *  * `descriptionInfo` - long description.
+     */
+    items: PropTypes.arrayOf(
+      PropTypes.shape({
+        value: PropTypes.number.isRequired,
+        caption: PropTypes.string,
```
argshook: all of the `PropTypes.string` should be `PropTypes.node`. `node` is also
a `string` so passing just string will work too.


`node` is useful when users want to use some non visual component, like
for experiment or translation, or some provider etc.

src/BarChart/BarChart.js
```diff
@@ -0,0 +1,146 @@
...
+        captionInShort: PropTypes.string,
```
argshook: let's rename to `labelShort`

src/BarChart/BarChart.js
```diff
@@ -0,0 +1,146 @@
...
+    );
+  };
+
+  _renderItem = (
+    { value, caption, captionInShort, description, descriptionInfo },
+    key,
+  ) => {
+    const { width, calculatedTotal } = this.state;
+    const { total } = this.props;
+    const k = total ? calculatedTotal / total : 1;
```
argshook: please don't use single char var names, what does it mean `k`?

src/BarChart/BarChart.js
```diff
@@ -0,0 +1,146 @@
...
+          <Heading ellipsis data-hook={dataHooks.description} appearance="H5">
+            {showText && description}
+          </Heading>
+        </div>
+      </div>
+    );
+  };
+
+  render() {
+    const { dataHook, items = [], total = 0 } = this.props;
```
argshook: set default values in `defaultProps`, this way they will also be visible in docs

src/BarChart/BarChart.js
```diff
@@ -0,0 +1,146 @@
...
+    const { calculatedTotal } = this.state;
+    const width = total ? (calculatedTotal / total) * 100 : 100;
+
+    return (
+      <div ref={elem => (this.node = elem)} className={styles.wrapper}>
+        <div
+          className={styles.root}
+          data-hook={dataHook}
```
argshook: `data-hook` should be on root element

src/BarChart/test/BarChart.visual.js
```diff
@@ -0,0 +1,105 @@
...
+import BarChart from '../BarChart';
+
+const tests = [
+  {
+    describe: 'sanity',
+    its: [
+      {
+        it: 'many items',
+        props: {
+          items: [
```
argshook: :)
```
[1,2,3,4,5,6,7,8,9].map(n => ({ value: '250', description: `Option ${n}` }))
```

src/BarChart/BarChart.js
maxvektor: I am not 100% sure about it. We have this trick with 2 headers with long and short text. Do you have an example so I can write test for node use case?

src/BarChart/BarChart.js
maxvektor: well, this is tricky. This is a multiplier we need to apply to the child node's width to calculate child. 
Should I call it `coefficient` or `multiplier`?

argshook: can be `coefficient`

src/BarChart/BarChart.js
argshook: not sure we need a test for that. I'd prefer `node` instead of `string` just cause sooner or later someone will get prop warning and complain to us and maybe do a PR. It happened many times already, so whenever i see `string` i always suggest `node`.

shlomitc: @arijus - let’s write this once and for all in our contribution guide

argshook: @shlomitc added https://github.com/wix/wix-style-react/commit/bfaa5f916958e2bcfab92990de0d3f2cd8f0ccaa

src/BarChart/BarChart.js
maxvektor: WDYT about `labelInShort` like in the `StatisticsWidget`?
https://github.com/wix/wix-style-react/blob/master/src/StatisticsWidget/StatisticsWidget.js#L16

maxvektor:
@argshook there is some questions left.

- `labelShort` vs `labelInShort`
- colors of the bars
- UX review?

argshook:
> * `labelShort` vs `labelInShort`

i would call it `labelShort`, cause the `in` for me doesn't give additional info.

> * colors of the bars

you should use colors defined here:

https://github.com/wix/wix-style-react/blob/master/src/Foundation/stylable/colors.st.css#L81-L94

> * UX review?

@milkyfruit pretty please

src/BarChart/BarChart.js
```diff
@@ -0,0 +1,149 @@
...
+
+  static propTypes = {
+    /** Applied as data-hook HTML attribute that can be used to create driver in testing */
+    dataHook: PropTypes.string,
+
+    /**
+     * Array of items
+     *  * `value` - This prop is used for sorting bars. Displayed as big text on a bar, when there is no caption prop.
+     *  * `label` -  Displayed as big text on a bar.
+     *  * `labelShort` - Is shown instead of a caption(value) when there is not enough space.
```
argshook:
```suggestion
     *  * `labelShort` - Is shown instead of a `label` when there is not enough space.
```

maxvektor:
Hey @milkyfruit
> A. Switch their z-indexes.

I am using Heading with an ellipsis, so this is the only possibility. 

Also, I made new colors. This looks odd. Maybe you should discuss it with our UX?

src/BarChart/BarChart.st.css
```diff
@@ -0,0 +1,77 @@
+:import {
+  -st-from: "wix-ui-core/dist/src/themes/backoffice/colors.st.css";
+  -st-named: D30, G05, Y10, R00, B10, D40, G10, Y20, R20, B40;
```
argshook:
```suggestion
  -st-named: B40;
```

argshook:
@milkyfruit current colors used are these (same as in avatar):

2019-09-23-160113_screenshot


please confirm if this is ok

src/BarChart/BarChart.st.css
maxvektor: sure, forgot to delete. Thanks!

milkyfruit:
@maxvektor  Right now if `labelShort` is enabled, but `descriptionInfo` is null it shows the shortened label, but without ellipsis:
image

Maybe, it should be displayed with ellipsis same as when there's only `label` value?
Then we would have a very clear logic:
A. `descriptionInfo` is enabled – text that doesn't fit is hidden
B. `descriptionInfo` is null – text that doesn't fit is shown with ellipsis

@wuwa Israel requests to downgrade color palette to old colors to match other internally made stat charts. In my opinion, we should make some temporary `downgrade` prop, release this component only with old colors or push to update internal components to new colors. What do you think?

argshook:
suggestion:

- `<BarChart/>` component with design system colors, for new users
- `<BarChart deprecatedColors/>` component with old colors, for existing users

this is an opt-in approach for existing users with clear indication that
colors will be changed.

once all statistics related components are migrated to wix-style-react,
existing users, who now require old colors, will be able to just remove
`deprecatedColors` prop and have new ones.

maxvektor:
I made changes after the comments. There is a new visual test for old colors, also it appears in API reference, but I haven't done an example for the main page. I think we should not suggest it to a user as a regular option. WDYT?

Please take a look one more time. @argshook @milkyfruit

src/BarChart/BarChart.js
```diff
@@ -0,0 +1,156 @@
...
+        label: PropTypes.node,
+        labelShort: PropTypes.node,
+        description: PropTypes.node,
+        descriptionInfo: PropTypes.node,
+      }),
+    ),
+
+    /** Used to calculate space for bars inside a widget. Should be specified if the actual total is different from the sum of values of all items */
+    total: PropTypes.number,
+    /** Use old color scheme */
```
argshook: here in the comment you can use `@deprecated`, this way in the api table this prop will be shown separately, with some red text :)

```
/** use old color scheme
 * @deprecated
 */
```

src/BarChart/docs/index.story.js
```diff
@@ -0,0 +1,307 @@
...
+                  {
+                    value: 72000000,
+                    label: '$72,000,000',
+                    labelShort: '$72M',
+                    description: 'Other',
+                  },
+                ]}
+              />`,
+            }),
+          ]),
```
milkyfruit: There should be a new example:
>Hiding values or showing ellipsis
>By default values that don't fit show ellipsis. If "description" prop is added those values are hidden and can be exposed on hover in a tooltip.

src/BarChart/docs/index.story.js
```diff
@@ -0,0 +1,307 @@
...
+                    descriptionInfo: '250 of 1200 users visiting site from mobile',
+                  },
+                ]}
+              />`,
+            }),
+          ]),
+
+          code({
+            compact: true,
+            description: 'Usage with one item.',
```
milkyfruit: Not sure if this example is valuable.

src/BarChart/docs/index.story.js
```diff
@@ -0,0 +1,307 @@
...
+                    descriptionInfo: '570 of 1200 visitors are uniq for current page',
+                  }
+                ]}
+              />`,
+          }),
+
+          code({
+            compact: true,
+            description:
+              'When total is not specified widget will take all the place',
```
milkyfruit: we could merge this example to the first example of 'total'

src/BarChart/docs/index.story.js
maxvektor: We have exactly this case on a project. I would like to have it.

src/BarChart/docs/index.story.js
```diff
@@ -0,0 +1,307 @@
...
+                    value: 450,
+                  },
+                ]}
+              />`,
+            }),
+          ]),
+
+          columns([
+            description({
+              title: 'Description and additional information',
```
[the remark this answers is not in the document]
maxvektor: We have similar props on `StatisticsWidget`. It would be great to have consistency between this two. [SatisticsWidget API](https://wix-wix-style-react.surge.sh/?activeTab=Description&addons=0&full=0&panelRight=0&selectedKind=Components&selectedStory=StatisticsWidget&stories=1)

Also I don't like to have `value`, `label` and `title`. Not sure what is a main think and what is secondary. 

@argshook WDYT?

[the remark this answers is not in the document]
argshook: for the sake of consistency i'd say to keep the names as they are now.

maxvektor:
@milkyfruit 
I hope, I haven't miss something.

PR #4098, opened 2019-09-15
<Page/> - add `height` prop

nitayneeman:
This PR adds a prop that changes the height of the page


shlomitc:
what about the documentation? we recommend and apply a wrapper div. It's better to change them as well

nitayneeman:
> what about the documentation? we recommend and apply a wrapper div. It's better to change them as well

I already did that for `src/Page/docs/index.story.js`:
https://github.com/wix/wix-style-react/pull/4098/files#diff-2b0fa85bbd661e4f85f7ff4822649ac0

What am I missing?

shlomitc:
@nitayneeman all `Page Layout` examples
https://wix-wix-style-react.surge.sh/?selectedKind=2.%20Layout&selectedStory=2.1%20Page%20Layout&full=0&addons=0&stories=1&panelRight=0

PR #4102, opened 2019-09-15
`<ColorPicker/>` - add support for mobile dragging

daphnakaplan:
### 🔦 Summary:
Add event listeners for mobile in ColorPickerHue and ColorPickerHsb to make them functional in mobile
changes made are:
add touchmove touchcancel and touchend events,
find the x and y position of the events accordingly 

### ✅ Checklist
- [ ] 👨‍💻 changes are approved by the UI Infra Developers  @mykas @shlomitc 
- 🔬 Change is tested 
   ColorPicker tested in storybook locally 
  - [ ColorPicker ] Component


mykas:
Screenshot 2019-09-16 at 08 59 25
I don't know how its for you, but it seems that it doesn work on chrome with mobile mode.

daphnakaplan:
Hey Mykolas,
So: there seems to be a problem in chrome with the selection area of the
mouse.
The problem is solved for Iphone (testing in the simulator) and for many
galaxy devices (tested in browserStack), but yes : there is a problem in
chrome in mobile mode in some cases.

the problem is :
because the area of the colorPicker in the storybook is so small, it seems
like the buttons underneath (HEX.. ) are somehow hovered/clicked even
though they shouldn't be :
for example even if the mouse is positioned like that , the HSB is clicked
instead of the slider.
[image: image.png]

Of course if I make a margin between the buttons and slider it solved the
problem. (or if I make the slider much bigger)

I would suggest, since the change didn't affect desktop and no one is using
this component on mobile, to allow the change and then we can test it in
mobile view.
otherwise, maybe we can make a new storybook page that is better for testig
this component, (since its very hard to test the component in the current
storybook that is not fitted for mobile)

what do you think?
Thanks,
Daphna

mykas:
naah I think all good if its just a problem in chrome with mobile mode. If it works on real simulator then thats good

PR #4103, opened 2019-09-15
<Table> -  improve `disabled` behavior

splintor:
### 🔦 Summary
Improve handling of `disabled` in `<Table>`:
1. When there are no items in the list, makes `SelectionContext.disabled` `true` (even when `selectionDisabled` is not set)
2. When the table `disable` or `data` props are updated, re-render bulk selection.
3. Update (and improve) documentation

### ✅ Checklist
- [X] 👨‍🎨 UX change is approved by the Design System UX @ayeletr 
- 📚 Change is documented
  - [X] API description
- 🔬 Change is tested
  - [X] Component


nitayneeman:
@ayeletr Do we want to support this behavior as part of the design system?

src/Table/BulkSelection/BulkSelection.js
```diff
@@ -78,6 +78,18 @@ export class BulkSelection extends React.Component {
         id => !this.state.notSelectedIds.includes(id),
       );
       this.setSelectedIds(selectedIds, undefined, nextProps);
+    } else if (
+      this.props.disabled !== nextProps.disabled ||
+      !this.areSelectedIdsEqual(this.props.allIds, nextProps.allIds)
+    ) {
+      const { selectedIds, notSelectedIds } = this.state;
```
splintor: I wonder what the correct behavior should be when `allIds` is updated. Shouldn't we reset the selected ids when the table data is changed? What do you think?

splintor: No, it doesn't happen in lines `86-`92`, These lines use the existing `selectedIds` and `notSelectedIds`, so bulk selection will not be reset. `allIds` is used in `BulkSelection` to do `toggleAll` and to determine the `All`/`SOME`/`NONE` state. `selectedCount` is only affected by `selectedIds` and `notSelectedIds` which are currently not reset.

Should I add this reset, or should we wait for @ayeletr?

splintor:
@nitayneeman Good point. I think that yes, we should treat the case of "no selectable lines" as an empty table and disable the bulk selection checkbox?

src/Table/BulkSelection/BulkSelection.js
nitayneeman: We should wait for @ayeletr

nitayneeman:
> @nitayneeman Good point. I think that yes, we should treat the case of "no selectable items" as an empty table and disable the bulk selection checkbox?

@ayeletr What do you think?

splintor:
@nitayneeman 
Here some more notes about the implementation:
1. Regarding `unselectable` items - this is the current behavior implemented in this PR, as `allIds` contains only selectable items, and if it is empty, it means we have no **selectable** items. If we want bulk selection to be enabled when there are only unselectable items, we will have to think of the proper way to implement it.

2. Regarding the behavior when data change: 
This is actually a bit more complex than I thought. The data can change because the user changed context (e.g. moved to the trash and we are now showing trashed sites), or because the user filtered the data - and then we need to think if we want to keep the selected sites that matched the filter, or because we are in `infiniteScroll` and the user scrolled down and more data was loaded (and then we obviously need to keep the selection).

So I'm not really sure what the correct implementation should be.
Maybe the code should check and only reset the selection if the old `allIds` is not a prefix of the new `allIds`. What do you think?

nitayneeman:
> @nitayneeman
> Here some more notes about the implementation:
> 
> 1. Regarding `unselectable` items - this is the current behavior implemented in this PR, as `allIds` contains only selectable items, and if it is empty, it means we have no **selectable** items. If we want bulk selection to be enabled when there are only unselectable items, we will have to think of the proper way to implement it.
> 2. Regarding the behavior when data change:
>    This is actually a bit more complex than I thought. The data can change because the user changed context (e.g. moved to the trash and we are now showing trashed sites), or because the user filtered the data - and then we need to think if we want to keep the selected sites that matched the filter, or because we are in `infiniteScroll` and the user scrolled down and more data was loaded (and then we obviously need to keep the selection).
> 
> So I'm not really sure what the correct implementation should be.
> Maybe the code should check and only reset the selection if the old `allIds` is not a prefix of the new `allIds`. What do you think?

Regarding 1 - do you think that having only `unselectable` items is a common use-case or just hypothetical? If it's common, assuming UX agrees, we will probably need to make the `BulkSelection` be aware about these `unselectable` items (I would not change `allIds` to avoid a breaking change, which means we need to collect them or compare the props based `notSelectedIds` somehow).

Regarding 2 - I see we've here two questions - first, should we reset? and if so, what are the triggers? Maybe we should think of how to differentiate between the situations so that `allIds` might be changed because of the consumer, and also the case it would be done internally because of the context. Either way, assuming UX agrees and that's prioritized, I think we should cover that in a separate PR.

splintor:
@nitayneeman Following our slack discussion:
Regarding 1: `allIds` is already without the `unselectable` items, so nothing needs to be done to disable the bulk selection checkbox when there are only `unselectable` items.

Regarding 2: It seems that the best way is to let the consumer decide when selection should be reset, and do it by updating the `key` prop of the `<Table>` component. We only need to decide if this needs to be documented.

PR #4104, opened 2019-09-15
`<Modal/>` - improving Modal's documentation

TalZa1810:
### 🔦 Summary
Improving Modal's documentation

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


src/Modal/docs/index.story.js
```diff
@@ -0,0 +1,91 @@
...
+  componentPath: '..',
+
+  componentProps: {},
+
+  exampleProps: {},
+
+  sections: [
+    header({
+      component: <Button>Button</Button>,
+      sourceUrl:
```
shlomitc: @argshook - is this and `issueUrl` fields required? I think they still appear in the generator

argshook: `issueUrl` is not required, it's taken from `.storybook/webpack.config.js`

`sourceUrl` is also not required, but it's not automated, there will be
no link without it.

it is possible to automate `sourceUrl` now and it closely relates to "link `wuf` and `autodocs`" effort

---

i'll remove `issueUrl` from generator

TalZa1810: Should I remove them both from this file? 
@shlomitc ?

argshook: You can remove `issueUrl`, but not `sourceUrl`

src/Modal/Modal.driver.js
```diff
@@ -8,13 +8,18 @@ const modalDriverFactory = ({ element }) => {
   const getCloseButton = () =>
     document.body.querySelector('[data-hook="modal-close-button"]');
   return {
+    /** true if the modal is on the DOM */
     exists: () => !!getPortal(),
     element: () => element,
+    /** true when to module is open */
```
mosheKerbel:
```suggestion
    /** true when the module is open */
```

src/Modal/Modal.js
```diff
@@ -13,31 +13,53 @@ const CHILDREN_WRAPPER_DIV_ID = 'modal-children-container';
...
     shouldCloseOnOverlayClick: PropTypes.bool,
+    /** Displays a close button on the top right corner of the overlay  */
     shouldDisplayCloseButton: PropTypes.bool,
     onRequestClose: PropTypes.func,
+    onOk: PropTypes.func,
     onAfterOpen: PropTypes.func,
-    horizontalPosition: PropTypes.oneOf(Object.keys(flexPositions)),
-    verticalPosition: PropTypes.oneOf(Object.keys(flexPositions)),
+    /** horizontal position of the modal*/
+    horizontalPosition: PropTypes.oneOf(['start', 'center', 'end']),
```
mosheKerbel: was it intended? becuase i see thatt flexPositions has different values
https://github.com/wix/wix-style-react/blob/89372bb359cba56c43c526d3fa15f7f4f88904bd/src/Modal/ModalConstants.js#L1

src/Modal/Modal.js
```diff
@@ -13,31 +13,53 @@ const CHILDREN_WRAPPER_DIV_ID = 'modal-children-container';
...
+    /** vertical position of the modal*/
+    verticalPosition: PropTypes.oneOf(['start', 'center', 'end']),
```
mosheKerbel: was it intended? becuase i see thatt flexPositions has different values
https://github.com/wix/wix-style-react/blob/89372bb359cba56c43c526d3fa15f7f4f88904bd/src/Modal/ModalConstants.js#L1

src/Modal/Modal.js
```diff
@@ -13,31 +13,53 @@ const CHILDREN_WRAPPER_DIV_ID = 'modal-children-container';
...
+    /** Specifies if modal portal supports scroll*/
     scrollable: PropTypes.bool,
+    /** Specifies if modal content should become scrollable when modal size will fit the window */
     scrollableContent: PropTypes.bool,
+    /** maxHeight of modal(when it has scrollableContent)*/
     maxHeight: PropTypes.string,
     height: PropTypes.string,
-    overlayPosition: PropTypes.oneOf(Object.keys(positions)),
+    /** css position of the modal overlay */
+    overlayPosition: PropTypes.oneOf([
```
mosheKerbel: why not use the modalConstants?

src/Modal/Modal.js
```diff
@@ -13,31 +13,53 @@ const CHILDREN_WRAPPER_DIV_ID = 'modal-children-container';
...
     parentSelector: PropTypes.func,
+    /** selector specifying where to apply the aria-hidden attribute */
+    appElement: PropTypes.string,
   };
 
   static defaultProps = {
     onOk: () => {},
     borderRadius: 0,
-    theme: colors.blue,
+    theme: 'blue',
```
mosheKerbel: modal constants?

src/Modal/docs/index.story.js
```diff
@@ -0,0 +1,118 @@
...
+  divider,
+  columns,
+  title,
+  playground,
+  code as baseCode,
+  description,
+  importExample,
+  testkit,
+} from 'wix-storybook-utils/Sections';
+import LinkTo from '@storybook/addon-links/react';
```
mosheKerbel:
```suggestion
```

src/Modal/Modal.js
TalZa1810: It's `Object.keys` and not `Object.values`. The keys are  `start`, `center` and `end`. The reason I used this instead of  `Object.keys` is because of this issue:
https://github.com/wix/wix-ui/issues/784

src/Modal/Modal.js
TalZa1810: see comment above

src/Modal/Modal.js
TalZa1810: the same issue

src/Modal/Modal.js
TalZa1810: same issue

mosheKerbel: in that case ModalConstants should be deleted.

TalZa1810: It is used in the component itself:
https://github.com/wix/wix-style-react/blob/master/src/Modal/Modal.js#L77

This issue is only related to the autodocs. I see that `colors` and `positions` are used only there so I guess I can delete those two but not `flexPositions`

TalZa1810: I've removed theme. If the tests pass I'll merge, ok?

mosheKerbel: yes, approved.

TalZa1810:
failed on applitools tests which are not related.

Issue #2313, opened 2018-10-11
React 16 Migration


shlomitc:
# React 16

## Strategy
Stay backward compatible with React 15 first, in one of the next major version we will drop support

## Work on React 16 branch
- this branch https://github.com/wix/wix-style-react/tree/migrate-to-react-16
- or this PR https://github.com/wix/wix-style-react/pull/2390/files

## Known issues
| # | scope | description | PR | status
| --- | --- | --- | --- | ---
| 2 | `<BulkSelectionConsumer/>` | component works but tests throws error  | | 
| 3 | `<DatePicker/>` | broken due to wrong `ref`s usage | [possible fix](https://github.com/wix/wix-style-react/pull/2378/commits/2dc33f494370017706f94fb829d7c9314dd3d485) | 

## Done
| # | scope | description | PR | status
| --- | --- | --- | --- | ---
| 1 | `<DataTable/>` | minor issue | https://github.com/wix/wix-style-react/pull/2382 | fixed
| 2 | `<Tooltip/>` | Fix potential null dereferencing | https://github.com/wix/wix-style-react/pull/2391 | fixed
| 2 | `<Tabs/>` | Modify Tooltip test to work with class component  | https://github.com/wix/wix-style-react/pull/2398 | fixed
| 3 | `Backoffice <Tooltip/>` | component works but tests are wrong  | removed in V6 | 
| 4 | `<RichTextArea/>` | broken - requires `slate` upgrade and code fixes | https://github.com/wix/wix-style-react/pull/2239 | will be deprecated, so no action needed

shlomitc:
`BulkSelectionConsumer` - Tests fail due to React printing to console and our test setup fails tests that write to console. React warns that we throw inside render. Possible solutions:
- Delete these tests
- Change the component - Instead of throwing, render a descriptive message for the developer

shlomitc:
`Backoffice Tooltip` https://github.com/wix/wix-style-react/blob/master/src/Backoffice/Tooltip/Tooltip.spec.js#L269
`show with delay and immediately hide` test (and possibly others too) is flaky due to wrong timeout implementation in https://github.com/InterAl/wait-for-cond/blob/master/index.js#L107
Should be fixed in `wait-for-cond`.

shlomitc:
`DatePicker`
[...]

PR #4106, opened 2019-09-15
Migrate to React 16

mosheKerbel:
1. Majority of the changes to support react 16 were done a while ago and merged already. described [here](https://github.com/wix/wix-style-react/issues/2313#issuecomment-513542027)
2. Changed the actual dev dependencies and peer dependencies to be on version ^16
3. Business manager is injecting `react 16.8.3` - [here](https://github.com/wix-private/business-manager/blob/ba4f8a44b52fff3fb7047e54606448e381f8a310/business-manager-web/src/partials/common-third-party.ejs)
4. Added aliases for the unsafe lifecycles, UNSAFE_*
using: `npx react-codemod rename-unsafe-lifecycles``
5. Removed the deprecated `richTextArea`, `richTextAreaComposite` which were replaced with the new `richTextInputArea`
6. Added migration docs
7. Added `px` units to `dropDownLayout` & `box` tests that were missing it and started failing due to React16 policy of ignoring unspecified units for css props.
8. Fixed `DatePicker` focus & click behaviour which started to explode in React16.
@shlomitc


mosheKerbel:
test3 - flaky async test
test4 - approved manually the layout diffs
test2 - approved manually the box diffs

PR #4107, opened 2019-09-15
`<ComposerHeader/>` - new component

Barackos:
### 🔦 Summary
Composer Header
...

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


src/ComposerHeader/ComposerHeader.js
```diff
@@ -0,0 +1,146 @@
...
+            </TextButton>
+          ),
+        )) ||
+        (React.isValidElement(actions) && actions))
+    );
+  }
+
+  render() {
+    const { backButtonValue, children } = this.props;
+    const backButton = this._renderBackButton(backButtonValue);
```
mykas: I suggest getting props inside the `_renderBackButton` and so this part wouldn't be needed anymore.

src/ComposerHeader/ComposerHeader.js
```diff
@@ -0,0 +1,146 @@
...
+  dataHook,
+  alignSelf = 'start',
+  saveStatus,
+  saveStatusValue,
+  children,
+}) => '';
+
+export const ComposerHeaderMainActions = ({ dataHook }) => <div />;
+
+ComposerHeaderActions.tag = 'Actions';
```
mykas: I'd say use `displayName` here instead of `tag`. We usually use `displayName`.

src/ComposerHeader/ComposerHeader.scss
```diff
@@ -0,0 +1,51 @@
+:import {
+  -st-from: "wix-ui-core/dist/src/themes/backoffice/colors.st.css";
```
mykas: import from `../Foundation/stylable/colors.st.css`.

Why I see that this file is `scss` isntead of `st.css`?

Barackos: Fixed (debris of `npm run generate`)

mykas:
> @mykas @Barackos
> 
> ### Descriptions
> Please update example descriptions.
> 
> * [ ]  **Description**
>   ComposerHeader contains main navigation actions – back button on the left and confirm actions on the right. It also has a "save" state preview and supports multiple custom actions. Use to build interfaces to edit photos, create albums, newsletters and similar.
> * [ ]  **Save Status**
>   Component has a predefined style to display a save status – saved, saving, unsaved, failed.
> * [ ]  **Dividers**
>   Component automatically creates dividers between Back Action, Main Actions and Regular Actions if they are used together.
> * [ ]  **Size**
>   ComposerHeader supports two sizes – small and medium. Small is for modal and medium for page layouts.
> 
> ### Other changes
> * [ ]  Please change Undo and Redo buttons to `<ToggleButton/>`
>   image
> * [ ]  A bug appears when a single action item is expected, all 3 are shown instead:
>   image

Regarding second point. We dropped the idea to predefine save statuses. What we see is two different styles: grey and destructive. We let our users set the states whatever, but in examples we will show the correct way.

PR #4114, opened 2019-09-16
re-coloring-avatar

alonbt:


src/Avatar/Avatar.js
```diff
@@ -101,7 +108,7 @@ Avatar.propTypes = {
     'size18',
   ]),
   /** Background color when text (initials) content is displayed */
-  color: PropTypes.oneOf(['blue', 'green', 'grey', 'red', 'orange']),
+  color: PropTypes.oneOf([...deprecatedColorList, ...avatarColorList]),
```
nitayneeman: We have a guide for how deprecating prop values:
https://github.com/wix/wix-style-react/blob/master/docs/internal/DEPRECATION_GUIDE.md#deprecating-props-values

Basically it means:
1. Removing these values from `PropTypes`
2. Adding `deprecationLog` for these values within the render function

src/Avatar/string-to-color.spec.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { PhotoCamera } from 'wix-ui-icons-common';
```
nitayneeman: Redundant?

src/Avatar/test/Avatar.visual.js
```diff
@@ -75,3 +75,19 @@ storiesOf(`Avatar/Props`, module).add('Presence and Indication', () => (
     })}
   </div>
 ));
+
+storiesOf(`Avatar/Props`, module).add('Colors', () => (
+  <div>
+    <div style={{ display: 'flex' }}>
```
nitayneeman: Pretty minor, but maybe `<Box>` that aligned with `space-between` will be better instead of this `&nbsp;`s

src/Avatar/string-to-color.js
```diff
@@ -0,0 +1,15 @@
...
+const stringToColor = name => {
+  if (!name) {
+    return;
+  }
+  const hash = stringToNumber(name);
+  const index = hash % avatarColorList.length;
+  return avatarColorList[index];
+};
+
+const stringToNumber = str =>
```
nitayneeman: I would explain very shortly how this logic works, or just link in a comment to the place it's taken from

nitayneeman:
It seems that this random background color by the text behavior breaks a few visual tests, which makes sense because the default is not blue anymore.

This could probably be a visual breaking change for some consumers (those who use the default color). Are we OK with that? @wuwa

src/Avatar/string-to-color.js
```diff
@@ -0,0 +1,15 @@
+import { avatarColorList } from './Avatar';
```
nitayneeman: This seems like a circular flow - string-to-color.js invokes Avatar.js and vice versa.
I would avoid this case by exporting the colors an external constant file and then use that in both files.

wuwa:
@nitayneeman yes for default color breaking change

src/Avatar/Avatar.js
```diff
@@ -24,8 +28,11 @@ const Avatar = props => {
     dataHook,
     className,
     shape,
+    text,
+    name,
     ...rest
   } = props;
+  const selectedColor = color || stringToColor(text || name);
```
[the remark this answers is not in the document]
alonbt: @nitayneeman I like the idea, the thing is `colorByText` is not exactly what it means. the logic is - if the component has a color then display it, otherwise generate a unique color.
here is the updated code:
```
{
  color: colorProp
} = this.props;

 const color = colorProp || stringToColor(text || name); //if color is provided as a prop use it, otherwise, generate a color based on the text
```

It's not ideal but I think it's more readable, do you have a better name or is it okay as it is ? I also added comments so it will be more descriptive. you can check new code in the updated PR

src/Avatar/Avatar.js
```diff
@@ -101,7 +113,7 @@ Avatar.propTypes = {
     'size18',
   ]),
   /** Background color when text (initials) content is displayed */
-  color: PropTypes.oneOf(['blue', 'green', 'grey', 'red', 'orange']),
+  color: PropTypes.oneOf(avatarColorList),
```
nitayneeman: I think our AutoDocs doesn't support this case, because:
image

Try to:
```suggestion
  color: PropTypes.oneOf([...avatarColorList]),
```

And if it still happens, I would add the values explicitly, like:
```suggestion
  color: PropTypes.oneOf(['A1', 'A2', 'A3', 'A4', 'A5', 'A6',]),
```

src/Avatar/Avatar.js
```diff
@@ -101,7 +113,7 @@ Avatar.propTypes = {
     'size18',
   ]),
   /** Background color when text (initials) content is displayed */
```
nitayneeman: Please improve this comment and explain the way the color is determined (what happens when it isn't provided?)

src/Avatar/Avatar.js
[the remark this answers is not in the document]
nitayneeman: Fine by me

src/Avatar/test/Avatar.visual.js
```diff
@@ -5,14 +5,16 @@ import Avatar from '..';
 import { PhotoCamera } from 'wix-ui-icons-common';
```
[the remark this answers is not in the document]
alonbt: this is not part of my PR... and it appears that the test uses it... so I don't think it is a good idea to remove it..

src/colors.scss
```diff
@@ -143,6 +150,13 @@ $Shadow40: 0 8px 8px 0 $S10, 0 3px 24px 0 $S40;
   Y40: $Y40;
   Y50: $Y50;
   Y60: $Y60;
+
+  A1: #3370FB;
```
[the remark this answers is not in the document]
alonbt: right!

nitayneeman:
Let's wait for the CI before merging

PR #4134, opened 2019-09-19
feat(colors): add "category" values

nitayneeman:
This PR adds the rest of the new category colors:
image


nitayneeman:
It seems that the value of `B4` is not blue whereas `C1` is not green. 

What are the correct values? Should we just switch them? @b3llash

wuwa:
take the colors from actual color squares: https://zpl.io/amW783e

PR #4135, opened 2019-09-19
Move storybook:build to where its really needed

mykas:
This PR is suppose to improve a little bit our parallel testing.


package.json
```diff
@@ -38,28 +38,27 @@
     "test:unit": "yoshi test --jest",
     "test:watch": "yoshi test --jest --watch",
     "pretest:e2e": "./scripts/ensure-dist.sh",
-    "test:e2e": "npm run test:e2e-runtime && concurrently 'npm run test:e2e-only' 'node scripts/e2e-server.js' --names 'protractor,e2e-server' --success first --kill-others",
+    "test:e2e": "npm run storybook:build && npm run test:e2e-runtime && concurrently 'npm run test:e2e-only' 'node scripts/e2e-server.js' --names 'protractor,e2e-server' --success first --kill-others",
```
shlomitc: since we have a split build - this command won't run on the "master" aggregating CI.
so when the "publish" step comes and we want to publish the storybook, storybook won't be built.
if you think it's that important, then you need to run it again in the publish step.

best approach - move our e2e tests away. for those that we must - have a separate storybook (that is different from the documentation one).
then we split documentation from testing

shlomitc: oh, now I see the postpublish, so the first comment is fine

mykas: well sure ideally we should test all of this in different storybook. What I am concerned now is about the `pr-postbuild` which actually publishes pr storybook on every parallel test build. Not sure where to put it. Maybe add another batch just for publishing storybook, but in doesn't make sense in master branch, right?

mykas: Or just add this: `if [ ! \"$agentType\" = \"pullrequest\" ];` in one of the batches.

mykas: Fun fact: in `test3:e2e` and `test4:e2e` we build storybook twice.

mykas: one storybook build is aprox: 2 minutes

mykas: Because every e2e parallel build run: 
```
npm run build (builds storybook)
npm run test3:e2e (build storybook again)
```

PR #4139, opened 2019-09-19
`<Tag/>` - remove redundant e2e when we have .visual already

mykas:


mosheKerbel:
are you sure it's redundant?
take for example this test https://github.com/wix/wix-style-react/blob/bda8d185c8744052f3729f0a064259dc1fd193af/src/Tag/Tag.e2e.js#L79
i don't see it exists in the visual

mykas:
We are not testing tooltip show ups on hover. This should be covered in unit tests.

mykas:
What we could add is just a screenshot that it wen to `...`

mykas:
Actually it can't be covered in unit tests too. So the only way to move this to .visual and se that it goes to `...`.

PR #4144, opened 2019-09-21
Cheatsheet fixes

TalZa1810:
### 🔦 Summary
Small cheatsheet fixes: 
1. changing directory name from `ComponentsFamilies` to `componentsFamilies`.
2. removing redundant imports.
3. changing the API of `FamilyStructure`- the default value of `showPreview` prop was changed to false.
4. Fixing a bug of using `Box` component with `width={54}` to `width="54px"` (which was caused by the React 16 migration) .
5. Changing the name of `ComponentStorybookName` to `SymbolName` and instaed of using `if else` statement using inline condition. 
6. Fixing examples of `Avatar` & `Swatches`.


TalZa1810:
failed on async issues which are not related to the cheatsheet

PR #4149, opened 2019-09-23
[Infra] - Zeit storybook deploy

mosheKerbel:
This is temporary experiments for zeit deployment
Once we decide to go for it, this code will be changed and moved to the appropriate repo.

@shlomitc @argshook 
The motivation for it, discussed already with Shlomi, is to be able to serve faster storybook using http://zeit.co

This is working well for the PR
https://wix-style-react-pr-4149.now.sh

image

[Full log](http://pullrequest-tc.dev.wixpress.com/viewLog.html?buildId=4751893&buildTypeId=FedInfra_WixStyleReac_WixStyleReactNew_Parallel_Tests_Test5Types&tab=buildLog&_focus=2046&state=251%2C1224#_state=251,1224)

How do you think we should progress now?
Merge it and test it a bit for master also and after that go a head with a more infra solution?


scripts/zeit-deploy.js
```diff
@@ -0,0 +1,25 @@
...
+const token = ZEIT_TOKEN;
+
+const pr =
+  (process.env.VCS_BRANCH_NAME &&
+    process.env.VCS_BRANCH_NAME.replace(/\D+/g, '')) ||
+  false;
+const domain = pr ? `wix-style-react-pr-${pr}` : 'wix-style-react';
+console.log(`params: token=${token} pr=${pr} domain=${domain}`);
+
+const deploymentCmd = `now alias --token=${token} $(now ./storybook-static --no-clipboard --token=${token} --public) ${domain}`;
```
argshook: @mosheKerbel perhaps `now` should be `npx now`, then you wont need `npm install` in `package.json`?

mosheKerbel: i'm not adding it package.json, i'm installing it like this, you can take a look at my changes
`npm install now --no-save`
i did it this way because i don't think that `npx` is installed on the ci machines.

argshook: `npx` comes with `npm`, so it should also be in ci. I saw install script is not as dependency but it's still in package.json.

Anyway, not critical at all, just a thought

mosheKerbel: in that case i'll use it

mosheKerbel:
failed due to flaky unrelated tests, merged.

Issue #4150, opened 2019-09-23
`<DropdownBase/>`- testkit functions don't work as expected


TalZa1810:
# 🐛 Bug Report

### 🏗 Relevant Components

`
<DropdownBase/>`
### 😯 Current Behavior

**App file:**
```
import React from 'react';
import DropdownBase from 'wix-style-react/DropdownBase';
import IconButton from 'wix-style-react/IconButton';
import ChevronDown from 'wix-style-react/new-icons/ChevronDown';

const options = [
  {
    id: 0,
    value: 'First option',
  },
  {
    id: 1,
    value: 'Second option',
  },
];

class App extends React.Component {
  render() {
    return (
          <DropdownBase
            showArrow
            dataHook={'dropdownbase-datahook'}
            options={options}
            onSelect={selectedOption =>
              'Triggered with: ' + JSON.stringify(selectedOption)
            }
          >
            {({ open, close }) => {
              return (
                <IconButton
                  skin="inverted"
                  dataHook={"myOpenButton"}
                  onMouseEnter={open}
                  onMouseLeave={close}
                >
                  <ChevronDown />
                </IconButton>
              );
            }}
          </DropdownBase>
    );
  }
}
```

**Test file:**
```
import React from 'react';
import App from './App';
import ReactTestUtils from 'react-dom/test-utils';
import { dropdownBaseTestkitFactory } from 'wix-style-react/dist/testkit';

describe('App', () => {
  it('should open dropdown once it is clicked', async () => {
    const div = document.createElement('div');
    const dataHook = 'dropdownbase-datahook';
    const component = (
      <div>
        <App />
      </div>
    );

    const wrapper = div.appendChild(
      ReactTestUtils.renderIntoDocument(component, { dataHook }),
    );

    const testkit = dropdownBaseTestkitFactory({ wrapper, dataHook });

    await testkit.clickTargetElement();
    expect(await testkit.isDropdownShown()).toBe(true);
  });
});

```

### 🤔 Expected Behavior
The `clickTargetElement` should open the `DropdownBase`.

Moreover, we've tried to use the `optionsCount` function and it's also doesn't work as expected to it seems like it's not specifically an issue just with the `clickTargetElement`  function. 

### 👣 Steps to Reproduce

Described above

  - **WSR version**: `^6.54.2`
  - **Yoshi version**: `...`

### 👀 Severity

- Critical / Major / Low

mosheKerbel:
[...]

PR #4152, opened 2019-09-23
<DropdownBase/> - fix testkit 'clickTargetElement'

mosheKerbel:
fixes [issue](https://github.com/wix/wix-style-react/issues/4150)
@TalZa1810 @zeevgg


mosheKerbel:
breaking change, but we're ok with it because after search in github, no-one is using it actively.

mosheKerbel:
i preferred to use hard-coded data-hooks instead of extracting it, because these tests are working the old way of story examples - i want the consumers to have the data-hooks explicitly.

mosheKerbel:
Failed due flaky test
Merged.

Issue #3441, opened 2019-05-20
<DatePicker/> - Exclude current and past dates


antanaswix:
# ✨ Exclude current and past dates

### 🔦 Scope

Exclude current and past dates for the date picker. 

...

### 💁 Explanation

We are adding a feature that will allow users to change the Valid until date for Purchases of Paid Plans app. We need users to select the date which is after the current date and prevent users from selecting past date and current date.

...

### 🐾 Possible solution 

...

### 👀 Severity

We couldn't release this feature.

- Critical

shlomitc:
it is possible with the `filterDate` predicate. I think we should add it to our docs

shlomitc:
did it work out for you?

PR #4159, opened 2019-09-24
<DatePicker/> - Add documentation to functions

EgoziE:
fix bug - https://github.com/wix/wix-style-react/issues/3441


src/DatePicker/docs/storySettings.js
```diff
@@ -4,3 +4,7 @@ export const storySettings = {
   category: Category.COMPONENTS,
   storyName: 'DatePicker',
 };
+
+export const testStories = {
```
nitayneeman: We can remove that because it's mostly related to the E2Es

PR #4162, opened 2019-09-25
<ImageViewer/> - fixing image zero height

gilmeir:
When providing a width/height of 100%, since the images are `absolute` positioned their height becomes 0


mosheKerbel:
failed due to flaky tests, merging.

PR #4168, opened 2019-09-26
<MessageBoxMarketerialLayout/> - add buttons padding removal prop

mosheKerbel:


mosheKerbel:
failed due to unrelate flaky tests

Issue #3986, opened 2019-08-20
Multiselect: layout breaks with long Tags


b3llash:
# 🐛 Bug Report

### 🏗 Relevant Components
`Multiselect`

### 😯 Current Behavior
Text overflows Tag (https://pasteboard.co/ItwJktP.png)

### 🤔 Expected Behavior
Limit the Tag width and show an ellipsis(?)

### 👀 Severity
Low

nitayneeman:
Talked to @LittleTal.

In the case of `MultiSelect` the tags will have `maxWidth` according to the parent width (the input) with an ellipsis.

FYI @shlomitc

EgoziE:
PR - https://github.com/wix/wix-style-react/pull/4180

PR #4180, opened 2019-10-02
<Tag/> - Add ellipsis to long tags in a container

EgoziE:
fix - https://github.com/wix/wix-style-react/issues/3986

Was verified in Chrome, FireFox, and Safari


src/Tag/Tag.js
```diff
@@ -161,7 +161,7 @@ Tag.defaultProps = {
   size: 'small',
   removable: true,
   theme: 'standard',
-  wrap: false,
+  wrap: true,
```
shlomitc: Isn’t this a breaking change? How will it affect users of it?

src/Tag/tests/Tag.visual.js
```diff
@@ -130,9 +120,13 @@ const tests = [
 tests.forEach(({ describe, its }) => {
   its.forEach(({ it, props, container }) => {
     storiesOf(`Tag/${describe}`, module).add(it, () => (
-      <div {...container}>
-        <Tag {...commonProps} {...props} />
-      </div>
+      <Box direction={'vertical'}>
```
shlomitc: Changing the tests meaning you won’t get indication for the previous one that was done. Maybe it worth a separate PR

EgoziE: Good idea!
I splitted the PR - https://github.com/wix/wix-style-react/pull/4188

src/Tag/Tag.js
```diff
@@ -21,21 +22,28 @@ const tagToTextSize = {
 class Tag extends WixComponent {
   static displayName = 'Tag';
 
+  componentDidMount() {
+    if (this.props.hasOwnProperty('wrap'))
```
nitayneeman: I think that the following is enough:
```suggestion
    if (this.props.wrap)
```

src/Tag/Tag.js
EgoziE: Finally, we decided to deprecate "wrap".
A deprecation message is displayed when used.

Issue #3947, opened 2019-08-14
<Accordion> - missing expand label for icon


mosheKerbel:
Items with **icon** toggle do not show the `expandLabel` when they are closed
for example - the second item should have it
![image](https://user-images.githubusercontent.com/7388217/63024616-3be69700-beb0-11e9-85c7-5fab55edaad2.png)

This is the code to reproduce:
```
<Accordion
  multiple
  items={[
    {
      title: 'First Initially Open Row',
      children: (
        <Text>
          Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
          eiusmod tempor incididunt ut labore et dolore magna aliqua
        </Text>
      ),
      open: true,
      collapseLabel: 'Less',
    },
    {
      title: 'Second Row',
      children: (
        <Text>
          Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
          eiusmod tempor incididunt ut labore et dolore magna aliqua
        </Text>
      ),
      collapseLabel: 'Less',
      expendLabel: 'More',
    },
    {
      title: 'Third Row',
      children: (
        <Text>
          Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
          eiusmod tempor incididunt ut labore et dolore magna aliqua
        </Text>
      ),
      collapseLabel: 'Less',
    },
  ]}
/>;
```

argshook:
closed by https://github.com/wix/wix-style-react/pull/4192

PR #4192, opened 2019-10-03
<Accordion> - missing expand label for icon

skad0:
### 🔦 Summary
Resolves this one: https://github.com/wix/wix-style-react/issues/3947

The problem was in buttonType. Every case was built thinking that buttonType always present, but sometimes it is not. So, while be still have here textButton as fallback, I just set it to be defaultProp and expected behaviour works fine.
Also added spec for this case and small example in story.
...

### ✅ Checklist
- 📚 Change is documented
  - [X] Story
- 🔬 Change is tested
  - [X] Component


argshook:
This is great, thank you for contribution!

PR #4232, opened 2019-10-08
chore(package.json): remove unused deps & devDeps

argshook:
A bunch of deps & devDeps has accumulated over time, this is a cleanup PR.


argshook:
So far looks ok, please revert if any issues arise

PR #4233, opened 2019-10-08
fix(dropdownlayout): keyboard navigation with infinite scroll

audriuss:
If `infiniteScroll` is applied to  `<Dropdown />` or `<DropdownLayout />` keyboard navigation does not work properly and error `Uncaught TypeError: Cannot read property 'getBoundingClientRect' of undefined` is thrown. It is reproducible in `<Dropdown />` or `<DropdownLayout />` storybooks examples with `infiniteScroll`:

Screen Shot 2019-10-08 at 16 54 20

It's happening because: when options container is wrapped into `infiniteScroll` additional `div` appears and options container selector `this.options.childNodes[markedIndex]` becomes incorrect.

Tried to add a test case for this one, but not successfully: when using `infiniteScroll` the component is rendered differently in unit tests, most probably it should be e2e test.


argshook:
failed on unrelated timeout, otherwise good to merge.

Thanks @audriuss!

Issue #4197, opened 2019-10-03
`<PopoverMenu/>` - internal component DropdownLayout has max-height: 260px. Shadow is not extending further.


gerdasi:
# 🐛 Bug Report

### 🏗 Relevant Components

`<PopoverMenu />`

...

### 😯 Current Behavior

When popover has more elements than height of 260px its shadow disappears. 
![Screenshot 2019-10-03 at 14 23 56](https://user-images.githubusercontent.com/54804606/66123982-ef842180-e5eb-11e9-8235-256c8a788691.png)

...

### 🤔 Expected Behavior

All popover should be covered in shadow.

...

### 👣 Steps to Reproduce

  - **WSR version**: `6.58.0`
  - **Yoshi version**: `4.16.3`

### 👀 Severity

- Critical

mykas:
Fixed.

PR #4235, opened 2019-10-09
`<PopoverMenu/>` - add maxheight support and set default value

mykas:
This PR is adding maxHeight value to DropdownBase and then its consumer PopoverMenu in order to control default value of DropdowLayout maxHeight. The default value has been set to be `auto` just to disable the maximum height constraints as we really don't have such constraints.


mykas:
https://github.com/wix/wix-style-react/issues/4197

PR #4239, opened 2019-10-09
fix(DropdownLayout): fixed drivers to work with infinite scroll

audriuss:
Related to: https://github.com/wix/wix-style-react/pull/4233

When `<DropdownLayout />` or `<Dropdown />` used with `infiniteScroll` testkits' methods to interact with options were not working correctly. `infiniteScroll` adds additional `div` and testkits' selectors did not respect that


argshook:
thanks! good to merge

PR #4269, opened 2019-10-10
<BadgeSelect/> - Add unidriver

EgoziE:
Created a uni driver for `<BadgeSelect/>` component


src/BadgeSelect/BadgeSelect.uni.driver.js
```diff
@@ -0,0 +1,31 @@
...
+  const badgeDriver = badgeUniDriverFactory(
+    base.$(`[data-hook="${DATA_ATTR.DATA_BADGE}"]`),
+  );
+  const dropdownLayoutDriver = dropdownLayoutDriverFactory(
+    base.$(`[data-hook="${DATA_ATTR.DATA_DROPDOWN}"]`),
+  );
+
+  const driver = {
+    ...baseUniDriverFactory(base),
+    clickOutside: popoverDriver.clickOutside,
```
mosheKerbel: you should use private drivers for some of these methods like in the normal driver

mosheKerbel:
we are good to merge, tests passed already.

PR #4298, opened 2019-10-15
Force buble version to fix issues with react-live code examples

simast:
There seems to be an issue with `buble` > 0.19.6 where JSX syntax is not transpiled and an error is thrown. Latest `react-live` package already has buble fixed at 0.19.6 - probably because of the same issues. We can remove this when we update to latest `react-live` package.


shlomitc:
shouldn't this fix be done in `wix-storybook-utils`?

mykas:
Yes, but for now we ship this to fix wix-style-react builds. Arijus is working on updating react-live which will have internally buble frozen.

simast:
@shlomitc  Implemented/moved this fix to `wix-storybook-utils` package.

PR #4307, opened 2019-10-15
<TimeTable/> - new component

AndriiTsarenko:


package.json
```diff
@@ -250,8 +250,7 @@
   "lint-staged": {
     "linters": {
       "*.{js,scss}": [
-        "yoshi lint --fix",
-        "git add"
+        "yoshi lint --fix"
```
mykas: Is this somehow related?

src/DragAndDrop/Draggable/components/Container.js
```diff
@@ -32,6 +32,12 @@ const target = {
       index: monitor.getItem().index,
     };
   },
+  canDrop(props, monitor) {
```
mykas: This feature news to be added on a separate PR. You are creating TimeTable component on this PR and it shouldn't modify other code.

src/SortableList/SortableList.js
```diff
@@ -8,19 +8,34 @@ import Container from '../DragAndDrop/Draggable/components/Container';
 import times from '../utils/operators/times';
 import withDNDContext from './withDNDContext';
 
+export const INSERTION_RULES = {
```
mykas: Again we have a feature added on the same PR as TimeTable. Should be a separate pr

package.json
AndriiTsarenko: I wonder of this because seem `git add` is not valid operation maybe `git add .`. But I it's not related to my PR and I committed this accidentally.

AndriiTsarenko: I've reverter.

src/TimeTable/docs/examples/TimeTableStaticItems.js
```diff
@@ -0,0 +1,81 @@
...
+  };
+
+  ensureDataForDay = (daysData, dayId) => {
+    if (!daysData[dayId]) {
+      daysData[dayId] = [];
+    }
+  };
+
+  handleDrop = (dropInfo) => {
+    const {payload, addedIndex, removedIndex, addedToContainerId, removedFromContainerId} = dropInfo;
```
mykas: So this information is too much for our users. If we think about this interface it is not clear for me what it is about. For example what is `payload` in the context of TimeTable.Card ?

src/TimeTable/TimeTable.uni.driver.js
```diff
@@ -0,0 +1,48 @@
+import { baseUniDriverFactory } from 'wix-ui-test-utils/base-driver';
+import { DataAttributes, DataHooks } from './constants';
+
+export const timeTableDriverFactory = base => {
+  const getDataHookSelector = dataHook => `[data-hook="${dataHook}"]`;
+  return {
+    ...baseUniDriverFactory(base),
+    /** returns displayed days count */
+    async daysCount() {
```
mykas: If you look at our other uni driver implementation the convention is not this. Its `daysCount: () => etc`

src/TimeTable/TimeTable.js
```diff
@@ -0,0 +1,263 @@
...
+            return (
+              <div
+                key={`${column.title}${column.subtitle}`}
+                data-hook={DataHooks.Header}
+                {...{
+                  [DataAttributes.HeaderActive]: active.toString(),
+                  [DataAttributes.HeaderDisabled]: disabled.toString(),
+                }}
+                className={classNames(styles.header, {
+                  [styles.dayActive]: active,
```
mykas: This should be used not with `classNames` library but rather implemented with stylable states. Contact me if you don't know how to do it. Or check this example: https://github.com/wix/wix-style-react/blob/master/src/Text/Text.js#L52

src/TimeTable/TimeTable.js
```diff
@@ -0,0 +1,263 @@
...
+    );
+  };
+
+  renderInsertButton = containerId => {
+    return (
+      <div
+        onClick={() => this.props.onAddButtonClick(containerId)}
+        className={styles.addItemButton}
+      >
+        <div>
```
mykas: Why this additional `<div>` ?

src/TimeTable/TimeTable.js
```diff
@@ -0,0 +1,263 @@
...
+    groupName: 'time-table',
+  };
+
+  state = {
+    showAddButtonDayIndex: null,
+    dragging: false,
+  };
+
+  static getMonday = date => {
+    date = new Date(date);
```
mykas: I noticed that you are mutating values all the time. In our code we are developing with immutability in practise in order to prevent overriding variables. Please change this to be immutable

src/TimeTable/TimeTable.js
```diff
@@ -0,0 +1,263 @@
...
+  };
+
+  getRenderData = () => {
+    const { renderOptions, groupName } = this.props;
+
+    if (!renderOptions) {
+      return [];
+    }
+
+    const daysInfo = [];
```
mykas: again you are mutating values with this variable. This whole block of code must be done with immutability in mind

src/TimeTable/TimeTable.js
```diff
@@ -0,0 +1,263 @@
...
+
+  render() {
+    const { dataHook, groupName } = this.props;
+    const renderData = this.getRenderData();
+    return (
+      <div className={styles.root} data-hook={dataHook}>
+        <div className={styles.headerContainer}>
+          {renderData.map(column => {
+            const {
+              active = false,
```
mykas: default values needs to stay in defaultProps - not in the code.

src/TimeTable/TimeTable.js
```diff
@@ -0,0 +1,292 @@
...
+    insertPosition: 'any',
+  };
+
+  state = {
+    addButtonColumnIndex: null,
+    dragging: false,
+    sortableListItems: getSortableListItems(this.props.columns),
+  };
+
+  UNSAFE_componentWillReceiveProps(nextProps) {
```
mykas: Isn't this gonna be deprecated?

simast: I thought they just deprecated old methods without `UNSAFE`, but I see they do call the `UNSAFE` methods as legacy now in official documentation as well. I will replace this with `getDerivedStateFromProps`.

PR #4312, opened 2019-10-15
<ToggleSwitch /> - Add types for component

Akemyshka:
Add prop types to ToggleSwitch component


mosheKerbel:
You raised the issue of the onChange(e) like the example inside storybook. This valid example would really break after using this d.ts

The solution as i see it should be inside wix-ui-core, 
add an optional event param to the onchange prop and add a unit test to the spec file - once passing onchange without a param and once with a param.

Issue #2862, opened 2019-01-31
<FormField/> - char counter changes the height when label is inline


nitayneeman:
Given a `FormField` with an inline label that gets an input child (which invokes `setCharactersLeft` on value change).

# The Issue
First of all, the character counter is shown only after typing a value. On top of that, when the counter appears, it changes the FormField's height surprisingly. This behavior doesn't happen when the label is placed on the top.

Here's an example:
![formfield](https://user-images.githubusercontent.com/7141972/52056221-512a5280-256a-11e9-9dba-618e56ff5da5.gif)

argshook:
> First of all, the character counter is shown only after typing a value.

this depends on when consumer calls `setCharactersLeft`, it does not necesarily have to be only during `onChange`.

initially i suggested to simply have `charactersLeft` prop on `FormField` consumer would simply pass a number and wouldn't need to bother calling function. buuuut lost the debate

shlomitc:
let's add a `charactersCounter` prop as another option to set it.
we can decide if to deprecate the render prop, or just leave it as it is.

argshook:
What about `maxLength`, same as in html?

shlomitc:
Whatever you think is better

argshook:
closed by https://github.com/wix/wix-style-react/pull/4350

PR #4350, opened 2019-10-21
Add charCount prop to FormField component

simast:
### 🔦 Summary
Add new `charCount` prop to FormField component. This is a replacement/alternative for existing render function method with `setCharactersLeft` interface (should we deprecate this method?). I think both methods complement each other as the "render prop" interface does not require any state.

Related issue: #2862 

### ✅ Checklist
- 📚 Change is documented
  - [x] Story
  - [x] API description
- 🔬 Change is tested
  - [x] Component
  - [x] Visual test


argshook:
> should we deprecate this [`setCharactersCount`] method

i think it can stay as it would be breaking change for no real benefit.
It doesn't weight much.

we can perhaps stop advertising it in docs

PR #4357, opened 2019-10-21
`<Swatches/>` - add clickEmptySwatch method to testkit

bildja:
### 🔦 Summary
Add `getEmptySwatch` to swatches testkit

### ✅ Checklist
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


mykas:
Hi Dima, we usally don't return unidriver based element. What do you try to achieve with such method?

bildja:
@mykas there is `getSwatch` method which returns the same driver element.
What I need is to click "clear" button from tests. There is a test inside WSR which does this https://github.com/wix/wix-style-react/blob/master/src/Swatches/test/Swatches.spec.js#L68

mykas:
I would rather add then: `clickEmptySwatch` . Method `getSwatch` was wrongly exported.

bildja:
@mykas changed that

src/Swatches/Swatches.uni.driver.js
```diff
@@ -14,6 +15,12 @@ export const swatchesDriverFactory = base => {
     /** Get swatch at given index */
     getSwatch: async index => fillPreviewDriverFactory(await getSwatch(index)),
 
+    /** Click empty swatch */
+    clickEmptySwatch: async () =>
+      fillPreviewDriverFactory(
```
mykas: Forgot to pass `body`. It will fail on you. Get the body from the second argument from swatchesDriverFactory

bildja: @mykas I didn't forget it, I just removed it, because `fillPreviewDriverFactory` doesn't even take this param.

mykas: Oh, you are right. Ok

Issue #4267, opened 2019-10-10
Virtualized <Table /> memoized row rendering


audriuss:
# ✨ Feature Request

### 🔦 Scope
`<Table />` with Virtualization

### 💁 Explanation
When one item of all input array is changed, all rows are remounted together with containing components, that have negative impact on performance and UX.
For example we have `<Avatar />` rendered as row cell, which indicates task assignee. If user change the assignee all rows are remounted together with other avatars which causes all avatars to flicker while loading images.

### 🐾 Possible solution
Since virtual `<Table />` is using  `react-window` under the hood, row memoization could be implemented: https://react-window.now.sh/#/examples/list/memoized-list-items

### 👀 Severity
- Major

Our project will use virtualized Tables quite much, so it's important to us have a good performance

audriuss:
PR: https://github.com/wix/wix-style-react/pull/4358

simast:
Fixed with #4462 and #4455

PR #4358, opened 2019-10-22
fix(virtualized DataTable): when data updated scroll keeps position + rows memoization

audriuss:
### 🔦 Summary
It is relevant only when table is virtualized. If user scrolls down the table and updates some data on the line (for example change the label) the scroll position is set to top, and sometimes blank table appears - rows are not rendered. It's reproducible on story book as well: scroll down virtualized table example and click "Show code" below the table.

Added prop `virtualizedUseRowsMemo` to virtualized `<Table />`. If it's `true`, used `React.memo` for virtualized table row rendering. Since `react-window` is used for the virtualized, my solution based on https://react-window.now.sh/#/examples/list/memoized-list-items). Fixes: https://github.com/wix/wix-style-react/issues/4267


simast:
Thanks for the contribution! As discussed, lets improve this PR with the following changes:

- We could split `renderVirtualizedTableElementWithRefForward` changes into a separate PR as this seems to be a significant catch/fix already.
- Let's try to add `.memo` support without extra table props (so the optimisation is enabled by default). For this we need to shallow compare with the table props as well (e.g. if a new column is added on the table the `.memo` should detect this - currently this is not handled).
- Let's use existing lib `shallowequal` for comparing shallow prop equality.

PR #4370, opened 2019-10-22
Add focusable prop to radio button

naamagr:
### 🔦 Summary
This enables setting a radio button as unfocusable. It is required for team-managment page where the radio buttons are for display only and not functional (selection is made by another component).

### ✅ Checklist
- [X ] 👨‍💻 API change is approved by the UI Infra Developers @shlomitc 
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [ ] Visual test

* There is no storybook page for single radio buttons (only radio groups) so I wasn't sure where this should be documented.


src/RadioGroup/RadioButton/RadioButton.js
```diff
@@ -23,6 +23,7 @@ class RadioButton extends WixComponent {
     style: PropTypes.object,
     type: PropTypes.oneOf(['default', 'button']),
     lineHeight: PropTypes.string,
+    focusable: PropTypes.bool,
```
shlomitc: why not just pass `tabIndex`? this makes a more flexible and consistent API across the system

src/RadioGroup/RadioButton/RadioButton.driver.js
```diff
@@ -10,6 +10,7 @@ const radioButtonDriverFactory = ({ element }) => {
     isChecked: () => radioButton.checked,
     isDisabled: () => radioButton.disabled,
     getLabel: () => label.textContent,
+    getButtonTabIndex: () => element.getAttribute('tabIndex'),
```
shlomitc: calling it `getTabIndex` should be enough

src/RadioGroup/RadioButton/RadioButton.spec.js
```diff
@@ -59,6 +59,24 @@ describe('RadioButton', () => {
     expect(onChange).not.toBeCalledWith(value);
   });
 
+  it('should be focusable by default', () => {
```
shlomitc: as you will change prop to be just `tabIndex`, I think it's better for the tests to not mentions "focusable"

argshook:
> radio buttons are for display only and not functional 

this essentially says that radio buttons are disabled. For that only `disabled` prop should be used without any additional things.

if when `disabled={true}` RadioButton is still focusable, then that should be fixed. No new prop needed imo

shlomitc:
The case Naama is trying to solve is a RadioButton inside a Thumbnail.
It is a valid case for the Thumbnail to be focusable and tab-able, while the RadioButton isn't and will be set as "checked" in a controlled way.

shlomitc:
@naamagr - please add a screenshot showing your use-case

naamagr:
image
@argshook 
@shlomitc I've pushed a fix. please re-review :)

PR #4371, opened 2019-10-22
explain npm package.json tasks

shlomitc:
Trying to make sense in the jungle. Didn't change anything, just classified as a groundwork for simplifying.

1. not sure where `test:performance` is used and why it uses `proxy-wix-ui-icons-common`
2. `sanity` can be renamed to `test:sanity`
3. `test:types` includes `npm run pr-storybook:publish`. it should be included in `test6:sanity` instead
4. not sure where `bundles:analyse` is being used`
5. `lint` is run in different workflows and inconsistently


mykas:
`test:performance` is being run by perfer service in ci. Basically the checks you see in github are basically because test:performance has been run and the naming of the scripts comes from them. It uses proxy icons because perfer is not running `npm run build` in its ci. For the bundles to work we need icons to be generated.

`bundles:analyse` is used to generate and run webpack bundles analyser where we can see and work on our components bundles sizes.

PR #4379, opened 2019-10-23
add prop alwaysDisplayChevron

banuni:
- Added a prop to <SidebarSectionItem />
### 🔦 Summary
<SidebarSectionItem /> now accepts alwaysDisplayChevron that'll make it show the little arrow with a low opacity, in the case of a drillable item.

### ✅ Checklist
- [ v] 👨‍🎨 UX change is approved by the Design System UX => Yaya & Ben
- 📚 Change is not documented
  - [v ] Story
  - [ v] API description
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ v] Component
  - [ ] Visual test


EgoziE:
The code looks good.
I couldn't see a visible difference in the Playground.
Would you please add a `use example` and a `visual test`?

src/SidebarSectionItem/SidebarSectionItem.js
```diff
@@ -28,6 +28,8 @@ class SidebarSectionItem extends React.PureComponent {
     disabled: PropTypes.bool,
     /** Indicates whether to display an icon for drilling in on hover */
     drillable: PropTypes.bool,
+    /** Indicates whether to display a low-opacity icon for drilling even without hover */
```
EgoziE: Let's make sure other people will understand when to use this prop, my suggestion:

Indicates whether to display a low-opacity icon for drilling even without hover, relevant only when `drillable` is true.

PR #4389, opened 2019-10-24
Perfer - add production paralel test for component size

mykas:
This PR is adding production build for component size. It will fail on bundle sizes increase which we can happen if:
- Someone ignored their perfer pr check fail and merged the pr.
- One of our dependencies got unexpected or expected bundle increase.

### DIFF
--- a/.ci_config
+++ b/.ci_config
@@ -6,7 +6,7 @@
     "test4:e2e",
     "test5:types",
     "test6:sanity",
-    "pr-storybook:publish"
+    "test7:size",
   ],
   "depmonkey_ignore": true,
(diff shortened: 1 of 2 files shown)


.ci_config
```diff
@@ -6,7 +6,7 @@
     "test4:e2e",
     "test5:types",
     "test6:sanity",
-    "pr-storybook:publish"
+    "test7:size",
```
shlomitc: Giving it a more semantic name like “performance” would have been better

mykas: Well ok, maybe performance is good. Just thought that maybe if we have `test:type` so it would also be clear to write `test:size`

PR #4396, opened 2019-10-25
<MediaOverlay /> - new component

simast:
### 🔦 Summary

Add new component `<MediaOverlay />` - used to overlay any content on top of a media element. The content can be displayed in default state, on hover or in both states. Default and hover states can have different skins.

Remaining tasks:

- [x] UniDriver
- [x] Unit tests
- [x] Visual tests
- [x] TypeScript typings


src/MediaOverlay/MediaOverlay.js
```diff
@@ -0,0 +1,186 @@
...
+          {this._renderContentArea(layer, Placement.BottomEnd)}
+        </div>
+      </>
+    );
+  };
+
+  _renderContentArea = (layer, placement) => {
+    const { children } = this.props;
+    const contentElements = filterContent(children, {
+      display: layerToDisplayMap[layer],
```
argshook: i was looking for a while at the `display` and its `always`, `hover` & `no-hover`
properties and i can't say im sold.

partly it's because of naming, specifically `display="no-hover"`. for me
this tells that `no-hover` thingy is displayed, but it doesn't tell that
it relates to only `onMouseOut` of MediaOverlay (not MediaOverlay.Content
where the `display` property is set)

It also can be confusing with `always`. user must know, that `no-hover` is not going to be displayed during `mouseOver`.

moreover, if there are many `MediaOverlay.Content` children, it can be
hard to track which is displayed when, especially if these components are in random order.
Consider a case when `MediaOverlay.Content`s are created dynamically with, say
```
actions.map(action =>
  <MediaOverlay.Content placement="top-end" display={action.important ? 'always': 'no-hover'}/>
)
```

at this point what `display` does is a little wtf to me. Cases like that can easily be reality in places like media gallery

if you think my concerns are valid, i then propose to invert control, and use a function:

```js
<MediaOverlay>
  {({ hovered }) => {
    { hovered &&
      <MediaOverlay.Content>
        this is hovered content
      </MediaOverlay.Content>
    }

    { !hovered &&
      <MediaOverlay.Content>
        this is NOT hovered content
      </MediaOverlay.Content>
    }

    <MediaOverlay.Content>
      this is always visible
    </MediaOverlay.Content>
  }}
</MediaOverlay>
```

---

sorry for wall of text, i have two keyboards

src/MediaOverlay/MediaOverlay.uni.driver.js
```diff
@@ -0,0 +1,18 @@
+import { baseUniDriverFactory } from 'wix-ui-test-utils/base-driver';
+
+export const mediaOverlayDriverFactory = base => {
+  return {
+    ...baseUniDriverFactory(base),
+
+    /** Get the current count */
+    getCountText: async () => base.$('[data-hook="mediaOverlay-count"]').text(),
```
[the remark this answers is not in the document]
argshook: fixed while i typed this comment :)

src/MediaOverlay/MediaOverlay.js
```diff
@@ -0,0 +1,186 @@
...
+        {...this._getFocusProps()}
+        {...styles('root', { clickable: !!onClick }, this.props)}
+        style={
+          isMediaImageUrl ? { backgroundImage: `url(${media})` } : undefined
+        }
+      >
+        {!isMediaImageUrl && React.isValidElement(media) && media}
+        {hasSingleSkin && this._renderSingleSkinBackgroundOverlay(skin)}
+        {this._renderOverlay(Layer.Default, !hasSingleSkin ? skin : Skin.None)}
+        {this._renderOverlay(
```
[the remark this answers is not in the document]
simast: Should be fixed, overlay layers are not rendered if they are empty/not required right now. There is one exception where hover layer will be rendered as an empty div (to support `hoverSkin` transitions).

src/MediaOverlay/MediaOverlay.js
```diff
@@ -0,0 +1,186 @@
...
+  // workaround we create this background overlay that has no content or animations
+  // and render here only the common skin background.
+  _renderSingleSkinBackgroundOverlay = skin => {
+    return <div {...styles('overlay', { layer: Layer.SingleSkin, skin })} />;
+  };
+
+  _renderContent = layer => {
+    return (
+      <>
+        <div {...styles('contentRow', { row: 'top' })}>
```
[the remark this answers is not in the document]
simast: I have added some checks to skip some empty `<div>`s in overlay layer content - however there is still a case where top/bottom row divs are kept - this is to support simple flex layout using `justify-content: space-between` and `flex-basis` for helping with centering middle content.

src/MediaOverlay/MediaOverlay.js
simast: We have discussed this internally and decided to rename `display` prop to `visible` and the `no-hover` value to `default`.

src/MediaOverlay/MediaOverlay.js
```diff
@@ -0,0 +1,249 @@
...
+        data-hook={dataHook}
+        onMouseEnter={this._onMouseEnter}
+        onMouseLeave={this._onMouseLeave}
+        onClick={onClick}
+        {...this._getFocusProps()}
+        {...styles('root', { clickable: !!onClick }, this.props)}
+        data-skin={skin}
+        data-hoverskin={this._getHoverSkin()}
+        style={
+          isMediaImageUrl ? { backgroundImage: `url(${media})` } : undefined
```
mykas:
```suggestion
         { backgroundImage:  isMediaImageUrl && `url(${media})` } 
```

src/MediaOverlay/MediaOverlay.uni.driver.js
```diff
@@ -0,0 +1,37 @@
...
+    /** Get custom media element node. */
+    getMediaNode: async () => {
+      const mediaUrl = await getMediaUrl();
+      if (mediaUrl !== null) {
+        return null;
+      }
+
+      const mediaNode = await base.$(':first-child');
+      // eslint-disable-next-line no-restricted-properties
+      return (await mediaNode.exists()) ? mediaNode.getNative() : null;
```
mykas: Lets not check if element exists or not. Leave it to unidriver to handle this

src/MediaOverlay/MediaOverlay.js
```diff
@@ -0,0 +1,247 @@
...
+
+    /**
+     * Click handler (when this property is provided the component will be clickable
+     * on hover and will have a pointer cursor).
+     */
+    onClick: PropTypes.func,
+
+    /**
+     * `<MediaOverlay.Content>` elements describing overlay content.
+     * Each element has the folllowing properties:
```
argshook:
```suggestion
     * Each element has the following properties:
```
:)

PR #4422, opened 2019-10-30
<DataTable> - use row reference for rowDetails instead of row number

giladsegal:
### 🔦 Summary

`rowDetails` is displayed according to row numbers saved in the component's state.
When row are appended to the beginning of the data, the state is not updated accordingly.
As a result, if a row is appended when another row's `rowDetails` is expanded, the `rowDetails` is shifted to a different row (due to invalid row numbers state)

### Checklist

- [ ] Change is tested - This fix is related to state changing over time - the only way to test it is to implement a consuming component that updates the state. I'm not sure if it's worth it, WDYT?
- [ ] Animation flickering - When appending rows, the `rowDetails` is positioned correctly but for some reason it re-triggers the expand animations

### DIFF
--- a/src/DataTable/test/DataTable.spec.js
+++ b/src/DataTable/test/DataTable.spec.js
@@ -1,11 +1,12 @@
 import React from 'react';
+import eventually from 'wix-eventually';
 import { mount } from 'enzyme';
 import {
   createRendererWithDriver,
   createRendererWithUniDriver,
   cleanup,
 } from '../../../test/utils/react';
-
+import { enzymeUniTestkitFactoryCreator } from 'wix-ui-test-utils/enzyme';
 import DataTable from '../DataTable';
 import dataTableDriverFactory from '../DataTable.driver';
 import { dataTableUniDriverFactory } from '../DataTable.uni.driver';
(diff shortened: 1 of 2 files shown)


giladsegal:
If I test don't you think it's better to implement a wrapping component than using enzyme overkill methods?

ezgif com-video-to-gif

src/DataTable/test/DataTable.spec.js
```diff
@@ -385,11 +387,45 @@ describe('Table', () => {
         // When we clicking second time to collapse content will disappear after a while (based on animation speed)
         expect(await driver.getRowDetailsText(0)).not.toBe('');
 
-        return new Promise(resolve => {
-          setTimeout(async () => {
```
giladsegal: The `setTimeout` was a side effect which was executed in my test
I had to switch it to `eventually`

src/DataTable/test/DataTable.spec.js
```diff
@@ -385,11 +387,45 @@ describe('Table', () => {
...
+        expect(await dataTableTestkit.getRowDetailsText(0)).toBe('');
+        await dataTableTestkit.clickRow(0);
+        expect(await dataTableTestkit.getRowDetailsText(0)).toBe('foo');
+
+        wrapper.setProps({
+          data: [{ a: 'value 5', b: 'value 6' }, ...props.data],
+        });
+
+        await eventually(async () => {
+          wrapper.update();
```
shlomitc: why is `update` inside the evntually?

src/DataTable/test/DataTable.spec.js
```diff
@@ -1,10 +1,12 @@
 import React from 'react';
+import eventually from 'wix-eventually';
 import { mount } from 'enzyme';
 import {
   createRendererWithDriver,
   createRendererWithUniDriver,
   cleanup,
 } from '../../../test/utils/react';
+import { dataTableTestkitFactory as dataTableEnzymeTestkitFactory } from '../../../testkit/enzyme';
```
shlomitc: please don't import from testkit. this file is auto-generated in build and not before the tests. use the driver directly

PR #4423, opened 2019-10-30
TableActionCell: Enable setting dataHook for each action

splintor:
### 🔦 Summary
We need our tests to be able to click a specific action in the list

### ✅ Checklist

- 📚 Change is documented
  - [X] API description
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


nitayneeman:
We typically avoid from exporting dataHooks for internal parts of the component. What is the scenario you try to test?

splintor:
Our QA (@dmitriweissmann) wants to write an automation test that clicks the Table Action Cell popover menu and then click one of the options.

We somehow found a way to click the internal popover menu (using the internal data-hook `table-action-cell-popover-menu` - although I would prefer being able to set it.

But we want to have a way to click an action in the popover menu, not based on index (which can change), or on text (which can be translated) - that's exactly what `data-hooks` are for...

nitayneeman:
> Our QA (@dmitriweissmann) wants to write an automation test that clicks the Table Action Cell popover menu and then click one of the options.
> 
> We somehow found a way to click the internal popover menu (using the internal data-hook `table-action-cell-popover-menu` - although I would prefer being able to set it).
> 
> But we want to have a way to click an action in the popover menu, not based on index (which can change), or on text (which can be translated) - that's exactly what `data-hooks` are for...

I understand your opinion with the dataHook - the thing is our guidelines define that we usually export the dataHook for the root element and covers the internal parts through the testkit (getters, actions and so on). 

The way we currently provide to click on the PopoverMenu child is by index. I think that TableActionCell is supposed to be aligned with that way... Perhaps we can better the way, but that's what we provide at the moment. 

@mykas Maybe you have an opinion?

src/TableActionCell/TableActionCell.js
```diff
@@ -58,10 +63,10 @@ function renderHiddenActions(actions, popoverMenuProps) {
       appendToParent
       {...popoverMenuProps}
     >
-      {actions.map(({ text, icon, onClick, disabled }, index) => (
+      {actions.map(({ text, icon, onClick, disabled, dataHook }, index) => (
         <PopoverMenuItem
           key={index}
-          dataHook="table-action-cell-popover-menu-item"
+          dataHook={dataHook || 'table-action-cell-popover-menu-item'}
```
splintor: I now realize that this doesn't really do anything, because `<PopoverMenuItem>` doesn't have a `dataHook` prop, and can't change the `data-hook` [it uses when rendering](https://github.com/wix/wix-style-react/blob/master/src/PopoverMenuItem/PopoverMenuItem.js#L57).

splintor: @nitayneeman

splintor:
@shlomitc @lidanh

mykas:
Hey @splintor , it would be much better to just upgrade the PopoverMenu to the newest one, because we don't do any bug fixed for the previous one.

splintor:
@mykas Thanks!
@nitayneeman Following @mykas comment, I now see that the beta `<PopoverMenu.MenuItem>` [allows overriding the items' `dataHook`](https://github.com/wix/wix-style-react/blob/master/src/beta/PopoverMenu/PopoverMenu.js#L240) (see https://github.com/wix/wix-style-react/pull/3822), so maybe it will also be OK to allow it for `<TableActionCell>` items? 

If you say it is OK, I'll update the PR to upgrade the `<PopoverMenu>` used by `<TableActionCell>` to the beta one.

src/TableActionCell/TableActionCell.js
```diff
@@ -29,15 +29,20 @@ function renderPrimaryAction({ text, theme, onClick, disabled }) {
...
-  return actions.map(({ text, icon, onClick }, index) => (
+  return actions.map(({ text, icon, onClick, dataHook }, index) => (
     <Tooltip
       key={index}
-      dataHook="table-action-cell-visible-action-tooltip"
+      dataHook={
+        dataHook
+          ? `${dataHook}-tooltip`
+          : 'table-action-cell-visible-action-tooltip'
+      }
```
nitayneeman: Assuming you have the `dataHook` prop for the PopoverMenuItem - isn't it enough?

src/TableActionCell/TableActionCell.js
```diff
@@ -29,15 +29,20 @@ function renderPrimaryAction({ text, theme, onClick, disabled }) {
...
       content={text}
       theme="dark"
     >
       <Button
         theme="icon-greybackground"
+        dataHook={dataHook || 'table-action-cell-action'}
```
nitayneeman: Same question here

src/TableActionCell/TableActionCell.js
nitayneeman: As you suggested, let's use the new `<PopoverMenuItem>`

splintor:
@nitayneeman If you give me OK to go with my suggestion, I'll update the driver to use custom `dataHook` when specified.
Specifically, I need this for our QA automation tests, which are written in Java, so I won't currently have use of the drivers.

src/TableActionCell/TableActionCell.js
splintor: No, because if the action is visible, it is not listed in the `<PopoverMenu>`. It is added as a separate `<Button>` within a `<Tooltip>`.

src/TableActionCell/TableActionCell.js
splintor: No, because if the action is visible, it is not listed in the `<PopoverMenu>`. It is added as a separate `<Button>` within a `<Tooltip>`.

src/TableActionCell/TableActionCell.js
splintor: Sure, just waiting for your green light.

nitayneeman:
> @nitayneeman If you give me OK to go with my suggestion, I'll update the driver to use custom `dataHook` when specified.
> Specifically, I need this for our QA automation tests, which are written in Java, so I won't currently have use of the drivers.

Let's go with your suggestion.

By the way, I'm not sure there's a case right now - I just wonder how we should treat that when the time comes.

mykas:
We had some discussion before in the team and I think that we should automation-friendly library so I tend to follow this kind of idea that our all compound component should be able get datahooks.

splintor:
@mykas @nitayneeman I pushed another commit that moves `<TableActionCell>` to use the beta `<PopupMenu>` and also added tests that use the new option of setting `dataHook` for `<PopupMenu.MenuItem>`s.

Please review again.

nitayneeman:
We also need to fix this broken e2e:
```Module not found: Error: Can't resolve '../../new-icons/More' in '/home/builduser/agent00/work/dbb751eb2b5c5b2e/dist/es/src/TableActionCell'```

**Comment by nitayneeman** [comment]
I see we've some broken visual tests - @amitrch1  could you take a look so that we can promote this PR?
Also, there's a tiny conflict.

**Comment by splintor** [comment]
@nitayneeman I merged the conflict, but I have no experience with `eyes` tests, so we might need your help with this.

I guess the visual tests break because we upgraded to the beta version of `<PopoverMenu>`, which might look differently.

Is this considered a breaking change, as it might break visual tests of other products that use `<TableActionCell>`?

File: src/TableActionCell/TableActionCell.js [hunk 4]
```diff
@@ -55,14 +56,22 @@ function renderHiddenActions(actions, popoverMenuProps) {
     <PopoverMenu
```
nitayneeman: The `<PopoverMenu/>` was always opened to the top before. Personally I prefer it automatically, but I would stick with the previous behaviour:
```suggestion
     <PopoverMenu
      placement="top"
```

nitayneeman:
> @nitayneeman I merged the conflict, but I have no experience with `eyes` tests, so we might need your help with this.
> 
> I guess the visual tests break because we upgraded to the beta version of `<PopoverMenu>`, which might look differently.
> 
> Is this considered a breaking change, as it might break visual tests of other products that use `<TableActionCell>`?

Yes, it might break them. In our case it's just some adjustments we need to apply (right button skin, margins, `PopoverMenu` position) in order to maintain the same behavior as before.

splintor:
@nitayneeman Thanks for your review. I pushed fixes to your comments - please review them.

Regarding the failing `eyes` test - I have no credentials to open https://wixeyes.applitools.com/, so I can't even see where the problem is - how can we proceed from now? Do you want to give me credentials? Or send me a screenshot and details of the failing scenario so I can check what went wrong?

splintor:
Using [this slack message] I was able to login to `eyes`. It seems that [the previous problem](https://wixeyes.applitools.com/app/test-results/00000251828707885066/00000251828707709478/steps/3?accountId=mE99eP8CUn9UTSXSv98L3OSaJ0NZmAcY6Wndx6WSHODdY110&mode=step-editor) of auto-opening `<PopupMenu>` no longer happens.

It seems that currently, the failing `eyes` tests fail because of style differences between the old `<PopoverMenu>` and the new beta one. How do we want to handle these failures?

src/TableActionCell/TableActionCell.js
```diff
@@ -55,14 +56,23 @@ function renderHiddenActions(actions, popoverMenuProps) {
...
       buttonTheme="icon-greybackground"
       dataHook="table-action-cell-popover-menu"
-      appendToParent
+      appendTo="parent"
+      placement="top"
```
nitayneeman: Let's add the following to be aligned with the previous design:
```suggestion
      placement="top"
      textSize="small"
```

nitayneeman:
> Using [this slack message](https://wix.slack.com/archives/C056JMHJ5/p1476785983000104) I was able to login to `eyes`. It seems that [the previous problem](https://wixeyes.applitools.com/app/test-results/00000251828707885066/00000251828707709478/steps/3?accountId=mE99eP8CUn9UTSXSv98L3OSaJ0NZmAcY6Wndx6WSHODdY110&mode=step-editor) of auto-opening `<PopupMenu>` no longer happens.
> 
> It seems that currently, the failing `eyes` tests fail because of style differences between the old `<PopoverMenu>` and the new beta one. How do we want to handle these failures?

After we adjust the `fontSize` I guess we will approve the rest of the failures (if any).

splintor:
@nitayneeman text size changed. 6 tests [are still failing](https://wixeyes.applitools.com/app/test-results/00000251828316789977).

nitayneeman:
It turns out we broke some of our consumers because of the Popover upgrade.

The thing is, we expose a testkit method called `getHiddenActionsPopoverMenuDriver` that returns the Popover driver as is (I don't really like that, but that's the situation). Anyway, now we return a driver with different API...

These are the affected consumers:
https://github.com/search?q=org%3Awix-private+getHiddenActionsPopoverMenuDriver&type=Code

So I see these options:
1. Revert the feature
2. Polyfill somehow the method above and keep the same API
3. Change the feature so it won't use the new Popover
4. Fix all issues of our consumers 

@splintor  What do you think?

splintor:
We moved to the beta `<PopoverMenu>` mainly because the old one didn't allow setting the data-hooks of its items (although it also helps us in fixing [WOS3-1061](https://jira.wixpress.com/browse/WOS3-1061))

1. Reverting is always an option, but we really want to be able to set dataHooks for various actions.
2. Implemented in https://github.com/wix/wix-style-react/pull/4568 - please review.
3. As mentioned above, the old `<PopoverMenu>` doesn't allow setting `data-hooks` for its items, so I don't see how this can be an option (other than updating the old `<PopoverMenu>`.
4. It seems like too much work, and also, some of the consumers' tests use `itemContentAt`, which is not currently implemented in the beta `<PopoverMenu>` testkit, so a change in it is needed anyway to make the consumers' tests work. (This is handled in https://github.com/wix/wix-style-react/pull/4568).

All-in-all, I guess #2 is a bit ugly but is probably the way to go - let's work on https://github.com/wix/wix-style-react/pull/4568 and make it merged, so you can release a fix.

PR #4442, opened 2019-11-03
<ClickOutside/> - Click outside render prop

EgoziE:
Created a new render prop `<ClickOutside/>`
Added unit test
Migrating `<DatePicker/>` component to use it for the example


mykas:
Hey Egozi, sorry to disturb your here, but it seems you are trying to implement a onClickOutside feature which is already implemented in Popover. DatePicker component is a component that still uses `popper.js` library directly. It would be much better to just migrate this to use Popover

EgoziE:
@mykas Thank you for noticing!
This is true, the sole reason for this new implementation is that both the _previous_ one inside `WixComponent`, and the one in the external library `react-onclickoutside` (used in `popover.jsx`) are using a deprecated React API `React.findDOMNode`.

This PR is one of many to come in order to fully migrate from React 15 to 16+

mykas:
Hey, but how are you planning to replace it ?

EgoziE:
After a short conversation with @mykas, we decided it's best to implement `<ClickOutside/>` in a way that will makes it possible to migrate every component that need this behaviour.

These are the steps:
1. Instead of "render" as a prop, the component will render its children
Like so:
```
<OnClickOutside rootRef={this.rootRef} onClickOutside={this.onClickoutside}>
   <MyComponent ref={ ref => (this.rootRef = ref)} >
</OnClickOutside>
```

2. Support click exclusion by class name, adding a prop `excludeClass`.
Any excluded elements will not trigger the `onClickOutside` callback.

3. Move this PR to `wix-ui-core` in order to use in components like `<Popover/>`

I'd love to hear your opinions @shlomitc, @nitayneeman, @mosheKerbel

EgoziE:
Moved to wix-ui-core

PR #4456, opened 2019-11-04
modifications for <Selector /> component 

brumeregan:
changes needed for ModalSelectorLayout
**subtitleNode** - for custom subtitleNode, when it is not simple text
**belowNode** and **showBelowNodeOnSelect** were added to achieve this design
Screen Shot 2019-11-04 at 11 13 33 AM
I changed layout styles for component, because flex styles not allowed to insert below component. So not to break existed styles, I created additional wrapper


milkyfruit:
@brumeregan @mykas I discussed with Vadym we should try 2 changes:
1. Change Hover color to B60
2. Don't show hover style when hovered on `extraNode` area.

milkyfruit:
@brumeregan discussed with Vadym, agreed to udpate mouse cursor style depending where it points.
Original list item area must use pointer, while `ExtraNode` if contains static content must show default mouse cursor.
Video example attached: [selectmodal.mov.zip](https://github.com/wix/wix-style-react/files/3835338/selectmodal.mov.zip)

brumeregan:
@milkyfruit updated PR )

PR #4458, opened 2019-11-04
`<Swatches/> -  add isEmptySwatchSelected to testkit

vladik7244:


src/Swatches/Swatches.uni.driver.js
```diff
@@ -5,7 +5,12 @@ export const swatchesDriverFactory = base => {
   const getSwatch = async index =>
     base.$$('[data-hook="color-swatches-swatch"]').get(index);
 
+  const getEmptySwatch = async () =>
+    base.$(`[data-hook="${dataHooks.empty}"]`);
+
   return {
+    getEmptySwatch,
```
mykas: This should not be here. We do not expose unidriver based elements for our users as in this case they would need know how unideriver api is defined.

mykas: It would be better to just add `isEmptySwatchSelected` method

PR #4496, opened 2019-11-10
`<MobilePreviewWidget/>`- new component

TalZa1810:
### 🔦 Summary
New component- MobilePreviewWidget. 
Related Zeplin file: https://zpl.io/2jMDQLr

### ✅ Checklist

- 📚 Change is documented
  - [x] Story
  - [x] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [x] Visual test


src/MobilePreviewWidget/docs/index.story.js
```diff
@@ -0,0 +1,116 @@
...
+export default {
+  category: storySettings.category,
+  storyName: 'MobilePreviewWidget',
+
+  component: MobilePreviewWidget,
+  componentPath: '..',
+
+  componentProps: {
+    dataHook: '',
+    children: childNode,
```
nitayneeman: The output of the playground isn't valid:
```
<MobilePreviewWidget>
  <Box
    align="center"
    backgroundColor="Y30"
    height="100%"
    padding="20px"
    verticalAlign="middle"
  >
    <[object Object]>
      Content goes here
    </[object Object]>
  </Box>
</MobilePreviewWidget>
```

src/MobilePreviewWidget/MobilePreviewWidget.js
```diff
@@ -0,0 +1,70 @@
...
+    backgroundColor: string,
+
+    /** Sets the height of the component.*/
+    height: string,
+
+    /** Sets the width of the component */
+    width: string,
+
+    /** Node to preview */
+    children: node.isRequired,
```
nitayneeman: I prefer to duplicate in this case, but just for your information - we could reuse and take these propTypes out of `PreviewWidget.propTypes`

src/MobilePreviewWidget/docs/index.story.js
```diff
@@ -0,0 +1,116 @@
...
+
+          code({
+            compact: true,
+            source: `<Layout cols={2}>
+                    <MobilePreviewWidget skin='gradient'>${childNodeString}</MobilePreviewWidget>                  
+                    <MobilePreviewWidget skin='custom' backgroundColor='B40'>${childNodeString}</MobilePreviewWidget>     
+                   </Layout>`,
+          }),
+        ],
+      }),
```
nitayneeman: I think I would only demonstrate the custom `skin` with a backgroundColor. We've the playground to see how each `skin` value looks, but the custom is actually different because it depends on `backgroundColor` - so it's worth an example IMO

PR #4501, opened 2019-11-11
EllipsisHOC - add `showTooltip` prop

nitayneeman:
This PR adds a new prop that allows components such as `<Text/>` to render ellipsis without a tooltip.

There are consumers that have containers with tooltips that contain ellipsed text inside, and the problem they're faced with is a tooltip that overrides the second tooltip.

@wuwa FYI


mykas:
But does this make sense that they have double ellipsis on text component? What would consumer need to do to get the full text view then?

wuwa:
@mykas it is for some edge cases where consumer adds custom tooltip (ie more content than the <Text>) that has nothing to do with space limitations. in those cases this tooltip covers also the ellipsised state.  
image

mykas:
so basically they want `...` without our tooltip

mykas:
I would then rather say that our ellipsis hoc should accept custom `content` to be passed instead of wrapping two tooltips on top. Maybe it is possible. @nitayneeman can you check?

wuwa:
that would not solve the case of custom tooltip in "no ellipsis" state

mykas:
We have prop `ellipsis` which can be enabled or disabled. Thats the case for ellipsis state

src/common/EllipsisHOC/index.js
```diff
@@ -87,6 +87,8 @@ Comp.propTypes = {
   hideDelay: number,
   /** `ellipsis` prop. Tooltip show delay. */
   showDelay: number,
+  /** `ellipsis` prop. Renders the ellipsis when necessary without a tooltip */
+  hideTooltip: bool,
```
mykas: Maybe lets call it `disableTooltip`?

nitayneeman: I thought about that as well, but decided to be aligned with `showTooltip` param of the `withEllipsedTooltip`.

Actually @wuwa suggested renaming semantically to `showTooltip` (with `true` as a default value) instead of `hideTooltip` (so that we just pass the param as is). WDYT?

mykas: yep, lets do it like that then. Use `showTooltip`

PR #4509, opened 2019-11-12
<ImageViewer/> - remove @testing-library from the private driver

nitayneeman:
This PR removes `@testing-library` import from the private driver, basically as done here:
https://github.com/wix/wix-style-react/pull/4171

Because we've consumers who need this fix, it's a quick one.

Also, the right way to handle IMO is:
1. Migrate the private driver to Unidriver
2. Refactor the public drivers to use the privates (like the generator generates)
3. Use the private drivers within the spec so that we invoke `render` only once for each test

@adimorwix FYI


nitayneeman:
The problem the consumers are faced with, is when they use any Protractor testkit while `@testing-library` isn't installed (it's a devDependency for us), they get the following:
`Cannot find module '@testing-library/react'`

That's because we import the private driver here:
https://github.com/wix/wix-style-react/blob/fix/imageviewer-private-driver/src/ImageViewer/ImageViewer.protractor.driver.js#L1

And obviously is exported here:
https://github.com/wix/wix-style-react/blob/fix/imageviewer-private-driver/testkit/protractor.js#L95

PR #4517, opened 2019-11-13
<Table/> - Fixed missing separator border for Table.Content when column titles are hidden

### LINKED ISSUE #4309: Table separator issue
# 🐛 Bug Report

### 🏗 Relevant Components
Table

### 😯 Current Behavior
If we render a table content without visible title bar (`<Table.Content titleBarVisible={false} />`) it has no separator between table header and table content. However, for the empty state this separator suddenly appears. See the video:
https://drive.google.com/open?id=1Y7b4JyQeQcjl_M7D6chKLWFQMJGXJPZP

### 🤔 Expected Behavior
Separator should always be there.

### 👣 Steps to Reproduce
Create a Table with empty state and `<Table.Content titleBarVisible={false} />`

### 👀 Severity
 Low (workaround applied)

simast:
### 🔦 Summary

Closes #4309

### ✅ Checklist

- 🔬 Change is tested
  - [x] Visual test


src/Table/components/TableContent.js
```diff
@@ -19,28 +20,33 @@ export const TableContent = ({ titleBarVisible, dataHook }) => {
...
-                  columns={createColumns({ tableProps, bulkSelectionContext })}
-                />
-              )}
-            </BulkSelectionConsumer>
-          );
-        } else {
-          return <DataTable {...dataTableProps} />;
-        }
+        return (
+          <div {...styles('content', { titleBarVisible })}>
```
argshook:
```suggestion
          <div {...styles('content', { titleBarVisible }, this.props)}>
```

this helps to avoid the notorious `red block of error` and also allows users to extend styles in `.st.css` files.

if i remember correctly, there's a promise that it's no longer required in stylable v2, but we're not using that yet

src/Table/Table.st.css
```diff
@@ -25,6 +25,14 @@
   background-color: value(B50);
 }
 
+.content {
+  -st-states: titleBarVisible;
+}
+
+.content:not(:first-child):not(:titleBarVisible) {
```
argshook: why `:not(:first-child)`?

argshook:
since this PR addresses #4309, i think it would be good to add a visual test covering that case with `<Table.EmptyState/>`

src/Table/Table.st.css
simast: we don't want to add this separator border when we just have `<Table.Content />` - that would be an undesired visual change from previous behavior. This selector will only apply if table content is preceded by other content - such as  `<Table.Toolbar />` for example.

src/Table/components/TableContent.js
simast: added props - slightly different since this is a functional component

PR #4524, opened 2019-11-13
Change root element to be inline-block

EgoziE:
Change root element to be from `inline` to `inline-block`
The reason for this change is that an `inline-block` allows to set a width and height on the element.
According to Zeplin, the height of the text should be fixed height.
After a short conversation with @wuwa we decided to change it.

Here's an example for the bug:
image
Size: medium
left side as `inline` - **height: 18.86px** 👎🏼
right side as `inline-block`- **height: 24px** 👍🏼

### DIFF
--- a/src/Text/Text.st.css
+++ b/src/Text/Text.st.css
@@ -20,6 +20,7 @@
     light,
     weight(enum(thin, normal, bold)), disabled;
 
+  display: inline-block;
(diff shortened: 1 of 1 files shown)


EgoziE:
Closing for now, we might just misuse `<Text/>` in some places..

PR #4536, opened 2019-11-15
`<SocialPreview/> - resolve visual flakiness

mykas:
This pr will use storybook stepper to resolve image onload flakiness on visual tests.


src/SocialPreview/tests/SocialPreview.visual.js
```diff
@@ -18,40 +18,42 @@ const defaultProps = {
...
-      return ComponentWrapper ? (
-        <ComponentWrapper>{component}</ComponentWrapper>
-      ) : (
-        component
-      );
+class AsyncStoryWrapper extends React.Component {
+  componentDidMount() {
+    setTimeout(() => {
+      this.props.onDone();
+    }, 3000);
```
argshook: @mykas whole 3 second? two tests, so it's 6 seconds. whoa

mykas: Eh this is copy paster from snapper example so yes its bit too much 🤔

mykas: But this is still a bit hackish as we assume that it will take certain amount of time to onload the image :/ It would be best to have some sort of hook to listen when image has been rendered.

PR #4559, opened 2019-11-18
Table: Enable setting origin for selection change event

splintor:
### 🔦 Summary
Our table header has a move action that removes items from the table, and when it ends, we deselect all items from the list using `SelectionContext.deslectAll()`, but this triggers the `selectionChanged` event in which we report BI event, but we only want to report this BI event when the user changed the selection. We want to be able to identify what caused the selection change and act accordingly.

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [X] API description
- 🔬 Change is tested
  - [ ] Component


splintor:
@nitayneeman Fixed. Please review again.

PR #4563, opened 2019-11-19
<SidebarSectionItem /> - change partial chevron opacity

banuni:


src/SidebarSectionItem/SidebarSectionItem.st.css
```diff
@@ -24,7 +24,7 @@
 :vars {
   paddingWithoutSuffix: 6px 24px 6px 24px;
   paddingWithSuffix: 6px 18px 6px 24px;
-  chevronPartialOpacity: 0.3;
+  chevronPartialOpacity: 0.6;
```
nitayneeman: @ayeletr says it should without opacity, like the chevron when hovering.
So I guess we don't need `chevronPartialOpacity` at all.

banuni: It's for an AB test we're conducting.it's under a "feature toggle (prop)" so it's supposed to be ok

banuni: @nitayneeman, I spoke with @ayeletr , she said it's ok

PR #4568, opened 2019-11-20
Provide backward compatability to old PopupMenu testkit

splintor:
https://github.com/wix/wix-style-react/pull/4423#issuecomment-555554972


nitayneeman:
I consulted with @shlomitc and we have a suggestion.

The problem right now is that we still have a breaking change, although we keep the same driver methods (it returns a `Promise`). In practice it means we cannot use a Unidriver (the new PopoverMenu driver) inside a regular driver (the TableActionCell driver).

So we've thought of solving the issue using `upgrade` prop for the TableActionCell component. It's a practice we use sometimes, which constitutes an intermediate stage containing the breaking change, so that our consumers can use it by their decision (and it becomes the default behavior later, in the next major release). This prop will enable the "setting dataHook for each action" feature and use the new PopoverMenu as well.

When `upgrade` is enabled, we can apply your changes which use the new PopoverMenu along with the "setting dataHook for each action" feature. Existing driver methods like `clickPopoverMenu` will have to be duplicated or get another parameter (`upgrade` maybe?), indicating to use the new PopoverMenu driver instead. When `upgrade` isn't passed, we'll use the old PopoverMenu as it was before (obviously the driver methods you added aren't going to work so we need to comment that). This guarantees us we don't break the consumers for sure, as long as they don't need the feature. Using the feature means to do appropriate adjustments (use driver methods that might have a completely different API).

An example of usage:
```jsx
  <TableActionCell
  	upgrade
    primaryAction={{
      text: 'Edit'
    }}
    secondaryActions={[
      {
        text: 'Star',
        icon: <Icons.Star />,
        onClick: () => {},
        dataHook: 'my-datahook'
      }
    ]}
  />
```

Another approach is to stay with the current solution but reimplementing the broken driver methods synchronously using dataHooks, as you suggested. This means we make a workaround for methods that are going to be legacy soon.

The `upgrade` prop sounds to me as the best solution we can offer considering our constraints - and I'm not sure it will take much effort than reimplementing the old methods.

@splintor What do you think?

splintor:
Sounds reasonable. I'll try to see how complex it is to implement the feature under the `upgrade` flag. Will update.

PR #4579, opened 2019-11-21
`EmptyState` - Add textAlign prop to EmptyState component

charmingelle:
#4391 

✅ 👨‍🎨 UX change is approved by the Design System UX (@milkyfruit)

- 📚 Change is documented
  - ✅ Story
  - ✅ API description
  
- 🔬 Change is tested
  - ✅ Component


charmingelle:
Done, please check)

argshook:
hello!

i'm taking this over with a few improvements:

- `align` prop should support not `left` & `right` but `start` & `end`
much like css does. The reason is that in RTL `left` would be displayed
on right, and naming is confusing. Therefore the standard name is `start/end`
- add visual tests
- a little less css

once build is green, i will merge

PR #4580, opened 2019-11-21
`<Proportion>`  - add new type to disabled aspectRation holder

daphnakaplan:
Add disable prop to Proportion

Add disable prop - if disabled child returns in original size (size must be set explicitly)
if not disabled - returns in desired aspect ratio

added in order to improve usability for future components such as Color Palette  

- ✅ 👨‍💻 API change is approved by the UI Infra Developers 
- 📚 Change is documented
✅Story
✅ API description
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


shlomitc:
Do we really need a prop for that?
if you don't want a proportion - why not just not rendering it?
It also doesn't really make sense as an API. `disabled` is making something not interactive, not "not working"

mykas:
@shlomitc so it all depends on the writing style. Proportion is sort of wrapping Component and user would need to do a hoc that either will wrap it or not. It is much easier to pass prop `disabled` and not care about it.

shlomitc:
so why not change the aspect ratio to `auto` or `none` or something more semantic instead of introducing another prop?

mykas:
Well we are open for suggestions. I didn't see any problem with `disabled`.

mykas:
I think that `none` is pretty good :) @daphnakaplan can you please change that?

shlomitc:
@daphnakaplan i will also be happy to see how you use it in the code (what is the trigger and what provides the size instead)

mykas:
@shlomitc this is part of ColorPallete initiative. It was meant to reuse the FillPreview component that already uses Proportion inside, but in ColorPallete the proportion is not needed the the width should flow with available space

daphnakaplan:
@Mykolas Šindeikis <mykolass@wix.com> @Shlomi Toussia Cohen
<shlomitc@wix.com>
so 'none' it is -
the downside is that FillPreview would then have to receive aspectRatio and
also pass it to Proportion instead of FillPreview remaining oblivious to
this scenario.
The scenario is : Color palette has a list of FillPreview - it overrides
FillPreview size with explicit size, FillPreview has Proportion that is set
to None aspect ratio therefore returns original Children without size
manipulation.
Other solution was to pass a custom aspect ratio to Proportion and to use
it as it.
I do like to mention that my main goal here is to develop Color Palette and
we did not set much time for it.

mykas:
@daphnakaplan but in this case it would be the same. You would still need to expose `aspectRatio` to FilLPreview component. Then in ColorPallete you just do:

```
<Layout>
  <FillPreview aspectRatio="none"
   etc.
</Layout>
```

daphnakaplan:
@mykas 
done changes : 'none' added to Ratio options and if 'none' aspectRatio than just return child without the wrapping

src/Proportion/Proportion.js
```diff
@@ -13,32 +13,38 @@ class Proportion extends React.PureComponent {
...
     const wrapperClass = classnames(styles.root, className);
     const aspectRatioHolder = this._getAspectRatioHolder();
-    const content = this._getContent();
+    const disable = aspectRatio === PREDEFINED_RATIOS.none ? true : false;
+    const content = this._getContent(disable);
 
     return (
       <div className={wrapperClass} data-hook={dataHook}>
-        {aspectRatioHolder}
+        {!disable && aspectRatioHolder}
```
mykas: We usually call this parameter `disabled` . Please change it so it would semantically same as in other components

src/Proportion/Proportion.uni.driver.js
```diff
@@ -10,5 +12,6 @@ export const proportionDriverFactory = base => {
       const height = await base.attr('offsetHeight');
       return width / height;
     },
+    isSvgExists: async () => byHook('proportion-aspect').exists(),
```
mykas: what is SVG? our user does not know what is that. I'd say it should be something: `isAspectRatioDisabled`

daphnakaplan:
KK, anything else?

mykas:
Nope, all good. Lets wait for the build to get green

PR #4588, opened 2019-11-24
Generator- adding types to generator

TalZa1810:
### 🔦 Summary
Adding types to generator


.wuf/generator/templates/Component/index.d.ts
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react';
+
+export interface {%ComponentName%}Props {
+  dataHook?: string;
+  buttonText?: string;
+}
+
+declare const {%ComponentName%}: React.ComponentClass<{%ComponentName%}Props>;
```
giladsegal: should be 
```ts
export default class {%ComponentName%} extends React.PureComponent<{%ComponentName%}Props>;
```

PR #4592, opened 2019-11-25
add node to title types

brumeregan:
currently title for <Selector /> component needs to be only string, but it also can render node
so to prevent type errors in console, it is better to add this type also :)


src/Selector/Selector.js
```diff
@@ -21,7 +21,7 @@ class Selector extends WixComponent {
       'cinema',
     ]),
     imageShape: PropTypes.oneOf(['rectangular', 'circle']),
-    title: PropTypes.string.isRequired,
+    title: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired,
```
argshook:
```suggestion
    title: PropTypes.node.isRequired,
```

PropType.node is also PropType.string, so there's no need of `oneOfType`

brumeregan: hmm.. but I have error when run my tests..

argshook: what kind of error? Prop incompatibility should be a warning

argshook: here's the implementation of `PropTypes.node` checker: https://github.com/facebook/prop-types/blob/master/factoryWithTypeCheckers.js#L474-L479

as you see `string` returns `true`. If you get a warning i would assume it's from somewhere else

brumeregan: I see this:

> console.error node_modules/prop-types/checkPropTypes.js:20
      Warning: Failed prop type: Invalid prop `title` of type `object` supplied to `Selector`, expected `string`.

argshook: okay, so this happens before this PR.

with `PropTypes.node.isRequired` you will not have such warning

PR #4607, opened 2019-11-27
<Card /> & <Card.Content /> controls and size props added

mantaspa:
### 🔦 Summary

- `controls` property added to `<Card />`. It allows passing in various control elements e.g. `<CloseButton />`.
- `size` property added to `<Card.Content />`. It adjusts `<Card.Content />` padding.

### ✅ Checklist

- 📚 Change is documented
  - [x] Story
  - [x] API description
- 🔬 Change is tested
  - [x] Visual test


src/Card/Content/Content.scss
```diff
@@ -1,5 +1,14 @@
 .root {
   padding: 24px 30px 30px;
+
+  &.small,
```
argshook: i think `small` shouldn't be here

src/Card/Card.scss
```diff
@@ -1,10 +1,21 @@
...
 .card {
+  position: relative;
   background: $D80;
   border-radius: $cardBorderRadius;
 }
 
+.controls {
+  position: absolute;
+  top: 0;
+  left: 0;
```
argshook:
```suggestion
```
let's not set `left`, cause if we do then controls div will take all available width. For now we're ok only with top right

src/Card/Card.scss
```diff
@@ -1,10 +1,21 @@
...
+  right: 0;
+  display: flex;
```
argshook:
```suggestion
```

src/Card/Card.scss
```diff
@@ -1,10 +1,21 @@
...
+  justify-content: flex-end;
```
argshook:
```suggestion
```
without `left: 0` this is also not needed

Issue #4484, opened 2019-11-07
<Table/> - row select checkbox have small click area


dorduch:
the clickable area around the row checkbox is too small, and it's easy to mistakenly click the row instead of selecting it

mykas:
@milkyfruit I guess we need to expand the area, right?

milkyfruit:
yes, I think it could be expanded to whole area up until the first value item (in given example it's avatar):
![image](https://user-images.githubusercontent.com/1684587/68395084-bd7c5700-0177-11ea-9210-a0d96dc1996f.png)

@dorduch how urgent is this?

dorduch:
we got some complaints about it

dorduch:
@milkyfruit  do you mean expanding it to include the padding and margin? or just the available space? (in this case it's above and below the checkbox)

milkyfruit:
@dorduch I menat this area:
![image](https://user-images.githubusercontent.com/1684587/69236949-44c8c200-0b9d-11ea-9724-2ecf7d88c697.png)

shlomitc:
should be fixed by now, closing

PR #4616, opened 2019-11-28
<Table> Enlarge table select checkbox click area

liorur:
This PR resolves issue #4484 
It done via CSS so no tests are applicable
The solutiuon is a bit hacky because there is no direct way to apply styles to <td> and there is no way to know in <Table> weather or not titleBarVisible is true or false


liorur:
@argshook pls review

liorur:
i would be glad to enlarge the title bar click area as well but i have no access to the css. in any case the problem is less serious there since the user only clicks one place and not many, and there is no issue with click being caught by onClick of row

nitayneeman:
> i would be glad to enlarge the title bar click area as well but i have no access to the css. in any case the problem is less serious there since the user only clicks one place and not many, and there is no issue with click being caught by onClick of row

Basically it's possible if we apply your solution on this part:
https://github.com/wix/wix-style-react/blob/enlarge_checkbox_click_area/src/DataTable/DataTable.js#L435

I agree it's pretty minor compared to the rows, so let's take it out of the scope

Issue #4611, opened 2019-11-28
ColorInput no colour chosen bug


gerdasi:
# 🐛 Bug Report

### 🏗 Relevant Components

<ColorInput />

### 😯 Current Behavior

I choose a colour in `<ColorInput />` then the colour changes. After I delete the colour code and press enter there is no colour value in input field but chosen colour is still the old one. After pressing enter again colour is set to transparent.

### 🤔 Expected Behavior

I choose a colour in `<ColorInput />` then the colour changes. After I delete the colour code and press enter the colour value in input and chosen swatch are transparent.

### 👣 Steps to Reproduce

  - **WSR version**: `7.19.0`
  - **Yoshi version**: `4.21.11`

### 👀 Severity

- Critical

argshook:
closed by #4624

PR #4624, opened 2019-11-29
<ColorInput /> bug fix: color swatch will display the previous color after input is cleared

mantaspa:
### 🔦 Summary
This fixes the following issue: https://github.com/wix/wix-style-react/issues/4611
...

### ✅ Checklist

- 🔬 Change is tested
  - [x] Component


src/ColorInput/ColorInput.js
```diff
@@ -80,10 +80,15 @@ class ColorInput extends React.Component {
     };
   }
 
+  static _normalizeValue = value => {
```
argshook: i'd put this in `hex-helpers` file and use as regular function, not sure of the purpose of it being a static method

mantaspa: Well, this function is very related to this component. I don't know if it is ok to move it outside of the module.

argshook:
I'll leave this to be merged later, we have some timoeut troubles with visual tests :(

PR #4632, opened 2019-12-02
<AutoCompleteWithLabel /> - fix reset options after select

saarkuriel:
### 🔦 Summary

`<AutoCompleteWithLabel />` - fix reset options after select and not on input blur

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [X] Component
  - [ ] Visual test


src/AutoCompleteWithLabel/AutoCompleteWithLabel.js
```diff
@@ -108,12 +104,13 @@ class AutoCompleteWithLabel extends React.PureComponent {
       maxLength,
       placeholder,
       native,
+      onBlur,
```
mykas: Here you add a new API for our users. I think it should be documented

src/AutoCompleteWithLabel/docs/examples.js
```diff
@@ -1,12 +1,50 @@
...
+<AutoCompleteWithLabel 
+  label="my label" 
+  options={[
+    { id: 0, value: 'First option' },
+    { id: 1, value: 'Second option' },
+    { id: 2, value: 'Third option' },
+    { id: 3, value: 'Fourth option' },
+    { id: 4, value: 'Fifth option' },
+  ]}
+  onSelect={() => {}}
```
mykas: Why these noop functions? This can be covered inside the component by using defaultProps :)

src/AutoCompleteWithLabel/test/AutoCompleteWithLabel.spec.js
```diff
@@ -61,11 +61,15 @@ describe('AutoCompleteWithLabel', () => {
...
+      <AutoCompleteWithLabel
+        label="my autocomplete"
+        options={options}
+        onSelect={jest.fn()}
+      />,
     );
     await driver.enterText('a');
     expect(await driver.optionsLength()).toEqual(2);
-    await driver.blur();
+    await driver.clickAtOption(0);
```
mykas: So you are basically changing the logic of the `isEditing` , right?

src/AutoCompleteWithLabel/AutoCompleteWithLabel.js
saarkuriel: I didn't add new API for users, I got rid of a custom onBlur that used to be before

src/AutoCompleteWithLabel/test/AutoCompleteWithLabel.spec.js
saarkuriel: yes, I will exit `isEditing` on select and not on blur

src/AutoCompleteWithLabel/AutoCompleteWithLabel.js
mykas: Ok, got it. Did not see that it was already in the chaining function

src/AutoCompleteWithLabel/test/AutoCompleteWithLabel.spec.js
mykas: and what happens if user blurs out? will it be still in editing mode? does that make sense?

saarkuriel: I think so, since he didn't select any option... 

The problem with `<InputWithOptions />` is that `onBlur` is being called on the input blur on not on the entire component, which causes `onBlur` to fire before `onSelect`. In this situation, `isEditing` is turning to `false` and resetting the options, which results in the wrong option being selected. That's what I was attempting to fix.

PR #4654, opened 2019-12-04
<SidebarSectionTitle/> - Fix issue with wrong text color

### LINKED ISSUE #4653: <Sidebar/> - Title color is wrong
# 🐛 Bug Report

### 🏗 Relevant Components

<SidebarSectionTitle />

### 😯 Current Behavior
image

SidebarSectionTitle is almost transperant

### 🤔 Expected Behavior
SidebarSectionTitle should look as before
image

### 👀 Severity
- Critical , BLOCKER

simast:
Closes #4653


src/SidebarSectionTitle/SidebarSectionTitle.js
```diff
@@ -28,7 +28,7 @@ class SidebarSectionTitle extends React.PureComponent {
           return (
             <Text
               data-hook={dataHook}
-              {...styles('root', { skin }, this.props)}
+              {...styles('root', { theme: skin }, this.props)}
```
nitayneeman: I wonder what the cause of the issue is?

Because it seems to not be reproduced on Storybook without this change anyway:
https://wix-style-react.now.sh/?activeTab=Description&addons=0&full=0&panelRight=0&selectedKind=Components&selectedStory=SidebarSectionTitle&stories=1

shlomitc:
Are we sure that's the right solution?
I think that it's missing some specificity. If there would be another classname specifying that this `Text` is a child of `SidebarSectionTitle`, it will work in a more predictable way.

I suggest trying splitting this styling into two. The first part will be creating the `.root` which can be an empty selector and then have a child called `.text` and extends `Text`

then in the code, apply both `styles.root` and `styles.text`

simast:
I am closing this PR as it does not fix the related issue - we have verified this. It's really strange issue as I was not able to reproduce this in Storybook nor with an empty/basic Yoshi project.

It could be a Stylable bug: https://github.com/wix/stylable/issues/770

mykas:
Hey @shlomitc actually your solution is sort of workaround that works but the idea is that stylable is working in unpreditctable way. The whole idea of stylable was the when you extend the component and add some overriding styles - stylable should take care of specificity. And this is what they were saying in their presentations and docs. 

Here we have a situation when user is extending text and adding additional same name states on top - the states from Text are not getting overriden and then we have css specificity issue. We have two classnames that have same css complexity: 

text-[skin-standard]
sidebarsection-title-[skin-dark]

Then according to css specs the order of css matters. For sidebar issue it was text-[skin-standard]

shlomitc:
@mykas I see. However, most of our components are specified with a `root` + an extending class.
These specific cases that the root element extending some stylesheet are less frequent, so it might even be consistent change.

PR #4661, opened 2019-12-05
`<TableActionCell/>` -  changed the disabled visible items behaviour

amitrch1:
### 🔦 Summary

TableActionCell component does not respect secondaryItems' disabled prop,
the visible items are displayed as usual and are clickable,
the hidden items are not clickable but are displayed as not disabled..

also, the tooltip of a disabled item needs to reflect that it is disabled (maybe add a reason to why it is disabled)

### ✅ Checklist
member -->
- [Partly (mostly a bug fix)] 👨‍🎨 UX change is approved by the Design System UX 
- 📚 Change is documented
  - [X] API description
  - [X] Other ( the proptypes documentation)
- 🔬 Change is tested
  - [X] Component


src/TableActionCell/TableActionCell.js
```diff
@@ -33,25 +33,33 @@ function renderPrimaryAction({ text, theme, onClick, disabled }) {
...
+  return actions.map(
+    (
+      { text, icon, onClick, dataHook, disabled, disabledTooltipText },
+      index,
+    ) => (
+      <Tooltip
+        key={index}
+        dataHook={dataHook || 'table-action-cell-visible-action-tooltip'}
+        content={
+          disabled && Boolean(disabledTooltipText) ? disabledTooltipText : text
```
mykas: so you are making this mandatory to show tooltip when button is disabled?

src/TableActionCell/test/TableActionCell.spec.js
```diff
@@ -32,7 +32,8 @@ const secondaryActionsProps = ({ actionTriggers, actionDataHooks } = {}) => {
   };
 };
 
-describe('Table Action Cell', () => {
+// eslint-disable-next-line
+describe.only('Table Action Cell', () => {
```
mykas: This should not be committed

src/TableActionCell/TableActionCell.js
```diff
@@ -33,25 +33,33 @@ function renderPrimaryAction({ text, theme, onClick, disabled }) {
...
-          onClick();
-          event.stopPropagation();
-        }}
-        withNewIcons
+  return actions.map(
+    (
+      { text, icon, onClick, dataHook, disabled, disabledTooltipText },
+      index,
+    ) => (
+      <Tooltip
```
mykas: Actually it would be great to migrate this Tooltip to the new one. So for this case you would have `disabled` prop available for you and then you can control it with maybe additional prop like: `tooltipDisabled` and reuse the same `text` for showing information.

src/TableActionCell/TableActionCell.js
amitrch1: it is paired to current behaviour...
what do you suggest?

src/TableActionCell/test/TableActionCell.spec.js
amitrch1: 👍

src/TableActionCell/TableActionCell.js
amitrch1: besides the fact i have no idea where to find the new Tooltip i have no problem updating the Tooltip
although, i think this makes less sense with regards to the usage of the component
seems less intuitive to change the text for disabled case,
and also i would expect (as a user) that not providing a tooltipText when disabled would just not show the tooltip

mykas: You can read up the documentation in storybook. By passing `upgrade` prop you should align to new api.

Regarding the api yes ok it might be more intuitive. But lets not call it `disabledTooltipText`. Something more simple like: `disabledDescription`.

amitrch1: 👍

splintor:
@mykas What's the status of this PR?
Amit is out for a month, and I'm replacing it.
If there are no more comments, can you please approve and merge this PR? 🙏

PR #4667, opened 2019-12-05
<InputWithOptions /> - change blur invocation

saarkuriel:
### 🔦 Summary

This is a suggestion to change the way `<InputWithOptions />` invokes `blur`.
The current behaviour is that blur is actually the `input` blur which I think is wrong. It invokes when selecting an option but it invokes **before** the selection happen. This limits the way of using this component in controlled mode.

My suggestion is to invokes the blur only if you click outside the component or selecting an option.


EgoziE:
Using this code:
```
<InputWithOptions
  onBlur={() => {console.log('yay')}}
  options={[
    {
      id: 0,
      value: 'First option'
    }
  ]}
 />
```

The function `onBlur` is invoked twice.

PR #4686, opened 2019-12-10
`<Table/>`- fixing table with action cell example

TalZa1810:
### 🔦 Summary
Fixing Table with ActionCell example to have a hover effect when hovering a table row.


src/Table/docs/examples/TableActionCellExample.js
```diff
@@ -50,14 +50,15 @@ class TableActionCellExample extends React.Component {
                 ]}
                 // How many actions are visible, whereas the rest would be shown within a popover menu
                 numOfVisibleSecondaryActions={2}
-                alwaysShowSecondaryActions={false}
```
TalZa1810: it's false by default:
https://wix-style-react.now.sh/?activeTab=API&addons=0&full=0&panelRight=0&selectedKind=Components&selectedStory=TableActionCell&stories=1

src/Table/docs/examples/TableActionCellExample.js
```diff
@@ -50,14 +50,15 @@ class TableActionCellExample extends React.Component {
...
               />
             ),
           },
         ]}
+        onRowClick={(rowData, rowNum) => console.log(`Row data is: ${JSON.stringify(rowData)}. Row num is: ${rowNum}`)}
```
nitayneeman: We should explain why this prop usage enables the hover

nitayneeman:
After merging this PR, maybe it's worth to create an issue to separate the hover functionality from `onRowClick` - so that the row would have hover whether it's actually clickable or not.

TalZa1810:
It is implemented according to the UX spec: https://zpl.io/bo8Kl7v

PR #4690, opened 2019-12-10
<ColorPicker/> - Add dataHook to prop types

EgoziE:
`<ColorPicker/>` - Add dataHook to prop types

And changed every `toBeTruthy` / `toBeFalsy` to `toBe(true/false)`


src/ColorPicker/ColorPicker.js
```diff
@@ -25,6 +32,9 @@ export default class ColorPicker extends React.PureComponent {
   static displayName = 'ColorPicker';
 
   static propTypes = {
+    /** Data attribute for testing purposes */
```
TalZa1810:
```suggestion
    /** Applied as data-hook HTML attribute that can be used in the tests  */
```

PR #4691, opened 2019-12-11
<DatePicker/> - Remove font usage

EgoziE:
<DatePicker/> - Remove font usage


src/Calendar/DatePickerHead/index.js
```diff
@@ -47,16 +44,25 @@ const DatePickerHead = ({
...
         />
       ) : (
-        caption(
-          getMonthName(localeUtils.getMonths(), date.getMonth()),
-          'datepicker-month-caption',
-        )
+        <Text
+          className={styles.caption}
+          weight="normal"
+          dataHook={'datepicker-month-caption'}
```
TalZa1810: We usually have a `constants.js` file for data hooks and other constants. For example:
https://github.com/wix/wix-style-react/blob/master/src/LinearProgressBar/constants.js 

It can be used for both the component, driver and tests when necessary :)

EgoziE: I prefer not to make this refactor bigger than it is right now.
The point of this task is just to replace `<div/>`s with `<Text/>`
But it's a good idea and we can do it in another PR

PR #4692, opened 2019-12-11
Remove caret from ui core

EgoziE:
Removing the caret from `wix-ui-core` dependency.
Making version 2.0.542 its final.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "wix-style-react",
   "description": "wix-style-react",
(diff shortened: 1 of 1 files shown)


shlomitc:
@EgoziE - what's the status of this PR?

EgoziE:
Too many errors and warnings, this effort is bigger than anticipated.
I suggest that if someone using wsr6 encounters a problem in the future, we'll help them upgrade to 7...
Closing

PR #4694, opened 2019-12-11
`<FloatingNotification/>`- refactoring uni driver and constants file

TalZa1810:
Refactoring constants file to be similar to the library convention and updating component and uni driver accordingly.


src/FloatingNotification/index.js
```diff
@@ -1,2 +1 @@
 export { default } from './FloatingNotification.js';
-export { NOTIFICATION_TYPES } from './constants';
```
TalZa1810: https://github.com/search?p=1&q=org%3Awix-private+NOTIFICATION_TYPES&type=Code
unused exported variable.

PR #4695, opened 2019-12-11
<MessageBoxMarketerialLayout/> - Remove font usage

EgoziE:
`<MessageBoxMarketerialLayout/>` - Remove font usage


EgoziE:
Visual tests will fail for 2 pixels difference in height
approved by Ayelet

PR #4699, opened 2019-12-12
<GooglePreview/> | #PROM-3311 | Adding ellipsis to url

MayaHamo:
Adding ellipsis to url section in GooglePreview component


src/GooglePreview/GooglePreview.js
```diff
@@ -41,6 +41,7 @@ class GooglePreview extends React.PureComponent {
           light={false}
           className={style.googlePreviewUrl}
           data-hook={'googlePreview-previewUrl'}
+          ellipsis
```
shlomitc: 1. Should this be the default behavior?
2. What about the other texts, should they be ellipsed as well?
3. Please add another test simulation for the visual tests in `./test/GooglePreview.visual.js` according to [this guide](https://github.com/wix/wix-style-react/blob/master/docs/contribution/VISUAL_TESTING.md)

MayaHamo: 1. Yes
2. The title is already ellipsed and the description shouldn't be ellipsed.
3. I'll add

Thanks @shlomitc !

PR #4718, opened 2019-12-17
<RadioGroup /> add key to radio items

shamit6:
i have a problem in RadioGroup of duplicated radio items.
this is my use case.
https://github.com/wix-private/ecom/blob/5414cf7973220c2c453bd9efc3046cf3f1e714ff/client/wixstores-dashboard/wixstores-dashboard-tax/src/components/Modals/SelectTaxCode/TaxCodeCatalog/TaxCodeCatalog.tsx#L74

i fixed it by adding a key prop to the radio group list item.

radioGroup


shamit6:
fix it in my code

PR #4725, opened 2019-12-18
Add rtl/ltr to Storybook

EgoziE:
Added an addon that changes ltr/rtl of the page
Changed the name to `Wix Style React ${version}`


shlomitc:
Can this be added to wix-storybook-utils? That way it can be a shared across all libraries

argshook:
> Can this be added to wix-storybook-utils? That way it can be a shared across all libraries

technically yes, but because this is storybook version specific, i'm not sure `wix-storybook-utils` should be aligned to that.

but that's only a concern, it should be doable easily anyway

shlomitc:
Why not align versions? I think it’s a proper do decide that since some major version, we support only storybook@5 versions

argshook:
> Why not align versions? I think it’s a proper do decide that since some major version, we support only storybook@5 versions

at this moment all our libs are on storybook@5 but there's great chance it's wont be the case in future.

we can do this and try to sync versions but my hunch is that it could become problematic

for example, storybook@6 is released and we upgrade wix-ui-core to use
it, but the rest is still on storybook@5.

in such case wix-storybook-utils might work for storybook@5 but not storybook@6 and
we would also need to update/fix that.

anyway, i'm not against, just pondering if it's worth it

shlomitc:
I think it worth it

argshook:
added jira https://jira.wixpress.com/browse/DSM-79

PR #4735, opened 2019-12-20
<ImageViewer/> - do not render img html at all when imageUrl prop is not defined

mykas:
This PR is fixing issue when we have race condition for css calculation when switching from image to additem state.


src/ImageViewer/ImageViewer.js
```diff
@@ -106,7 +108,12 @@ class ImageViewer extends Component {
   };
 
   _renderImage = () => {
-    const { imageLoading } = this.state;
+    const { imageLoading, isImageLoaded } = this.state;
+
+    if (!this.props.imageUrl) {
```
simast: I think we need to improve this further. We cannot return here as we will be introducing breaking changes to testkit methods (they will throw as required elements/dataHooks will not exist).

src/ImageViewer/ImageViewer.js
```diff
@@ -123,6 +130,7 @@ class ImageViewer extends Component {
           /** hide container when no image provided, so AddItem behind it can be clickable */
           shouldDisplay: shouldDisplayContainer,
         })}
+        style={{ display: isImageLoaded ? 'flex' : 'none' }}
```
simast: I think we are duplicating `shouldDisplay` stylable state here (note that it is also used to set `display: none`)

src/ImageViewer/ImageViewer.js
mykas: Solved in testkits

Issue #4710, opened 2019-12-16
<InputArea/> - missing status prop


MayaHamo:
Currently `<InputArea/>` has only 2 status options:
1. regular
2. error

Unlike very similar component, `<Input/>` which has 4 status options:
1. regular
2. error
3. warning
4. loading

_Possible solution:_
Replace **error** (bool) and **errorMessage** InputArea props, in **status** (string) and **statusMessage** props, like in Input.

EgoziE:
PR - https://github.com/wix/wix-style-react/pull/4736/

shlomitc:
Merged, closing

PR #4736, opened 2019-12-22
<InputArea/> - Adding `status`  and `statusMessage` props

MayaHamo:
Solving this issue: https://github.com/wix/wix-style-react/issues/4710

Adding `status` and `statusMessage` props instead of current `error` and `errorMessage`
to support also in `warning` status (and maybe more statuses in the future)


src/InputArea/test/InputArea.spec.js
```diff
@@ -195,6 +209,20 @@ describe('InputArea', () => {
       });
     });
 
+    describe('status attribute', () => {
+      it('should display an error icon if status="error"', async () => {
+        const driver = createDriver(<InputAreaForTesting status="error" />);
+
+        expect(await driver.hasError()).toBeTruthy();
```
EgoziE: Please don't use `toBeTruthy()`
because not only `true` is truthy
Use `toBe(true)` instead

src/InputArea/test/InputArea.spec.js
```diff
@@ -195,6 +209,20 @@ describe('InputArea', () => {
...
+      });
+
+      it('should display an warning icon if status="warning"', async () => {
+        const driver = createDriver(<InputAreaForTesting status="warning" />);
+
+        expect(await driver.hasWarning()).toBeTruthy();
```
EgoziE: Please don't use `toBeTruthy()`
because not only `true` is truthy
Use `toBe(true)` instead

src/InputArea/InputArea.js
```diff
@@ -14,6 +15,9 @@ import { dataHooks } from './constants';
  * General inputArea container
  */
 class InputArea extends WixComponent {
+  static StatusError = 'error';
+  static StatusWarning = 'warning';
+
   constructor(props) {
```
EgoziE: Please add to the constructor a deprecation log.
Here's an example from `<DatePicker/>`:
```
import deprecationLog from '../utils/deprecationLog';
...
deprecationLog('DatePicker isOpen is deprecated. Please use initialOpen');
```

src/WarningIndicator/index.js
```diff
@@ -0,0 +1 @@
+export { default } from './WarningIndicator';
```
EgoziE: About the `<WarningIndicator/>` component.
It's an exact copy of `<ErrorIndicator/>` which is bad...

But, we decided not to let you refactor a lot of code, so I'll merge it and we'll refactor it later in another PR

src/InputArea/test/InputArea.spec.js
```diff
@@ -128,11 +128,25 @@ describe('InputArea', () => {
...
+          <InputAreaForTesting disabled status="error" />,
+        );
+        expect(await driver.hasExclamation()).toBeFalsy();
+      });
+
+      it('should not display a status icon even if status="warning"', async () => {
+        const driver = createDriver(
+          <InputAreaForTesting disabled status="warning" />,
+        );
+        expect(await driver.hasExclamation()).toBeFalsy();
```
EgoziE: Please don't use `toBeFalsy()`
because not only `false` is falsy
Use `toBe(false)` instead

src/InputArea/test/InputArea.spec.js
```diff
@@ -128,11 +128,25 @@ describe('InputArea', () => {
...
+      it('depreacted - should not display an error icon even if the error is true', async () => {
         const driver = createDriver(<InputAreaForTesting disabled error />);
         expect(await driver.hasExclamation()).toBeFalsy();
       });
 
+      it('should not display a status icon even if status="error"', async () => {
+        const driver = createDriver(
+          <InputAreaForTesting disabled status="error" />,
+        );
+        expect(await driver.hasExclamation()).toBeFalsy();
```
EgoziE: Please don't use `toBeFalsy()`
because not only `false` is falsy
Use `toBe(false)` instead

src/InputArea/docs/index.story.js
```diff
@@ -29,12 +32,30 @@ export default {
...
     theme: 'normal',
     hasCounter: false,
     resizable: false,
     disabled: false,
   },
 
+  exampleProps: {
+    status: [
+      {
+        label: 'Valid',
```
EgoziE: You can remove the `Valid` option
We don't want our consumers to mistake it as part of the API
There's a `Clear` button for removing status
image

src/InputArea/docs/ExampleWarning.js
```diff
@@ -0,0 +1,22 @@
...
+
+export default () => (
+  <div>
+    <div style={style}>
+      Left to right
+      <InputArea status="warning" statusMessage="Warning" />
+    </div>
+    <div className="rtl" style={style}>
+      Right to left
+      <InputArea rtl status="warning" />
```
EgoziE: Please add a `statusMessage`
(and for error too please :P)

src/WarningIndicator/index.js
MayaHamo: Yes I know :/ 
first I thought reusing `<Input/>` suffix components, but after talking with Shlomi I understand it's not a good idea.

but I still can unified `<ErrorIndicator/>` and `<WarningIndicator/>` to just `<StatusIndicator/>` or something.

EgoziE: Exactly
In order not to block you, we'll do it in another PR

MayaHamo:
@EgoziE pushed fixes :)

src/WarningIndicator/index.js
MayaHamo: Thanks!

PR #4742, opened 2019-12-23
Typography - Define Text API

EgoziE:
Defined 2 typography files (one scss, one st.css)
Rearranged `<Text/>` typography API


src/typography.scss
```diff
@@ -0,0 +1,24 @@
+:root {
```
shlomitc: This is a recreation of `src/fonts.scss` and we don't need it now

src/typography.st.css
```diff
@@ -0,0 +1,24 @@
+:vars {
```
shlomitc: replace the existing `Foundation/stylable/typography.st.css`.

shlomitc:
Goal:
Create 3 possible entries with the same API for both `Text` and `Heading`. All components in the library should use one of these options. In pseudo:
```
<Text/> (Public API to consumers)
Text.st.css (Internal API to in the library)
Text.scss => (Kind of a public API to consumers)
  1. through css modules <div className={Typography.text}/> 
  2. through plain scss @import "../Typography.scss" .myClass {@include text}
```

Things to remember:

Public APIs we should still to maintain and can't break:
1. Components: Text, Heading, Typography
2. styling: everything that in `src/*.scss`

Future thoughts:
We can decide how to publically allow consuming `.st.css` and other stylesheets if required. The mandatory thing is not breaking existing APIs.

EgoziE:
@shlomitc 
I changed the PR a bit, this is only a small refactoring
I don't want to waste more time on it :)

Now when `<Text/>` API is well defined, I'll start replacing all the messy places that use typography explicitly instead of using `<Text/>` component or `Text.st.css`/ `Text.scss`

src/Text/Text.st.css
```diff
@@ -1,113 +1,103 @@
 :import {
-  -st-from: "../Foundation/stylable/typography.st.css";
-  -st-named: fontLight, fontRoman, fontMedium, fontBold;
+  -st-from: '../Foundation/stylable/typography.st.css';
+  -st-named: sizeTiny, sizeSmall, sizeMedium, weightThin, weightNormal, weightBold, weightXBold;
+}
+:import {
```
shlomitc: redundant

src/Avatar/Avatar.st.css
```diff
@@ -304,7 +304,8 @@
 
 /* Sizes text */
 .avatar:contentType(text)::content {
-  -st-mixin: sizeSmall, weightBold, light;
+  -st-mixin: sizeSmall, weightBold;
+  color: value(D80);
```
shlomitc: Don't you want to extend `Text.st.css` here?

PR #4745, opened 2019-12-24
Storybook - fix all links in all stories

EgoziE:
All links were broken :)


src/ModalMobileLayout/docs/index.story.js
```diff
@@ -137,7 +139,10 @@ export default {
               title: 'Description',
               text: [
                 'Use this component together with ',
-                <LinkTo kind="Components" story="Modal">{`<Modal />`}</LinkTo>,
+                <LinkTo
+                  kind={Category.COMPONENTS}
+                  story="Modal"
```
shlomitc: As a next refactor - this can be taken from `stories/symbolsComponentsMapping/components.js` or `stories/symbolsComponentsMapping/symbols.js`.
Let's open a task for it

src/EmptyState/docs/index.story.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { linkTo } from '@storybook/addon-links';
+import { LinkTo, linkTo } from '@storybook/addon-links';
```
TalZa1810: no need for `LinkTo`. Only `linkTo`.

TalZa1810: Only `linkTo` is used in this file (no `LinkTo`).

TalZa1810:
```suggestion
import { linkTo } from '@storybook/addon-links';
```

stories/Introduction/Cheatsheet/componentsFamilies/FoundationFamily.js
```diff
@@ -1,4 +1,5 @@
 import React from 'react';
+import { Category } from '../../../storiesHierarchy';
```
TalZa1810: So because of the API change, we're using both `Category` and `symbolsGroup` in each Family file. 
The `SymbolsGroup` is for the UI (for the title- line 44) and the `Category` for the linking itself.

WDYT? @shlomitc

src/ModalMobileLayout/docs/index.story.js
EgoziE: https://jira.wixpress.com/browse/DSM-89

stories/Introduction/Cheatsheet/componentsFamilies/FoundationFamily.js
shlomitc: Ideally, we should use the up-to-date mapping in `symbolsGroup` and `storybookMapping`.

`storiesHierarchy` should be deprecated but if it's making the PR too big, we can do it later

EgoziE: I opened another task in Jira

PR #4749, opened 2019-12-25
using correct imports 

amiryonatan:
when examining our product bundle we noticed that we are bundling lodash libraries again even though we're externalizing it. 
we did the same practice in wix-ui-adi in order for webpack to recognize that this is the same lodash as we are externalizing


shlomitc:
There are a few issues with this approach:

1. It will immediately increase the bundle size of whoever is using the library and not externalizing `lodash`.
2. This will cause an increase of bundle size also because the library is not consumed in a tree-shakable way - `wix-style-react/Button` points to `dist/src/Button` which is not an "es-module". So this will cause a bad impact on everyone
3. In general, I'd like to avoid the approach that projects need to know the internal implementation of the library. However, you can decide to externalize the cherry-picked imports (e.g. 'lodash/clamp'`)

amiryonatan:
@shlomitc are you planning on changing the output of the library to es-module? 
this will definitely be better for all the consumers .. as you will be able to remove the `wix-style-react/Button` hack

shlomitc:
I don't see it as a hack, but I will definitely want to support esmodules officially.
In general, we plan the next major version to include performance improvements and this is one of them. It will probably take a couple more months to create this release...

shlomitc:
cc @mykas

PR #4750, opened 2019-12-25
Commiting auto generated files after new component

EgoziE:
Commiting auto generated files after new component
+ adding a line to the `<Box/>` docs


shlomitc:
I think all the testkit files can be git-ignored.
they are rebuilt anyway in a build step and are should not have any impact if changed or committed

@argshook - any other thoughts?

argshook:
> I think all the testkit files can be git-ignored.
they are rebuilt anyway in a build step and are should not have any impact if changed or committed

@shlomitc i wish to `.gitignore` those files too.

i'm hesitant to do that because git changes are currently "last line of defense", we're able to visually see if there's anything bad in there.

There can be something bad in there, because those files are generated
from a template, which can be edited (and broken) and sanity test might
not catch it.

but it's new year, i'm willing to take the risk :) (and fix if we're breaking)

i'll jira this

PR #4759, opened 2019-12-29
add onEditClick prop and use it for selection

splintor:
### 🔦 Summary
Enable passing `onColumnClick` field with a column definition, to make it easy handling clicking in the entire `<td>` area. This is continuing [this slack discussion](https://wix.slack.com/archives/C39FTGUBZ/p1577024853042400)

Currently, the mail problem we had (https://github.com/wix/wix-style-react/issues/4484) is fixed in https://github.com/wix/wix-style-react/pull/4616, but:
a. https://github.com/wix/wix-style-react/pull/4616 is a little hacky
b. it doesn't address other use cases, where users might want to be able handling other columns' clicks.

This PR currently doesn't address header clicks. If needed, this can easily be added.

### ✅ Checklist

- [?] 👨‍💻 API change is approved by the UI Infra Developers - still being discussed in https://wix.slack.com/archives/C39FTGUBZ/p1577024853042400
- [x] 👨‍🎨 UX change is approved by the Design System UX - [approved by @LittleTal](https://wix.slack.com/archives/C39FTGUBZ/p1577086351051600?thread_ts=1577024853.042400&cid=C39FTGUBZ)
- 📚 Change is documented
  - [x] API description
- 🔬 Change is tested
  - [x] Component


src/Table/Table.js
```diff
@@ -31,19 +31,24 @@ export function createColumns({ tableProps, bulkSelectionContext }) {
       ) : (
         <TableBulkSelectionCheckbox dataHook="table-select" />
       ),
+      onColumnClick: (column, row, rowNum, event) => {
```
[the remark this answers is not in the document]
splintor: I didn't understand why `row` and `rowNum` are redundant, and how it is related to the name of this field, but I have no problem chaning it to `onCellClick`, so [I did](https://github.com/wix/wix-style-react/pull/4759/commits/9c3c93b3103e63cadbd59da15d57908ff5b86c70).

splintor:
Not sure exactly what you meant by "https://github.com/wix/wix-style-react/pull/4616 doesn't work anymore", but I did find a problem when clicking on the `<label>` near the checkbox and fixed it (by calling `preventDefault`). I also removed the now-not-needed styles added in https://github.com/wix/wix-style-react/pull/4616 and set the `cursor` to be `pointer` for the entire `<td>`.

I also didn't understand what you meant by a `a suggestion for a non-hacky solution` - I thought this PR is such a suggestion... 😮 

Following [your comment](https://github.com/wix/wix-style-react/pull/4616#pullrequestreview-336152468), I'm thinking of adding code to also call `onCellClick` when header cells are clicked (passing `null` as `row` and `-1` as `rowNum`) - let me know what you think about this idea.

splintor:
Another thought about calling `onCellClick` for header clicks - it might be better to pass `{}` for header, to avoid exceptions in case someone forgot to address header clicks. Again, would love to hear your opinion.

src/Table/Table.js
```diff
@@ -31,25 +31,31 @@ export function createColumns({ tableProps, bulkSelectionContext }) {
...
               dataHook="row-select"
               checked={isSelected(id)}
-              onChange={() => toggleSelectionById(id, 'Checkbox')}
             />
           </div>
         );
       },
       width: '12px',
-      style: { padding: 0, height: '60px', width: '50px' },
+      style: { cursor: 'pointer' },
```
nitayneeman: Two things regarding this change:
1. We still need the previous styles (take a look at the visual tests, they are broken)
2. `cursor: 'pointer'` is supposed to be conditioned by `row.unselectable` (currently it always seems as selectable, no matter what)

src/Table/Table.js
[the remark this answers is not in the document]
nitayneeman: Maybe it was only me. Conceptually, it sounds to me that method that's called "onColumnClick" doesn't really need the row data as part of its signature but rather just the column index

nitayneeman:
> Not sure exactly what you meant by "#4616 doesn't work anymore", but I did find a problem when clicking on the `<label>` near the checkbox and fixed it (by calling `preventDefault`). I also removed the now-not-needed styles added in #4616 and set the `cursor` to be `pointer` for the entire `<td>`.
> 

That's exactly the problem I meant. 

> I also didn't understand what you meant by a `a suggestion for a non-hacky solution` - I thought this PR is such a suggestion... 😮

I treated actually this PR more generally - adding a new callback prop for the column. But yes, it also simplifies https://github.com/wix/wix-style-react/pull/4616, which is great :)

> Following [your comment](https://github.com/wix/wix-style-react/pull/4616#pullrequestreview-336152468), I'm thinking of adding code to also call `onCellClick` when header cells are clicked (passing `null` as `row` and `-1` as `rowNum`) - let me know what you think about this idea.

How would it work with `onSortClick` that's also triggered by the header when `sortable` is true?

nitayneeman:
> Another thought about calling `onCellClick` for header clicks - it might be better to pass `{}` for header, to avoid exceptions in case someone forgot to address header clicks. Again, would love to hear your opinion.

Not sure exactly what you mean (default value? for which prop?). Anyway, in case you've the capacity to do that, let's discuss and implement it in a separate PR.

splintor:
Good question about `onSortClick`. I guess we can decide that `onCellClick` will be called only when `onSortClick` is not defined.

In my second comment, I was referring to the value of the `row` parameter we will pass when calling `onCellClick` for header clicks. Passing `null` or `undefined` might cause an exception if `onCellClick` tries to access `row.id`.

Deferring it to a separate PR is a great idea. So, is there something still waiting for me in this PR?

nitayneeman:
> Good question about `onSortClick`. I guess we can decide that `onCellClick` will be called only when `onSortClick` is not defined.
> 
> In my second comment, I was referring to the value of the `row` parameter we will pass when calling `onCellClick` for header clicks. Passing `null` or `undefined` might cause an exception if `onCellClick` tries to access `row.id`.
> 
> Deferring it to a separate PR is a great idea. So, is there something still waiting for me in this PR?

Only that:
https://github.com/wix/wix-style-react/pull/4759#discussion_r362320918

src/Table/Table.js
splintor: Oops, I missed `unselectable`. The problem is currently, `style` is a column field, shared by all cells of this column.
I'm thinking of changing `style` to be `string | () => string`, so we can specify a different style for each cell. What do you think? Do you want me to do it in a separate PR?

nitayneeman: Why do we need a different style for each cell? (what's different besides `cursor`?)

The only case I can think of is the header cell, and if that's the case, let's take it out of the scope of this PR

splintor: As you said - `unselectable`. I want to set `cursor: 'pointer'` only for selectable rows, and the current `style` syntax doesn't allow that.

nitayneeman: Oh, because we don't have the `row` - makes sense. 

Personally I would prefer it would be part of `style.checkboxContainer` and not inline (this sounds the ideal solution but apparently would take more effort, I guess?)

Another approach - setting `cursor: 'pointer'` on the cell itself when `onCellClick` is provided, because we already have this:
https://github.com/wix/wix-style-react/blob/d97d4f91f0732a8108956cb12b63d0354dbc1f7b/src/Table/Table.js#L35

Which means here:
https://github.com/wix/wix-style-react/blob/d97d4f91f0732a8108956cb12b63d0354dbc1f7b/src/DataTable/DataTable.js#L282

Either way, in case we choose to add a new column prop - I think it's worth a different PR.

splintor: > We still need the previous styles
No, we don't. We just need to remove the extra classes added by the hacky https://github.com/wix/wix-style-react/pull/4616 - [Done](https://github.com/wix/wix-style-react/pull/4759/commits/43cc87d8001d16e31f34ff4d76c3d4278cb30b0f).

> Personally I would prefer it would be part of style.checkboxContainer and not inline 
The entire idea of this commit was to make the `<td>` in charge of clicking the checkbox, so it would be less hacky, and continue to work even if the size of the `<td>` changes and doesn't match the `checkboxContainer` element. I think this addition of the hacky `checkboxContainer` should be removed, and we should be able to style the entire cell as we like.

> Another approach - setting cursor: 'pointer' on the cell itself when onCellClick is provided
I'm trying to think in a more general way, and allow other developers that might need to style `<td>` differently for different rows.

I pushed [a commit to implement it](https://github.com/wix/wix-style-react/pull/4759/commits/79ebe74b7c602e3afd1b0578e05cf2b93235e911). Note that I'm not talking about adding a new column prop, but to enable developers to pass a function in the `style` prop.

Issue #4688, opened 2019-12-10
<MultiSelectInput/> - Selection on click issues


b3llash:
# 🐛 Bug Report
@tzabarc @BnayaZil 

### 🏗 Relevant Components
`<MultiSelectInput/>`

### 😯 Current Behavior

**Issue 1:**
Checkbox only gets into `hover` state when hovered directly, even though clicking anywhere on the item will check/uncheck it.
[GIF](https://i.gyazo.com/516737a3621bf27d8880d6aa7a0ba00c.gif)

**Issue 2:**
Item is not being checked/unchecked when clicking directly on the label.
[GIF](https://i.gyazo.com/877bc1496d56fbc1a8eaebe68d211718.gif)

### 🤔 Expected Behavior

**Issue 1:**
Checkbox should get into hover state when the user hovers any active area (that is: any area that will check/uncheck the item).

**Issue 2:**
Item should be checked/unchecked when clicking any active area of the item.

----

**WSR version**: `7.22.0`

### 👀 Severity
Low

BnayaZil:
The issue caused by the label.

EgoziE:
Hey @b3llash 
Thanks for opening this issue!
These bugs will be fixed once we're done with - https://github.com/wix/wix-style-react/issues/4656

Issue #4656, opened 2019-12-04
Use <ListItemSelect/> and <ListItemSection> in our OOTB dropdowns


EgoziE:
1. Check which dropdowns can use our new `<ListItemSelect/>` component
2. Refactor code

EgoziE:
When done, please close - https://github.com/wix/wix-style-react/issues/4688

PR #4762, opened 2019-12-29
stop unwanted click event

Jefoss:
### 🔦 Summary
MultiSelectCheckbox component has an annoying bug where you cant check the checkbox by clicking the node itself.
there is a click event on the input that causes the option that was just selected to get unselected right after.

you can reproduce the problem in the component [playground](https://wix-style-react.now.sh/?path=/story/components-api-components--multiselectcheckbox) - try to check the option by clicking the text/icons.

### ✅ Checklist
- 🔬 Change is tested
  - [x] Visual test

### DIFF
--- a/src/Checkbox/Checkbox.js
+++ b/src/Checkbox/Checkbox.js
@@ -130,6 +130,7 @@ class Checkbox extends WixComponent {
           checked={checked}
           disabled={disabled}
           onChange={disabled ? null : onChange}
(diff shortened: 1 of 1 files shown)


EgoziE:
Your solution made broke clicking on the checkbox box itself.
This issue is known - https://github.com/wix/wix-style-react/issues/4688
and will be fixed soon when https://github.com/wix/wix-style-react/issues/4656 is done

PR #4763, opened 2019-12-29
<NoBorderInput/> - Remove explicit use of font

EgoziE:
`<NoBorderInput/>` - Remove explicit use of font


src/NoBorderInput/NoBorderInput.js
```diff
@@ -55,16 +56,20 @@ class NoBorderInput extends React.Component {
...
 
     const renderStatusLine = () =>
-      !disabled && (
+      !disabled &&
+      status &&
+      statusMessage && (
         <div
           data-hook="status-message"
-          className={classNames(statusClass, styles.message)}
+          {...textStyles('root', {
```
TalZa1810: Why not using the `<Text/>` component itself?

EgoziE: 👍

PR #4772, opened 2019-12-30
Make tests more robust

EgoziE:
Refactor our unit tests by replacing `toBeTruthy` and `toBeFalsy` with `toBe(true/false)`

This time I had to change some tests and drivers
I made sure not to break any API
If a driver did something it wasn't supposed to (like `isSomething` return 4 instead of true) and someone used it wrong, it's their problem.


src/Input/tests/Input.spec.js
```diff
@@ -663,7 +663,7 @@ describe('Input', () => {
           );
           await driver.clickClear();
           expect(onClear).toHaveBeenCalledTimes(1);
-          expect(onClear.mock.calls[0][0]).toBeTruthy;
+          expect(onClear.mock.calls[0][0]).toBeTruthy();
```
nitayneeman: Maybe better?
```suggestion
          expect(onClear.mock.calls[0][0]).toBe(true);
```

EgoziE: **Expected:** true

**Received:** {"_dispatchInstances": null, "_dispatchListeners": null, "_targetInst": {"_debugHookTypes": null, "_debugID": 2227, "_debugIsCurrentlyTiming": false, "_debugOwner": [FiberNode], "_debugSource": null, "actualDuration": 0, "actualStartTime": -1, "alternate": [FiberNode], "child": [FiberNode], "childExpirationTime": 0, "contextDependencies": null, "effectTag": 128, "elementType": "button", "expirationTime": 0, "firstEffect": null, "index": 0, "key": null, "lastEffect": null, "memoizedProps": [Object], "memoizedState": null, "mode": 0, "nextEffect": [FiberNode], "pendingProps": [Object], "ref": [Function anonymous], "return": [FiberNode], "selfBaseDuration": 0, "sibling": null, "stateNode": <button … />, "tag": 5, "treeBaseDuration": 0, "type": "button", "updateQueue": null}, "bubbles": undefined, "button": 0, "cancelable": undefined, "currentTarget": null, "defaultPrevented": undefined, "dispatchConfig": {"dependencies": [Array], "isInteractive": true, "phasedRegistrationNames": [Object]}, "eventPhase": undefined, "isDefaultPrevented": [Function functionThatReturnsFalse], "isPersistent": [Function functionThatReturnsTrue], "isPropagationStopped": [Function functionThatReturnsFalse], "isTrusted": undefined, "nativeEvent": {"target": <button … />, "type": "click"}, "target": <button aria-disabled="false" class="buttonnext1194329874--root Focusable834259523--root closebutton53188677--closeButton buttonnext1194329874--medium" data-hook="input-clear-button" tabindex="0" type="button"><span … /></button>, "timeStamp": 1577780969070, "type": "click"}

PR #4773, opened 2019-12-31
Infra: remove .st.css and .scss imports from drivers

shlomitc:
## Why?
1. Drivers should be lean and simple javascript modules
2. Adding dependencies like styles force the consumer to transpile it wit, which adds a level of complication
3. `stylable` class names and data-attributes gets hash which is different between production code and dev code - see this issue https://github.com/wix/wix-style-react/issues/3160

## What was done
1. At first, changed only `unidriver` drivers, as they are relevant for E2E testing
2. Every `.st.css` import that looked for `st-state`, is now looking for `data-xxx` in the dom
3. Every `scss` class import was replaced with a simple string (it didn't work in e2e before, but now will not break builds for not transpiling scss)

## Pros and Cons on `data-xxx` attributes
+ no more imports in drivers, predictable tests
- increases dom size

## Other alternatives
Since it's only drivers code, we can still use the stylable states in the dom `data-123asd-skin="red"`, get the `dataset` and go over it looking for attribute names that end with `skin`.
This could work but requires modifications in `unidriver`. Also, in `stylable@2`, `data-*` are no longer used, so we will need to replace the strategy to class-names...

## Now what
Let's see how big is the bundle size after doing this change.
If it's not that bad - I think we should merge it and do the same in `wix-ui-core` and `wix-ui-backoffice`


giladsegal:
btw if builds are breaking now, it won't change because wix-style-react imports components from wix-ui-core (as you mentioned) and they use the same strategy so unless we apply it in all endpoints we will experience the same result

src/Text/Text.js
```diff
@@ -28,6 +28,12 @@ export const WEIGHTS = {
   bold: 'bold',
 };
 
+const getStyleDataAttributes = styleAttributes =>
```
giladsegal: maybe we can make this a util and re-use it? WDYT

giladsegal:
one more thing - I don't think in E2E tests we need to rely on class names, because in E2E we want to act as the user, e.g - I don't care that the notification has a success theme, I either want to see that the text changed or take a screenshot

src/Text/Text.js
shlomitc: In most of the components, there are 1 or 2 props so it was straightforward just to write it.
Also, it might cause some performance issues due to data manipulation and looping. Do you think it will be a problem?

shlomitc:
@giladsegal - I agree that all drivers' methods that check for styles are redundant. We should focus on the abstraction of behavior (clicking, editing, etc...)

Regarding `wix-ui-core`, you're right - that's the next step

shlomitc:
I'm merging in order not to have conflicts, we can keep discussing it on the thread anyway

nitayneeman:
Looks good (I know you already merged that).

I think it's worth documenting this practice (maybe adding to the generator an example spec or mention inside an internal best-practices doc we don't have yet).

shlomitc:
@nitayneeman - I would make this even stronger - a lint rule forbidding to add such imports in the drivers

PR #4776, opened 2020-01-01
revert: "Fix issue with EllipsisHOC creating new components on each r…"

nitayneeman:
This PR reverts https://github.com/wix/wix-style-react/pull/4734 to allow editor-based consumers (it turns out they exist!) to use the library, even though they're contained within an application uses a React version which lower than v16.8


nitayneeman:
This PR is waiting for verification from one of the consumers so it actually solves their problem.

simast:
Should we implement the same functionality using alternative solution? For example we could use lodash `_.memoize()` for the same effect.

nitayneeman:
This PR was intended to publish a package consumed by editor-based consumers, and so they do (at the moment, until we'll decide a better solution).

Closing.

PR #4784, opened 2020-01-02
Infra: add a sanity E2E test for puppeteer testkit import

shlomitc:
Following this PR https://github.com/wix/wix-style-react/pull/4773

This real E2E test should simulate consumers using our testkits as exported by the library.

**It approves that the puppeteer testkit exists and can be consumed**

**Note**:  this test does not check if `.st.css`/`.scss` imports were accidentally done because we use [`jest-preset-yoshi`](https://github.com/wix/wix-style-react/blob/master/test-runtime/imports/package.json#L26) which adds [`@stylable/jest plugin`](https://github.com/wix/yoshi/blob/4ff07e24fb84deb4bf91ec2d084355a135631cdc/packages/jest-yoshi-preset/jest-preset.js#L181) and other loaders for `scss`.


shlomitc:
CC @yavorsky, it was really easy to add a new test, can't believe we didn't use it so far

PR #4785, opened 2020-01-02
Fix clickRowCheckbox in <Table> teskit

splintor:
### 🔦 Summary
In https://github.com/wix/wix-style-react/pull/4759, we moved the click handler from the `<Input>` handle to the cell's `<td>` element.
This causes [contacts-web build](http://tc.dev.wixpress.com/viewLog.html?buildId=113652212&buildTypeId=Wix_CRM_Contacts_ContactsWeb) and [sites-list build](http://tc.dev.wixpress.com/viewLog.html?buildId=113655850&buildTypeId=SitesList_SitesList) to break.

This PR is meant to fix it, by making the testkit methods click the cell and not the checkbox.

I have no idea how come `Table.spec.js`, which uses the same testkit methods, didn't help. I'd love to hear your view on this.

### DIFF
--- a/src/Table/Table.uni.driver.js
+++ b/src/Table/Table.uni.driver.js
@@ -5,8 +5,10 @@ import deprecationLog from '../utils/deprecationLog';
 
 export const tableUniDriverFactory = base => {
   const dataTableDriver = dataTablePrivateUniDriverFactory(base);
+  const getRowSelectionCell = async index =>
+    await dataTableDriver.getCell(index, 0);
   const getRowCheckbox = async index =>
-    (await dataTableDriver.getCell(index, 0)).$('[data-hook="row-select"]');
+    (await getRowSelectionCell(index)).$('[data-hook="row-select"]');
   const getRowCheckboxDriver = async index =>
     checkboxUniDriverFactory(await getRowCheckbox(index));
   const getBulkSelectionCheckboxDriver = async () => {
@@ -53,9 +55,9 @@ export const tableUniDriverFactory = base => {
       deprecationLog(
         '"clickRowChecbox" method is deprecated (because of typo) and will be removed in next major release, please use "clickRowCheckbox" driver method',
       );
-      return (await getRowCheckbox(index)).click();
+      return (await getRowSelectionCell(index)).click();
     },
-    clickRowCheckbox: async index => (await getRowCheckbox(index)).click(),
+    clickRowCheckbox: async index => (await getRowSelectionCell(index)).click(),
     /** Click the bulk-selection checkbox */
     clickBulkSelectionCheckbox: async () =>
       (await getBulkSelectionCheckboxDriver()).click(),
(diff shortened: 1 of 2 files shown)


splintor:
@nitayneeman I checked this further and realized that the table driver code has two problems.
First, the `clickRowCheckbox` method calls the `element.click()` method which might be problematic with JSDOM - that's we we try to use `Simulate` when running React tests. Not sure why WSR doesn't fail because of it.
Second, the misspelled `clickRowChecbox` method calls the checkbox driver `click()` method and this triggers a `change` event rather than `click` event, which works well on the checkbox, but doesn't bubble up to the `<td>` as a click event, so project that still use it (like `contacts-web`) will fail after our change.

To fix it, I changed the two methods to use `eventTrigger.click(element)`, and in the UniDriver, mimic the Checkbox UniDriver and call `eventTrigger.click(element)` for React, or `event.click()` otherwise.

I checked it with `sites-list` (which was failing without this PR), and it passed.

src/Table/Table.uni.driver.js
```diff
@@ -2,13 +2,19 @@ import { baseUniDriverFactory, getElement } from '../../test/utils/unidriver';
...
 export const tableUniDriverFactory = base => {
   const dataTableDriver = dataTablePrivateUniDriverFactory(base);
   const getRowCheckbox = async index =>
     (await dataTableDriver.getCell(index, 0)).$('[data-hook="row-select"]');
   const getRowCheckboxDriver = async index =>
     checkboxUniDriverFactory(await getRowCheckbox(index));
+  const simulateClick =
+    base.type === 'react'
+      ? // eslint-disable-next-line no-restricted-properties
+        async element => Simulate.click(await element.getNative())
```
nitayneeman: It seems that Unidriver already invokes `Simulate.click` for React:
https://github.com/wix-incubator/unidriver/blob/master/adapters/jsdom-react/src/index.ts#L150

So I wonder why just invoking `async element => base.click()` is not enough?

splintor: I'm far from being a `UniDriver` expert - I just copied what was done for the [Checkbox UniDriver](https://github.com/wix/wix-style-react/blob/master/src/Checkbox/Checkbox.uni.driver.js#L22,L31), but maybe this was done because for some reason, `change` event was needed instead of `click`.

Giving another look at my code after your comment, I realized it has another problem - it should not call `base.click()` but `element.click()`.

Pushed a fix - please review again.

src/Table/Table.uni.driver.js
```diff
@@ -2,6 +2,7 @@ import { baseUniDriverFactory, getElement } from '../../test/utils/unidriver';
 import { dataTablePrivateUniDriverFactory } from '../DataTable/DataTable.private.uni.driver';
 import { checkboxUniDriverFactory } from '../Checkbox/Checkbox.uni.driver';
 import deprecationLog from '../utils/deprecationLog';
+import { Simulate } from 'react-dom/test-utils';
```
nitayneeman: This is redundant:
```suggestion

```

splintor:
Removed unneeded import. `contacts-web` should work. `sites-list` doesn't currently use `Table UniDriver` so no point in re-testing it.

PR #4787, opened 2020-01-03
<ColorPicker/> - Remove use of WixComponent

EgoziE:
Remove use of WixComponent from `<ColorPicker/>` with a small refactor


src/ColorPicker/ColorPicker.js
```diff
@@ -107,8 +99,16 @@ export default class ColorPicker extends React.PureComponent {
 
     return (
       <div className={css.root} data-hook={dataHook}>
-        <ColorPickerHsb current={current} onChange={this.change} />
-        <ColorPickerHue current={current} onChange={this.change} />
+        <ColorPickerHsb
+          dataHook="color-picker-hsb"
```
TalZa1810: I think that these strings should be added to a `constants` file so that they can be used in both the component and the drivers.

src/ColorPicker/ColorPicker.js
```diff
@@ -107,8 +99,16 @@ export default class ColorPicker extends React.PureComponent {
...
+          current={current}
+          onChange={this.change}
+        />
+        <ColorPickerHue
+          dataHook="color-picker-hue"
```
TalZa1810: same here

src/ColorPicker/ColorPicker.js
EgoziE: done

PR #4788, opened 2020-01-04
Infrastructure: fix named-importing issues

mykas:
## Infrastructure: fix named-importing issues (support es-modules)

This PR is fixing named imports and introduce a new storybook that will serve us in finding those ugly stylable errors during runtime when our users using our library through named imports.
Screenshot 2020-01-16 at 13 18 17

### How our es modules are created

We have this script `npm run transpile` which basically copies our src folder to `dist/es/src` and modifies some components and stylesheets that are importing from `wix-ui-core`. 

The modifications for our js file for example Popover:
from
```js
import Popover from 'wix-ui-core/dist/src/components/popover'
```
to 
```js
import Popover from 'wix-ui-core/dist/es/src/components/popover'
````

The modifications for our st.css files for example Popover:
from:
```css
:import {
    -st-from: 'wix-ui-core/indext.st.css';
    -st-default: Popover;
}
```
to
```css
:import {
    -st-from: 'wix-ui-core/indext.es.st.css';
    -st-named: Popover;
}
```

### So why do we still have those stylable errors?

The reason for this is because the script that changes those paths for stylable is limited to only finding the imports that are used like this: `-st-from: 'wix-ui-core/indext.st.css';` , but in our codebase we have many stylesheets that import with the path:
```
-st-from: `wix-ui-core/dist/src/components/popover/Popover.st.css'
```
And thats ok, because we have been never informed to write otherwise.

### What did I do to fix this

So fixing these kind of imports is not so easy because we have many different cases:
```
:import {
    -st-from: "wix-ui-core/dist/src/components/button-next/button-next.st.css";
    -st-default: Button;
    -st-named: suffix, prefix;
}
```

Somes ideas that came up and got crushed by the reality:
- why don't we have a lint rule for that? nope. Stylable does not have linter that we can use.
- maybe stylable 2/3 already solved this? nope. They only suggested to do some workaround which would effort for us to change things. And also said that workaround would only work in stylable 3.

So whats left? 

Of course throwing error on our `npm run transpile` when we detect such paths and explain it how to change it.

```
onst wrongPaths = [
            {
              regexp: /wix-ui-core\/dist\/src\/hocs\/.*\/[A-Za-z]*\.st\.css/,
              correct: 'hocs.st.css',
            },
            {
              regexp: /wix-ui-core\/dist\/src\/components\/.*\/[A-Za-z]*\.st\.css/g,
              correct: 'index.st.css',
            },
          ];

          wrongPaths.forEach(({ regexp, correct }) => {
            if (regexp.test(results)) {
              const message = [
                'This stylesheet',
                filepath,
                'includes import path',
                results.match(regexp),
                'which is not compatible with our es-modules infrastructure.',
                'Makes sure to change your wix-ui-core import to this: -st-import:',
                `wix-ui-core/${correct}`,
                'and to named import: -st-named: ComponentName;',
                '',
              ];
              throw new Error(message.join(' '));
            }
          });
```


shlomitc:
I think it might be a bit of an overkill.
What you're doing is testing ALL the components visually (which is expensive) just to make sure we didn't do an error while writing the code.

It sounds to me like this should throw an error during build time, and not test time.
Can you think of some lint rule to force it?

mykas:
Yes, I thinking on improving Artems script that transpiles the path. With linting its a bit hard because there is no lint for stylable or at least I am not aware of how to extend it.

mykas:
The stylable error is problematic as it throws only in runtime. Thats why I went for visuals path.

shlomitc:
AFAIK stylable should break during build time for errors. Mind talking with Tom Raviv about this?

mykas:
Sure, maybe it is already solved in stylable.

package.json
```diff
@@ -38,6 +38,7 @@
     "storybook": "npm run prestorybook && start-storybook -p 6006",
     "storybook:visual": "start-storybook -c ./.storybook/visual -p 6007",
     "storybook:e2e": "start-storybook -c ./.storybook/e2e -p 6008",
+    "storybook:visual-es": "start-storybook -c ./.storybook/visual-es -p 6009",
```
shlomitc: who's running this command?

test-runtime/imports-visual/es-named.visual.js
```diff
@@ -0,0 +1,38 @@
...
+];
+
+components.forEach(name => {
+  if (ignoreList.includes(name)) {
+    return;
+  }
+
+  const props = DEFINITIONS[name] && DEFINITIONS[name].props;
+
+  const { [name]: Component } = require(`../../`);
```
shlomitc: where is this import leading to? the root of the project?

testkit/component-definitions.js
```diff
@@ -1,18 +1,15 @@
...
-const {
-  Popover,
-  MultiSelect,
-  AutoComplete,
-  Input,
-  Notification,
-  FloatingHelper,
-  GoogleAddressInput,
-} = AllComponents;
+import { GmapsTestClient } from '../src/GoogleAddressInput/gmapsTestClient';
```
shlomitc: This file should not sit in the `testkit` folder.
since it's test-related metadata, it should not be exposed eventually.
it should sit near to the test files

.storybook/visual-es/webpack.config.js
```diff
@@ -0,0 +1,22 @@
+const merge = require('lodash/merge');
```
shlomitc: What you're actually doing is testing the build tool.
So running these tests (and also the storybook) as part of the development flow is just wrong.

you're using the `test-runtime` which is great because you test the library's output, however, you're running the test at the library level (like a uni-test).

This whole thing should be treated separately from the regular testing flow of components

It might be better to move this whole configuration to the `test-runtime` folder

package.json
mykas: Nobody, we will leave this as a tool to check the components that are being used with named imports. If somebody complaints that they have stylable error when they changed to named imports - we have a way to check it here.

test-runtime/imports-visual/es-named.visual.js
mykas: Yes, and finds the `index.js`

testkit/component-definitions.js
mykas: I think we can move it to `.wuf` folder.

test-runtime/imports-visual/es-named.visual.js
shlomitc: I don't think we generate one in the root. it should go to `dist`

.storybook/visual-es/webpack.config.js
mykas: So I am not sure if we gonna run visuals tests for now. Now that we have some sort of script that covers our stylable issues maybe lets leave this storybook just for testing purposes.

package.json
shlomitc: so following to my other comments - this should move as well to the `test-runtime` domain and not be part of this component development flow

.storybook/visual-es/webpack.config.js
shlomitc: so if you don't use it - let's remove it from the PR

test-runtime/imports-visual/es-named.visual.js
mykas: Sorry I am wrong. It actually reads package.json and reads either `module` or `main`. Depending if user needs commons js or es modules

mykas: This is actually how it works with webpack based builders. For example jest uses `main` for finding the paths

.storybook/visual-es/webpack.config.js
mykas: But this is webpack-config file that is being used to build storybook that shows if components are having issues with stylable while named imported

shlomitc: buy it's working only on the output (after `npm run bulid` is run)

test-runtime/imports-visual/es-named.visual.js
shlomitc: interesting. why can't you `import` btw?

shlomitc: just `import * as components from '../../'` and then loop over it

testkit/component-definitions.js
shlomitc: It looks more like a test asset and not part of the build tool.
Probably it should sit next to each one of the components (like the `.meta` file we used to have)

mykas: thats another topic actually! I've talked with Arijus about it and we agreed that we should have somewhere a place near component that will define what are the required components for simple example that could be used for showcasing component or doing teskit sanity tests :) for example remember that application Arijus did where you drag n drop components? Thats where it can benefit too for required props

shlomitc: It’s not just only required props - it’s a valid simulation. Simulations can be used for multiple purposes, one of them is sanity/default rendering

mykas: For now I am gonna leave this in testkit folder but I will open an issue that we should think about adding required props to our simulation files instead of having one file that covers all this.

package.json
mykas: renamed to `storybook:named-imports` to be more descriptive what it is about

test-runtime/imports-visual/es-named.visual.js
mykas: done

.storybook/visual-es/webpack.config.js
mykas: I've added `npm run transpile` in the `storybook:named-import` script.

test-runtime/imports-visual/es-named.visual.js
shlomitc: You just keep on patching it.
Again - all of these tests/debug mode/storybook visuals are testing the **output** of the library.

All of this should not be part of the regular development flow, hence not part of the package at all.

What you are doing is testing the build tool, and it should not be part of the regular development flow, hence not in the package json. We’re just keep on pushing there stuff without control and it has a cost of maintenance, understanding and unclear workflow.

I’d insist that all of this should go out of the package json and that the storybook should not be in the root of the project.

Use the e2e-runtime as a platform to test the dist output of the library

mykas: But e2e is costly to test out all the components and we agreed that visual testing is also an overkill, what I was saying that until we really are sure about our named-imports - we have to have some kind of tool to see where are the problems to cover those with maybe more script runners. I believe this storybook will be deleted in the future as we gonna have a pretty solid infrastructure that will cover most our cases.

mykas: Knowing that we might actually fix all the stylable errors when we migrate to stylable 3 - I don't want to spend any time on adding additional testing infrastructure for testing named-imports, because we have already a script that suppose to throw errors when we have wrong import paths in stylable.

mykas: Or do you mean that you want to move named-imports storybook config from .storybook because it will confuse our users on whats the purpose? I mean I could move it to test-runtime just for making it clear that its test-runtime, but why not to have the script in npm package.json? is it because it can confuse people?

mykas: ok, I think I have an idea.

mykas: Screenshot 2020-01-17 at 11 19 17
Do you mean this?

shlomitc: Yes, exactly

mykas: Anyway, running internally in that project is not an option because I need to target then `start-storybook` from node_modules of wix-style-react and it start to fail on some files for no reason. So I still need to have a npm script in our package.json :)

src/Button/Button.st.css
```diff
@@ -1,7 +1,6 @@
 :import {
-    -st-from: "wix-ui-core/dist/src/components/button-next/button-next.st.css";
-    -st-default: Button;
-    -st-named: suffix, prefix;
```
simast: just so we don't forget - let's double check suffix/prefix changes, tests are showing diffs

package.json
```diff
@@ -61,7 +61,7 @@
     "lint": "yoshi lint --fix",
     "copy-public-styles": "cp src/*.scss .",
     "bundle": "webpack --config ./node_modules/yoshi/config/webpack.config.client.js --debug --display-error-details --progress",
-    "transpile": "node ./scripts/transpile.js",
+    "transpile": "rm -rf dist && node ./scripts/transpile.js",
```
simast: `rimraf dist && ...` - for cross-platform (we have rimraf as dependency)

scripts/replaceStylableImports.js
```diff
@@ -20,11 +20,54 @@ module.exports = () => {
             reject();
             return;
           }
-          const codeWithEsImport = content.replace(
-            /wix-ui-core\/index\.st\.css/g,
-            'wix-ui-core/index.es.st.css',
+
+          let results = content; //eslint-disable-line
```
simast: don't seem to have any eslint warning, do we need this comment?

scripts/replaceStylableImports.js
```diff
@@ -20,11 +20,54 @@ module.exports = () => {
...
+            .replace(
+              /wix-ui-core\/index\.st\.css/g,
+              'wix-ui-core/index.es.st.css',
+            )
+            .replace(
+              /wix-ui-core\/hocs\.st\.css/g,
+              'wix-ui-core/hocs.es.st.css',
+            );
+
+          foundStylesheets.forEach(
```
simast: as discussed - we can just throw an error during build time if we see hoc imports from dist the same way as with components.

scripts/replaceStylableImports.js
```diff
@@ -20,11 +20,54 @@ module.exports = () => {
...
+            name =>
+              (results = results.replace(
+                `-st-default: ${name}`,
+                `-st-named: ${name};`,
+              )),
           );
-          fs.writeFile(filepath, codeWithEsImport, function(err) {
+
+          if (results.match(/wix-ui-core\/dist\/src\/components/g)) {
```
simast: `.match()` returns an array while `regexp.test(string`) returns a boolean - would suggest to use `.test` here instead.

```js
if (/wix-ui-core\/dist\/src\/components/g.test(results)) {
```

scripts/replaceStylableImports.js
mykas: yes, it changes to `const`

mykas: oh wait, today it stopped showing..never mind then

mykas:
Blocked by this in order to fix `prefix` and `suffix` issues in Button: https://github.com/wix/wix-style-react/pull/4886

Issue #4000, opened 2019-08-23
Beta <PopoverMenu/> - deprecate old popover from the storybook


shlomitc:
# ✨ Feature Request
`<PopoverMenu/>`

### 🔦 Scope
old PopoverMenu should be moved from the storybook documentation into the deprecated section.

Also - the new popover should be available through the playground and not just in its page doc.

### 💁 Explanation
Components/PopoverMenu => Deprecated/PopoverMenu

Also, the test file should be moved
Tests/7. Tooltip/ 7.3 PopoverMenu => try to move to visual or just rename to Tests/Deprecated/PopoverMenu

regarding the playground, either create a new `Beta` scope or decide to merge it with the `Components` and create a `Deprecated` scope for old components

mykas:
Fixed

PR #4789, opened 2020-01-04
<PopoverMenu/> - deprecate story

mykas:
This PR is deprecating old PopoverMenu story with a message to use the one from beta section. https://github.com/wix/wix-style-react/issues/4000


simast:
By the way, maybe we can add `PopoverMenu` to Playground's LiveExample `scope={...}` prop? So the new PopoverMenu can be used in playground (this was mentioned in issue #4000).

PR #4790, opened 2020-01-04
Generator- adding constants file

TalZa1810:
1. Adding constants file to the generator and using this in both the component & driver.
2. Moving the stylable to the root div and updating st.css file accordingly.


.wuf/generator/templates/Component/Component.js
```diff
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types';
 import Text from '../Text';
 import Button from '../Button';
 import styles from './{%ComponentName%}.st.css';
+import { dataHooks } from './constants';
```
EgoziE: How about:
```
import { dataHooks } from './{%ComponentName%}.const';
```

What do you think is better name?

TalZa1810: We talked about it in our last `design system` meeting and have decided to leave this as `constants` file. Btw, most constants file in the library are named as `constants`.

.wuf/generator/templates/Component/constants.js
```diff
@@ -0,0 +1,6 @@
+
+
+export const dataHooks = {
```
simast: Just a tiny observation regarding enum naming in general. In TypeScript `enum`s are both types and values. So if you have a component accepting a prop that should be an enum value - you will type it using the same enum. For example:

```ts
interface Props {
  size: Size;
}
```

That's why I personally like naming enums using PascalCase and not using plural forms. Otherwise the enum-like constants will look odd if we ever migrate to TypeScript and will need additional renaming then.

TalZa1810: As far as I know, currently `dataHooks` are named as camal case in most of the components `constants` files . I think that if we decide to change it in the generator to `DataHooks`, we should create another PR to align all the components constants to look the same. 
WDYT?

simast: Yep, just noticed that most components use `dataHooks` naming, so let's keep it this way for consistency.

TalZa1810: ok, so can I merge this?

PR #4791, opened 2020-01-05
Variable input

tomershk:
### 🔦 Summary

Adding a new component: `<VariableInput />`
Allowing to mix text with tags UI (`<Tag />`), add them using API or text placeholders.

### ✅ Checklist

- 📚 Change is documented
  - [x] Story
  - [x] API description
- 🔬 Change is tested
  - [x] Component
  - [x] Visual test


src/VariableInput/test/EditorUtilities.spec.js
```diff
@@ -0,0 +1,135 @@
...
+  };
+
+  let editorState;
+
+  beforeEach(() => {
+    editorState = EditorState.createEmpty();
+  });
+
+  describe('String to and from ContentState', () => {
+    it('should convert placeholder to entity', () => {
```
nitayneeman: The terminology question is also relevant here (and actually along this whole file)

src/VariableInput/test/EditorUtilities.spec.js
```diff
@@ -0,0 +1,135 @@
...
+        strPrefix.length,
+      );
+
+      editorState = _moveSelectionTo(editorState, strPrefix.length + 3);
+      editorState = EditorUtilities.moveToEdge(editorState);
+      expect(editorState.getSelection().getAnchorOffset()).toEqual(
+        strPrefix.length + variableEntity.text.length + 2, // Entity with space on each side
+      );
+    });
+  });
```
nitayneeman: I think we should add a test or two that use `hasUnparsedEntity`

src/VariableInput/docs/index.story.js
```diff
@@ -0,0 +1,112 @@
...
+              }}
+            />`,
+            }),
+          ]),
+
+          code({
+            title: 'Full Interactive Preview',
+            description: 'A non compact version of same code example as above',
+            source: '<VariableInput buttonText="Hello World!"/>',
+          }),
```
nitayneeman: It seems redundant, let's remove that

src/VariableInput/docs/index.story.js
```diff
@@ -0,0 +1,112 @@
...
+  storyName: 'VariableInput',
+
+  component: VariableInput,
+  componentPath: '..',
+
+  componentProps: {
+    initialValue: 'Hello World!',
+    size: 'small',
+    status: '',
+    statusMessage: '',
```
nitayneeman: We don't support these props yet

src/VariableInput/docs/index.story.js
```diff
@@ -0,0 +1,112 @@
...
+          ]),
+
+          divider(),
+
+          title('Examples'),
+
+          columns([
+            description({
+              title: 'Simple Usage',
+              text: 'A simple example with compact preview',
```
nitayneeman: The description should be more informative, something like: "This example shows a value that contains plain text and known variable"

src/VariableInput/test/VariableInput.spec.js
```diff
@@ -0,0 +1,127 @@
...
+      const text = 'Some text {{page.name}}';
+
+      const driver = createDriver(
+        <VariableInput onSubmit={callback} variableParser={variableParser} />,
+      );
+      driver.enterText(text);
+
+      expect(callback).not.toHaveBeenCalled();
+    });
+  });
```
nitayneeman: I think we should add a test that changes the prefix and suffix (using `variableTemplate`) and then parses the value correctly according to these affixes

src/VariableInput/VariableInput.scss
```diff
@@ -0,0 +1,70 @@
...
+      border-color: $B10;
+      box-shadow: 0 0 0 3px $F00;
+    }
+  }
+}
+.tag-wrapper {
+  line-height: initial;
+  padding-left: 2px;
+  padding-right: 2px;
+  > span{
```
nitayneeman: Let's replace it with a dedicated class selector :)

Like:
```suggestion
  > .tag {
```

src/VariableInput/test/VariableInput.visual.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import VariableInput from '../VariableInput';
+
+const commonProps = {
+  variableParser: variable => (variable === 'page.name' ? 'Page name' : false),
+};
+
+const tests = [
```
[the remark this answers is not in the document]
tomershk: Will add them in the next PR once `size` will be added

nitayneeman:
I asked @ayeletr to do a UX review

src/VariableInput/VariableInput.js
```diff
@@ -0,0 +1,166 @@
...
+    className: string,
+    /** Applied as data-hook HTML attribute that can be used in the tests */
+    dataHook: string,
+    /** Initial value to display in the editor */
+    initialValue: string,
+    /** Callback funciton when focusing out. Also, after calling `insertVariable()`
+     * `onSubmit(value: String, invalidKeys: Array<String>): void` */
+    onSubmit: func,
+    /** Specifies the size of the input and variables*/
+    size: oneOf(Object.keys(sizeTypes)),
```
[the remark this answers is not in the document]
tomershk: prop removed, will do once we add the `size` in the next PR

src/VariableInput/VariableInput.st.css
```diff
@@ -0,0 +1,25 @@
+:import {
```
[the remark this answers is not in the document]
nitayneeman: A better decision - let's move the styles from the SCSS file and use Stylable instead :)

src/VariableInput/VariableInput.scss
```diff
@@ -0,0 +1,70 @@
...
+  :global(.public-DraftEditor-content) {
+    background-color: $backgroundColor;
+    min-height: $minHeight;
+    max-height: var(--max-height);
+    padding: $padding;
+    border: $border;
+    border-radius: $borderRadius;
+    overflow: auto;
+    box-sizing: border-box;
+    line-height: 30px;
```
nitayneeman: Basically we use the `Text` mixin (line 43), so why do we need to specify `line-height` manually?

src/VariableInput/EditorUtilities.js
```diff
@@ -0,0 +1,290 @@
...
+            <span className={styles.textWrapper}> </span>
+            <Tag
+              id={entityKey}
+              removable={false}
+              theme={disabled ? 'neutral' : 'dark'}
+              size={size}
+            >
+              {text}
+            </Tag>
+            <span className={styles.textWrapper}> </span>
```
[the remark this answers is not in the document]
tomershk: When adding just the `<Tag />` cursor is placed inside the tag while we are right after the tag.
Also when we are before the tag. 
This will allow us proper UI while moving with the cursor

src/VariableInput/VariableInput.uni.driver.js
```diff
@@ -0,0 +1,24 @@
...
+  return {
+    ...baseUniDriverFactory(base, body),
+    getContent: () => getContent(base).text(),
+    getPlaceholder: () => getPlaceholder(base).text(),
+    enterText: async text => {
+      const contentElement = await getContent(base).getNative(); // eslint-disable-line no-restricted-properties
+
+      // TODO: implement for puppeteer. Throw error if type is not handled
+      if (base.type === 'react') {
+        ReactTestUtils.Simulate.beforeInput(contentElement, { data: text });
```
nitayneeman: We've a utility called `ReactBase` that wraps `ReactTestUtils`:
https://github.com/wix/wix-style-react/blob/master/test/utils/unidriver/ReactBase.js

So we can add a new method that simulates `beforeInput` and then reuse it, like here:
https://github.com/wix/wix-style-react/blob/master/src/Tooltip/Tooltip.uni.driver.js#L51

src/VariableInput/test/EditorUtilities.spec.js
```diff
@@ -0,0 +1,135 @@
...
+
+describe('EditorUtilities', () => {
+  const variableEntity = {
+    text: 'Page name',
+    value: 'page.name',
+  };
+  const variableParser = value =>
+    value === variableEntity.value ? variableEntity.text : false;
+  const strPrefix = 'test string ';
+  const strPlaceholder = `${strPrefix}{{${variableEntity.value}}} `;
```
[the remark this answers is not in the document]
tomershk: Yes, with placeholder, will rename it

src/VariableInput/test/EditorUtilities.spec.js
tomershk: It is a wrapper function for `getMatchesInString`

src/VariableInput/VariableInput.scss
```diff
@@ -0,0 +1,70 @@
+@import '~draft-js/dist/Draft';
+
+@import '../common';
+@import '../typography-helpers';
+
+$minWidth: 378px;
+$minHeight: 30px;
```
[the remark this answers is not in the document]
tomershk: Correct, no minimum width size, height will be handled when pushing the `size` prop

test/utils/unidriver/ReactBase.js
```diff
@@ -61,6 +61,8 @@ export function ReactBase(base) {
       Simulate.mouseEnter(await htmlElement(), eventData),
     mouseLeave: async eventData =>
       Simulate.mouseLeave(await htmlElement(), eventData),
+    enterText: async eventData =>
```
nitayneeman: I see the convention is to use the same method name, so maybe:
```suggestion
    beforeInput: async eventData =>
```

src/VariableInput/test/EditorUtilities.spec.js
nitayneeman: I think it's worth to test because it's sort of an API of this file (that we would like to cover somehow). But, we cover the method indirectly so maybe it's pretty minor. 

Your choice :)

src/VariableInput/EditorUtilities.js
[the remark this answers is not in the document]
nitayneeman: I would like to avoid rendering redundant DOM elements, but I don't have an idea for a better solution at the moment. Let's keep it as is for now (we can improve it in a separate PR).

Moreover, we should add a comment in the code that explains why we render them.

src/VariableInput/test/storySettings.js
```diff
@@ -0,0 +1,7 @@
+import { Category } from '../../../stories/storiesHierarchy';
+
+export const storySettings = {
+  category: Category.COMPONENTS,
```
nitayneeman: Until we'll have the UX review, let's mark as WIP:
```suggestion
  category: Category.WIP,
```

Issue #4707, opened 2019-12-15
<Accordion> - Doesn't collapse when clicking on an item


b3llash:
# 🐛 Bug Report

### 🏗 Relevant Components
`<Accordion>`

### 😯 Current Behavior
When collapsed, clicking anywhere on the item makes it expand.
When expanded, the item itself is inactive and the user must click on the icon to collapse.
[GIF](https://i.gyazo.com/8f45bbbbbc255199360916236949172c.gif)

### 🤔 Expected Behavior
Clicking anywhere on the item should expand/collapse.

  - **WSR version**: `7.22.0`

### 👀 Severity
Major

@tzabarc @BnayaZil

argshook:
Hello, @b3llash this is actually intended behaviour according to spec.

![2019-12-16-104436_screenshot](https://user-images.githubusercontent.com/4284659/70892039-10f08900-1ff1-11ea-94c2-dd0a61cb44dd.png)

but perhaps it's not a good behaviour and should be revisited. @LittleTal please advise

LittleTal:
Updated Zeplin spec
Open item should have Shadow30
and there is a divider at the bottom of the item's content area 

![image](https://user-images.githubusercontent.com/23282301/71369662-2139d280-25b4-11ea-9e64-8a16957b29e2.png)

Issue #4715, opened 2019-12-17
<Accordion /> missing divider 


LittleTal:
# 🐛 Bug Report

### 🏗 Relevant Components
`<Accordion />`

...

### 😯 Current Behavior
After selecting an item title, there is no divider between the content and the next item title.
Instead, the divider appears only at the bottom of the open item's title

![image](https://user-images.githubusercontent.com/23282301/70976679-10212b00-20b5-11ea-9677-9f40c5306d44.png)

While hovering the second item the bg from the first item is merged with the hover of the next item : 

![image](https://user-images.githubusercontent.com/23282301/70976722-24652800-20b5-11ea-802d-2dbcee878ee0.png)

...

### 🤔 Expected Behavior
Divider should appear also at the end of the content area when item is open:

![image](https://user-images.githubusercontent.com/23282301/71076015-b04a8300-218d-11ea-9188-dac7bd5b62eb.png)

On hover 

![image](https://user-images.githubusercontent.com/23282301/71076084-ca846100-218d-11ea-96de-1b2c69610e0f.png)

...

### 👣 Steps to Reproduce

  - **WSR version**: `...`
  - **Yoshi version**: `...`

### 👀 Severity

Low

milkyfruit:
other related issue: https://github.com/wix/wix-style-react/issues/4721

PR #4793, opened 2020-01-05
<Accordion/> - change expand behavior to happen when title is clicked

bamai:
### 🔦 Summary
Fixes issues https://github.com/wix/wix-style-react/issues/4707 & #4715.
Added a divider and shadow to an expanded object, and changed the item to be clickable when expanded.
...

### ✅ Checklist

- ✅  👨‍💻 API change is approved by the UI Infra Developers  @jonathanadler 
- [✅] 👨‍🎨 UX change is approved by the Design System UX 
- 📚 Change is documented
  - [✅] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [✅] Visual test

### DIFF
--- a/src/Accordion/Accordion.spec.js
+++ b/src/Accordion/Accordion.spec.js
@@ -6,7 +6,6 @@ import { accordionPrivateDriverFactory } from './Accordion.private.uni.driver';
 
 import { eventually } from '../../test/utils/unit/eventually';
 import { buttonTypes } from './constants';
-import { rejects } from 'assert';
 
 describe('Accordion', () => {
   const FakeIcon = () => <div>fake icon</div>;
--- a/src/Accordion/Accordion.spec.js
+++ b/src/Accordion/Accordion.spec.js
@@ -109,7 +109,7 @@ describe('Accordion', () => {
       await driver.clickHeaderAt(0);
       expect(await driver.isItemExpandedAt(0)).toBe(true);
       await driver.clickHeaderAt(0);
-      return eventually(
+      await eventually(
         async () => {
           expect(await driver.isItemExpandedAt(0)).toBe(false);
         },
--- a/src/Accordion/Accordion.spec.js
+++ b/src/Accordion/Accordion.spec.js
@@ -109,8 +109,12 @@ describe('Accordion', () => {
       await driver.clickHeaderAt(0);
       expect(await driver.isItemExpandedAt(0)).toBe(true);
       await driver.clickHeaderAt(0);
-      await new Promise(resolve => setTimeout(resolve, 350));
-      expect(await driver.isItemExpandedAt(0)).toBe(false);
+      return eventually(
+        async () => {
+          expect(await driver.isItemExpandedAt(0)).toBe(false);
+        },
+        { timeout: 350 },
+      );
     });
 
     it('should accept an expand and collapse button labels', async () => {
--- a/src/Accordion/Accordion.spec.js
+++ b/src/Accordion/Accordion.spec.js
@@ -100,15 +100,15 @@ describe('Accordion', () => {
 
     it('should expand an item on click', async () => {
       const driver = createDriver(<Accordion items={singleItem} />);
-      await driver.clickHeaderAt(0);
+      await driver.clickItemAt(0);
       expect(await driver.isItemExpandedAt(0)).toBe(true);
     });
 
     it('should collapse an expanded item on click', async () => {
       const driver = createDriver(<Accordion items={singleItem} />);
-      await driver.clickHeaderAt(0);
+      await driver.clickItemAt(0);
       expect(await driver.isItemExpandedAt(0)).toBe(true);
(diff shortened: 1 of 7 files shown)


src/Accordion/Accordion.spec.js
```diff
@@ -6,6 +6,7 @@ import { accordionPrivateDriverFactory } from './Accordion.private.uni.driver';
 
 import { eventually } from '../../test/utils/unit/eventually';
 import { buttonTypes } from './constants';
+import { rejects } from 'assert';
```
jonathanadler: I think this line was left here by accident (probably when I wrote the promise examples)

src/Accordion/Accordion.spec.js
```diff
@@ -100,8 +101,17 @@ describe('Accordion', () => {
...
+      await driver.clickHeaderAt(0);
+      expect(await driver.isItemExpandedAt(0)).toBe(true);
+    });
+
+    it('should collapse an expanded item on click', async () => {
+      const driver = createDriver(<Accordion items={singleItem} />);
+      await driver.clickHeaderAt(0);
       expect(await driver.isItemExpandedAt(0)).toBe(true);
+      await driver.clickHeaderAt(0);
+      await new Promise(resolve => setTimeout(resolve, 350));
```
jonathanadler: Before we merge this PR, I would ask someone to take a look at this line, and tell us what they think.
Maybe it will be better to add an `onClose` method nevertheless.

jonathanadler: @TalZa1810 what do you think?

TalZa1810: Maybe using `wix-eventually` would be a nicer way to test this instead of `setTimeOut`

src/Accordion/Accordion.uni.driver.js
```diff
@@ -25,6 +25,11 @@ const accordionDriverFactory = base => ({
       .$('[data-hook="toggle-accordion-wrapper"]')
       .click(),
 
+  clickHeaderAt: idx =>
+    getItemAt(idx, base)
+      .$('[data-hook="header"]')
```
TalZa1810: Adding the dataHook to the `constants` file and using in both driver & component.

src/Accordion/AccordionItem/AccordionItem.js
```diff
@@ -103,49 +102,53 @@ class AccordionItem extends React.PureComponent {
...
     return (
-      <div
-        data-hook={dataHooks.item}
-        onMouseEnter={this._onMouseEnter}
-        onMouseLeave={this._onMouseLeave}
-        onClick={!open && !disabled ? onToggle : null}
-      >
+      <div className={style.itemWrapper}>
         <div
+          data-hook={dataHooks.item}
```
TalZa1810: I see that there's already a constants file, so just adding all the data hooks to the `dataHooks`object.

src/Accordion/AccordionItem/AccordionItem.js
```diff
@@ -103,49 +102,53 @@ class AccordionItem extends React.PureComponent {
     const { icon, title, open, children, onToggle, disabled } = this.props;
 
     return (
-      <div
-        data-hook={dataHooks.item}
-        onMouseEnter={this._onMouseEnter}
-        onMouseLeave={this._onMouseLeave}
-        onClick={!open && !disabled ? onToggle : null}
-      >
+      <div className={style.itemWrapper}>
```
TalZa1810: We usually add the stylable in the root div so the styles should be moved from line 110 here and align the `st.css` accordingly.

src/Accordion/Accordion.spec.js
```diff
@@ -100,10 +100,23 @@ describe('Accordion', () => {
...
+      await driver.clickHeaderAt(0);
       expect(await driver.isItemExpandedAt(0)).toBe(true);
     });
 
+    it('should collapse an expanded item on click', async () => {
+      const driver = createDriver(<Accordion items={singleItem} />);
+      await driver.clickHeaderAt(0);
+      expect(await driver.isItemExpandedAt(0)).toBe(true);
+      await driver.clickHeaderAt(0);
+      return eventually(
```
TalZa1810: Instead of `return eventually...` use `await eventually...`

You can have a look in another test using eventually as a reference:
https://github.com/wix/wix-style-react/blob/master/src/Accordion/Accordion.spec.js#L130

src/Accordion/Accordion.uni.driver.js
```diff
@@ -25,6 +25,11 @@ const accordionDriverFactory = base => ({
       .$('[data-hook="toggle-accordion-wrapper"]')
       .click(),
 
+  clickHeaderAt: idx =>
```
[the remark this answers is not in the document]
jonathanadler: Yeah, I agree.
We should probably go with item, since header is part of the implementation itself.

PR #4800, opened 2020-01-06
VariableInput - disabled prop

tomershk:
### 🔦 Summary

VariableInput - disabled prop

### ✅ Checklist

- 📚 Change is documented
  - [x] Story
  - [x] API description
- 🔬 Change is tested
  - [x] Component
  - [x] Visual test


src/VariableInput/VariableInput.uni.driver.js
```diff
@@ -6,6 +6,8 @@ export const getContent = base => base.$('.public-DraftEditor-content');
 export default (base, body) => {
   return {
     ...baseUniDriverFactory(base, body),
+    isDisabled: async () =>
+      Boolean(await getContent(base).attr('contenteditable')),
```
nitayneeman: I think it should be:
```suggestion
      !Boolean(await getContent(base).attr('contenteditable')),
```

Because when you've `contenteditable` it means the content isn't readonly, right?

src/VariableInput/docs/index.story.js
```diff
@@ -92,6 +92,22 @@ export default {
             />`,
             }),
           ]),
+          columns([
+            description({
+              title: 'Disabled',
+            }),
```
nitayneeman: Please add some description about happens in this example (using `text`)

src/VariableInput/VariableInput.uni.driver.js
tomershk: Actually, I took this example from RichTextInputArea, we always got `contenteditable` attribute, when disabled it is set to `false`,  but it is not clear.
Will push a fix, because: `Boolean('false') === true`

src/VariableInput/VariableInput.st.css
```diff
@@ -1,14 +1,15 @@
 
 :import {
     -st-from: "../Foundation/stylable/colors.st.css";
-    -st-named: D80, B30, B10, F00, B50
+    -st-named: D80, B30, B10, F00, B50, D10-30
```
nitayneeman: This import is missing:
```suggestion
    -st-named: D80, B30, B10, F00, B50, D10-30, D60
```

nitayneeman:
It seems the tag styles aren't aligned with the spec. 

It should be:
image

The solution is to passing the `disabled` prop to Tag when VariableInput is disabled.

tomershk:
> It seems the tag styles aren't aligned with the spec.
> 
> It should be:
> image
> 
> The solution is to passing the `disabled` prop to Tag when VariableInput is disabled.

Looks like when i pass `disabled` together with `theme=dark` it shows wrong:
image

Should i remove the theme tag while adding the disabled tag?

nitayneeman:
> > It seems the tag styles aren't aligned with the spec.
> > It should be:
> > image
> > The solution is to passing the `disabled` prop to Tag when VariableInput is disabled.
> 
> Looks like when i pass `disabled` together with `theme=dark` it shows wrong:
> image
> 
> Should i remove the theme tag while adding the disabled tag?

Both `disabled` and `theme` props of Tag should be conditioned by the `disabled` prop of the VariableInput, so basically yes.

We're doing something similar here (not exactly the same, but the props are conditioned by props from the upper level):
https://github.com/wix/wix-style-react/blob/master/src/SidebarSectionItem/SidebarSectionItem.js#L89

nitayneeman:
We had a few conflicts, I fixed them.

Waiting for the CI.

nitayneeman:
We need to fix the conflicts

tomershk:
Conflicts resolved

PR #4801, opened 2020-01-06
VariableInput - onChange prop

tomershk:
### 🔦 Summary

VariableInput - onChange prop

### ✅ Checklist

- 📚 Change is documented
  - [x] API description
- 🔬 Change is tested
  - [x] Component


src/VariableInput/VariableInput.uni.driver.js
```diff
@@ -12,7 +12,7 @@ export default (base, body) => {
 
       // TODO: implement for puppeteer. Throw error if type is not handled
       if (base.type === 'react') {
-        ReactBase(base).enterText({ data: text });
+        Simulate.beforeInput(contentElement, { data: text });
```
nitayneeman: Why not using ReactBase? 

We can pass the `contentElement` if needed

tomershk: couldn't manage to make it work as expected
Tried: `ReactBase(contentElement).beforeInput({ data: text });`
But test are failing, as method not being called, any suggestions?

nitayneeman: I think you want to pass the Unidriver content element, right?

Something like:
```suggestion
        ReactBase(base.$('.public-DraftEditor-content')).beforeInput({ data: text });
```

Or even better:
```suggestion
        ReactBase(getContent(base)).beforeInput({ data: text });
```

tomershk: Added a TODO for this, as discussed

Issue #4436, opened 2019-11-01
Does default "textAlign" value for tooltip should be "start"?


rudnitskih:
# ✨ Feature Request

### 🔦 Scope

I am from Premium Domains team and in all our mocks all tooltips have content alignment to the left. Developers sometimes forget to set `textAlign="start` in Tooltip and our designer is upset. If `textAlign="start` is default behaviour in our design system, maybe we should change it in WSR component. I understand this is breaking change, but WDYT?

Some examples with our designs:
- https://share.goabstract.com/f12458f1-1418-41a0-b6bd-4b842050b12e
- https://app.goabstract.com/projects/05528230-bb19-11e8-9f9f-c92dc976b27b/branches/master/collections/c51e88bc-42f4-4e7c-baa4-97a70d41eaae

### 👀 Severity
- Low

mykas:
Yes, this was intentional and defined by Design Systems. @milkyfruit please comment on this on Monday

milkyfruit:
@mykas by design default text alignment should be left (if LTR). I agree with @rudnitskih, I think we should change `textAlign="start` as a default value.

Zeplin specs don't indicate text alignment, however two line examples show it as left: https://zpl.io/boDpzxW

Issue #4437, opened 2019-11-01
Allow change textAlign for status message 


rudnitskih:
# ✨ Feature Request

### 🔦 Scope

`<Input/>` doesn't allow control properties for `statusMessage` (`<Tooltip />`. 

### 💁 Explanation

Our design: 
![image](https://user-images.githubusercontent.com/8848858/68019658-15f9b300-fca5-11e9-9f41-267e085882c8.png)

We don't know how to achieve this design with current `<Input />` implementation.

### 🐾 Possible solution 

1. Change our design with alignment to center
2. Allow to control `<Tooltip />` props.
3. Fixing #4436 will automatically fix this issue

### 👀 Severity

- Low

rudnitskih:
@mykas I think this is a bug according to design system - https://app.zeplin.io/project/5c8f6f1f9f489505cac756dc/screen/5c90f6ab3754f40dd0464ff2

rudnitskih:
Can we fix it in 7.x.x branch or we should for WSR 8?

mykas:
It is a bug! I will do a pr to fix this.

rudnitskih:
Any updates?

nitayneeman:
@mykas I just wonder how did you planned to fix this issue because there are two options:

1. Exposing a prop that allows customizing the `statusMessage` tooltip
2. Changing the default value of `textAlign` for the tooltip to `start` (as described here -> https://github.com/wix/wix-style-react/issues/4436)

I really prefer the second option, but it's sort of a visual breaking change (or might it still look centered for regular text length and affects only a very short text?).

mykas:
Hey @nitayneeman so your concern make sense, but I've talked with Domas and he said that mostly it should be with `start` position at all times so it make sense to align everybody because it is kind of how design systems wants it to be.

nitayneeman:
Fixed as part of https://github.com/wix/wix-style-react/pull/4813, closing.

rudnitskih:
Thanks guys!

PR #4813, opened 2020-01-07
<Tooltip/> - change `textAlign` default value

nitayneeman:
This PR fixes https://github.com/wix/wix-style-react/issues/4436 by setting the `textAlign` prop to be `start` defaultly.

According to the design system, the text is supposed to be aligned to the start by default (and currently it's centered).

Discussion:
https://github.com/wix/wix-style-react/issues/4437#issuecomment-571504861


src/Tooltip/TooltipNext/Tooltip.js
```diff
@@ -73,7 +71,7 @@ class Tooltip extends React.PureComponent {
     maxWidth: 204,
     flip: true,
     fixed: false,
-    textAlign: 'center',
+    textAlign: 'start',
```
simast: I think `start` value is not supported in Edge right now: https://caniuse.com/#feat=mdn-css_properties_text-align_flow_relative_values_start_and_end

nitayneeman: Good point.

According to MDN - if `start` isn't supported by the browser, the value will act as left if the direction is `ltr` and right for `rtl`.:
image

So I think this case will bring the same result, which is good :)

simast: ah, seems to be good then

PR #4820, opened 2020-01-07
<RichTextInputArea/> - Expose plain text value with onChange event

### LINKED ISSUE #4543: <RichTextArea/> - improve usage of char count
# 🐛 Bug Report

### 🏗 Relevant Components
`<RichTextArea/>`
...

### 😯 Current Behavior
Char limit in `<RichTextArea/>` starts with 89 characters instead of 100
image
...

### 🤔 Expected Behavior
image

...

### 👣 Steps to Reproduce

  - **WSR version**: `...`
  - **Yoshi version**: `...`

### 👀 Severity

Low

simast:
Closes #4543

Added another parameter to change event: `onChange(htmlText, { plainText })`. The `plainText` value can be used to easily implement character count functionality for example.

```jsx
<FormField label="Rich Text Area label">
  {({ setCharactersLeft }) => (
    <RichTextInputArea
      onChange={(htmlText, { plainText }) => setCharactersLeft(100 - plainText.length)}
    />
  )}
</FormField>
```


nitayneeman:
I wonder if we want to add an example for composition inside the FormField story page:
https://wix-style-react.now.sh/?path=/story/components-api-components--formfield

simast:
> I wonder if we want to add an example for composition inside the FormField story page:

I feel like this should be done not in a component story but in a separate design guidelines story for Rich Input. Note how we have "3.1. Text Input" and "3.2 Text Area" stories - they already demonstrate composite use of FormField and input controls with working character count implemented. I think we just need something similar for Rich Input. I believe this is what @LittleTal  was working on and where the issue comes from originally.

LittleTal:
@simast `<RichTextInputArea/>` story is still WIP by me, once I get to the task I'll update the example as you suggested.

PR #4824, opened 2020-01-07
fix(Notification.ActionButton): remove default onClick

### LINKED ISSUE #3889: <Notification.ActionButton> is broken with defualt onClick
# 🐛 Bug Report

### 🏗 Relevant Components

 <Notification.ActionButton>

...

### 😯 Current Behavior

if we pass
```
 <Notification.ActionButton
            target={'_blank'}
            link={someLinl}
            type="textLink">
            {someText}
 </Notification.ActionButton>
```
then the default on click prevents the link from working

### 🤔 Expected Behavior

the default onClick should return true, or be removed altogether so the link can work

...

### 👣 Steps to Reproduce

just render the component

  - **WSR version**: latest
  - **Yoshi version**: latest

### 👀 Severity

 Low

argshook:
closes #3889

also add `type="button"` for `<button>`

this is to prevent accidental form submit,
because default `button.type` in HTML DOM is `submit`


argshook:
hiccup on applitools, otherwise ok to merge

PR #4825, opened 2020-01-07
Variable input - rows & size

tomershk:
### 🔦 Summary

VariableInput - rows and size props

### ✅ Checklist

- 📚 Change is documented
  - [x] Story
  - [x] API description
- 🔬 Change is tested
  - [x] Component
  - [x] Visual test


src/VariableInput/test/VariableInput.visual.js
```diff
@@ -22,6 +23,34 @@ const tests = [
...
+          size: sizeTypes.medium,
+        },
+      },
+      {
+        it: 'Large size',
+        props: {
+          initialValue: 'Welcome to my {{page.name}} ',
+          size: sizeTypes.large,
+        },
+      },
```
nitayneeman: I think we can combine these specs into one screenshot (we do that to save some resources in the CI).

We don't have a well-defined convention, but here are some examples of ways to do that:
https://github.com/wix/wix-style-react/blob/master/src/Button/test/Button.visual.js#L32
https://github.com/wix/wix-style-react/blob/master/src/Tag/tests/Tag.visual.js#L116
https://github.com/wix/wix-style-react/blob/master/src/Breadcrumbs/test/Breadcrumbs.visual.js#L52

src/VariableInput/test/VariableInput.visual.js
```diff
@@ -22,6 +23,34 @@ const tests = [
...
+      {
+        it: '3 Row',
```
nitayneeman: Because the height calculation depends both on `rows` and `size` I would test a few cases:
1. Just X rows (medium is the default) - basically what you already did
2. Small size & X rows 
3. Large size & X rows 

X could be 3, doesn't really matter, but what that does matter - is to combine them into one screenshot (as done for `size`)

src/VariableInput/VariableInput.st.css
```diff
@@ -24,10 +34,9 @@
...
 .root :global(.public-DraftEditor-content) {
-  -st-mixin: sizeMedium, weightNormal;
   background-color: value(D80);;
-  min-height: 36px;
-  padding: 2px 7px;
+  height: calc(var(--rows) * var(--rowHeight) + 12px);
+  padding: 5px 7px;
   border: 1px solid value(B30);
   border-radius: 6px;
   overflow: auto;
```
nitayneeman: It seems like we get an overflow in small & medium sizes:
image

src/VariableInput/index.d.ts
```diff
@@ -6,6 +6,8 @@ export interface VariableInputProps {
   dataHook?: string;
   disabled?: boolean,
   initialValue?: string,
+  rows?: number,
+  size?: string,
```
nitayneeman: I would create a type for `size` with the possible values

src/VariableInput/test/VariableInput.private.uni.driver.js
```diff
@@ -1,11 +1,19 @@
-import { variableInputDriverFactory as publicDriverFactory } from '../VariableInput.uni.driver';
+import publicDriverFactory from '../VariableInput.uni.driver';
+import { tagUniDriverFactory } from '../../Tag/Tag.uni.driver';
+import { dataHooks } from '../constants';
 
 export const getPlaceholder = base =>
   base.$('.public-DraftEditorPlaceholder-root');
+// const textTestkitFactory = testkitFactoryCreator(textDriverFactory);
```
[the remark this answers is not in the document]
tomershk: Yes, I removed it while resolving the conflicts

src/VariableInput/VariableInput.uni.driver.js
```diff
@@ -12,7 +12,7 @@ export default (base, body) => {
```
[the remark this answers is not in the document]
tomershk: Not sure, because once you set them when you initial the component, they will not change.

src/VariableInput/VariableInput.st.css
```diff
@@ -24,10 +34,9 @@
   left: 10px;
 }
 .root :global(.public-DraftEditor-content) {
-  -st-mixin: sizeMedium, weightNormal;
   background-color: value(D80);;
-  min-height: 36px;
-  padding: 2px 7px;
+  height: calc(var(--rows) * var(--rowHeight) + 12px);
```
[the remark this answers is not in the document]
tomershk: Yes, no auto grow

src/VariableInput/test/VariableInput.visual.js
tomershk: Actually from what i understand, currently it is all in one screenshot, and by doing that i will create additional screenshot correct? 
(Each describe is a new screenshot?)

nitayneeman: Each `it` creates a different story page, which means a different screenshot:
https://github.com/wix/wix-style-react/blob/6b3507204a/src/VariableInput/test/VariableInput.visual.js#L59

You can run the visual test by `npm run storybook:visual` and then navigate to `http://localhost:6007`

If you want to combine them into one screenshot, the line I attached above should be smarter (similar to the examples I sent you before).

PR #4844, opened 2020-01-09
<VariableInput /> - UX page

tomershk:
### 🔦 Summary

VariableInput  adding an UX page

### ✅ Checklist

- 📚 Change is documented
  - [x] Other (explain) - UX page


nitayneeman:
Waiting for @ayeletr to do the UX review.

tomershk:
Great, one small thing, thinking about making the `+ Insert Variable` functional, might make the example code a bit too big.

nitayneeman:
> Great, one small thing, thinking about making the `+ Insert Variable` functional, might make the example code a bit too big.

I would make only the "Action" example to work (and also mention that inside the relevant description).

ayeletr:
Hey,
Here are my UX review comments.
Please let me know if something isn't clear.
Ping me when you're done so I can take another look :)

-1 Description: Variable Input enables the user to insert given tags (variables) as well as free text inside an input. This component is used to build various forms.
Variable Input is a composition of 2 individual components – <FormField/> and <InputArea/>. 

-2 The input should expand vertically when the content gets longer, just like Text Aria behaves. As described in the spec - should use Text Aria component for for that to happen (not input).

-3 The “Insert Variable” should open a do with a list. Once clicking on an item from the list, it should then be added to the field.

-4 Variable Input button space from field should be 9 px.

-5 The ‘Insert Variable’ button works only on one filed in the examples. Can we make it work on all of them.

-6 Large size Variable Input should be 42px height and use medium size tags.

-7 When focused, at the beginning of the input the cursor doesn’t have the correct padding for left, and is not vertically centered in the input.

-8 When the tag appears there’s extra space on it’s right for some reason.

-9 Sometimes the cursor disappears:
https://www.screencast.com/t/CotTVhvyGc
(Here I click the right arrow)

stories/VariableInput/examples.js
```diff
@@ -0,0 +1,116 @@
...
+  { id: 2, value: 'Third Option' },
+  { id: 3, value: 'Fourth Option' },
+]}
+onSelect={(selectedOption) => {
+  this._ref.insertVariable(selectedOption.value);
+}}
+>
+{({ open, close }) => {
+  return (
+    <TextButton onMouseEnter={open} onMouseLeave={close} size="tiny" prefixIcon={<Icons.Add />}>Insert Variable</TextButton>
```
nitayneeman: I think the dropdown is supposed to be toggled `onClick` (not `onMouseEnter` & `onMouseLeave`)

stories/VariableInput/index.story.js
```diff
@@ -0,0 +1,118 @@
...
+      ),
+
+      issueUrl: 'https://github.com/wix/wix-style-react/issues/new',
+    }),
+
+    columns([
+      description({
+        title: 'Description',
+        text: `Variable Input enables the user to insert given tags (variables) as well as free text inside an input. This component can be used to build various forms.
+        Variable Input is a composition of 3 individual components: FormField, VariableInput`,
```
nitayneeman: What did we decide regarding the description?

tomershk: Will add the sentence:

> it might include Dropdown to insert new variables

stories/VariableInput/examples.js
tomershk: Will change to `onClick`
(Took the example of `Uncontrolled example with an icon` from here: https://wix-style-react.now.sh/?path=/story/components-api-components--dropdownbase)

tomershk:
- Added spacing when we got multiple lines
- Replace DropDownBase with PopoverMenu
@nitayneeman @ayeletr

src/VariableInput/VariableInput.st.css
```diff
@@ -80,19 +78,17 @@
...
   display: inline;
-  vertical-align: middle;
-}
-.root:size(large) .textWrapper {
-  line-height: 30px;
 }
 .root:size(small) .textWrapper {
   -st-mixin: text-small-normal;
 }
+.root :global(.public-DraftEditor-content > div) {
```
nitayneeman: I would do the following instead of relying on an internal direct child:
```suggestion
.root :global(.public-DraftStyleDefault-block) {
```

stories/Introduction/Cheatsheet/componentsFamilies/InputFamily.js
```diff
@@ -40,6 +40,9 @@ import GoogleAddressInput from 'wix-style-react/GoogleAddressInput';
 import Search from 'wix-style-react/Search';
 import ImageViewer from 'wix-style-react/ImageViewer';
 import { Category } from '../../../storiesHierarchy';
+import VariableInput from 'wix-style-react/VariableInput';
+import { DropdownBase } from '../../../../src';
```
nitayneeman: We need to change the PopoverMenu here (and use below)

tomershk: Done

Issue #4835, opened 2020-01-08
<Dropdown/> | <DropdownLayout/> - divider item is clickable.


etaybarzilay89:
# 🐛 Bug Report

### 🏗 Relevant Components
<Dropdown/>
<DropdownLayout/>

...

### 😯 Current Behavior

1. When using  listItemSectionBuilder to build an option of Dropdown, and this options is of type "divider", the option is clickable (i.e. you can click on the divider itself).
2. When using "isOptionADivider"  from the testkit, it returns false for the divider item.
3. missing types for ListItemSectionBuilder

...

### 🤔 Expected Behavior

1. divider item should not be clickable.
2. isOptionADivider should return true for divider item.
3. should have types.

...

### 👣 Steps to Reproduce

code snippet used in WSR playground:

`class MyComponent extends React.Component {
  render() {
    return (
      <Container>
        <Dropdown options={[
  {
     id: 1,
     value: <span>hello</span>
  },
  listItemSectionBuilder({ type: 'divider' }),
  {
     id: 2,
     value: <span>hello</span>
  },
]}/>
      </Container>
    );
  }
}`

  - **WSR version**: `7.30.0`
  - **Yoshi version**: `4.27.0`

### 👀 Severity

Major - We can't use a divider in the dropdown with this behavior.

EgoziE:
Hey Etay
Thanks for opening this issue

Point 1 - this PR https://github.com/wix/wix-style-react/pull/4848
Point 2 is a bit more complicated, it will be resolved when we finish with https://github.com/wix/wix-style-react/issues/4656
Point 3 - I'll do it in another PR next week

PR #4848, opened 2020-01-09
<ListItemSection/> - Disable click propogation on divider ans whitespace

EgoziE:
`<ListItemSection/>` - Disable click propogation on divider ans whitespace

Resolves https://github.com/wix/wix-style-react/issues/4835


src/ListItemSection/ListItemSection.st.css
```diff
@@ -4,7 +4,7 @@
 }
 
 .root {
-  -st-states: subheader;
+  -st-states: subheader, divider, whitespace;
```
TalZa1810: if the type can be one of the following: [divider , whitespace] , I think it can be an enum:
```
-st-states: type(enum( divider , whitespace ))
```

EgoziE: Thanks - I'll do it in a second PR :)

PR #4864, opened 2020-01-13
VariableInput - multilne prop

tomershk:
### 🔦 Summary

VariableInput - multiline prop:
Allow to use one line with horizontal scrolling, like we got in Input component

### ✅ Checklist

- 📚 Change is documented
  - [x] Story
  - [x] API description
- 🔬 Change is tested
  - [x] Component


src/VariableInput/VariableInput.js
```diff
@@ -26,6 +26,8 @@ class VariableInput extends React.PureComponent {
     disabled: bool,
     /** Initial value to display in the editor */
     initialValue: string,
+    /** When set to true, component will allow multiple lines, otherwise will scroll horizontaly and ignore return key*/
+    multiline: bool,
```
nitayneeman: I wonder why do we actually need a completely different prop - because it's possible to indicate this situation when `rows` is just been set to 1...

The problem with `multiline` is that now we can create weird cases when combining that with the `rows` prop.

For example:
```jsx
    <VariableInput
      rows={5}
      multiline={false}
    />
```

nitayneeman: Well, the native TextArea behaves like multiline even when it's set to 1, so maybe it makes sense to distinguish between TextArea behavior to a simple Input using a dedicated prop, assuming the component's supposed to follow TextArea (or InputArea more specifically) in case of multiline.

src/VariableInput/VariableInput.js
```diff
@@ -129,11 +142,20 @@ class VariableInput extends React.PureComponent {
           placeholder={placeholder}
           onBlur={() => setTimeout(this._onBlur, 0)}
           readOnly={disabled}
+          {...oneLineProps}
```
nitayneeman: What do you think about:
```suggestion
          {!multiline && ...singleLineProps}
```

When:
```js
   const singleLineProps = {
          handlePastedText: this._handlePastedText,
          handleReturn: () => 'handled',
    };
```

I think it's a little more readable

src/VariableInput/VariableInput.js
```diff
@@ -129,11 +142,20 @@ class VariableInput extends React.PureComponent {
...
         />
         {this._renderIndicator(status, statusMessage)}
       </div>
     );
   }
+  _handlePastedText = (text, html, editorState) => {
+    if (/\r|\n/.exec(text)) {
+      text = text.replace(/(\r\n|\n|\r)/gm, '');
```
nitayneeman: Please explain with a comment above why we need to do that

nitayneeman:
I would add a visual test for the scenario that `multiline` is false

src/VariableInput/index.d.ts
```diff
@@ -6,6 +6,7 @@ export interface VariableInputProps {
   dataHook?: string;
   disabled?: boolean,
   initialValue?: string,
+  multiline?: bool,
```
nitayneeman: Should be:
```suggestion
  multiline?: boolean,
```

PR #4867, opened 2020-01-13
Cheat sheet readme/file structure

tomershk:
### 🔦 Summary
While adding `VariableInput` to the Cheetsheat I was missing a summery for files needed to be change, this is my suggestion.


TalZa1810:
Hi @tomershk 
Thank you so much for the feedback! :)

PR #4878, opened 2020-01-15
<MessageBoxMarketerialLayout/> - add width prop

Jefoss:


src/MessageBox/MarketerialLayout/MessageBoxMarketerialLayout.scss
```diff
@@ -66,14 +65,12 @@ $headerImageHeight: 185px;
 }
 
 .title {
-  width: 384px;
```
simast: This width value seems to have provided some paddings before. The content and title is currently too close to modal borders. Let's maybe remove this fixed width and replace with fixed left/right paddings?

PR #4879, opened 2020-01-15
test(GeneratedTestComponent): add visual test

argshook:
This PR adds generated test component before running visual tests.

Idea is that we want to test also generated component, in order to make sure it's actually generated.

It's merely a naive smoke test.


package.json
```diff
@@ -47,6 +48,7 @@
     "test:e2e": "npm run test:e2e-runtime && build-storybook -c ./.storybook/e2e && npm run test:e2e-only",
     "test:e2e-only": "NODE_ENV=production yoshi test --protractor",
     "test:e2e-runtime": "cd ./test-runtime/imports && NODE_ENV=production npm run test",
+    "pretest:visual": "npm run generate-test-component",
```
nitayneeman: So this means that each time we run `test:visual`, whether locally or by the CI, would be side-effect files?

Maybe it would be better to differentiate to something like?
```suggestion
    "test:generator": "npm run generate-test-component",
```

argshook: my idea is:
1. run visual test on CI
2. generated component is not part of release
3. locally no generated files should be visible

current state of PR supports 1 & 2

3 is only true when running `npm run test:visual` locally, which to my mind should be a rare case. To debug visual tests users should do `npm run storybook:visual`.

if one would need to do `npm run test:visual` locally, it means there's something serious to debug, in which case a `pretest:visual` shouldn't be a surprise. If you debug something serious, you probably check what npm scripts are happening (im assuming tho)


---

your suggested separate CI build would work, but it would only test build phase of generated component. It would test that running CLI throws no errors but it wouldn't test that component is actually generated and well integrated (all the templates, codemods) are ok.

It's not a must for test to be visual. I just think visual test covers most corners with least effort. wdyt?

nitayneeman: I think it's worth adding the visual test mainly because that's very easy to accomplish. My only concern was that the generated files might be a little annoying when you debug with the remote, but maybe you're right it doesn't happen so often.

I say we will give it a try and get some feedback

argshook:
> this would potentially lead into committing files which are not needed
> - I prefer the generator to have a sandbox mode that generates the
> file to a temp folder which can be deleted or ignored

generator is not only for creating new files, it also runs codemods to
adjust existing ones.

A sandbox would mean a imitating whole wix-style-react (smaller
footprint but still with all the files that codemods run on).

too complicated for future maintenance imo. adding/editing some codemod
would require to adjust the sandbox too.

that's whole lot more to worry than a visual test with a slim chance it
might get committed.

slim chance, because:

* contributor would run `npm run test:visual` (instead of `npm run storybook:visual`)
* contributor would commit files without looking
* no one reviews PR and it gets merged into master

i'd say this scenario is rare enough

shlomitc:
After some thought, I'm closing this one.
I think that it's ok for now to manually verify it. The generator is actively used and the responsibility of each one that changes it to check it.
I prefer this over more complexity of our build

PR #4882, opened 2020-01-16
<MessageBoxMarketerialLayout/> - add noBodyPadding prop

Jefoss:


src/MessageBox/MarketerialLayout/MessageBoxMarketerialLayout.js
```diff
@@ -36,6 +37,7 @@ class MessageBoxMarketerialLayout extends WixComponent {
     // instead of introducing a breaking change for padding removal for non buttons existence, we add this prop
     const shouldRemoveButtonsPadding =
       removeButtonsPadding && !primaryButtonLabel && !secondaryButtonLabel;
+    const contentPaddingStyle = noBodyPadding ? {} : { padding: '0 108px' };
```
simast: I would suggest to use classNames (like in `MessageBoxFunctionalLayout`) or change this to:
```js
const contentStyle = noBodyPadding ? { padding: 0; } : undefined;
```
and leave the default `padding: 0 108px;` in stylesheet.

src/MessageBox/MarketerialLayout/MessageBoxMarketerialLayout.js
```diff
@@ -141,12 +147,14 @@ MessageBoxMarketerialLayout.propTypes = {
   primaryButtonTheme: PropTypes.oneOf(['blue', 'purple']),
   removeButtonsPadding: PropTypes.bool,
   width: PropTypes.string,
+  noBodyPadding: PropTypes.bool,
```
simast: could you also add this prop to documentation (as a new table row):
https://github.com/wix/wix-style-react/blob/master/src/MessageBox/README.md

src/MessageBox/MarketerialLayout/MarketrialLayout.visual.js
```diff
@@ -57,6 +57,17 @@ const tests = [
           primaryButtonLabel: 'Button',
         },
       },
+      {
+        it: 'no body padding',
+        props: {
+          title: 'No Body Padding',
```
simast: could you also make this text longer? so we capture with visual test that the paddings are removed for the title as well.

PR #4887, opened 2020-01-19
<FormField/> - add `suffix` prop

nitayneeman:
This PR adds a new prop that allows rendering a custom node at the end of the label row. We need this prop to simplify the VariableInput composition.

Also, the types:
https://github.com/wix/wix-ui/pull/1728


src/FormField/FormField.uni.driver.js
```diff
@@ -48,5 +49,12 @@ export const formFieldUniDriverFactory = (base, body, { dataHook }) => {
       );
       return await testkit.getTooltipText();
     },
+    getSuffix: async () => {
```
nitayneeman: The better way is to do:
```suggestion
    getSuffix: () => base.$(`[data-hook="${dataHooks.suffix}"]`)
```

However, because we support the legacy driver and it returns the native node - the method needs to be compatible and imitate the legacy as long as we run the same tests using `runTests` approach on both drivers.

mosheKerbel:
@nitayneeman plz modify `wsr-types` as well

nitayneeman:
The broken visual is related to the test I added, and the change is OK

PR #4888, opened 2020-01-19
<MultiSelect/> close multi select dropdown when prop is true

ronnyrin:
### 🔦 Summary

Allow closing multi select when prop is true. It's already written in the docs but never worked because it was hard-coded to `false`

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ x] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x ] Component
  - [ ] Visual test


mosheKerbel:
@ronnyrin 
Even though the code makes sense and also your test, it does not work.
Try to test it through the storybook examples of Multiselect, it stays opened.
Need to figure out why.

ronnyrin:
@mosheKerbel it does work on the example of "Tag Input + Suggestions" in the usage tab. In all other cases you can't enter custom text so it behaves the same as before.

EgoziE:
Hey @ronnyrin 
According to UX, any Multiselect should **not** be closed on selection, the problem is in our docs and I will fix it ASAP.
If you need a dropdown that closes in selection you can use `<Dropdown/>` - https://wix-style-react.now.sh/?path=/story/components-api-components--dropdown

Is there any UX need for it, you can discuss it in wsr-ux channel :)

PR #4889, opened 2020-01-19
<MultiSelectCheckbox/> - Upgraded story page + added visuals

EgoziE:
Upgraded story page for `<MultiSelectCheckbox/>`
And added visuals

This is a preparation for replacing list items with `<ListItemSelect/>`

PR #4890, opened 2020-01-19
<MultiSelectCheckbox/> - Replaced list item with <ListItemSelect/>

EgoziE:
Replacing list item with `<ListItemSelect/>`

DO NOT MERGE BEFORE PR - https://github.com/wix/wix-style-react/pull/4889


src/MultiSelectCheckbox/MultiSelectCheckbox.js
```diff
@@ -2,26 +2,37 @@ import React from 'react';
...
+        return {
+          ...option,
+          overrideStyle: true,
+          value: (
+            <ListItemSelect
+              checkbox
+              selected={this.isSelectedId(option.id)}
+              disabled={option.disabled}
+              title={option.value}
+              onClick={e => e.preventDefault()} // This is prevented because there's an event listener wrapping the option
```
shlomitc: where is the event listener coming from? the `DropdownLayout`?

EgoziE: Fixed in checkbox component

EgoziE:
There are few more things missing in this PR:
1. Setting focus to `<ListItemSelect/>` - Use FocusableHOC (not developed)
right now there's a problem that the focus goes to the checbox-box itself
2. Provide a way to use builders instead of the component itself
3. Check a composite value of a node as a list item.

In order to achieve this globally in `<DropdownLayout/>` we should test:
- InputWithOptions
- BadgeSelect
- DropdownPicker
- CalendarPanel
- Dropdown/DropdownBase
- MultiSelect/MultiSelectCheckbox

We need to pay attention to:
- Make all list item components usable here: `<ListItemSelect/>` `<ListItemSection/>` `<ListItemAction/>` and `<ListItemEditable/>` (not yet developed)
- And their builders.
- Focus to relevant components (e.g. not deviders)
- How custom nodes behave with list items

I think there's a week (5-6 days) worth of dedicated work here.
With developing `<ListItemEditable/>` even two weeks.

shlomitc:
@EgoziE - let's try to take it separately from this PR.
There are two topics you mentioned - 1) Bugs in the `ListItemSelect`. 2) Refactoring the DropdownLayout.

The first one should be straightforward and actionable (fix a11y issues), so it can be a separate issue.
The second issue is debatable - we can decide to keep the `DropdownLayout` slim, or we can consider integrating the `ListItemSelect`. In any case, let's start this as a separate discussion.

src/MultiSelectCheckbox/MultiSelectCheckbox.js
```diff
@@ -2,49 +2,39 @@ import React from 'react';
...
+            />
```
EgoziE, line 33: Here add highlighed

TalZa1810: done

src/ListItemSelect/ListItemSelect.js
```diff
@@ -31,7 +31,7 @@ class ListItemSelect extends React.PureComponent {
     prefix: PropTypes.node,
 
     /** Title of the list item */
-    title: PropTypes.string,
+    title: PropTypes.node,
```
EgoziE, line 34: Don't forget the d.ts

src/MultiSelectCheckbox/MultiSelectCheckbox.js
```diff
@@ -2,26 +2,61 @@ import React from 'react';
...
+      deprecationLog(
+        '<MultiSelectCheckbox/> - help and helpMessage props are deprecated. Please use <FormField/> as a wrapper instead.',
+      );
+    }
+
+    if (props.hasOwnProperty('theme') && props.theme !== 'normal') {
+      deprecationLog(
+        '<MultiSelectCheckbox/> - theme prop is deprecated, please contact us or your UX if needed.',
+      );
+    }
```
EgoziE, line 32: This was removed in wsr8

src/ListItemSelect/ListItemSelect.js
TalZa1810: I just rebased the PR. I think that this change I'll make in the ListItemSelect PR.

PR #4896, opened 2020-01-20
Typography - Use helveticaNeue from parastorage in Storybook

EgoziE:
Prepare Storybook (and visual/e2e tests) for the new font.
Replacing the declaration of the old Helvetica Neue font with the new one from wix-fonts.
In the future all it needs is replacing only the link line in `preview-head.html` files and that's it!


EgoziE:
This change is breaking, will change to Madefor in time will be easier.

EgoziE:
Changed the PR to use the old static.parastorage.com/services/third-party/fonts/Helvetica/fontFace.css
Because my initial change was too breaking.
Still, this is a good preparation to use Madefor from wix-fonts

shlomitc:
What was breaking?

EgoziE:
The weights
there are two possible solutions:
1. use the old fontFace.css file
2. change the API in wix-fonts to be specific for wsr

I don't mind changing it, it's gonna be deprecated anyway...
what do you think?

shlomitc:
So you mean the helvetica font css file introduce broken API? I’d make sure that our component support both the wix-fonts new api and the existing parastorage api

EgoziE:
Broken of some sort...
I developed helvetica in the new standard like Madefor for a possible future use.
Now I understand there's no need for it, so I'll change it to be like the old one :)
Thanks

PR #4905, opened 2020-01-21
<Text/> - Fix nested <ul> styles being applied to incorrect elements

simast:


src/Text/Text.js
```diff
@@ -53,6 +53,7 @@ const Text = ({
     skin,
     light,
     weight,
+    withList: true,
```
shlomitc: Why is this state needed?

simast: we reuse text Stylable styles in other components, for example: https://github.com/wix/wix-style-react/blob/master/src/MessageBox/FunctionalLayout/MessageBoxFunctionalLayout.js#L8

This causes new Text ul/li styling to be applied to whatever is nested within `MessageBoxFunctionalLayout` (such as tabs). There are other components that do the same thing (Checkbox, Slider).

PR #4906, opened 2020-01-21
<Cell/> Layout - Fix Firefox bug on grid text overflow

frontned:
https://bugzilla.mozilla.org/show_bug.cgi?id=1528030

### 🔦 Summary
Fix: Firefox bug on grid text overflow
Text overflow breaks Layout Cell.
columns


frontned:
@mosheKerbel it should be fixed now. I replace min-width: 0 with max-width: 100% since here: https://bugzilla.mozilla.org/show_bug.cgi?id=1528030 it says that there are two fixes for this problem either min-width or max-width and since max-width does not break any other layout I think this should be the solution for the problem.

PR #4911, opened 2020-01-22
<Dropdown/> - Improve dropdown list constraints documentation

simast:


simast:
@shlomitc I have split up the examples into a separate "Constraints" section. Also used the compact live code section so I don't think we need to combine both width/height examples into a single one right now.

PR #4917, opened 2020-01-23
Typography - Rename variables

EgoziE:
just a rename


src/Foundation/stylable/typography.st.css
```diff
@@ -31,17 +31,17 @@
 /* Weights */
 .thin {
   font-weight: 400; /* IE11 */
-  font-weight: var(--wsr-font-weight-thin, 400);
+  font-weight: var(--wix-font-weight-thin, 400);
```
shlomitc: are you sure that's correct?
This means that the library's public api is now `--wix-font-weight-thin`.
I think the previous state is ok

PR #4923, opened 2020-01-26
`<Dropdown/>`- adding an example of clear button

TalZa1810:
### 🔦 Summary
Adding an example of clear button

### ✅ Checklist

- 📚 Change is documented
  - [x] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test

### DIFF
--- a/src/Dropdown/docs/examples.js
+++ b/src/Dropdown/docs/examples.js
@@ -209,3 +209,23 @@ export const overflow = `
   />
 </div>
 `;
+
+export const clearValue = `
+  () => {
+  const [val,setVal] = React.useState(-1);
+
+  return <Dropdown
+  updateControlledOnClear
+    clearButton
+    selectedId={val}
+    onClear={()=> setVal(-1)}
+    onSelect={({id}) => setVal(id)}
+    placeholder="Select an option"
+    options={[
+      { id: 0, value: 'Option 1' },
+      { id: 1, value: 'Option 2' },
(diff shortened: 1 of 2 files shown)


shlomitc:
I don’t think it makes sense for a Dropdown to have a clear button. Clear is related to typing usually and not for selection. Is this in the specs? Ux approves?

mosheKerbel:
> I don’t think it makes sense for a Dropdown to have a clear button. Clear is related to typing usually and not for selection. Is this in the specs? Ux approves?

If this is the case, then this is really a high priority issue that we need to solve.
Users get props like this `clearButton` in the API due to inheritance, so we need not to include them in auto-docs.

TalZa1810:
I asked the user for his `spec` to see if this was part of his design. I wait for his reply.

shlomitc:
image
@milkyfruit - does this make sense for our `<Dropdown/>` component? I understand it when we're in "Autocomplete" mode, but don't think it should be for the "select" mode

TalZa1810:
The user said that he just played with the component. It's not something he got from the designer.

shlomitc:
Then let’s close this PR and suggest alternatives if needed

milkyfruit:
there was a similar github issue a while ago for time or date picker...
Nor then, nor now any designer requested for it.
Let's close this issue until we have a real use case.

PR #4925, opened 2020-01-27
`<Sidebar/>`- changing scrollbar colors

TalZa1810:
### 🔦 Summary
Changing Sidebar scrollbar colors for both light and dark skins 

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [x] Visual test


src/Sidebar/test/Sidebar.visual.js
```diff
@@ -106,3 +117,73 @@ storiesOf('Sidebar', module).add('Hidden', () => (
...
+            <Sidebar.Item itemKey={'mailboxes'}>
+              <SidebarSectionItem>Mailboxes</SidebarSectionItem>
+            </Sidebar.Item>
+
+            <SidebarDivider />
+
+            <Sidebar.Item itemKey={'hire-an-expert'}>
+              <SidebarSectionItem>Hire an Expert</SidebarSectionItem>
+            </Sidebar.Item>
+          </Sidebar>
```
nitayneeman: If we're already testing the scrollbar, I would also add a simple persistent footer (not related to your changes but just to improve the domain this test covers)

src/Sidebar/test/Sidebar.visual.js
```diff
@@ -106,3 +117,73 @@ storiesOf('Sidebar', module).add('Hidden', () => (
...
+
+            <Sidebar.Item itemKey={'mailboxes'}>
+              <SidebarSectionItem>Mailboxes</SidebarSectionItem>
+            </Sidebar.Item>
+
+            <SidebarDivider />
+
+            <Sidebar.Item itemKey={'hire-an-expert'}>
+              <SidebarSectionItem>Hire an Expert</SidebarSectionItem>
+            </Sidebar.Item>
```
nitayneeman: Same feedback for the content - we can replace it with an element that higher than 500px

src/Sidebar/test/Sidebar.visual.js
```diff
@@ -106,3 +117,73 @@ storiesOf('Sidebar', module).add('Hidden', () => (
...
+                  <Avatar size="size72" />
+                  <Box direction="vertical" align="center" marginTop={2}>
+                    <Text size="small" secondary weight="bold">
+                      Owner
+                    </Text>
+                    <Text size="tiny" secondary light weight="normal">
+                      owner@something.com
+                    </Text>
+                  </Box>
+                </Box>
```
nitayneeman: I think we can simplify the header by replacing it with a shorter example (`div` with text for example).

The thing is that this test might fail from the wrong reason when SidebarHeader/TextButton/Box/Avatar/Text is broken - generally it's OK but that's not the purpose of **this** test so we would not get any benefit...

TalZa1810:
Ayelet approved this. Waiting for tests to pass to merge.

PR #4930, opened 2020-01-28
named-imports - export missing components

mykas:
This PR is fixing missing named exports after testing the codemode of named-imports on events project.


src/index.d.ts
```diff
@@ -1,7 +1,14 @@
 import * as React from 'react';
 
-export { IconElement, InjectedFocusableProps, Omit } from './common';
 
+import { default as PopoverMenuBeta } from './beta';
+
+// NEXT section to rexport as named import from NEXT namespace
+export const NEXT = () => ({});
```
shlomitc: Maybe instead of adding an additional entry, just name it as `PopoverMenuBeta`

simast: alternatively maybe we should not expose beta/next components form the main index file but keep it in /beta/index.js? The users would then use (and our codemod could convert/handle this as well to):

```js
import { PopoverMenu } from 'wix-style-react/beta'
```

Note that we already have `index.js` in `/beta`.

src/index.d.ts
```diff
@@ -1,7 +1,14 @@
...
+NEXT.PopoverMenu = PopoverMenuBeta;
+
+
+export { IconElement, InjectedFocusableProps, Omit } from './common';
```
simast: is this a public api? have you found where users are using types from `/common`?

src/index.d.ts
mykas: Well, I think it is easier to just go with `PopoverMenuNext` instead. I am not sure how to handle it there properly.

src/index.d.ts
mykas: it is public api, but we are not promoting yet  named imports so we might delete those just before users start using it? what do you think @shlomitc

simast: my bad, I see the types were already exposed before your changes

PR #4936, opened 2020-01-30
<MarketingLayout/> - adjusted paddings, added image background

simast:
- All paddings are now part of `MarketingLayout` (previously they were on `<Card.Content>`). The recommended way now is to use `<Card><MarketingLayout/></Card>` (skipping `<Card.Content>` entirely).
- There is a new prop `imageBackgroundColor` to allow changing the color for image area.
- `image` prop was adjusted in a backwards compatible way to accept image URL directly (will be rendered as ``).
- Adjusted documentation.
- Added "12.9 Marketing Card Layout" symbol to cheatsheet.


argshook:
> recommended way now is to use <Card><MarketingLayout/></Card> (skipping <Card.Content> entirely).

not sure this is ok, because it technically is a breaking change. Users
who currently use `<Card.Content/>` will suddenly get too much padding,
potentially breaking their layouts.

what's the reaosn of need to omit `<Card.Content/>`?

src/MarketingLayout/MarketingLayout.st.css
```diff
@@ -1,57 +1,93 @@
...
+  padding-left: 42px;
+}
+:global([dir="rtl"]) .root:inverted:withImageBackground .contentContainer {
   padding-left: 0;
+  padding-right: 42px;
 }
 
 .title {
-  padding-bottom: 18px;
+  margin-bottom: 18px;
```
argshook: if `margin` is not required is usually choose to use padding, as that one doesn't have the collapse effect, a tad bit safer for future changes

simast:
> what's the reason of need to omit <Card.Content/>?

There were two reasons for this change:

1. Card.Content padding does not match MarketingLayout paddings (card has 24px top padding and 30px bottom padding)
2. We need to show custom color for background area image (that also includes the padding area). Here is screenshot of what I mean (this is when MarketingCard will be used with Modals):

Screenshot 2020-01-30 at 16 04 42

As for backwards compatibility - I think there are not many users yet. Global github search shows that there are two users of MarketingLayout component so far. I am thinking of approaching them before release and informing about the changes. WDYT?

argshook:
>  I think there are not many users yet. Global github search shows that there are two users of MarketingLayout component so far.

if it's just few users, then sure, that's no problem

simast:
@argshook Merging this. Contacted the three users of `<MarketingLayout>` on Slack and they all agreed to make required changes when new WSR version is published.

PR #4939, opened 2020-02-01
docs(FAQ) - update importing component docs

mykas:
This should be merged after we test out named-imports next week.

### DIFF
--- a/docs/FAQ/README.md
+++ b/docs/FAQ/README.md
@@ -28,11 +28,10 @@ Please create a pull request with the solution if it worth sharing.
 
 ### How should I import a component properly?
 
-Currently, as the library does not support tree-shaking, the recommended way to `import` a component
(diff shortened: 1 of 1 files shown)


docs/FAQ/README.md
```diff
@@ -28,11 +28,10 @@ Please create a pull request with the solution if it worth sharing.
 
 ### How should I import a component properly?
 
-Currently, as the library does not support tree-shaking, the recommended way to `import` a component
-is using the entry file of the component, and **not** importing from the index file:
+Our library supports named-imports:
 
 ```js
-import Button from 'wix-style-react/Button';
```
shlomitc: we can still mention that we're allowing both import ways. 
cherrypicking is "good" because you don't bundle the entire.
treeshaking is "best" because it eliminates dead code

PR #4941, opened 2020-02-02
CloseButton - add large size

itaytay:
### 🔦 Summary

...

### ✅ Checklist

- 📚 Change is documented
  - [x] Story
  - [x] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [x] Visual test

### DIFF
--- a/src/CloseButton/CloseButton.js
+++ b/src/CloseButton/CloseButton.js
@@ -29,7 +29,7 @@ class CloseButton extends PureComponent {
       'transparent',
     ]),
     /** size of closebutton */
-    size: oneOf(['small', 'medium']),
+    size: oneOf(['small', 'medium', 'large']),
     /** Click event handler  */
     onClick: func,
     /** applies disabled styles */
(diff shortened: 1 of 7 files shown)


src/CloseButton/CloseButton.js
```diff
@@ -44,6 +44,19 @@ class CloseButton extends PureComponent {
...
 
+  getCloseIcon(size) {
+    let CloseIcon;
+    if (size === SIZES.small) {
+      // fallback to Close icon if children not provided (current behavior)
+      CloseIcon = <Close data-hook="close" />;
+    } else if (size === SIZES.medium) {
+      CloseIcon = <CloseLarge data-hook="close-medium" />;
+    } else {
+      CloseIcon = <CloseLarge data-hook="close-large" size="12" />;
```
nitayneeman: I would expect to get a prepared SVG instead of enlarging manually

src/CloseButton/CloseButton.js
```diff
@@ -44,6 +44,19 @@ class CloseButton extends PureComponent {
     disabled: false,
   };
 
+  getCloseIcon(size) {
```
simast: Note that if you don't prefix a method with an underscode `_` - it will appear in our "API" documentation tab and is a suggestion that you can call this method with a `ref` reference. We should probably rename this to `_ getCloseIcon` or `_renderCloseIcon`.

itaytay: Thanks @simast - I didn't know that.
And I agree that _renderCloseIcon is a better name.
Will change it of course

src/CloseButton/CloseButton.js
itaytay: The problem is that in wix-ui-icons-common there's already Close.svg, and CloseLarge.svg, which oddly enough, represent the 'small' and 'medium' close buttons respectively. Adding a 'large' icon would require renaming, and therefore refactoring, across several different projects, which are using this at the moment.
If it wasn't an SVG, I'd refactor, but scaling an SVG is better than this refactor imo

PR #4951, opened 2020-02-04
`<Slider/>`- replacing tooltip with wsr tooltip

TalZa1810:
### 🔦 Summary
Replacing the custom tooltip with the `wix-style-react` tooltip. 

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test

### DIFF
--- a/src/Slider/SliderHandle.js
+++ b/src/Slider/SliderHandle.js
@@ -1,7 +1,8 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { withFocusable } from 'wix-ui-core/dist/src/hocs/Focusable/FocusableHOC';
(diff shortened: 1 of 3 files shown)


TalZa1810:
Replacing the existing Tooltip with the wsr Tooltip means also updating the driver functions. Moreover, the tooltip didn't work as expected when dragging the Knob. Therefore, I've decided to simply change the st.css to have a black background. Should consider in the future whether to implement this if there is a use case where it is necessary to use the actual Tooltip component.

PR #4958, opened 2020-02-05
<SidePanel/> - Export TS types for testkits and add types to index.d.ts

simast:
The `src/index.d.ts` part is now automated using component generator (implemented before SidePanel was generated).


EgoziE:
Thanks @simast 
I missed this in the review...
Is it something that can be added to the generator?

simast:
The `src/index.d.ts` was already added yesterday to generator. However I am not sure if it is possible to automate this for testkits right now since they are generated slightly differently (as ejs template, can't parse it with codemod). But maybe we could check for *.driver.d.ts files and somehow map this automatically (instead of current manual array map).

PR #4960, opened 2020-02-05
Typography - fixed headings weights

EgoziE:


.storybook/e2e/stories.scss
```diff
@@ -1,8 +1,8 @@
 :root {
   --wsr-font-family: var(--wix-font-stack);
-  --wsr-font-weight-thin: var(--wix-font-weight-thin);
-  --wsr-font-weight-normal: var(--wix-font-weight-normal);
+  --wsr-font-weight-regular: var(--wix-font-weight-regular);
   --wsr-font-weight-bold: var(--wix-font-weight-bold);
+  --wsr-font-weight-xbold: var(--wix-font-weight-xbold);
```
shlomitc: I don't think that's correct. we don't have xbold in our system

shlomitc: also, is this needed also in the e2e storybook?

EgoziE: These are the 3 weights 
regular - 400
bold      - 700
xbold    - 800

unless we have it in the e2e file, the font won't change.
We can use the new `<FontUpgrade/>` instead 😉

EgoziE: I'll delete the xbold in another PR because it requires some work

PR #4961, opened 2020-02-05
Generator - rearrange

EgoziE:
Added small missing stuff to the generator after `<SidePanel/>` component contribution


.wuf/generator/templates/$ComponentName/$ComponentName.js
```diff
@@ -59,4 +49,21 @@ class <%= ComponentName %> extends React.PureComponent {
...
+  dataHook: PropTypes.string,
+
+  /** A css class to be applied to the component's root element */
+  className: PropTypes.string,
+
+  /** Text for the button */
+  buttonText: PropTypes.string,
+};
+
+<%= ComponentName %>.defaultProps = {
```
shlomitc: I think we can ditch this. We never put default content values. You can move it to the story config

PR #4962, opened 2020-02-05
`<Slider/>`- adding custom marks feature

TalZa1810:
### 🔦 Summary
Exposing `marks` props to allow having custom marks.

### ✅ Checklist

- 📚 Change is documented
  - [x] Story
  - [x] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [x] Visual test

### DIFF
--- a/src/Slider/Slider.js
+++ b/src/Slider/Slider.js
@@ -61,7 +61,8 @@ export default class Slider extends Component {
   }
 
   _renderHandle = props => {
-    const { displayTooltip, disabled, value } = this.props;
+    const { displayTooltip, disabled } = this.props;
+    const { value } = props;
     let tooltipValue;
 
     if (this._isCustomMarks()) {
--- a/src/Slider/Slider.js
+++ b/src/Slider/Slider.js
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import Slide from 'rc-slider';
-
+import { dataHooks } from './constants';
 import { generateID } from '../utils/generateId';
 import SliderHandle from './SliderHandle';
 import styles from './Slider.scss';
@@ -54,7 +54,11 @@ export default class Slider extends Component {
       <div>
         <div className={styles.markLine} />
         <div className={styles.markValue}>
-          {shouldRenderText && <div className={styles.markText}>{value}</div>}
+          {shouldRenderText && (
+            <div data-hook={dataHooks.sliderMark} className={styles.markText}>
+              {value}
+            </div>
+          )}
         </div>
       </div>
     );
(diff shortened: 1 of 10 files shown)


TalZa1810:
Hi @ZilvinasMaciulis ,
Should I force the user by code to provide us an object of:
 `{ number: string }`  or `{ number : number }` to prevent cases where they customize the label node (screenshot above) or do we want to allow this?

src/Slider/Slider.js
```diff
@@ -19,35 +19,77 @@ const range = ({ min, max, step }) => {
...
   _renderHandle = props => {
     const { displayTooltip, disabled } = this.props;
+    const { index } = props;
+    let tooltipValue;
+
+    if (this._isCustomMarks()) {
+      const { marks } = this.props;
+      const { value: key } = props;
+      if (marks.hasOwnProperty(key)) {
+        tooltipValue = marks[key].toString();
```
jonathanadler: In your example you passed a `div` as the value of the mark
so i'm not sure it will work with this `toString` 
no? 🤔

src/Slider/tests/Slider.visual.js
```diff
@@ -39,6 +39,20 @@ const tests = [
...
+        props: {
+          marks: {
+            0: '$0',
+            2: '$79',
+            4: '$129',
+            6: '$199',
+            8: '$349',
+            10: '$499',
+          },
+          displayMarks: true,
```
jonathanadler: Is this prop used in the component?

src/Slider/Slider.js
TalZa1810: Do you mean the example I've posted above?
Basically, in my opinion, we should **not** allow passing div. We want to pass only a string or a number. A div won't be good for the tooltip. The question is whether throwing an error for this case or not:
https://github.com/wix/wix-style-react/blob/663a7bfffee0a33d629fdfd0d20fd81a7daab3f1/src/Slider/Slider.js#L160

In the storybook example, I've used a string, not a div.

src/Slider/Slider.js
```diff
@@ -19,35 +19,77 @@ const range = ({ min, max, step }) => {
...
+
+    return marksLabels;
+  }
 
-      return acc;
-    }, {});
+  _isCustomMarks() {
+    const { marks } = this.props;
+    return !(
+      Object.entries(marks).length === 0 && marks.constructor === Object
```
TalZa1810: no need for `marks.constructor === Object` when using propTypes

src/Slider/tests/Slider.visual.js
TalZa1810: Yes. In the `commonProps` it is set to `false` but in the component, the defaultValue is set to `true`. Therefore, in the visual tests I've set this to true.

PR #4963, opened 2020-02-06
FontUpgrade + MadeforToggle addon

EgoziE:
Added a `<FontUpgrade/>` wrapper to support Madefor font.
With that I added a small addon to storybook to help us toggle between Helvetica Neue and Madefor:

Switched off:
image

Switched on:
image


EgoziE:
Now we can control JS and CSS with the font context.
For example:

```
import { FontUpgradeContext } from '../../FontUpgrade/FontUpgrade';

<FontUpgradeContext>
  {(context = {}) => (
    <Text size="medium" weight={context.active ? 'bold' : 'thin'}>
      This is a text!
    </Text>
  )}
</FontUpgradeContext>
```

src/FontUpgrade/FontUpgrade.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import createReactContext from 'create-react-context';
+import styles from './FontUpgrade.css';
+
+const Context = createReactContext({});
```
shlomitc: I think this should be in a separate file `FontUpgradeContext` and other components should import the common files

PR #4964, opened 2020-02-06
Fix grid css for chrome 80 problem

EgoziE:
Our visuals started breaking in Chrome 80.
After some investigation, it seems something changed in the way Grid works. (probably calculation of fr unit)
So I found a quick solution, but we should investigate this some more to figure out if we're using css grid wrong.

@argshook, can you please take a look, I'll continue investigating this next week.
Thanks!


jonathanadler:
Might be related to this:
https://bugs.chromium.org/p/chromium/issues/detail?id=1049573&q=grid&can=2

shlomitc:
Can you provide a reproducible example?

shlomitc:
reproducible example (just put in the playground)
this will break in Chrome 80 and work in older Chromes / FF / Safari
```jsx
<Layout>
  <Cell span={4}>
    <Text ellipsis>Hello World Hello World Hello World Hello World Hello World Hello World Hello WorldHello World Hello World Hello World Hello World Hello World</Text>
  </Cell>
    <Cell span={4}>
    <Text ellipsis>Hello World Hello World Hello World Hello World Hello World Hello World Hello WorldHello World Hello World Hello World Hello World Hello World</Text>
  </Cell>
    <Cell span={4}>
    <Text ellipsis>Hello World Hello World Hello World Hello World Hello World Hello World Hello WorldHello World Hello World Hello World Hello World Hello World</Text>
  </Cell>
</Layout>
```

@giladsegal - is there a chance we're misusing CSS Grid in our `<Layout/>` component?

EgoziE:
I can see there are still few visuals failing , although it's less than before.
So this solution is not perfect... 😞 
this requires further investigation

simast:
Related Chromium issue for some context: https://bugs.chromium.org/p/chromium/issues/detail?id=1038546

argshook:
@EgoziE i think the `minmax` addition is a good fix.

there are two failing visual tests:

1) ColorInput examples:

in there `<Layout/>` is used but it misses `<Cell/>`. So it's a misuse
of `Layout` component which should be fixed. 

2) Layout Holy grail example:

before chrome 80 using `<Cell><div style="width: 400px"></div></Cell>`
would render all 400px width and push other cells.
after chrome 80 this is no longer the case other cells are not pushed.
if inner content of a `Cell` is wider than grid column, it simply
overflows.

This is a bugfeature. We should probably updat our holy grail example
and perhaps not even use `Layout` for it but rely on flexbox. That's
because holy grail example is not a fixed but a fluid grid.

---

all in all i think `minmatch` is a good fix and the broken visuals of
ours are misuses of CSS grid.

Chrome 80 can, however, have impact on our users. For example, i've seen
users just use `<Layout>` component without any `<Cell>`. After chrome
80 such layouts for them most likely will fail

shlomitc:
@argshook - thanks for the investigation. Can you push a fix for these two failing tests?
Also, I think we should provide this fix also to v6 as it's still very popular across our projects

EgoziE:
Some conclusions:
so apparently `1fr` means `minmax(**auto**, 1fr)` this means that the track to which it is applied cannot shrink below the **content** size which is the min value in the minmax() function is auto, meaning **content-based** size.
In order to limit the content to the layout cell width, the right fix is `minmax(**0**, 1fr)` which means that the track to which it is applied can shrink up to a minimum of 0.

Also I fixed the faulty visuals.

EgoziE:
Source - https://www.w3.org/TR/css-grid-1/#min-size-auto
and the specific line - https://www.w3.org/TR/css-grid-1/#valdef-grid-template-columns-flex

PR #4965, opened 2020-02-08
`<ModalPreviewLayout/>`- adding tooltip props for close button & navigation buttons

TalZa1810:
### 🔦 Summary
1. Having the option to custom the tooltips of the close button & navigation buttons.
2. Fixing issue with tooltip flakiness when clicking on prev & next buttons.

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [x] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [ ] Visual test


src/ModalPreviewLayout/NavigationButton/NavigationButton.js
```diff
@@ -19,27 +19,35 @@ const tooltipProps = {
...
-    <Tooltip {...tooltipProps[direction]}>
-      <IconButton as="button" skin="transparent" {...props}>
-        {iconButtonArrow[direction]}
-      </IconButton>
-    </Tooltip>
-  </div>
-);
+const NavigationButton = ({ direction, tooltipText, ...props }) => {
+  //TODO: adding tests for checking tooltip text
+  if (typeof tooltipText !== 'string') {
```
shlomitc: Same here

TalZa1810:
Hi @EgoziE ,
Do you think that I should also visual test each of the tooltip props? WDYT?

src/ModalPreviewLayout/ModalPreviewLayout.js
```diff
@@ -25,10 +25,19 @@ class ModalPreviewLayout extends React.PureComponent {
...
+    prevButtonTooltipText: PropTypes.string,
+    /** Tooltip next button text */
+    nextButtonTooltipText: PropTypes.string,
   };
 
   static defaultProps = {
     shouldCloseOnOverlayClick: true,
+    closeButtonTooltipText: 'Close',
+    prevButtonTooltipText: 'Previous',
+    nextButtonTooltipText: 'Next',
```
EgoziE: Shlomi raised a point an another PR, adding default values for "user props" is not recommended.
For example, if I don't want a tooltip, I'm getting the default, which is not even translated.

I think these 3 lines should be deleted 🗡

TalZa1810:
Hi @EgoziE ,
Can you please review the changes? I've removed the default props and updated the code and unit tests accordingly. Also, after merging and releasing, the types PR can be merged:
https://github.com/wix/wix-ui/pull/1756

FYI @michalavraham

src/ModalPreviewLayout/ModalPreviewLayout.js
```diff
@@ -55,26 +61,40 @@ class ModalPreviewLayout extends React.PureComponent {
     };
   }
 
-  _renderNavigationButtons() {
+  _renderNavigationButtons(hasLeft, hasRight) {
+    const { prevButtonTooltipText, nextButtonTooltipText } = this.props;
```
mosheKerbel: looks like it's not up to the user if nav buttons should be rendered
in this case, shouldn't `prevButtonTooltipText` & `nextButtonTooltipText` be mandatory props?

mosheKerbel: my bad, i see it's part of `NavigationButton`

TalZa1810: If they are mandatory, we should have default props for this, no?
That's why I changed the logic: 
https://github.com/wix/wix-style-react/pull/4965#discussion_r376956592

should we make them mandatory and have default props for this or not?

TalZa1810: or just have them required without default props?

PR #4969, opened 2020-02-10
<Avatar/> - Added onClick prop

simast:
- Added `onClick` prop and accessibility features/styling
- Fixed avatar testkit methods (was not working properly)
- Updated avatar TypeScript typings


src/Avatar/Avatar.uni.driver.js
```diff
@@ -7,7 +7,9 @@ export default base => {
     iconButtonDriverFactory(base.$(`[data-hook="${dataHooks.indication}"]`));
 
   return {
-    ...avatarDriverFactory(base),
+    ...avatarDriverFactory(
+      base.$(`[data-hook="${dataHooks.avatarCore}"] > :first-child`),
```
argshook: is this needed? is `.click` not working without such selector?

simast: click is not working without setting proper base element, and even more core avatar testkit methods right now are broken without this patch expecting to find data attributes and such (https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/avatar/avatar.uni.driver.ts#L19)

as an alternative I can add `data-hook` prop support to core avatar so we don't need `> :first-child` here. WDYT?

argshook: ok i see

> as an alternative I can add `data-hook` prop support to core avatar so we don't need `> :first-child` here. WDYT?

yes that'd be ideal i think.

Some (but not all) `wix-ui-core` components support `dataHook` prop
which is then translated to `data-hook` html attribute. I guess `Avatar`
should do the same

PR #4971, opened 2020-02-10
move the css variables to body

EgoziE:


src/FontUpgrade/FontUpgrade.js
```diff
@@ -1,16 +1,25 @@
...
 import { FontUpgradeContext } from './context';
 
+const styles = {
+  '--wsr-font-family': 'var(--wix-font-stack)',
+  '--wsr-font-weight-thin': 'var(--wix-font-weight-thin)',
+  '--wsr-font-weight-normal': 'var(--wix-font-weight-normal)',
+  '--wsr-font-weight-bold': 'var(--wix-font-weight-bold)',
+};
+
 const FontUpgrade = ({ dataHook, active = true, children }) => {
```
shlomitc: `children` is missing in `propTypes`

src/FontUpgrade/FontUpgrade.js
```diff
@@ -1,16 +1,25 @@
...
+  if (active) {
+    Object.entries(styles).forEach(([key, value]) =>
```
shlomitc: this loop will run each time the component is rendered, which will cause some performance issues.
It should be set only once when rendering the component and removed when unmounted

src/FontUpgrade/FontUpgrade.js
shlomitc: you misspelled `FontUpgrade.Proptypes = {`
it should be `FontUpgrade.propTypes = {`

src/FontUpgrade/FontUpgrade.js
```diff
@@ -1,16 +1,25 @@
 import React from 'react';
-import styles from './FontUpgrade.scss';
 import PropTypes from 'prop-types';
 import { FontUpgradeContext } from './context';
 
+const styles = {
```
shlomitc: this can still be a class name which is placed (portaled) on a different node

PR #4972, opened 2020-02-10
Add dynamic minutes step in <TimePicker />

DanielEliraz:
Adding minutesStep prop in timePicker.
This prop set the number of minutes to be changed on arrow click.

- 📚 Change is documented
  - [X] Story
  - [X] API description
- 🔬 Change is tested
  - [X] Component
  - [ ] Visual test


DanielEliraz:
Hi @itzikkit, please tag Or also here.
And let's try to push it..

itzikkit:
Added @orgol99, our PM

src/TimeInput/TimeInput.js
```diff
@@ -45,6 +48,7 @@ export default class TimePicker extends Component {
     disableAmPm: false,
     disabled: false,
     dashesWhenDisabled: false,
+    minutesStep: 20,
```
EgoziE: Is it approved by our UX that the default step is 20 min?
I seems right to me that if I don't provide this prop (or sets it to 0) it gos back to 1 hour step as before.

DanielEliraz: It was 20 minutes before, only if the user selected the hour in the picker than it was 1 hour

DanielEliraz: `const timeUnit = this.state.lastFocusedTimeUnit || 'minutes';
const amount = timeUnit === 'hours' ? 1 : 20;`

EgoziE: there was not default value before

EgoziE: I'm talking about `defaultProps`

DanielEliraz: OK, but it was default (and only) in the code

EgoziE: I talked with our UX, this behaviour will change anyway when we rewrite this component in wsr8
So we'll merge it for now :)

simast:
Let's not forget to add TypeScript typings for this new prop as well:
https://github.com/wix/wix-ui/blob/851eebff3b0b67cef3f7a6c218b28909d947c7a4/packages/wsr-types/src/components/TimeInput.d.ts#L3-L12

PR #5000, opened 2020-02-14
Test waitBeforeScreenshot to reduce flakiness with eyes-storybook

simast:
Just testing `waitBeforeScreenshot` option as we seem to have more and more flaking tests (related to fonts not being loaded when screenshot is taken).


simast:
Closing as it doesn't look like this solves issues with flaking tests.

Issue #2990, opened 2019-02-19
`<Box/>` - Add `spacing` prop : space between with value (`px, %, ...`)


hepiyellow:
# ✨ Feature Request

### 🔦 Scope

`<Box/>`

### 💁 Explanation

Allow having fixed spacing between items (children)

### 🐾 Possible solution 

```
<Box spacing="tiny"/>
```
- Will render the children with tiny (6px) spacing.
- The `spacing` prop is independent of the `direction` and the `align` props.
- Expect, when `align=space_between` OR `verticalAlign=space_between`, in which case it overrides the `align|alignVertical` prop.

### 👀 Severity

- Low (Medium) - we should reuse this in many components, internally.

hepiyellow:
@nitayneeman @shlomitc

EgoziE:
Moved to Jira - closing

PR #5005, opened 2020-02-15
WIP: <Box /> - new feature gap

mykas:
This is still WIP it will be explained a bit in a separate RFC issue.

### DIFF
--- a/src/Box/Box.js
+++ b/src/Box/Box.js
@@ -126,7 +126,11 @@ const Box = ({
     marginTop: formatSpacingValue(marginTop),
     marginRight: formatSpacingValue(marginRight),
     marginBottom: gap
-      ? `-${formatSpacingValue(gap)}`
+      ? marginBottom
+        ? `calc(-${formatSpacingValue(gap)} + ${formatSpacingValue(
+            marginBottom,
+          )}`
+        : `-${formatSpacingValue(gap)}`
       : formatSpacingValue(marginBottom),
     marginLeft: formatSpacingValue(marginLeft),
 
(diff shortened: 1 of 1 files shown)


nitayneeman:
I think we should discuss in the RFC before implementing and also compare the pros/cons compared to the `Layout` component

mykas:
I agree. Created this just an example in RFC issue.

shlomitc:
@nitayneeman - so what's your opinion regarding this?
We also have an old issue suggesting something similar
https://github.com/wix/wix-style-react/issues/2990

nitayneeman:
At first I thought that we should not complicate `Box` for something that's currently possible with the Layout.

But, it seems that `gap` is going to be supported natively by `flexbox` in the near future: https://youtu.be/-oyeaIirVC0?t=820

It's already available in FF:
https://developer.mozilla.org/en-US/docs/Web/CSS/gap#Support_in_Flex_layout

This means the approach is to let the flex container to own and determine the spacing of the children. It's nice to see this demand also comes from the community - so that's why I think this feature is reasonable. 

I guess this PR became the RFC issue? 🤨

mykas:
Sorry, I just could not find the right time to open RFC :/

shlomitc:
So... do we plan to continue with this PR? @nitayneeman what do you suggest?

nitayneeman:
> So... do we plan to continue with this PR? @nitayneeman what do you suggest?

@mykas are you going to continue with the RFC you've planned soon?

If not, I suggest closing this PR in the meantime and prioritize a task in the backlog of this feature (we can also keep it for a guild week but we need to refine a solution before).

shlomitc:
So I'm closing this PR for now, it's Passover cleaning time.
@mykas - if you get some time to work on it or find someone who can contribute - please reopen

PR #5006, opened 2020-02-16
<AutoCompleteWithLabel />, <InputWithLabel /> - style fix

saarkuriel:
### 🔦 Summary

According to this Jira:
https://jira.wixpress.com/browse/PREM-10763?filter=32448
The following style fixes should be applied.
Thanks

### ✅ Checklist

- [x] 👨‍🎨 UX change is approved by the Design System UX @wuwa via Anat Steiner 
- 📚 Change is documented
  - [x] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test

### DIFF
--- a/src/AutoCompleteWithLabel/AutoCompleteWithLabel.scss
+++ b/src/AutoCompleteWithLabel/AutoCompleteWithLabel.scss
@@ -11,10 +11,16 @@
   display: flex;
   align-items: center;
   line-height: 24px !important;  
+  align-items: flex-start;
+  margin-top: 2px;
 }
 
 .statusMessageIcon {
   margin-right: 5px;
   display: flex;
   margin-left: -2px;
+}
+
+.errorMessageContent {
+  line-height: 18px;
 }
--- a/src/InputWithLabel/InputWithLabel.st.css
+++ b/src/InputWithLabel/InputWithLabel.st.css
@@ -46,10 +46,16 @@
   display: flex;
   align-items: center;
   line-height: value(status-message-height) !important;
+  align-items: flex-start;
+  margin-top: 2px;
 }
 
 .statusMessageIcon {
   margin-right: 5px;
   display: flex;
   margin-left: -2px;
+}
+
+.errorMessageContent {
+  line-height: 18px;
 }
--- a/src/AutoCompleteWithLabel/AutoCompleteWithLabel.js
+++ b/src/AutoCompleteWithLabel/AutoCompleteWithLabel.js
@@ -169,7 +169,12 @@ class AutoCompleteWithLabel extends React.PureComponent {
             <span className={styles.statusMessageIcon}>
               <StatusAlertSmall />
             </span>
-            <span data-hook={dataHooks.errorMessage}>{statusMessage}</span>
+            <span
+              data-hook={dataHooks.errorMessage}
+              className={styles.errorMessageContent}
+            >
+              {statusMessage}
+            </span>
           </Text>
         )}
       </div>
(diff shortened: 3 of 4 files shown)


mykas:
Hey from the JIRA ticket I see these requirements:
Screenshot 2020-02-17 at 13 51 34

what does it mean that content below should be pushed down? Maybe it meant to wrap when content is too long?

src/InputWithLabel/InputWithLabel.st.css
```diff
@@ -46,10 +46,16 @@
...
 }
 
 .statusMessageIcon {
   margin-right: 5px;
   display: flex;
   margin-left: -2px;
+}
+
+.errorMessageContent {
+  line-height: 18px;
```
EgoziE: This is not standard and should be approved by wsr ux

src/AutoCompleteWithLabel/AutoCompleteWithLabel.scss
```diff
@@ -11,10 +11,16 @@
...
 }
 
 .statusMessageIcon {
   margin-right: 5px;
   display: flex;
   margin-left: -2px;
+}
+
+.errorMessageContent {
+  line-height: 18px;
```
EgoziE: This is not standard and should be approved by wsr ux

saarkuriel:
> Hey from the JIRA ticket I see these requirements:
> Screenshot 2020-02-17 at 13 51 34
> 
> what does it mean that content below should be pushed down? Maybe it meant to wrap when content is too long?

That means that the error will not overflow if there's content below. That's sort of a given...

mykas:
But then it just goes on and on
Screenshot 2020-02-17 at 13 55 57

Is that intentional?

mykas:
> This PR should be reviewed by our UX

It seems that this component is not in our zeplin specs at all. I am not sure what is the status of these components. @nitayneeman @shlomitc can you tell us?

shlomitc:
@mykas - this is a legitimate component in our design system, I asked our UX to add it to our sketch file.
cc- @milkyfruit

saarkuriel:
> But then it just goes on and on
> Screenshot 2020-02-17 at 13 55 57
> 
> Is that intentional?
> It's not wrapping cause there are no spaces

What string will be that long without any spaces?

mykas:
> > But then it just goes on and on
> > Screenshot 2020-02-17 at 13 55 57
> > Is that intentional?
> > It's not wrapping cause there are no spaces
> 
> What string will be that long without any spaces?

You're right this is not a good example.

saarkuriel:
@mykas can we proceed with this? thanks

src/InputWithLabel/InputWithLabel.st.css
mykas: @milkyfruit please check when you can

milkyfruit:
If I'm correct this component was developed just for Premium, specs came from there too. There is no intention to promote this input to the rest of designers, so official specs don't exist too. Officially is promoted to use `<Input/>`.

To fix this component, Jira comments are correct. The component should behave like this:
- Component's height is auto
- Error message text should Wrap
- Component's bounding box should grow with wrapped text

saarkuriel:
> If I'm correct this component was developed just for Premium, specs came from there too. There is no intention to promote this input to the rest of designers, so official specs don't exist too. Officially is promoted to use `<Input/>`.
> 
> To fix this component, Jira comments are correct. The component should behave like this:
> 
> * Component's height is auto
> * Error message text should Wrap
> * Component's bounding box should grow with wrapped text

If this is the situation, I'd be more than happy to take these components into premium. It would be much easier for us. @shlomitc please advise.

shlomitc:
The component should stay part of `wix-style-react` and specs will be added in the near future.
At the moment, let's follow the guidelines as you receive

PR #5009, opened 2020-02-16
<VariableInput /> - onBlur, fix setValue issue

tomershk:
### 🔦 Summary

- New API for `onBlur` to allow client to use different logic than `onSubmit`.
- Fix issue when older value was with new line, focus was on the new line and after we call `setValue` new line was removed. (caused the component to crash)

### ✅ Checklist

- 📚 Change is documented
  - [x] API description
- 🔬 Change is tested
  - [x] Component


src/VariableInput/VariableInput.uni.driver.js
```diff
@@ -30,6 +30,20 @@ export default (base, body) => {
         contentElement.sendKeys(text);
       }
     },
+    focus: async () => {
```
nitayneeman: Do we really need `focus`? 
I wonder what the use-case is for the consumers (maybe it should be private)?

nitayneeman:
Let's wait for the CI

nitayneeman:
@tomershk We also need to adjust the testkit types:
https://github.com/wix/wix-style-react/blob/master/src/VariableInput/VariableInput.uni.driver.d.ts

tomershk:
Will that resolve this issue as well?
https://wix-style-react.now.sh/?activeTab=Testkit&path=%2Fstory%2Fcomponents-api-components--variableinput

nitayneeman:
> Will that resolve this issue as well?
> https://wix-style-react.now.sh/?activeTab=Testkit&path=%2Fstory%2Fcomponents-api-components--variableinput

This seems to be a different issue

PR #5019, opened 2020-02-17
Remove composite components

EgoziE:
I know this is a big PR but all of the composites were related in a way...

Removed:
- `<AutoCompleteComposite/>`
- `<InputAreaWithLabelComposite/>`
- `<MultiSelectComposite/>`
- `<FieldWithSelectionComposite/>`
- `<GoogleAddressInputWithLabel/>`


src/GoogleAddressInputWithLabel/GoogleAddressInputWithLabel.protractor.driver.js
```diff
@@ -1,7 +1,7 @@
-import inputAreaWithLabelCompositeDriverFactory from '../Composite/InputAreaWithLabelComposite/InputAreaWithLabelComposite.protractor.driver';
```
shlomitc: I think we can remove this composite as well

PR #5020, opened 2020-02-18
replaceStylalbeImports - return & resolve promise early if no changes needed to avoid redundant file writes

mykas:
This PR is fixing an unknown bug in stylable imports replacer which left a syntax issue in TimeTable.st.css that caused yoshi start command to hang when user used `enhanchedTpaStyle` configuration.

The script was written in such way that even some scripts that did not have things to be changed - the fs.writeFile did file writing anyway. This pr is fixing this and returning early if no changes are needed to be applied to the file.


scripts/replaceStylableImports.js
```diff
@@ -23,29 +45,31 @@ module.exports = () => {
...
+            resolve();
+            return;
+          }
 
-          wrongPaths.forEach(({ regexp, correct }) => {
-            if (regexp.test(results)) {
+          if (changes) {
+            results = results
+              .replace(
+                /wix-ui-core\/index\.st\.css/g,
```
simast: could we reuse `PATHS_TO_CHANGE ` here to not duplicate regular expressions?

mykas: When I iterate through that const - it will not replace the strings. If you have other ways to solve this - please suggest.

simast: hmm, I tested with:

```js
PATHS_TO_CHANGE.forEach(({ regexp, changeTo }) => {
  results = results.replace(regexp, changeTo);
});
```

Also added `/g` flag to `PATHS_TO_CHANGE` regexps so it would replace globally and it seems to work correctly.

mykas: Then the `test` part will file above as it does not work for some reason with flag `/g`

simast: both `.replace` and `.test` works for me with the same regexp and with `/g` flag. Have you noticed the issue for one particular component maybe?

PR #5029, opened 2020-02-19
<StatusIndicator/> - change driver

EgoziE:
Changed driver functions and added comments


src/StatusIndicator/StatusIndicator.uni.driver.js
```diff
@@ -6,15 +6,20 @@ export const statusIndicatorDriverFactory = (base, body) => {
   return {
     ...baseUniDriverFactory(base, body),
 
-    hasTooltip: async () => {
+    /** Returns the type of the status oneOf(error, warning, loading) */
+    getStatus: async () => base.attr('data-status'),
+
+    /** Returns true iff a message was provided */
```
simast:
```suggestion
    /** Returns true if a message was provided */
```

EgoziE: iff is "if and only if"...
I can change it to :
Returns true if a message was provided, false otherwise.

simast: got it, it's so minor let's keep it as is

PR #5033, opened 2020-02-19
<ListItemSelect /> - added hovered prop

audriuss:
### 🔦 Summary
Currently, if `listItemSelectBuilder` or `listItemActionBuilder ` are used to build `DopdownLayout` options there is no indication of current item when navigating with keyboard. It's reproducible in storybook: 
http://localhost:6006/?path=/story/components-api-components--listitemaction or
https://wix-style-react.now.sh/?path=/story/components-api-components--listitemselect (advanced example)
Just try to navigate with keyboard on the lists 

Solution:
`hovered` prop is passed by `<DopdownLayout />`, so just need to apply hover style based on it inside `<ListItemSelect />` and `<ListItemAction />`. So added this prop on the components and applied the style.


src/ListItemAction/ListItemAction.js
```diff
@@ -79,6 +82,7 @@ class ListItemActionComponent extends React.PureComponent {
     as: 'button',
     skin: 'standard',
     size: 'medium',
+    hovered: false,
```
mykas: Lets add a visual test for this case

src/ListItemSelect/ListItemSelect.js
```diff
@@ -42,6 +42,9 @@ class ListItemSelect extends React.PureComponent {
     /** If true, the item is selected */
     selected: PropTypes.bool,
 
+    /** If true, the item is hovered */
+    hovered: PropTypes.bool,
```
mykas: Lets add a visual test for this case

src/ListItemAction/ListItemAction.js
audriuss: 👍

src/ListItemSelect/ListItemSelect.js
audriuss: 👍

EgoziE:
I know there's a bug right now, but this solution is like a band aid with a side effect.
There's no real use for a hover prop in a list item. I don't want it to be exposed and used.
I'll discuss this with the team and think of a solution :/

EgoziE:
Nevertheless I appreciate the PR and your will to help!

EgoziE:
@audriuss We had a little talk about this issue, 
I'll change the prop to be "highlight" instead of "hover" and merge it.

audriuss:
@EgoziE  thanks! Agree that `highlight` make more sense

PR #5037, opened 2020-02-20
<InputArea/> - status & statusMessage

EgoziE:
<InputArea/>
Removed props 'error' and 'errorMessage'.
Now using only already developed 'status' and 'statusMessage'.

Note: in the driver I had to use tooltip driver from wix-ui-core because it has only uni-driver.


src/InputArea/test/InputArea.spec.js
```diff
@@ -201,25 +201,25 @@ describe('InputArea', () => {
...
-      it('should display an warning icon if status="warning"', async () => {
-        const driver = createDriver(<InputAreaForTesting status="warning" />);
-
-        expect(await driver.hasWarning()).toBe(true);
+      [
+        { status: 'error', message: 'Error Message' },
+        { status: 'warning', message: 'Warning Message' },
+        { status: 'loading', message: 'Loading Message' },
+      ].forEach(test => {
+        it('should display an error icon if status="error"', async () => {
```
simast:
```suggestion
        it('should display a status icon when status="${test.status}"`, async () => {
```

src/InputArea/InputArea.js
```diff
@@ -386,9 +348,6 @@ InputArea.propTypes = {
   onKeyDown: PropTypes.func,
   onKeyUp: PropTypes.func,
 
-  /** @deprecated onShow prop for the error tooltip */
-  onTooltipShow: PropTypes.func,
```
simast: should we remove this from `Input` component as well?

EgoziE: Yes, Input is my next PR

PR #5008, opened 2020-02-16
move types from wsr

mosheKerbel:
### 🔦 Summary

...

### ✅ Checklist

- [ ] 👨‍💻 API change is approved by the UI Infra Developers 
- [ ] 👨‍🎨 UX change is approved by the Design System UX 
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test

PR #5042, opened 2020-02-20
Create types-faq.md

mosheKerbel:
@shlomitc 
for some reason the links do not work in preview mode of PR
but the bullets on top are links to the sections.


mosheKerbel:
for some reason the links do not work in preview mode of PR
but the bullets on top are links to the sections.

shlomitc:
you can use this link to verify links work:
https://github.com/wix/wix-style-react/blob/temp-faq/types-faq.md

I'll go over it soon

types-faq.md
```diff
@@ -0,0 +1,70 @@
...
+ <Button onClick={onClick}/>
+```
+   
+Don't have time for it now and want a quick fix?
+```jsx
+ <Button onClick={onClick as any}/>
+```
+   
+
+
```
shlomitc:
```suggestion
   # Typescript type definitions in `wix-style-react`

## Background
`wix-style-react` was initially created as a Javascript library, using "PropTypes" to define the components APIs. In the past few years, many new projects were created with Typescript support, enjoying the benefits of types definitions. Due to that, we decided to gradually start adding type definitions to the library.

## Introducing the type definitions in the library
Adding type definitions to all the components in the library is not an easy task. Creating more than 400 type definitions for the components and their test kits, making sure props aren't missing, demystifying complex object structures and keeping them in sync while the library keeps evolving every day.

### Release

In order not to get code divergence, we decided to introduce the types in `wix-style-react@7` as a `minor` change.

### Type definitions correctness

All the types were tested and validated on projects heavily using the library, getting validation that they are very much accurate, however, there might still be some mismatches, so pull-requests for fixes will be highly apprecaiated

## Usage and common issues

### Using the new types

#### Should I do something in order to get the new type definitions?

**No!** all types are built-in in the library, so you get them out of the box.

#### I already use `wix-style-react` and my build got broken after types were added. What should I do?

First, we're really sorry about that, and our immediate solution is to **opt-out all types**. Simply add the following line to your `tsconfig.json`:
`"files": ["./node_modules/wix-style-react/ignore.d.ts"]`
Now that your build passes again, you can fix it whenever you are ready.

#### Opting-out specific components
If your components are imported like this `import Button from 'wix-style-react/Button'`, you can decide to **opt-out specific components**. This means you will get all of the component types, besides specific ones that break your project

1. Remove `./node_modules/wix-style-react/ignore.d.ts` from the `tsconfig.json` in case you added it.
2. Add a new file entry to your `tsconfig.json` files:
    `"files": ["./src/ignore-wsr-types.d.ts"]`.
3. Override only specific components
    ```ts
    //ignore-wsr-types.d.ts
    declare module 'wix-style-react/Button';
    ```
This option is great if most of the project work and you want to get the new types and only leave out those who are not working for you. When coming to fix your issue, simply comment out this override.

If you're consuming applying  "tree-shaking" by importing `ESModule` components (`import {Button} from 'wix-style-react'`), then this method will not work, unfortunately... you will need to continue opting-out the entire library.

#### Common reason for broken builds

1. You are using `string`, but the types are is too permissive.
Wrong:
     ```tsx
     const size: string;
     <Button size={size}/>
     ```
    Correct:
    
    ```tsx 
     const size: ButtonSize;
     <Button size={size}/>
     
     OR
     
     const size: 'tiny' | 'small' | 'medium' | 'large';   
     <Button size={size}/>
    ```

2. You are passing a function that is too permissive.
     Wrong:
    ```tsx
     const onClick: Function;
     <Button onClick={onClick}/>
     ```
     
    Correct:
    ```tsx
    const onClick: React.MouseEventHandler<HTMLElement>;
    <Button onClick={onClick}/>
    ```
    
    Workaround:
    ```jsx
     <Button onClick={onClick as any}/>
    ```
    
#### I found an issue with your types
We'll appreciate it if you could [open an issue](https://github.com/wix/wix-style-react/issues/new/choose), and will be very happy if you could make a **pull-request** fixing it. We'll do our best to be as responsive as possible to get it merged.
```

mosheKerbel:
> I did a bit of a makeover, please go over it.
> 
> Also - I'm not quite sure I understand the suggestion regarding the `Button` component fix:
> 
> ```tsx
> const size: ```jsx
> const size: ButtonSize
> OR
> const size: 'tiny' | 'small' | 'medium' | 'large';   
> <Button size={size}/>
> ```
> 
> What is `ButtonSize`? Where does it come from? also, do you really expect people to define const `size: 'tiny' | 'small' | 'medium' | 'large';` This looks like an overkill.
> Why not simply suggest to change it to:
> `<Button size="tiny"/>`

The situation is where you are passing a value that you already defined a type for,
For example:

```jsx
import { Button } from wix-style-react;
const CompProps = {
  buttonSize: string;
}
class MyComponent<CompProps, CompState> {
   render() {
      <Button size={this.props.buttonSize}/>
  }
}
```

where we declared the class props with `buttonSize: string` we were to permissive
correction: 
```jsx
import { Button, ButtonSize } from wix-style-react;
const CompProps = {
  buttonSize: ButtonSize;
}
class MyComponent<CompProps, CompState> {
   render() {
      <Button size={this.props.buttonSize}/>
  }
}
```

types-faq.md
```diff
@@ -0,0 +1,86 @@
...
+    ```
+This option is great if most of the project work and you want to get the new types and only leave out those who are not working for you. When coming to fix your issue, simply comment out this override.
+
+If you're consuming applying  "tree-shaking" by importing `ESModule` components (`import {Button} from 'wix-style-react'`), then this method will not work, unfortunately... you will need to continue opting-out the entire library.
+
+#### Common reason for broken builds
+
+1. You are using `string`, but the types are is too permissive.
+Wrong:
+     ```tsx
```
mosheKerbel: import { Button } from wix-style-react;
const CompProps = {
  buttonSize: string;
}
class MyComponent<CompProps, CompState> {
   render() {
      <Button size={this.props.buttonSize}/>
  }
}

@shlomitc 
It makes the code example a bit longer, but i realize that the idea is more cllear that way, u think i should change it?

types-faq.md
```diff
@@ -0,0 +1,86 @@
...
+     const size: string;
+     <Button size={size}/>
+     ```
+    Correct:
+    
+    ```tsx
```
mosheKerbel: import { Button } from wix-style-react;
const CompProps = {
  buttonSize: ButtonSize;
}
class MyComponent<CompProps, CompState> {
   render() {
      <Button size={this.props.buttonSize}/>
  }
}

types-faq.md
shlomitc: Wrong:
```tsx
import { Button } from 'wix-style-react';

const MyComp: React.SFC<{ size: string }> = ({ size }) => (
  <Button skin={size}>Click me</Button>
);
```

Good:
```tsx
import { Button } from 'wix-style-react';

const MyComp: React.SFC<{ size: 'small' | 'tiny' }> = ({ size }) => (
  <Button size={size}>Click me</Button>
);
```

##


I don't think that this is a realistic example since consumers will probably have only a subset of options, so I'm not sure we should recommend using it:

```tsx
import { Button, ButtonSize } from 'wix-style-react';
const MyComp: React.SFC<{ size: ButtonSize }> = ({ size }) => (
  <Button size={size}>Click me</Button>
);
```

giladsegal: if it is only a subset of available types and we want to remain typed we can use
`Pick<ButtonSize, 'small' | 'tiny'>`
Given that users will use this syntax, if we will change the available button sizes, their code won't compile

giladsegal:
I would add another example here, which is obvious but important - maybe the user has a typo or  uses a property that doesn't exist anymore (e.g. `upgrade` prop that was removed)

mosheKerbel:
moved into the main PR - https://github.com/wix/wix-style-react/pull/5008/commits/f24d119b329229bbfb6462e42fa1ea25a5955de1

PR #5043, opened 2020-02-21
(WSR8) Promote beta PopoverMenu

simast:
- `src/beta/PopoverMenu` becomes new `src/PopoverMenu`
- Changed testkit export name from `PopoverMenuTestkit` to `popoverMenuTestkitFactory` for consistency with other testkit exports.
- Cleaned up all `/beta` related infra.


stories/storiesHierarchy.js
```diff
@@ -22,7 +22,6 @@ export const Category = {
   PICKERS_AND_SELECTORS: 'Design Guidelines|Pickers and Selectors',
   OTHER: 'Design Guidelines|Other',
   COMPONENTS: 'Components API|Components',
-  BETA: 'Components API|Next API',
```
EgoziE: Do we want to remove Beta section for good?

shlomitc: Let's kill it

PR #5044, opened 2020-02-23
`<Slider/>` - fixing tooltip bootom calculation

TalZa1810:
### 🔦 Summary
There was an issue with `Slider` within  `FormField`. The tooltip didn't display correctly.
This PR is to fix this.


TalZa1810:
fails on visuals which are not related.

PR #5047, opened 2020-02-23
(WSR8) <DatePicker/> - Replace error and errorMessage with status and status…

EgoziE:


simast:
Let's update TypeScript types:

https://github.com/wix/wix-style-react/blob/master/src/DatePicker/index.d.ts#L18-L19

src/DatePicker/index.d.ts
```diff
@@ -2,23 +2,24 @@ import * as React from 'react';
 import {WixComponentProps} from "../BaseComponents/WixComponent";
 import {CalendarProps} from "../Calendar";
 
+export type DatePickerState = 'error' | 'warning' | 'loading';
```
simast: would call this type `DatePickerStatus` so it matches the prop name. Note that we will start exposing these type aliases after WSR types are merged.

PR #5056, opened 2020-02-24
(WSR 8) - <DatePicker/> - removed isOpen prop

EgoziE:
`<DatePicker/>` - removed deprecated `isOpen` prop


src/DatePicker/DatePicker.js
```diff
@@ -184,7 +160,7 @@ export default class DatePicker extends WixComponent {
           this.openCalendar(e);
         }}
         onKeyDown={this._handleKeyDown}
-        tabIndex={this.state.isDateInputFocusable ? 1 : -1}
+        tabIndex={1}
```
simast: I think this is an oversight and should be `tabIndex={0}`

PR #5059, opened 2020-02-24
Guild week test - changes with index.d.ts - Please ignore 

zivpiz:
### 🔦 Summary

...

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test

### DIFF
--- a/src/Box/Box.js
+++ b/src/Box/Box.js
@@ -155,7 +155,7 @@ Box.propTypes = {
   /** Define styles through a classname */
   className: PropTypes.string,
   /** Defines if the box behaves as an inline element */
-  inline: PropTypes.bool,
+  inline: PropTypes.string,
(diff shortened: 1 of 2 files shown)


simast:
I am not sure I understand this PR - the only place where we use `inline` in the component is as a boolean truthy value. Why do we want to make it a string? If there is an issue with TypeScript types, convert it to boolean where Box is used:

```jsx
<Box inline={!!someStringValue} />
```

zivpiz:
@simast Hey, sorry that the title was not clear but these are just some test PR's we created for the guild week task we're on, so you can ignore these.
Thanks for the review though :)

PR #5064, opened 2020-02-24
(WSR 8) <ImageViewer/> - Remove deprecated props and testkit

EgoziE:
Deprecated props:
  - removed `tooltipPlacement`
  - changed `error` to `status='error'`
  - changed `errorMessage` to `statusMessage`

Deprecated testkit:
  - removed `isErrorVisible`
  - removed `getErrorTooltipContent`
  - added `hasStatus`
  - added `getStatus`
  - added `hasStatusMessage`
  - added `getStatusMessage`


src/ImageViewer/ImageViewer.uni.driver.js
```diff
@@ -93,11 +93,5 @@ export const imageViewerUniDriverFactory = (base, body) => {
     hasStatusMessage: () => statusIndicatorTestkit().hasMessage(),
     /** If there's a status message, returns its text value */
     getStatusMessage: () => statusIndicatorTestkit().getMessage(),
-
-    // Error - Deprecated
-    /** @deprecated */
-    isErrorVisible: () => errorTooltip.exists(),
```
simast: minor - we can also remove `errorTooltip` var as it is no longer used

src/ImageViewer/ImageViewer.st.css
```diff
@@ -49,10 +48,6 @@
   background-color: transparent;
 }
 
-.root:error {
```
simast: In addition to this error state we should also remove entire `/** error styles **/` section below

PR #5066, opened 2020-02-24
(WSR 8) <Input/> - remove magnifyingGlass prop

EgoziE:
`<Input/>`

- Props:
  - removed `magnifyingGlass`

- Testkit:
  - removed `hasMagnifyingGlass`
  - removed `clickMagnifyingGlass`


src/Input/InputSuffix.js
```diff
@@ -103,22 +97,6 @@ const InputSuffix = ({
       ),
       isVisible: suffixRules.inputHelpSuffix({ help, disabled }),
     },
-    {
-      component: () => (
-        <div
-          className={styles.magnifyingGlass}
```
simast: let's clean up styles as well

src/Input/Input.js
```diff
@@ -57,11 +57,6 @@ class Input extends Component {
   };
 
   logDeprecations(props) {
-    if (props.magnifyingGlass) {
-      deprecationLog(
-        `Input's magnifyingGlass prop is deprecated and will be removed in the next major release, please use suffix property with '<Input suffix={<Input.Affix><Search /></Input.Affix>}/>' instead`,
```
simast: just an observation that if you try this in playground:

```jsx
<Input suffix={<Input.Affix><Icons.Search /></Input.Affix>}/>
```
the icon will not be blue by default as with old behaviour. Not sure if this is an issue?

EgoziE: This deprecation is in favour of `<Search/>` component.
Adding a magnifyingGlass to input is discouraged by UX (I asked Ayelet)

EgoziE: I'll update the migration guide

PR #5074, opened 2020-02-24
(WSR 8) - <Popover/> & <InputWithOptions/> remove prop disableClickOutsideWhenClosed

EgoziE:
- Props:
  - removed `disableClickOutsideWhenClosed`, this functionality is now permanent


src/InputWithOptions/InputWithOptions.spec.js
```diff
@@ -421,18 +421,6 @@ describe('InputWithOptions', () => {
       expect(await dropdownLayoutDriver.isShown()).toBe(false);
     });
 
-    it('should trigger callback function on clicking outside', async () => {
-      const handleClickOutside = jest.fn();
```
simast: I would maybe move this spy and `onClickOutside={handleClickOutside}` to the test above (`should hide options on outside click`). And expand this test with assertions so we still cover `onClickOutside` prop with unit tests for InputWithOptions component.

EgoziE: I added a test 👍

PR #5077, opened 2020-02-25
(WSR 8) <Tag/> - remove wrap prop

EgoziE:
- Props:
  - removed `wrap` - text is now wrapped by default


MIGRATION.md
```diff
@@ -69,6 +69,11 @@ Removed `<Input.Units/>` use `<Input.Affix>` instead.
   - changed `error` to `status='error'`
   - changed `errorMessage` to `statusMessage`
 
+#### `<Tag/>`
+
+- Props:
+  - removed `wrap` - text is now wrapped by default
```
simast: I think the default is using `ellipsis ` - so it is not wrapped (but cut with an ellipsis)

EgoziE: fixed comment

PR #5082, opened 2020-02-25
Change ActionButton.link type to string

splintor:
According to [the docs](https://wix-style-react.now.sh/?path=/story/components-api-components--notification), `link` is a `string`, not a `boolean` (as it is passed to `href`).


simast:
Thanks, could you update our TS type tests as well?

https://github.com/wix/wix-style-react/blob/95f0bafa109dcf8c264e0094739877d98ea8fa49/test/types/Notification.tsx#L24

PR #5087, opened 2020-02-25
chore(testkit/index.d.ts): generate file with `wuf`

argshook:
This PR addresses `testkit/*.d.ts` file generation with `wuf export-testkits`

tl;dr; `testkit/*.d.ts` files are now generated, same as `testkit/*.js`

---

Although the diff is rather large, it should be relatively safe for merge.

process of how this was done:

1. copy existing `testkit/index.d.ts`, `testkit/enzyme.d.ts` & `testkit/puppeteer.d.ts` to `testkit/bak`
1. run `nodemon --ext ejs --exec "npm run update-components" --watch ".wuf/testkits/*-typescript.template.ejs"`
    this way each change on any of `*-typescript.template.ejs` triggers `npm run update-components`
1. run `diff testkit/index.d.ts testkit/bak/index.d.ts`
1. adjust template until no diff

this way we move from manual `testkit/*.d.ts` to generated but still ensure that generated ones look exactly as manual.

---

* remove `.wuf/generator/codemods/testkit*.js`, no longer needed as `npm run update` takes care of it now
* update template files:
  * `.wuf/testkit/vanilla-typescript.template.ejs`
  * `.wuf/testkit/enzyme-typescript.template.ejs`
  * `.wuf/testkit/puppeteer-typescript.template.ejs`
* add `testkit/*.d.ts` to be part of `npm run test:types`


argshook:
@mosheKerbel @simast this is hairy to review and i don't expect you to go through  all the lines, just want you to be aware of this change

.wuf/testkits/puppeteer-typescript.template.ejs
```diff
@@ -1,10 +1,67 @@
...
+      const camelName = utils.toCamel(component.name);
+      const isLegacy = component.puppeteerLegacyTestkit;
+      const importName = component.puppeteerTypeImportName || component.name + (isLegacy ? 'Driver' : 'UniDriver')
+      const importPath = component.puppeteerTestkitPath || component.testkitPath || '../src/' + component.name + '/' + component.name + (isLegacy ? '.puppeteer.driver' : '.uni.driver')
+      const wrappedImport = defaultExport.includes(component.name) ? importName : '{' + importName + '}'
+      const exportName = component.puppeteerTestkitExportName || camelName + 'TestkitFactory'
+      const testkitType = isLegacy ? 'PuppeteerTestkitFactory' : 'PuppeteerUniTestkitFactory'
+%>
+import <%= wrappedImport %> from '<%= importPath %>';
+export const <%= exportName %>: <%= testkitType %><<%= importName %><%- withHTMLElement.includes(component.name) ?  '<HTMLElement>' : ''%>>;<% }) %>
```
simast: hmm, I don't think puppeteer will return `HTMLElement` here? Must be some other type, let's double check UniDriver implementation of `getNative` for puppeteer

shlomitc:
I recommend adding a test in the ‘runtime-e2e’ in order to verify the script output can be consumed currently and won’t break

argshook:
I added generated .d.ts files to be part of `npm run test:types` which runs `tsc`. I reckon if there's a mismatching export or typing, tac would complain. Already caught some bugs with it. Maybe it's enough wdyt? @shlomitc

shlomitc:
This testkit file is the "output" of the library.
The `test:types` runs on the "source" of the library (.d.ts of each component)
Mixing the two types of testing is not a good practice.

Technically, the entire `testkit` folder is a `dist` folder (besides the polyfills) can be removed and generated only in a `pre-publish` step.

You can decide to run `tsc` in the `runtime-e2e` on the output files

.wuf/testkits/puppeteer-typescript.template.ejs
argshook: as discussed, this `HTMLElement` comes from previously manual scripts, i left it to not introduce any change. Can be investgated in separate PR imo

argshook:
@shlomitc i moved .d.ts check to `npm run test:e2e-runtime`, please see if this is what you had in mind

test-runtime/imports/tsconfig.declarations.json
```diff
@@ -0,0 +1,21 @@
...
+    "jsx": "react",
+    "strict": true,
+    "noEmit": true,
+    "noImplicitAny": true,
+    "noImplicitThis": true,
+    "strictFunctionTypes": true,
+    "strictNullChecks": true
+  },
+  "include": [
+    "../../testkit/*.d.ts"
```
shlomitc:
```suggestion
    "../../dist/testkit/*.d.ts"
```
it should point the `dist` folder

shlomitc:
@argshook yes, that's great. I think that later one we can take it a few steps forward:
1. tests - the `test-runtime` could be a typescript project importing a testkit and infer the types of it 
2. templates - move [these entires](https://github.com/wix/wix-style-react/pull/5087/files#diff-62d59d314919f75ac5f602f575ecda37R25-R51) to the `wuf` model instead of hardcoding it in the template

PR #5091, opened 2020-02-25
<NumberInput/> - Fix onChange type

simast:


src/NumberInput/index.d.ts
```diff
@@ -1,9 +1,11 @@
 import * as React from 'react';
 
+import { OmitPolyfill } from '../common';
 import {InputProps} from '../Input';
 
-export interface NumberInputProps extends InputProps {
+export type NumberInputProps = OmitPolyfill<InputProps, 'onChange'> & {
   strict?: boolean;
+  onChange?: (value: number) => void;
```
mosheKerbel:
```suggestion
  onChange?: (value: number | null) => void;
```
https://github.com/wix/wix-style-react/blob/62bc64a783678060db52a3657bf9165a4d12c991/src/NumberInput/NumberInput.js#L75

test/types/NumberInput.tsx
```diff
@@ -43,7 +43,7 @@ function NumberInputOldWithAllProps() {
       noLeftBorderRadius
       noRightBorderRadius
       onBlur={_ev => {}}
-      onChange={_ev => {}}
+      onChange={(value: number) => {}}
```
giladsegal: We need to see that typescript manages to infer the type of value + use `_` prefix so the linter won't complain about unused variable
```suggestion
      onChange={_value => {}}
```

simast: Hmm, I am not sure I agree with this - the test will not catch any issues if we leave untyped code like this. I could go to typings and change onChange argument type unintentionally and the tests would still say it's ok.

PR #5094, opened 2020-02-26
Improve Notification.ActionButton type definition

splintor:
Looking at the code, I realized that `Notification.ActionButton` isn't really a `ButtonWithAsProp` - it actually expose a very limited set of props, that depend on the `type` field.

This PR updates the types to reflect the actual way `ActionButton` works.


mosheKerbel:
thanks @splintor looks accurate 
@giladsegal would u mind taking a second look?
it looks like we use only specific props here, so it's wrong to expose more to the user:
https://github.com/wix/wix-style-react/blob/8c998e7a9de693b31614ff7f1e990e73b0882609/src/Notification/ActionButton/ActionButton.js#L20

PR #5099, opened 2020-02-26
(WSR 8) <LinearProgressBar/> - remove <Loadable/> and shouldLoadAsync prop

EgoziE:
Remove internal use of `<Loadable/>`

- Props:
  - removed `shouldLoadAsync`

Testkit:
  - removed `getTooltip`
  - removed `isTooltipShown`
  - changed `getTooltipErrorMessage` in driver (not uni) returns a `string` instead of a `Promise<string>`


src/LinearProgressBar/LinearProgressBar.js
```diff
@@ -57,40 +29,43 @@ class LinearProgressBar extends React.PureComponent {
...
+  errorMessage: PropTypes.string,
+
+  /** Use light theme instead of dark theme */
+  light: PropTypes.bool,
+
+  /** Use to display a percentage progress.*/
+  showProgressIndication: PropTypes.bool,
+
+  /** The number of the percentage progress */
+  value: PropTypes.number || PropTypes.string,
```
simast: 😬 let's fix this with `oneOfType`

src/StatusIndicator/StatusIndicator.uni.driver.js
```diff
@@ -3,28 +3,24 @@ import { tooltipDriverFactory } from '../Tooltip/TooltipNext/Tooltip.uni.driver'
...
 export const statusIndicatorDriverFactory = (base, body) => {
+  const tooltipDriver = tooltipDriverFactory(
+    base.$(`[data-hook="${dataHooks.tooltip}"]`),
+    body,
+  );
+
   return {
     ...baseUniDriverFactory(base, body),
 
+    tooltipDriver,
```
simast: I think we are trying to avoid exposing nested drivers? One way to solve this would be to use `StatusIndicator.private.uni.driver.js` for this - and use it internally with `LinearProgressBar`, but don't expose this from the new StatusIndicator component.

I might be not aware of the latest agreement here, is it ok to have nested drivers in new components?

EgoziE:
I talked with Shlomi
We decided to remove: `getTooltip` and `isTooltipShown` from testkit and not expose tooltip driver

PR #5101, opened 2020-02-26
Enable passing either x or y to PopoverMenu.moveBy

splintor:
In [our code](https://github.com/wix-private/sites-list/blob/master/src/client/pages/app/site-list/components/folder-actions-menu/component.tsx#L133), we pass either `x` or `y` to move `<PopoverMenu>` and with the new types, this fails, as it requires passing both `x` and `y`. This PR is meant to fix it (while also adding missing tests to beta `PopoverMenuNext`.


src/beta/PopoverMenu/index.d.ts
```diff
@@ -3,14 +3,15 @@ import {
...
+import { Offset } from '../../common';
 
 export interface PopoverMenuNextProps {
   triggerElement: React.ReactNode;
   children?: React.ReactNode;
   maxWidth?: number;
   minWidth?: number;
   zIndex?: number;
-  moveBy?: { x: number; y: number };
+  moveBy?: Offset;
```
simast: I would say let's just type it as:

```ts
moveBy?: { x?: number; y?: number };
```

This will match our PropType checks and other TS typings we currently have for the same prop (DropdownBase/Tooltip).

splintor: Don't we want to force users to specify at least one of `x` or `y`?

simast: Ideally yes, but we are not doing this right with other identical `moveBy` TypeScript types and PropType checks.

If you want to change this I think we should unify this everywhere (for both TypeScript and PropTypes). Alternatively consider using `moveBy?: { x?: number; y?: number };` right now so we don't have inconsistent definitions.

splintor: :+1:
(Mainly because I don't know how to require **at least one** in `PropType`.)

PR #5105, opened 2020-02-26
(WSR 8) <Page/> - Remove upgrade prop

EgoziE:
- Props:
  - removed `upgrade` - component is now upgraded by default


src/Page/docs/PageTestStoriesDeprecated.js
```diff
@@ -1,102 +0,0 @@
-import React from 'react';
```
simast: we should also remove this from `stories/e2e.js`:
https://github.com/wix/wix-style-react/blob/c73627a8765d2453a8ca6b63855cb20b31be78e8/stories/e2e.js#L8

src/Page/Page.js
```diff
@@ -542,8 +542,6 @@ Page.propTypes = {
 
   /** z-index of the Page */
   zIndex: PropTypes.number,
-  /** When true the page will use height: 100% and not require a parent of `display: flex;flex-flow: column;`. Also Page.Content's may grow using `min-height: inherit`. Supports Page.Sticky. New header minimization approach.*/
-  upgrade: PropTypes.bool, // This Upgrade prop is only for documentation, the actual use is in index.js
```
simast: let's also clean up another `upgrade` prop passed to PageHeader in this file:
https://github.com/wix/wix-style-react/blob/c73627a8765d2453a8ca6b63855cb20b31be78e8/src/Page/Page.js#L299

src/PageHeader/PageHeader.scss
```diff
@@ -90,10 +90,7 @@ See: https://css-tricks.com/flexbox-truncated-text/#article-header-id-3 */
     height: 30px;
     margin-top: 6px;
     margin-left: auto;
-
-    &.animationEnabled {
-      transition: margin 0.2s cubic-bezier(0.4, 0, 0.2, 1);
-    }
+    transition: margin 0.2s cubic-bezier(0.4, 0, 0.2, 1);
```
simast: Hmm, we were only applying this animation for the old (non upgraded) Page before. Are we enabling this for the new page as well?

EgoziE: I asked UX, disabling animation for now, we might add this in the future

PR #5111, opened 2020-02-26
sync ts test

mosheKerbel:
sync ts test

### 🔦 Summary

...

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


package.json
```diff
@@ -40,6 +40,7 @@
     "storybook:visual": "start-storybook -c ./.storybook/visual -p 6007",
     "storybook:e2e": "start-storybook -c ./.storybook/e2e -p 6008",
     "------test": "",
+    "prebuild": "if [ \"$agentType\" = \"pullrequest\" ]; then sync-ts; fi",
```
zivpiz: @mosheKerbel notice that there's already a prebuild in line 59, this is probably the reason why `wix-style-react-install` is green

shlomitc: I think we should add it as a "test" step - `test:types`
doing that means also it will run only on one machine and not hold back other parallels tasks

PR #5114, opened 2020-02-27
Chore(syncTS): Add SyncTS script to pre-push husky hook

zivpiz:
Guild Week PR - Add `sync-ts` tool to WSR repo.
Selected hook: `pre-push`.

Once this PR is merged, `pre-push` hook will prevent users from adding / deleting / updating components, without handling the matching `d.ts` files.

[Tool spec](https://docs.google.com/document/d/1rS0bSDB3niDJDZ84uCyIpa82uyPwHp8HrPtQ7GhiBGo/edit)


mosheKerbel:
after talking to @shlomitc  about it we won't use the pre static analysis stuff, to avoid annoying delays to the user.
I will modify it and add it to the "test:types" cycle.

PR #5116, opened 2020-02-27
Small changes - preparation to wsr8

EgoziE:
Just some small changes to help with wsr8

Changes:
`<Input/>` - remove remains of previously deleted styles
`<Input/>` and `<InputArea/>` - remove `onDoubleClick ` that points to undefined function
`<MultiSelect/>` - fixed driver d.ts
`<RichTextInputArea/>` and `<VariableInput/>` - removed internal use of ErrorIndicator in favour of StatusIndicator - **No component or testkit API were changed but will in wsr8**

### DIFF
--- a/src/RichTextInputArea/RichTextInputArea.js
+++ b/src/RichTextInputArea/RichTextInputArea.js
@@ -10,7 +10,7 @@ import RichTextToolbar from './Toolbar/RichTextToolbar';
 import EditorUtilities from './EditorUtilities';
 import { RichTextInputAreaContext } from './RichTextInputAreaContext';
 import { defaultTexts } from './RichTextInputAreaTexts';
-import ErrorIndicator from '../ErrorIndicator';
+import StatusIndicator from '../StatusIndicator';
 
 const decorator = new CompositeDecorator([
   {
@@ -156,9 +156,10 @@ class RichTextInputArea extends React.PureComponent {
           />
           {hasError && (
             <span className={styles.errorIndicator}>
-              <ErrorIndicator
+              <StatusIndicator
                 dataHook="richtextarea-error-indicator"
-                errorMessage={statusMessage}
+                status="error"
+                message={statusMessage}
               />
             </span>
           )}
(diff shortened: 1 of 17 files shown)


EgoziE:
**Note:** Variable input visuals should fail, the status indicator was not according to zeplin

src/RichTextInputArea/RichTextInputArea.js
```diff
@@ -156,9 +119,10 @@ class RichTextInputArea extends React.PureComponent {
           />
           {hasError && (
             <span className={styles.errorIndicator}>
-              <ErrorIndicator
+              <StatusIndicator
```
simast: The new error icon does not have the same paddings as the old one and looks out of place/not aligned:
Screenshot 2020-02-28 at 09 09 58

src/RichTextInputArea/RichTextInputArea.uni.driver.js
```diff
@@ -1,26 +1,35 @@
 import { baseUniDriverFactory } from 'wix-ui-test-utils/base-driver';
 import ReactTestUtils from 'react-dom/test-utils';
 
-import { errorIndicatorDriverFactory } from '../ErrorIndicator/ErrorIndicator.uni.driver';
+import { statusIndicatorDriverFactory } from '../StatusIndicator/StatusIndicator.uni.driver';
+import { statusTypes } from '../VariableInput/constants';
```
simast: looks weird that this is imported from `VariableInput`, don't think these components are somehow related

src/RichTextInputArea/RichTextInputArea.uni.driver.js
```diff
@@ -1,26 +1,35 @@
...
   return {
     ...baseUniDriverFactory(base, body),
     isDisabled: async () =>
       Boolean(await getContent(base).attr('contenteditable')),
-    hasError: async () => await getErrorIndicator(base).exists(),
+    hasError: async () => {
+      const exists = await statusIndicatorDriver.exists();
+      if (exists) {
+        const status = await statusIndicatorDriver.getStatus();
+        return status === statusTypes.error;
```
simast: Note that we only support error status indicator in RichTextInputArea - we should not need this check. If we remove this we can also remove constant import from VariableInput above.

If the idea is to add more status types in the future or in wsr8 branch - let's do that there and leave here simple logic.

PR #5119, opened 2020-02-29
<FloatingHelper/> - Fix TypeScript types

simast:
This PR simplifies `FloatingHelper` TypeScript types. I think initially we copied types from backoffice and this included types for ClosablePopover - which probably should not be exposed at all as it is an internal implementation of FloatingHelper (as far as wix-style-react is concerned).

After this clean-up users can import only the following:

```ts
import { FloatingHelper, FloatingHelperProps, FloatingHelperAppearance } from 'wix-style-react';
```

Also modified the types to accept more Popover props - since we spread all rest props on Popover here (and there are users already relying on this): 
https://github.com/wix/wix-ui-backoffice/blob/a985fbc1dde62b22a9db731522990f03d84576b6/src/components/FloatingHelper/ClosablePopover/ClosablePopover.tsx#L151-L152

### DIFF
--- a/src/FloatingHelper/index.d.ts
+++ b/src/FloatingHelper/index.d.ts
@@ -1,129 +1,40 @@
 import * as React from 'react';
-import {FloatingHelperContent} from "./FloatingHelperContent/FloatingHelperContent";
+import { FloatingHelperContent } from './FloatingHelperContent/FloatingHelperContent';
+import { OmitPolyfill } from '../common';
+
 export type FloatingHelperProps = import('wix-ui-core/dist/src/createHOC').WixComponentProps &
   FloatingHelperPropsInner;
 
-export default class FloatingHelper extends React.PureComponent<FloatingHelperProps> {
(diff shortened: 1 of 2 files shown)


TalZa1810:
Hey @simast ,
I'm working on migrating this component from `wix-ui-backoffice` so this should probably be part of this PR: https://github.com/wix/wix-style-react/pull/5057 . I'll add the types to be part of the migration PR.

simast:
@TalZa1810 I think this can be merged independently. The type fixes are valid for FloatingHelper as in current master and for your PR changes as well.

mosheKerbel:
@simast there are places which require to modify tests like add `moveBy` prop
I would like to wait with it and to merge your new changes with the types that @TalZa1810 is definning in her refactor PR.
I recommend we wait with this, and i will review the types in her PR and add there the content of this PR.

mosheKerbel:
done here - https://github.com/wix/wix-style-react/pull/5057

PR #5123, opened 2020-03-01
`<Popover/>` - adding d.ts to popover protractor driver

TalZa1810:
### 🔦 Summary
Adding protractor d.ts file to the `<Popover/>` driver. I need this for the `<FloatingHelper/>` migration.


giladsegal:
We decided not to support protractor drivers at all, why do you need it?

TalZa1810:
Currently, we export protractor drivers for `<FloatingHelper/>`. I need this `d.ts` file for the protractor drivers of `<FloatingHelper/>`. 
I'm migrating this component from `wix-ui-backoffice` to `wix-style-react` and I don't want to break our users.

PR #5126, opened 2020-03-01
Tiny changes

EgoziE:
Removed upgrade prop from Button in some stories
and added key to loops


src/Badge/docs/ExampleBadges.js
```diff
@@ -8,8 +8,8 @@ const sizesString = sizes.join(', ');
 const types = Object.keys(TYPE);
 const typesString = types.join(', ');
 
-const renderBadge = props => (
-  <span style={{ padding: '5px' }}>
+const renderBadge = (props, key) => (
```
simast: it does not look like the `key` param is passed in where `renderBadge` is called

src/Badge/docs/index.story.js
```diff
@@ -70,27 +70,35 @@ export default {
...
-          ].map(code),
+
+          code({
+            title: 'With icon',
+            description:
+              'Badge can contain icon as a prefix/suffix Icon size should match badge size. For a medium sized badge use normal icons. For a small badge use small icons which end with the prefix Small',
+            source: examples.withIcon,
+          }),
+
+          code({
```
simast: looks like duplicated example? we have two identical `'With icon'` examples here

Issue #5048, opened 2020-02-23
[RFC] Consuming wix-style-react in SSR projects


mykas:
# Consuming wix-style-react in SSR projects

## The Problem

Currently our users that use SSR in their project are kind of ok with the solution we provide for them to make wix-style-react work for their project. (Our FAQ explains what kind of parameter needs to be enabled in yoshi config to export stylable css and attach a hook for imports to work).

BUT, now that we are planning to migrate to named-imports in order to support ES modules infrastructure - our "quick-hacks" from docs are not enough anymore.

#### 1. CSS-Modules (sass)

`Server` renders an app to generate html with classnames - it will pick commonjs modules from `wix-style-react/dist/src` and the reason for this is because server is NODE based. 

```
<div classname="_-node_modules-wix-style-react-dist-src-Box-Box__root__3wDCc">
```

`Client` assets are beeing created during webpack build and added to html as external stylesheet. During the webpack build it will use ES-modules infrastructure.

```
// app.css
_-node_modules-wix-style-react-dist-es-src-Box-Box__root__bjtF6
```

> Notice we have two problems here:
> 1. CommonJs: `dist/src`, ESmodules: `dist/es/src`
> 2. CommonJs: `root__3wDCc`, ESmodules: `root__bjtF6`

```error
Warning: Prop `className` did not match. Server: "_-node_modules-wix-style-react-dist-src-Box-Box__root__3wDCc _-node_modules-wix-style-react-dist-src-Box-Box__horizontal__VpebB _-node_modules-wix-style-react-dist-src-Box-Box__middle__KW85s" Client: "_-node_modules-wix-style-react-dist-es-src-Box-Box__root__bjtF6 _-node_modules-wix-style-react-dist-es-src-Box-Box__horizontal__2lMkn _-node_modules-wix-style-react-dist-es-src-Box-Box__middle__3WiyU"
```

#### 2. Stylable

Same as with css-modules, but stylable does not generate full path in the classname. The only problem that hash is dependend on stylesheet path.

Server:
```
<div classname="WSRAvatar1045097907--placeholder">
```

Client:
```
// app.stylable.css
WSRAvatar2846992615--placeholder
```

> CommonJs: `1045097907`, ESmodules: `2846992615`

### The output

1. Webpage will have no styles applied, only html because of classname mismatch.
2. React will throw an error in the console.

## Possible Solutions

[...]

PR #5128, opened 2020-03-02
`css-modules` - fix ssr issues when used with named-imports

mykas:
This is WIP pr to fix the issue for SSR. Read more here: https://github.com/wix/wix-style-react/issues/5048


scripts/transpile/plugins/babel-plugin-sass-to-es.js
```diff
@@ -0,0 +1,37 @@
+const resolve = require('path').resolve;
+
+// This plugin transpiles sass imports for commonjs to imports from dist/es
+// this solves the problem for ssr where we have classname mismatch.
+
+module.exports = function() {
```
mykas: This is mainly the logic I do to change the path of sass import.

mykas:
I agree. It is already fragile enough so I will make sure that we have some kind of tests for this.

scripts/transpile/transpileSrc.js
```diff
@@ -40,26 +46,9 @@ const writeFileAsync = (fileLoc, dir, code) => {
   });
 };
 
-const copyAsync = ({ src, dist, withESTransform }) => {
+const copyAsync = ({ src, dist }) => {
   return new Promise((resolve, reject) => {
     copy(src, dist, function(err, files) {
-      files.forEach(file => {
```
mykas: Removing this here, because it was a duplicate task. The `replaceStylableImports` already does string replacement for the same cases.

scripts/transpile/plugins/babel-plugin-sass-to-es.js
```diff
@@ -0,0 +1,45 @@
+const path = require('path');
+
+// This plugin transpiles sass imports to imports from dist/es
+// this solves the problem for ssr where we have classname mismatch.
+
+function transpileSASS(api, state) {
```
argshook:
```suggestion
function transpileSASS(path, state) {
```

i'd use `path` here (and also lines 40, 41), cause that's what babel is passing here, it's the api of path object

scripts/transpile/__tests__/babel-plugins-sass-to-es.spec.js
```diff
@@ -0,0 +1,60 @@
+import transpileSASS from '../plugins/babel-plugin-sass-to-es';
+
+const oneLevel = {
```
argshook: these tests are cool, but maybe it would be easier to test by running the whole babel with your plugins? This way you don't need to get into hairy details of plugin api, it would be more like e2e test

scripts/transpile/tasks/replaceStylableImports.js
```diff
@@ -2,7 +2,7 @@ const fs = require('fs');
 const glob = require('glob');
 const path = require('path');
 
-const targetDir = path.resolve(__dirname, '..', 'dist/es/src');
+const targetDir = path.resolve(__dirname, '../../../', 'dist/es/src');
```
argshook: this is meh comment, but for the purists, there should be no '/' in `path.resolve`. It will add slashes for you. This way it will also work well on windows

```suggestion
const targetDir = path.resolve(__dirname, '..', '..', '..', 'dist', 'es', 'src');
```

scripts/transpile/tasks/transpileSrc.js
```diff
@@ -8,11 +8,17 @@ const mkdirp = require('mkdirp');
 const { parse } = require('@babel/parser');
 const { transformFromAstAsync } = require('@babel/core');
 
-const srcDir = path.resolve(__dirname, '../dist');
-const esDir = path.resolve(__dirname, '../dist/es');
+const srcDir = path.resolve(__dirname, '../../../dist');
```
argshook: maybe it would make sense to split sass and js tasks into separate ones?

scripts/transpile/tasks/transpileCopyFiles.js
```diff
@@ -0,0 +1,20 @@
+const execa = require('execa');
+
+const options = { stdio: 'pipe', env: { FORCE_COLOR: true } };
+
+module.exports = function({ name }) {
+  return new Promise((resolve, reject) => {
+    try {
+      const promise = execa.shell(
+        `babel ${name} --out-dir dist/${name} --copy-files --plugins=@babel/plugin-transform-modules-commonjs`,
```
argshook: other tasks are calling `babel` programatically, through js. but this one here outsources `babel` to the shell, is this ok? maybe for the sake of consistency it should be js here too?

scripts/transpile/plugins/babel-plugin-sass-to-es.js
```diff
@@ -0,0 +1,45 @@
...
+  const originalPath = api.node.source.value;
+  if (/[A-Za-z]*\.scss/.test(originalPath)) {
```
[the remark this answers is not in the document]
argshook: or, even better, if `sass` transpilation is split into its own task, then glob should ensure babel is run only on scss files. then no need for if/regex here at all

scripts/transpile/index.js
```diff
@@ -0,0 +1,56 @@
...
+      console.log(`🚀 Done in ${Math.round(new Date() - START_TIME) / 1000}s`),
+    description: 'Done!',
+  },
+];
+
+tasks.reduce(
+  (promise, { task, description, name }) =>
+    promise
+      .then(() => task()({ name }))
+      .then(() => progress.tick(STEP_WIDTH, { dir: description }))
```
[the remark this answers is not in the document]
mykas: I thought that first too, but then what It does is basically show for example 80% done even though the part from 60-80% is still in progress. The description then make sense though.

scripts/transpile/index.js
```diff
@@ -0,0 +1,56 @@
+/* eslint no-console: 0 */
+const ProgressBar = require('progress');
+
+const STEPS = 5;
```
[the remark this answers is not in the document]
mykas: It probably can

scripts/transpile/tasks/transpileCopyFiles.js
mykas: I agree

scripts/transpile/index.js
```diff
@@ -0,0 +1,56 @@
...
+);
+
+const tasks = [
+  {
+    task: () => require('./tasks/cleanDist'),
+    description: 'cleaned ./dist',
+  },
+  {
+    task: () => require('./tasks/transpileCopyFiles'),
+    name: 'testkit',
```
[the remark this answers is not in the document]
mykas: This is needed for `transpileCopyFiles` function to know which folder to copy and transpile. I have renamed it to `folder` so it would make sense.

scripts/transpile/plugins/babel-plugin-sass-to-es.js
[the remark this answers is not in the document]
mykas: Will check for `scss` extension only. As we talked before this is not possible to split because this part is done on `.js` files with babel transformation tool. It make sense to have this in one place alongside the other babel plugin that changes `wix-ui-core/dist/src` to `wix-ui-core/dist/es/src`

scripts/transpile/__tests__/babel-plugins-sass-to-es.spec.js
mykas: I will try to refactor the testing part

mykas:
Hey, I am merging this and leaving some tests to implement later, because I was not able to solve issues with babel while running jest. 

But I double checked this solution with storybook:named-imports and also on wix-style-react-mobile project. Considering that this did not have any tests before - I think its good enough

PR #5133, opened 2020-03-03
(WSR 8) <AddItem/> - remove tooltipProps prop

EgoziE:
- Props:
  - removed `tooltipAppendTo` use `tooltipProps` instead
  - removed `tooltipContent` use `tooltipProps` instead
  - removed `tooltipFixed` use `tooltipProps` instead
  - removed `tooltipFlip` use `tooltipProps` instead
  - removed `tooltipPlacement` use `tooltipProps` instead

This PR is related to the Tooltip PR but I decided to split it


src/ImageViewer/ImageViewer.js
```diff
@@ -45,7 +50,10 @@ class ImageViewer extends Component {
         dataHook={dataHooks.addItem}
         disabled={disabled}
         tooltipContent={addImageInfo}
-        tooltipProps={{ ...tooltipProps, content: addImageInfo }}
+        tooltipAppendTo={tooltipProps.appendTo}
+        tooltipFixed={tooltipProps.fixed}
+        tooltipFlip={tooltipProps.flip}
+        tooltipPlacement={tooltipProps.tooltipProps}
```
simast: :bug:

simast: we should also update `ImageViewer/index.d.ts` types for `tooltipProps` using `Pick<>` - since not all tooltip props are accepted now.

EgoziE: I'm not sure we should limit this prop, it is used in two places.
1. Part of AddItem - line 47
2. Part of the main tooltip - line 166
What do you say?

simast: Oh I see now. Well this is really strange then - some of the props in `tooltipProps` will only apply to some of the tooltips in the ImageViewer component now.

I am curious, maybe you know more details, but why did we deprecate `tooltipProps` on AddItem? We have many other places in the library where we pass tooltip props like that and they are not marked as deprecated..

Also note that there is a bug here with:
> tooltipPlacement={tooltipProps.tooltipProps}

EgoziE: I'll ask UX if they know something
(and thanks for the bug)

testkit/protractor.js
```diff
@@ -50,7 +50,7 @@ const load = module => {
 };
 
 export const accordionTestkitFactory = protractorUniTestkitFactoryCreator(load(require('../src/Accordion/Accordion.uni.driver')));
-export const addItemTestkitFactory = protractorTestkitFactoryCreator(load(require('../src/AddItem/AddItem.protractor.driver')));
+export const addItemTestkitFactory = protractorTestkitFactoryCreator(load(require('../src/AddItem/AddItem.driver')));
```
simast: We are removing `src/AddItem/AddItem.protractor.driver`? I think we should leave it.. even if we want to remove it mapping it here to jsdom based `AddItem.driver` seems incorrect.

src/AddItem/AddItem.spec.js
```diff
@@ -68,7 +68,7 @@ describe('AddItem', () => {
 
     describe('Tooltip', () => {
       it('should appear [when] `theme` image prop is passed', async () => {
-        const props = { tooltipContent: 'content', theme: 'image' };
+        const props = { tooltipProps: { content: 'content' }, theme: 'image' };
```
simast: lets update examples and docs as well to use `tooltipProps: { content: 'content' }` instead of `tooltipContent `

testkit/protractor.js
EgoziE: I checked and no one uses it
what is the correct way?

simast: Have to use unidriver then. To enable unidriver for AddItem protractor testkit remove this line (and run `npm run build` again):
https://github.com/wix/wix-style-react/blob/671dcd691f51b69b2a381e348cc5f3bee6daf930/.wuf/testkits/definitions.js#L487

EgoziE:
@simast Last commit was to replicate the deleted Tooltip HOC file
I asked UX, in the future AddItem component will be re-developed with new responsive design.

src/AddItem/AddItem.js
```diff
@@ -125,22 +109,34 @@ class AddItem extends Component {
...
 
     const container = (
-      <div {...style('content', { theme, size, alignItems, disabled })}>
+      <div
+        {...style('content', {
+          theme,
+          size,
+          alignItems,
+          disabled,
+          tooltip: tooltipProps.content,
```
simast: I would convert this to boolean (`!!tooltipProps.content`). Just so so we are not adding some unnecessary stuff to DOM (in Stylable states are `data-*` attributes).

PR #5139, opened 2020-03-04
docs(*): use `allComponents` instead of `baseScope`

argshook:
This PR is a cleanup of `baseScope` reference leftover.

live code playground solutions, be it `code`, `example` sections or `<LiveCodeExample/>` require `components` or `scope` to be passed.

It is needed in order to render components in live playground. Without
this object, live playground would yell about undefined references.

Long long time ago this object was exported from `LiveCodeExample`
component source file as `baseScope`.

Later on a dedicated `stories/utils/allComponents.js` file was created
for this purpose.

However, both `allComponents` and `baseScope` were left.

To increase consistency, `baseScope` is removed in this PR and only `allComponents` is left


argshook:
if build is green, it's safe to merge. Because if any of the references would be broken, `build-storybook` would fail.

PR #5141, opened 2020-03-04
`<FloatingHelper/>`- improving FloatingHelper components stories

TalZa1810:
### 🔦 Summary
Improving the component story of both `<FloatingHelper/>` and `<FloatingHelper.Content/>`

### ✅ Checklist

- 📚 Change is documented
  - [x] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


TalZa1810:
Hi @mykas ,
I'm working on the `<FloatingHelper/>` documentation (both the `<FloatingHelper/>` and `<FloatingHelper.Content/>` documentation.
The implementation was changed a bit. Now we use the `wix-style-react` `<Popover/>` component. When implementing the `Controlled example`, I noticed that now it works slower than it used to be. Currently, we haven't released the migrated example so what we have live is the old implementation: https://wix-style-react.now.sh/?path=/story/components-api-components--floatinghelper

This is the migration PR: https://github.com/wix/wix-style-react/pull/5057/files

This is the `wix-ui-backoffice` implementation: https://github.com/wix/wix-ui-backoffice/blob/master/src/components/FloatingHelper/ClosablePopover/ClosablePopover.tsx

I'm not sure what causes this, maybe because we're using the `wix-style-react` Popover and not the one from `wix-ui-core`? wdyt?

I hope my explanation is clear :)

mykas:
Hey @TalZa1810 , I have an idea why it is so. It is related to `timeout` prop that is being passed to Popover.  

You see, this is actually not a problem but rather how it works when `timeout` prop works properly. It was meant for animation timeout and it works only if animation classnames has been implemented. If you check this in wix-style-react Popover: https://github.com/wix/wix-style-react/blob/master/src/Popover/Popover.st.css#L146  and its basically same thing but missing in wix-ui-backoffice. 

What it means is that timeout value never actually worked in wix-ui-backoffice and it was as timeout was set to 0.

Question is: do we really need this animation on FloatingHelper? For now you can just set timeout to 0 and it will be the same effect as in wix-ui-backoffice.

src/FloatingHelper/docs/index.story.js
```diff
@@ -1,134 +1,156 @@
-import * as React from 'react';
+import React from 'react';
```
mosheKerbel: description does not look good
image

description should look like this example (FilleButton)
image

src/FloatingHelper/docs/index.story.js
```diff
@@ -1,134 +1,156 @@
...
 import FloatingHelper from '..';
+import { placementOptions } from '../constants';
```
mosheKerbel: I would remove it and put an example of the `open` and `close` instance method as part of the examples and in the **bottom** of the examples.
This is not an intuitive way to use the component and i would not recommend use it, so yeah, keep maintaining it but not advertise it much.
image

mosheKerbel: to be more precise, i see that you already added such example
i'd add as a description of the `Works only in Uncontrolled mode`:
`This option of controlling the popover is not recommended, but is still supported for backward compatibility.`

src/FloatingHelper/FloatingHelperContent/docs/index.story.js
```diff
@@ -1,31 +1,130 @@
 import React from 'react';
 import FloatingHelperContent from '..';
+import Box from '../../../Box/Box';
```
mosheKerbel: same design comment as before
image

TalZa1810:
visual tests fail is irrelevant to the change. As for the types, I'm not sure why it fails. The propTypes was incorrect so I've fixed this but it seems like it fails on types.
wdyt @mosheKerbel  ? can i merge this?

PR #5151, opened 2020-03-08
Docs - <Multiselect/> tags prop

marik22312:
### 🔦 Summary

Fixed `Multiselect` tags docs

### ✅ Checklist
- 📚 Change is documented
  - [X] Story
  - [X] API description
- 🔬 Change is tested
  - [X] Component


marik22312:
@TalZa1810 Ok, so maybe we should expand the object definition...
the `MultiSelect` component doesn't work when there's no "label" key in the arrays objects

PR #5160, opened 2020-03-10
[Tag] allow Tag to accept any type of children

ipanasenko:
What was a reason for restricting it? In our code we render translated strings inside `Tag`, like this:

```ts
<Tag id="new-search">
  <FormattedMessage id="empty-state.media-from-wix-no-results-new-search" />
</Tag>
```


src/Tag/index.d.ts
```diff
@@ -1,7 +1,6 @@
 import * as React from 'react';
 
 export interface TagProps {
-  children: string;
```
itaytay: Why remove it and not just replace it with React.ReactNode ?
cc @mosheKerbel

ipanasenko: because it's ReactNode by default

ipanasenko: But it's optional by default, so we need to specify it explicitly as ReactNode if children are required

mosheKerbel: right, children is an exception and by default exist on the type.

mosheKerbel:
> What was a reason for restricting it? In our code we render translated strings inside `Tag`, like this:
> 
> ```ts
> <Tag id="new-search">
>   <FormattedMessage id="empty-state.media-from-wix-no-results-new-search" />
> </Tag>
> ```

you are right, indeed node should be the allowed type.

PR #5166, opened 2020-03-11
<ColorInput/> - error to status - preparation to wsr8

EgoziE:
Added status and statusMessage
in wsr8 error and errorMessage will be removed


src/ColorInput/ColorInput.js
```diff
@@ -1,5 +1,13 @@
 import React from 'react';
-import { node, bool, string, func, oneOf, oneOfType, object } from 'prop-types';
+import PropTypes, {
```
TalZa1810: Why destructing if you already import `PropTypes` ?

src/ColorInput/ColorInput.spec.js
```diff
@@ -289,6 +289,23 @@ describe('ColorInput', () => {
       });
     });
 
+    describe('with status', () => {
+      [
+        { status: 'error', statusMessage: 'Error Message' },
+        { status: 'warning', statusMessage: 'Warning Message' },
+        { status: 'loading', statusMessage: 'Loading Message' },
+      ].forEach(test => {
```
TalZa1810: using `it.each()` is a nicer way in my opinion. You can check out the example here as a reference:
https://github.com/wix/wix-style-react/blob/master/src/ToggleButton/test/ToggleButton.spec.js#L18

src/ColorInput/ColorInput.uni.driver.d.ts
```diff
@@ -1,15 +1,21 @@
...
-    confirm: () => Promise<void>,
-    clickColorViewer: () => Promise<void>,
-    enterText: (text: string) => Promise<void>,
-    getValue: () => Promise<string>,
-    hasError: () => Promise<boolean>,
-    getPlaceholder: () => Promise<string | null>,
-    getSize: () =>  Promise<InputSize | null>,
-    isDisabled: () => Promise<boolean>,
-    colorPickerVisible: () => Promise<boolean>,
+  cancel: () => Promise<void>;
```
TalZa1810: I think the syntax should be aligned like the one we have in the generator:
https://github.com/wix/wix-style-react/blob/master/.wuf/generator/templates/%24ComponentName/%24ComponentName.uni.driver.d.ts

I remember that @argshook explained to me in a different PR about the reason for this specific syntax but I don't remember what it was  :)

src/ColorInput/ColorInput.js
```diff
@@ -15,9 +23,13 @@ class ColorInput extends React.Component {
...
     disabled: bool,
-    /** sets error state */
+    /** Sets UI to indicate a status */
+    status: PropTypes.oneOf(['error', 'warning', 'loading']),
+    /** The status message to display when hovering the status icon, if not given or empty there will be no tooltip */
+    statusMessage: PropTypes.node,
+    /** @deprecated - use status prop instead */
     error: bool,
-    /** error message which appears in tooltip */
+    /** @deprecated - use statusMessage prop instead */
```
TalZa1810: I think that we should use the `deprecationLog ` for this prop as we've used in your favourite `Tooltip` component:
https://github.com/wix/wix-style-react/blob/master/src/Tooltip/index.js#L4

src/ColorInput/ColorInput.uni.driver.d.ts
argshook: i think it was just that eslint or prettier or something was fixing it to the simpler syntax, that's all :)

PR #5171, opened 2020-03-11
Build cleanup script

EgoziE:
As a preparation to wsr8 I created a **build cleanup script** to delete all files generated from `npm run build` and `npm run storybook:build`.

wsr8 moved, changed, and removed a lot of files and there's a lot unnecessary files that can fail a rebuild and should be deleted.

When the time comes and we merge wsr8, or you switch to wsr8 branch,  just run `npm run build:cleanup` and then `npm run build` and `npm run storybook:build`.


package.json
```diff
@@ -67,6 +67,7 @@
     "build-puppeteer-testkit": "webpack --config ./testkit/webpack.config.js",
     "export-components": "import-path --path src --dts",
     "storybook:build-visual": "build-storybook -c ./.storybook/visual",
+    "build:cleanup": "./scripts/build-cleanup.sh",
```
argshook: should it maybe be a a prebuild step? so that it runs automatically without users needing to know about it

scripts/build-cleanup.sh
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
+
+EXCLUDE_FILES="
```
argshook: i think all this script can be done with `git clean -fdx --exclude=node_modules`

`man git-clean`

```
git-clean - Remove untracked files from the working tree
```

```
-f, --force
    If the Git configuration variable clean.requireForce is not set to false, git clean will refuse to delete files or directories unless given -f or -i. Git will
    refuse to modify untracked nested git repositories (directories with a .git subdirectory) unless a second -f is given.
```

maybe we don't need the `-d` but it shouldn't be harmful

```
-d
    Normally, when no <path> is specified, git clean will not recurse into untracked directories to avoid removing too much. Specify -d to have it recurse into such
    directories as well. If any paths are specified, -d is irrelevant; all untracked files matching the specified paths (with exceptions for nested git directories
    mentioned under --force) will be removed.

```

```
-x
    Don’t use the standard ignore rules (see gitignore(5)), but still use the ignore rules given with -e options from the command line. This allows removing all
    untracked files, including build products. This can be used (possibly in conjunction with git restore or git reset) to create a pristine working directory to test a
    clean build.
```

scripts/build-cleanup.sh
```diff
@@ -0,0 +1,56 @@
...
+"
+VAR="-name "
+for ELEMENT in $EXCLUDE_FILES; do
+  VAR+="${ELEMENT} -or -name "
+done
+VAR+="."
+
+# Delete files on root directory, except excluded.
+echo Delete build files
+find . -depth 1 -type f -not \( $VAR \) -delete
```
argshook: for `find` with `-not` it's usually recommended (by my made-up recommendation
commitee) to also use `-prune`. That way whatever is specified in the
`-not` part will not even be visited by consecutive commands,
potentially making performance better.

for `-delete` it's not that important, but if you ever want to `grep` or
`sed` or something, this can considerbly make things faster.

for example

```
find . -not \( -path ./lib -prune \) -name '*.svg.js' 
```

package.json
EgoziE: I wasn't sure at the beginning because CI always makes a clean install, but you're right, I'll change it.

scripts/build-cleanup.sh
EgoziE: Unfortunately there are untracked files you do want to keep.
like `applitools.private.config.js`

EgoziE: Do you think there's a way to use it and keep some files?

argshook: yes, use `--exclude`

`git clean -fdx --exclude node_modules`

can also use it multiple times

`git clean -fdx --exclude node_modules --exclude applitools.private.config.js`

scripts/build-cleanup.sh
EgoziE: I'm not sure I understand you..
The $VAR is a long list separated by a space 
I tried:
```
find . -depth 1 -type f -not \( -path $VAR -prun \)
```
and it finds nothing.. maybe I misused it

the -delete is for removing the files found

EgoziE: anyway, with `git clean -fdx --exclude node_modules --exclude applitools.private.config.js` I might not need this at all

argshook: should be `-prune`, not `-prun`

but i see, you can ignore me, the `-prune` is more relevant when you do deep `find`s and possibly combine output of `find` with some other command through `xargs`.

so it's not really related in this case :)

PR #5176, opened 2020-03-12
<VariableInput/> types test

EgoziE:
Added types tests  for `<VariableInput/>` and `<StatusIndicator/>`


test/types/StatusIndicator.tsx
```diff
@@ -0,0 +1,19 @@
...
+  return (
+    <StatusIndicator
+      className="class-name"
+      dataHook="data-hook"
+      status="error"
+      message="message"
+      tooltipPlacement="top"
+    />
+  );
+}
```
TalZa1810: I think that you've forgot to test the testkits:
https://github.com/wix/wix-style-react/blob/master/test/types/Accordion.tsx#L34

PR #5177, opened 2020-03-12
<FloatingHelper/> - testkit definitions

EgoziE:
@TalZa1810


TalZa1810:
fails on visuals which are not related

PR #5182, opened 2020-03-12
contribution stories conventions

itaytay:
### 🔦 Summary

...

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


EgoziE:
It might help contributors understand this sentence better if you provide a link to an existing relevant component as an example.

itaytay:
@egozi - yeah I know I’ll add it on Sunday. Just opened this PR real quick before going home so I won’t forget.
I’ll probably rephrase it a bit too
Thanks

itaytay:
@EgoziE - I added an example, and I think the text is clear enough as is. Wdyt? anything you would change?

TalZa1810:
Hey @itaytay ,
As far as I understood from Arijus, we want to encourage our users to use the new storybook API - `example` rather than using the `code` API (the old one). Moreover, the convention is `index.story.js`. You can have a look in the generator as a reference:
https://github.com/wix/wix-style-react/blob/master/.wuf/generator/templates/%24ComponentName/docs/index.story.js

itaytay:
@TalZa1810  - regarding the new API you're right, but I didn't refactor it yet.
The example can be refactored of course.
Regarding the naming, note that this is a design guidelines story, and not a components api story.
I think the naming should also represent that, and @nitayneeman thought this naming convention could be nice for the design guidelines too. Wdyt?

TalZa1810:
@itaytay 
Currently, all UX story files are named as `index.story.js`. They are listed under the `stories` directory. For example the UX story `Text Input`:
https://github.com/wix/wix-style-react/blob/master/stories/TextInput/index.story.js

If we want UX stories to have a different naming, I think that we should align all files to have the same naming.

docs/contribution/DOCUMENTING_COMPONENTS.md
```diff
@@ -52,6 +52,10 @@ export default {
 }
 ```
 
+Every new comoponent requires a component story in the Components API section.
+If the UX symbol is a composite of more than 1 component, it also requires a Design Guidelines story.
+Example can be found [here] (https://github.com/wix/wix-style-react/blob/master/src/CustomModalLayout/docs/design.story.js)
```
shlomitc: as discussed, UX symbols and guidelines should sit in the `stories` folder, please change

itaytay: Updated. 
This example (Tag input) also uses the new 'example' api and was recently updated by @milkyfruit so I think it's a good one to use

itaytay:
@shlomitc - waiting for you response after the update

shlomitc:
I think the explanation still lacks some information. What is a UX symbol (better name it Component’s Sketch symbol)? What is a Design Guideline story and how does it looks like?

I think that the best thing to do is explaining with an example. For instance. Take 3.1 Input and explain the compositions of FormField and Input. What code examples should appear in each one of them.

I would also put this as a new title instead of appending it to the “components api” one

itaytay:
@shlomitc - I updated the PR. Is this what you meant?

docs/contribution/DOCUMENTING_COMPONENTS.md
```diff
@@ -52,6 +52,21 @@ export default {
...
+For example, [`Tag Input`](https://wix-style-react.now.sh/?path=/story/design-guidelines-inputs--3-12-tag-input) is a composition of 2 individual components – [`<FormField/>`](https://wix-style-react.now.sh/?path=/story/components-api-components--formfield), and [`<MultiSelect/>`](https://wix-style-react.now.sh/?path=/story/components-api-components--multiselect).
+Both of these individual components have their Component API stories.
+
+`Tag Input` has a Design Guidelines story, to provide info and exampls on how to implement this Component Sketch symbol as a composition.
+
+### Design Guidelines key parts
+ - Description
+ - Included Components: a list of all individual components used in the examples
+ - Examples: showcase the different basic variants
+Code example can be found [here] (https://github.com/wix/wix-style-react/blob/master/stories/MultiSelect/index.story.js)
```
shlomitc:
```suggestion
Code example can be found [here](https://github.com/wix/wix-style-react/blob/master/stories/MultiSelect/index.story.js)
```

itaytay:
@shlomitc - ok great, I committed your suggestion of course, and will merge once all tests are green

PR #5185, opened 2020-03-12
<InputArea> - propagate onEnterPressed event

### LINKED ISSUE #5184: <InputArea /> create newline on every Enter keypress
# ✨ Feature Request

### 🔦 Scope
```<InputArea />``` creates newline on every ```Enter``` keypress, even with ```onEnterPressed``` props.

### 💁 Explanation
We need add some way to have more control for that behaviour.(For example to not create new line)

### 🐾 Possible solution 
We can propagate event to ```onEnterPressed``` and then we can use ```e.preventDefault``` in callback handler if we need it

### 👀 Severity

- Low

kiriltaran:
### 🔦 Summary

Add event argument for onEnterPressed props(callback)

## Issue
Resolves #5184

### DIFF
--- a/src/InputArea/InputArea.js
+++ b/src/InputArea/InputArea.js
@@ -227,7 +227,7 @@ class InputArea extends React.PureComponent {
     this.props.onKeyDown && this.props.onKeyDown(e);
 
     if (e.key === 'Enter') {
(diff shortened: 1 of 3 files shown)


kiriltaran:
I have 1 failed test, but don't have access to applitools.

PR #5190, opened 2020-03-16
Avoid "any" in Table types

splintor:
### 🔦 Summary
Avoid using `any` in `<Table>` (and `<DataTable>`) type definition.
The main change is making `<Table>` and `<DataTable>` generic and enable to pass the row data type, to make it type-safer. To be backward compatible, `object` is used as the default type for row data.

Other changes are setting the type of `setSelectedIds` parameter, and making ref fields be on `HTMLElement` instead of `any`.

### ✅ Checklist

- 🔬 Change is tested
  - [X] Type


morsh:
[!image](https://media.giphy.com/media/3ov9kbdtuww5Jh74ac/giphy.gif)

mosheKerbel:
Thanks @splintor! 
It looks great.
the only thing I would like to keep the same and not change from `any` is the ref forwarding parts. It's a bit complicated to follow through the different code parts and I'm afraid others will get broken because of this change.

@giladsegal do you mind taking a second look?

giladsegal:
Prop types should never be generic because when we render a component we **never** pass the prop types as generic. So using generics with default type will always result in the default type being used. The only We to get around it is to use a highly non-standard syntax:
```tsx
<DataTable<MyPropsType>>
    ...
</DataTable>
```

mosheKerbel:
@splintor i changed the default row from `object` to  `any` as the fallback
and the refs as i said, keeping as any.

splintor:
@mosheKerbel Great, thanks!
Sorry I didn't have time to handle it myself.

BTW, while you are at it, want to handle [my other WSR PR](https://github.com/wix/wix-style-react/pull/5194) as well?

mosheKerbel:
> @mosheKerbel Great, thanks!
> Sorry I didn't have time to handle it myself.
> 
> BTW, while you are at it, want to handle [my other WSR PR](https://github.com/wix/wix-style-react/pull/5194) as well?

Thanks for the other PR, will review it.
We aim to review such small PRs in a day, maximum 2.
If it is not the case like in this one,  please ping us in the channel.

PR #5191, opened 2020-03-16
`<Button />` - ellipsed text

frontned:
### 🔦 Summary

Make Button text ellipsed when it is too long.

### ✅ Checklist

- 📚 Change is documented
  - [X] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test

### DIFF
--- a/src/Button/Button.js
+++ b/src/Button/Button.js
@@ -2,6 +2,7 @@ import React, { PureComponent } from 'react';
 import { ButtonNext } from 'wix-ui-core/dist/src/components/button-next';
 import { generateDataAttr } from '../utils/generateDataAttr';
 import { isMadefor } from '../FontUpgrade/utils';
+import ellipsisHOC from '../common/EllipsisHOC';
 
 import styles from './Button.st.css';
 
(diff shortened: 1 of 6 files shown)


src/Button/Button.js
```diff
@@ -85,10 +86,21 @@ class Button extends PureComponent {
         )}
         data-hook={dataHook}
       >
-        {children}
+        <EllipsedContent ellipsis>{children}</EllipsedContent>
```
mykas: We usually give our users ability to determine when he needs the `ellipsis` by exposing ellipsis prop. By this prop we have the mechanism to dynamically load only when `ellipsis={true}`

src/Button/Button.js
```diff
@@ -85,10 +86,21 @@ class Button extends PureComponent {
...
 
+const EllipsedContent = ellipsisHOC(({ children, ...rest }) => (
+  <span {...rest}>{children}</span>
+));
+
+EllipsedContent.propTypes = {
+  /** String based node */
+  children: node,
+
+  ...ellipsisHOC.propTypes,
```
mykas: This place is not gonna work properly. You should do it like this: `tooltipProps: PropTypes.shape(ellipsisHOC.propTypes)`

src/Button/docs/index.story.js
```diff
@@ -113,6 +113,13 @@ export default {
                 `,
             source: examples.custom,
           }),
+
+          example({
+            title: 'Ellipsis',
+            text:
+              'If a button container is too small or you set a fixed with to a button and the text does not fit, it will have an ellipsis with a tooltip.',
```
mykas: Change it to: `If a Button parent container is too small and the text does not fit then Button can have ellipsis enabled by passing ellipsis="true"`

mykas:
Ok, so because ellipsis adds some KB's to this bundle, please rise to 11 Kb in perfer config for the Button

mykas:
Closing for now as it require a bit more effort and talking within the team how it should be handled.

PR #5194, opened 2020-03-16
<Box/> - Enable passing numeric sizing values

splintor:
### 🔦 Summary
The documentation and types say sizing props (`*width` and `*height`) can have numeric values, which should be specified in pixels, but numeric values are actually ignored.

This PR fixes it by adding `px` suffix in case a numeric value is specified (similar to how spacing values are processed).

### ✅ Checklist

- 🔬 Change is tested
  - [X] Component
  - [X] Visual test


mosheKerbel:
@splintor you missed the `$` sign in the `format` func, luckily your tests caught it and failed.
I added a commit and rebased to master.
once it pass in CI I will merge.

mosheKerbel:
But now I see that all the Visual tests of `box` are failing, need to check that.

mosheKerbel:
My bad, visual tests are also passing, all green
Probably saw an old build.

splintor:
Thanks!

PR #5202, opened 2020-03-17
<ToggleButton/> - size and labelPlacement props

yevhen-kutishchev:
### 🔦 Summary

Add `size` and `labelPlacement` props according to spec https://app.zeplin.io/project/5c8f6f1f9f489505cac756dc/screen/5d80e274b4379354beebd33f

### ✅ Checklist

- 📚 Change is documented
  - [x] Story
  - [x] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [x] Visual test


src/ToggleButton/ToggleButton.js
```diff
@@ -43,37 +53,113 @@ class ToggleButton extends Component {
...
+      'data-selected': selected,
+      'data-skin': skin,
+    };
+
+    if (tooltipContent) {
+      deprecationLog(
+        '"tooltipContent" prop is deprecated. Use "labelValue" instead.',
+      );
+
+      if (!labelValue) {
```
mykas: I don't think this needs to be here. Just pass this to the Tooltip prop directly like this:

```js
   <Tooltip
          upgrade
          {...tooltipProps}
          size="small"
          content={labelValue || tooltipContent}
          data-hook={dataHook}
          data-placement={labelPlacement}
        >
```

src/ToggleButton/ToggleButton.js
```diff
@@ -43,37 +53,113 @@ class ToggleButton extends Component {
...
+            data-hook="togglebutton-trigger"
+            onClick={onClick}
+          >
+            {children}
+          </ButtonNext>
+        </Tooltip>
+      );
+    }
+
+    if (labelPlacement === 'end') {
```
mykas: I believe there are too many ifs here. Why not use stylable state instead of switching between `.root` and `.bottomLabelRoot` ? Two many duplicate `if`s can cause feature missing on of them.

src/ToggleButton/ToggleButton.js
```diff
@@ -43,37 +53,113 @@ class ToggleButton extends Component {
...
         >
-          {children &&
-            cloneElement(children, {
-              width: childSize,
-              height: childSize,
-            })}
+          {children}
+
+          {labelValue ? (
+            <span
```
mykas: Also I did not get why `end if` has a span and `bottom` uses <Text/>` component

mykas: ok I get, you thought that because when `end` is used - it suppose have common white background with a button, but actually this is only a background color which can be switched with stylable state

src/ToggleButton/ToggleButton.js
```diff
@@ -43,37 +53,113 @@ class ToggleButton extends Component {
...
+            data-hook="togglebutton-trigger"
+          >
+            {children}
+          </ButtonNext>
+          <Text
+            {...styles('bottomLabelText', { disabled })}
+            size="tiny"
+            weight="thin"
+            ellipsis
+            data-hook="togglebutton-label"
```
mykas: Here it should be `dataHook` instead

src/ToggleButton/ToggleButton.js
```diff
@@ -43,37 +53,113 @@ class ToggleButton extends Component {
...
+        >
+          <ButtonNext
+            {...styles('bottomLabelButton', { selected, size })}
+            {...commonButtonProps}
+            data-hook="togglebutton-trigger"
+          >
+            {children}
+          </ButtonNext>
+          <Text
+            {...styles('bottomLabelText', { disabled })}
```
mykas: Text component supports `disabled` skin. No need to use stylable on top

src/ToggleButton/ToggleButton.js
yevhen-kutishchev: In case with `end` I used span instead of `<Text />` to make label inherit color style from ButtonNext component. If I would use `<Text />` then I would have to add styles to make label have appropriate color (the way I did with `bottom` placement label text).

mykas: But can't you replicate the same styleing with Text props? from zeplin I see that it is possible

src/ToggleButton/ToggleButton.js
```diff
@@ -1,22 +1,82 @@
...
+    return (
+      children && (
+        <Tooltip
+          {...{ dataHook: 'toggleButtonTooltip', ...tooltipProps }}
+          upgrade
+          size="small"
+          content={labelValue}
+          disabled={tooltipDisabled}
+        >
+          {/* This additional span is needed make focus states work correctly on descendant <div> element. Otherwise Tooltip will override those props. */}
```
mykas: no need for the comment here

src/ToggleButton/ToggleButton.js
```diff
@@ -25,56 +85,93 @@ class ToggleButton extends Component {
...
-        content={tooltipContent}
+      <ButtonNext
+        {...styles('root', { disabled, selected, skin, labelPlacement })}
+        tabIndex={-1}
         data-hook={dataHook}
+        data-placement={labelPlacement}
+        data-selected={selected}
+        data-skin={skin}
+        onClick={disabled ? null : onClick}
+        {...rest}
```
mykas: this should be moved to the first place. Otherwise use will be able to override our props

src/ToggleButton/ToggleButton.uni.driver.js
```diff
@@ -2,10 +2,20 @@ import { buttonNextDriverFactory } from 'wix-ui-core/drivers/unidriver';
...
 export const toggleButtonDriverFactory = (base, body) => {
-  const tooltipDriver = tooltipDriverFactory(base, body);
+  const tooltipBaseElement = base.$('[data-hook="toggleButtonTooltip"]');
+  const tooltipDriver = tooltipDriverFactory(tooltipBaseElement, body);
+  const buttonDriver = buttonNextDriverFactory(base);
 
-  const buttonBaseElement = base.$('[data-hook="togglebutton-trigger"]');
-  const buttonDriver = buttonNextDriverFactory(buttonBaseElement);
+  function getLabelPlacement() {
+    return body
```
mykas: why do you use document.body hear? What if user will have many other components given during the tests?

src/ToggleButton/index.d.ts
```diff
@@ -5,11 +5,19 @@ import { OmitPolyfill } from '../common';
...
   selected?: boolean;
   disabled?: boolean;
   dataHook?: string;
+  labelValue?: React.ReactNode;
+  labelPlacement?: 'tooltip' | 'bottom' | 'end';
+  labelEllipsis?: boolean;
+  /** @deprecated use labelValue instead */
   tooltipContent?: React.ReactNode;
-  tooltipProps?: OmitPolyfill<TooltipNewProps, 'size' | 'content' | 'dataHook' | 'upgrade'>;
+  tooltipProps?: OmitPolyfill<
```
mykas: You can import tooltipCommonProps typings already from `./src/common.d.ts`. There you should find `popoverCommonProps` which applies also for tooltip

src/ToggleButton/ToggleButton.uni.driver.js
yevhen-kutishchev: It doesn't work with `base`, cause when I use `base` it's searches inside of `base` but data attribute I'm searching is in fact located on base element. That's why I decided to use body. But what you saying making sense.
What if I do something like `base.getNative().parentNode`?

src/ToggleButton/index.d.ts
yevhen-kutishchev: I don't see them there

src/ToggleButton/ToggleButton.uni.driver.js
yevhen-kutishchev: Forget what I wrote above :)
Using `base` totally works.

src/ToggleButton/index.d.ts
mykas: You should probably need to rebase with master as it was added recently

src/ToggleButton/ToggleButton.st.css
```diff
@@ -29,70 +28,155 @@
...
 }
 
-.root::content {
-  line-height: 0;
-  width: 100%;
-  height: 100%;
-  display: flex;
-  align-items: center;
-  justify-content: center;
+.root:skin(dark):active:not(:disabled),
```
mykas: Why do you need to target `:not(:disabled)` ?

src/ToggleButton/ToggleButton.st.css
```diff
@@ -29,70 +28,155 @@
...
+.root:skin(dark):labelPlacement(bottom):selected:not(:disabled) .label {
+  color: value(D10);
+}
+.root:skin(dark):active:not(:disabled) .icon,
+.root:skin(dark):selected:not(:disabled) .icon {
+  background-color: value(D10);
 }
 
-/* skin Dark */
+.root:skin(dark):active:hover:not(:disabled),
```
mykas: Here you target like this: when `active` when `hover` but not `disabled`. Is this really the case?

src/ToggleButton/ToggleButton.st.css
yevhen-kutishchev: Yes, I think so. I don't want to apply this style when it's disabled.

src/ToggleButton/ToggleButton.st.css
yevhen-kutishchev: Yes

mykas: You mean that there exist active + hover case? When its active its active

yevhen-kutishchev: Ok, I see what you mean. It already was there when I started this task. I just left it there. Should I remove it?

mykas: yep, lets clean it up a bit :) also not disabled too

src/ToggleButton/ToggleButton.st.css
mykas: there is no need, this should be handled by the browser anyway

yevhen-kutishchev: I've checked, if I remove `:not(:disabled)` then disabled button changes it's color when pressed.

mykas: ok, then leave it

mykas:
Hey @milkyfruit  please review this PR storybook where our contributor added examples for Label and Size

milkyfruit:
@yevhen-kutishchev @mykas the update looks good.

There's just one small tweak needed in the example:
- [ ] Update size code example
For tiny and small sizes the small icon must be used: `<Icons.CropRotateSmall />`

Other issues that might not be part if this PR, but nice if were updated:
- [ ] We must deprecate skin "dark", so let's hide it from 2 examples:
- Skin
"Toggle supports 1 skin style – `standard`
Skin `dark` is deprecated."
- Selected

- [ ] Update skin `standard` hover styles on mouse click ([remove blue background](https://zpl.io/VY73XWp))

PR #5209, opened 2020-03-18
Dynamic Chunks - naming

mykas:


src/CircularProgressBar/CircularProgressBar.js
```diff
@@ -85,7 +85,9 @@ class CircularProgressBar extends React.PureComponent {
             Tooltip: () =>
               // TODO: convert to WSR Tooltip
               shouldLoadAsync
-                ? import('../Tooltip/TooltipNext')
+                ? import(
```
shlomitc: I think @EgoziE removes this in WSR v8. also this might not be relevant anymore once the dynamically loaded Tooltip is integrated.
WDYT?

src/common/EllipsisHOC/index.js
```diff
@@ -45,28 +40,24 @@ const fallbackEllipsis = {
   whiteSpace: 'noWrap',
 };
 
-const LazyEllipsisHOC = loadable(() => retry(() => import('./EllipsisHOC')));
+const LazyEllipsisHOC = React.lazy(() =>
+  retry(() => import(/* webpackChunkName: "EllipsisHOC" */ './EllipsisHOC')),
```
shlomitc: In general, I think we should name all of our chunks with some common prefix like `wsr`, let's see if there are recommended naming convnetions

src/CircularProgressBar/CircularProgressBar.js
mykas: Its true, but for now lets have it in this version at least.

src/common/EllipsisHOC/index.js
mykas: I'll do a research on naming

src/CircularProgressBar/CircularProgressBar.js
mykas: Also one more thing. This here never actually worked. It always resolved to `require` because webpack can't parse `shouldLoadAsync` during build so it just leaves this untouched.

EgoziE: That's kinda funny...
Anyway we removed `shouldLoadAsync` completely in wsr8

PR #5210, opened 2020-03-18
<Tooltip/> - legacy driver

EgoziE:
After Moshe's review on migrating to wsr8 we decided to create a legacy driver to Tooltip next


mosheKerbel:
@EgoziE what am i missing here? 
it looks like a breaking change but on master
what was once a unidriver is now a driver?

shlomitc:
Same comments as in this one: https://github.com/wix/wix-style-react/pull/5211#issuecomment-600863666
What is the motivation?

EgoziE:
Closing for now

PR #5211, opened 2020-03-18
<Button/> - legacy driver

EgoziE:
After Moshe's review on migrating to wsr8 we decided to create a legacy driver to Button


mosheKerbel:
looks good
as i see it, users should not use this teskits on their own, but get it from our nested drivers in edge cases.
so, IMO, the drivers and types are necessary but the testkit is not.
@shlomitc wdyt?

shlomitc:
before I dive into the code - can you please explain what is the motivation for it?
As I see it - users are already using the existing driver, which is async (unidriver). Why do you want to change it?

EgoziE:
> before I dive into the code - can you please explain what is the motivation for it?
> As I see it - users are already using the existing driver, which is async (unidriver). Why do you want to change it?

I was trying to follow your comment in wsr8 branch, I might have had a different picture in my head.
When you suggested to write a legacy driver for Button, you meant a sync driver, but you don't want it to be exposed, only used internally?

EgoziE:
@mosheKerbel 
@shlomitc 
Can you please re-check it?

mosheKerbel:
you also need to switch it here to use it 
https://github.com/wix/wix-style-react/blob/fb03fd25907bf5bd8eef0ee96a44d1220507735a/src/ModalSelectorLayout/ModalSelectorLayout.driver.js#L30

mosheKerbel:
great
plz update the d.ts as well :)

PR #5215, opened 2020-03-19
<Checkbox/> - Fix getLabel in testkit

EgoziE:
GetLabel will now:
Returns the content of the checkbox. If the content is textual, a text will be returned. If the content is a single node, it will be returned as a node, else a list of nodes.


EgoziE:
fixed in another way

PR #5228, opened 2020-03-22
(WSR 8) migration guide

mosheKerbel:
Hey @mykas 
I saw that you changed some stuff and added some good notes.
Still, some of them i think we should skip
Because i want to use short and clear examples where possible and not write too much:

I think this one is clear enough, that's why i suggest to remove it
This codemode should be able to take care of these:	
```diff	
- import TextButton from 'wix-style-react/TextButton';	
- import TextButton from 'wix-style-react/dist/src/TextButton';	
- import TextButton from 'wix-style-react/TextButton';	
- import PopoverMenu, {PopoverMenuProps} from 'wix-style-react/beta/PopoverMenu';	
```

This example does not make sense
```
But it will not handle namespace based imports and this should be handled manually:
import * as SomeIconButton from 'wix-style-react/IconButton';
```

I don't know how useful this step is, since they are going to fail over a lot of other errors due to TS compliation.
```
Testing after migration
To double check that everything works well, just make sure to run yoshi build or any build tool like storybook to see if there are any errors thrown.
```


mykas:
Hey, sure if thats the goal to rely on ts compiler then you can remove it.

mykas:
I would like to though have a troubleshooting part in the migration guide, because lots of users will start getting stylable warnings about the duplicate namespaces. Mostly these are unhandled imports, testkits imported in the client code or having external dependencies that included commonjs imports. Is that possible to have in the migration guide? I expect many people asking about this and would like to have one place to link to

mykas:
This is important because otherwise users might get bigger bundle size because of unhandled imports in their external deps

mykas:
Hm now that I though more about it...unhandled imports are less of a problem in wsr8 because of deprecated cherry picking. 

And teskit part probably can be fixed if we migrate our teskits not to use stylable files 🤔

The only issue might be just external deps. Nor sure how to communicate that

mosheKerbel:
> I would like to though have a troubleshooting part in the migration guide, because lots of users will start getting stylable warnings about the duplicate namespaces. Mostly these are unhandled imports, testkits imported in the client code or having external dependencies that included commonjs imports. Is that possible to have in the migration guide? I expect many people asking about this and would like to have one place to link to

Yes i agree, it is indeed necessary to create a `troubleshooting` section in the bottom of the migration guide.

mosheKerbel:
> Hm now that I though more about it...unhandled imports are less of a problem in wsr8 because of deprecated cherry picking.
> 
> And teskit part probably can be fixed if we migrate our teskits not to use stylable files 🤔
> 
> The only issue might be just external deps. Nor sure how to communicate that

Do you mean drivers which are still looking for stylable selectors instead of data attirbutes?
`skin == styles.dark` instead of `data-skin='light'`?

mykas:
> > Hm now that I though more about it...unhandled imports are less of a problem in wsr8 because of deprecated cherry picking.
> > And teskit part probably can be fixed if we migrate our teskits not to use stylable files 🤔
> > The only issue might be just external deps. Nor sure how to communicate that
> 
> Do you mean drivers which are still looking for stylable selectors instead of data attirbutes?
> `skin == styles.dark` instead of `data-skin='light'`?

yes, exactly. I was able to find at least 4 testkits. So now I will try to migrate those in our current version. Shlomi last time migrated only the unidrivers for sled users, but regular drivers got left.

PR #5229, opened 2020-03-22
`<Badge/>` - migrate regular testkit to not use stylable && deprecate getSufixIcon and getPrefixIcon

mykas:
Hey, in this PR I am migrating regular driver to not use the stylable stylesheet for testing. Even though this is not hurting our Sled users anymore, but it does hurt our users that import regular drivers in their storybook as it might also import wrong stylesheet when users are actually using es modules.

Also this PR is deprecating `getSuffiIcon` and `getPrefixIcon`. This is usually a bad practise to return element for our users and in uni driver we returned a unidriver element which is not a known API that user understands.


mykas:
well, we can never know 🤷‍♂ Only by searching in wix-private org, but the abstraction level that user can have makes it hard to really know if it is used. Lets leave it with deprecation in version 7 and delete it in wsr8.

PR #5231, opened 2020-03-22
`<Text/>` - migrate regular driver to not use stylesheet for testing

mykas:
This PR is migrating regular driver to not use stylable stylesheets.


shlomitc:
If this is something to be done to all drivers, I suggest adding the legacy drivers glob to our [lint rule](https://github.com/wix/wix-style-react/blob/master/package.json#L363-L386) and just create one PR that removes them all

mykas:
There were only 4 elements that had stylabelDOMUtil usage.

PR #5233, opened 2020-03-22
(WSR 8) - <Label/> - remove component

EgoziE:
Removing label component
(next PR removing wix-ui-backoffice dependency)


EgoziE:
Checked - Failing tests are unrelated.

EgoziE:
> Don't forget to add this to the migration guide

done

MIGRATION.md
```diff
@@ -544,6 +545,10 @@ Props:
 - removed `disableClickOutsideWhenClosed`, this functionality is now permanent <br/>
   Just remove this prop, no other change required.
 
+## \<Label/>
+
+This component was deleted, use native label instead.
```
shlomitc: I don't think that's the appropriate solution.
We should advise on using either:
1. `<FormField/>`
2. `<Text tagName="label">`

EgoziE: Thanks! I'll write a better example.

PR #5234, opened 2020-03-22
Take content out of selection area in RadioButton

DanielEliraz:
🔦 Summary
This is the description of RadioGroup selectionArea prop:

> Selection area emphasises the clickable area, none means no emphasis, hover is when the mouse is on the component, and always will show constantly

While the RadioButton content prop description is:

> node to be rendered under label. Clicking it will not trigger onChange

So I fixed that the content will be out of the selectionArea.


EgoziE:
Under UX discussion

src/RadioGroup/RadioButton/RadioButton.driver.js
```diff
@@ -1,16 +1,19 @@
...
 const radioButtonDriverFactory = ({ element }) => {
-  const radioButton = () => element.childNodes[0];
-  const label = () => element.childNodes[1];
+  const radioWrapper = () => element.childNodes[0];
+  const radioButton = () => radioWrapper().childNodes[0];
+  const label = () => radioWrapper().childNodes[1];
 
   return {
-    exists: () => !!element,
+    exists: () => !!radioWrapper(),
```
EgoziE: function exists is reserved for checking the element is not null (which in this case element.childNodes[0] will throw an error
Why do you need to check this child element specifically?

src/RadioGroup/RadioButton/RadioButton.driver.js
```diff
@@ -1,16 +1,19 @@
...
     check: () => ReactTestUtils.Simulate.change(radioButton()),
     isChecked: () => radioButton().checked,
     isDisabled: () => radioButton().disabled,
     getLabel: () => label().textContent,
-    getTabIndex: () => element.getAttribute('tabIndex'),
+    getLabelElement: () => label(),
+    getValue: () => radioButton().value,
```
EgoziE: Would you please add tests demonstrating usage of these new testkit functions?

src/RadioGroup/docs/examples.js
```diff
@@ -19,3 +19,22 @@ export const selectionArea = `
     <RadioGroup.Radio value={2}>Option 2</RadioGroup.Radio>
   </RadioGroup>
 `;
+
+export const withContent = `
```
EgoziE: Nice example, I think that the use case here will be demonstrated better with a selection area and node content:
```
  <RadioGroup value={1} selectionArea="always">
    <RadioGroup.Radio
      key={0}
      value={1}
      content={<Box margin={2}>Another details for option 1, not clickable</Box>}
    >
      Option 1
    </RadioGroup.Radio>
    <RadioGroup.Radio
      key={1}
      value={2}
      content={<Box margin={2}>Another details for option 2, not clickable</Box>}
    >
      Option 2
    </RadioGroup.Radio>
  </RadioGroup>
```

src/RadioGroup/docs/index.story.js
```diff
@@ -45,6 +45,25 @@ const exampleChildren = [
       </RadioGroup.Radio>,
     ],
   },
+  {
+    label: 'Button with content',
+    value: [
+      <RadioGroup.Radio
+        key={0}
+        value={1}
+        content={'Another details for option 1, not clickable'}
```
EgoziE: Same comment as above:
<Box margin={2}>Another details for option 1, not clickable</Box>

src/RadioGroup/docs/index.story.js
```diff
@@ -45,6 +45,25 @@ const exampleChildren = [
...
+      >
+        <div>Option 1</div>
+      </RadioGroup.Radio>,
+      <RadioGroup.Radio
+        key={1}
+        value={2}
+        content={'Another details for option 2, not clickable'}
```
EgoziE: Same comment as above:
<Box margin={2}>Another details for option 2, not clickable</Box>

src/RadioGroup/test/RadioGroup.visual.js
```diff
@@ -88,17 +88,33 @@ const tests = [
...
+          },
+        },
+      },
+    ],
+  },
 ];
 
 tests.forEach(({ describe, its }) => {
-  its.forEach(({ it, props }) => {
+  its.forEach(({ it, props, radioButtonProps = {} }) => {
```
EgoziE:
```suggestion
  its.forEach(({ it, props = {}, radioButtonProps = {} }) => {
```

src/RadioGroup/test/RadioGroup.visual.js
```diff
@@ -88,17 +88,33 @@ const tests = [
       },
     ],
   },
+  {
+    describe: 'with content',
+    its: [
+      {
+        it: 'in 2 and 3 radios',
+        radioButtonProps: {
```
EgoziE: Good visual test, I recommend adding a selection area:

```suggestion
        props: {
          selectionArea: 'always',
        },
        radioButtonProps: {
```

src/RadioGroup/RadioGroup.scss
```diff
@@ -256,3 +256,7 @@ button.radioButton {
     background-color: $B60;
   }
 }
+
+.content {
+  margin-top: 6px;
```
EgoziE: Is this a mandatory margin by design?
I think this should be style-less and users should decide for themselves their margin according the their design

DanielEliraz: That's what Ayelet wanted

src/RadioGroup/RadioButton/RadioButton.driver.js
DanielEliraz: It is used in the RadioGroup.driver.js

src/RadioGroup/RadioGroup.scss
EgoziE: cool

PR #5235, opened 2020-03-22
(WSR 8) - <MessageBoxFunctionalLayout/> - remove green theme

EgoziE:
Props:
- removed value `green` in prop `theme` - will fall back to `blue`.


src/MessageBox/FunctionalLayout/HeaderLayout.js
```diff
@@ -34,7 +34,7 @@ HeaderLayout.propTypes = {
   title: PropTypes.node,
   onCancel: PropTypes.func,
   closeButton: PropTypes.bool,
-  theme: PropTypes.oneOf(['red', 'green', 'blue', 'lightGreen', 'purple']),
+  theme: PropTypes.oneOf(['red', 'blue', 'purple']),
```
mykas: I see that HeaderLayout is actually exposed in our index.js but we don't probably advertise this, right? Maybe worth deleting from main index.js  🤔

mykas: @shlomitc I don't have full story on our HeaderLayout component on reason why it is exposed. This PR is probably not the place to do it, but maybe we should consider exports like `HeaderLayout`, `HeaderLayout1`, `FooterLayout`, `FooterLayout1`  to be removed: 
Screenshot 2020-03-22 at 15 57 45

shlomitc: We don't want to expose them, so from my side, we can just remove it

EgoziE: can be done in master I guess

mykas: I am not sure about the master. I know at least two teams who uses it.

PR #5237, opened 2020-03-22
[wsr8] Migration.md - add an example of white Tooltip

mykas:


MIGRATION.md
```diff
@@ -1204,6 +1204,58 @@ The following props were changed/removed and require you to modify your code acc
 - `fixed` - whether to enable the fixed behaviour. This behaviour is used to keep the Tooltip at it's original placement even when it's being positioned outside the boundary.
 - `flip` - whether to enable the flip behaviour. This behaviour is used to flip the Tooltips placement when it starts to overlap the target element.
 
+### Deprecated theme `white`
+
+Tooltip no longer supports theme white, but white skin is available on Popover which is a building block for current Tooltip.
+
+Here is an example of same functionality as Tooltip but with white skin:
```
mykas: @shlomitc
We considered `white` Tooltip as bad practise (coming from UX), but there are still places where white tooltip is the only choice like on `dark` backgrounds. Though this is a rare case. Do you think we should write about white tooltip possible implementation?

PR #5238, opened 2020-03-22
(WSR 8) - <Badge/> - remove testkit functions

EgoziE:
Testkit:
- removed `getPrefixIcon` - use your own dataHook on prefix element instead.
- removed `getSuffixIcon` - use your own dataHook on suffix element instead.


src/Badge/Badge.spec.js
```diff
@@ -105,24 +105,38 @@ describe('Badge', () => {
...
+        expect(content.querySelector(`[data-hook="${dataHook}"]`)).toBeTruthy();
       });
 
       it('should have suffix icon', async () => {
+        const dataHook = 'dataHook';
         const driver = createDriver(
-          <Badge suffixIcon={<Email />}>Hello</Badge>,
+          <Badge
+            suffixIcon={
+              <div data-hook={dataHook}>
```
shlomitc: I think you can just pass `data-hook` to the `Email` without the div but not critical

src/Badge/Badge.driver.js
```diff
@@ -2,6 +2,8 @@ const badgeDriverFactory = ({ element, eventTrigger }) => {
   return {
     /** checks if element exists */
     exists: () => !!element,
+    /** returns the element */
+    element: () => element,
```
shlomitc: I don't think we need to expose this, why should we?

src/Badge/Badge.spec.js
EgoziE: for some reason it didn't work, I'll try it again

MIGRATION.md
```diff
@@ -238,6 +239,48 @@ This component was deleted, use `<FormField/>` instead.
   ```
 </details>
 
+## \<Badge/>
+
+This component was deleted, use `<FormField/>` instead.
```
shlomitc: That's incorrect

PR #5239, opened 2020-03-22
(WSR 8) - Remove wix-ui-backoffice dependency

EgoziE:
Remove wix-ui-backoffice dependency ❗️


testkit/all-components.js
```diff
@@ -10,7 +10,6 @@ const NON_COMPONENT_FOLDER = [
   'mixins',
   'common',
   'FieldLabelAttributes',
-  'Backoffice',
```
shlomitc: This one is not related I think, but I didn't find this in code so I don't think it will harm anything

EgoziE: 🎃

PR #5240, opened 2020-03-22
(WSR 8) - <ModalSelectorLayout/> - use button unidriver

EgoziE:
Use `<Button/>` unidriver in `<ModalSelectorLayout/>` driver


EgoziE:
tests failures are unrelated

src/ModalSelectorLayout/ModalSelectorLayout.protractor.driver.js
```diff
@@ -1,36 +1,17 @@
...
-    getButtonTextContent: () => element.getText(),
-    isButtonDisabled: () => hasAttribute(element, 'disabled'),
-    isPrefixIconExists: () => element.$('[data-hook="btn-prefix"]').isPresent(),
-    isSuffixIconExists: () => element.$('[data-hook="btn-suffix"]').isPresent(),
-    element: () => element,
-  };
-
-  return mergeDrivers(publicDriver, focusableDriver);
-};
+import { buttonDriverFactory } from '../Button/Button.uni.driver';
```
mykas: Actually our uni driver does not have  all the methods that were previously exposed. These methods are not present in current uni driver:
```
isPrefixIconExists: () => element.$('[data-hook="btn-prefix"]').isPresent(),	
isSuffixIconExists: () => element.$('[data-hook="btn-suffix"]').isPresent(),	
element: () => element,
```

- `element` method should be definitely.
- `isPrefixIconExists` and `isSuffixIconExists` also should be deprecated as we encourage our users to put their own data-hook on their elements and test it.

EgoziE: `isPrefixIconExists` and `isSuffixIconExists` are deprecated and were removed in another PR so it's ok
and `element` function exists in `baseUniDriverFactory` :)

mykas: oh, right `element` is from baseDriver

PR #5241, opened 2020-03-22
<Text/> - add onClick function to driver

mosheKerbel:
only existed on uniDriver but we still expose the old driver for the js-dom react
so i added it also there per consumer request.


mykas:
Just wondering - does it make sense to have a click on a span? 

I mean that suppose to be just a text block.  

Also it would interesting to see a case where user suppose to click on text without having underline or color of blue just like TextButton :)

mosheKerbel:
> Just wondering - does it make sense to have a click on a span?
> 
> I mean that suppose to be just a text block.
> 
> Also it would interesting to see a case where user suppose to click on text without having underline or color of blue just like TextButton :)

Yep
That's what Egozi also suggested

Closing for now and suggested user to use TextButton instead.

PR #5242, opened 2020-03-22
<AddItem/> - Added deprecation log

EgoziE:
Preparation to WSR8
Added deprecation log to `<AddItem/>`'s deprecated props


src/AddItem/AddItem.js
```diff
@@ -87,6 +87,24 @@ class AddItem extends Component {
...
+    [
+      'tooltipAppendTo',
+      'tooltipFlip',
+      'tooltipFixed',
+      'tooltipContent',
+      'tooltipPlacement',
+    ].forEach(prop => {
+      if (props.hasOwnProperty(prop)) {
+        deprecationLog(
+          `do not use '${prop}' prop, use 'tooltipProps' prop instead.`,
```
shlomitc:
```suggestion
          `do not use '${prop}' prop, use 'tooltipProps' prop instead.`,
          `<AddItem/> - do not use '${prop}' prop, use 'tooltipProps' prop instead.`,
```

PR #5243, opened 2020-03-22
<GoogleAddressInput/> - Added deprecation log and documentation

EgoziE:
Preparation to WSR8
Added deprecation log to <GoogleAddressInput/>'s deprecated props


src/GoogleAddressInput/GoogleAddressInput.js
```diff
@@ -39,6 +40,12 @@ class GoogleAddressInput extends React.Component {
     this.onFocus = this.onFocus.bind(this);
     this.onSet = this.onSet.bind(this);
     this.onManuallyInput = this.onManuallyInput.bind(this);
+
+    if (props.hasOwnProperty('error') || props.hasOwnProperty('errorMessage')) {
+      deprecationLog(
+        'props error and errorMessage are deprecated. Please use status="error" and statusMessage',
```
EgoziE:
```suggestion
        '<GoogleAddressInput/> - props error and errorMessage are deprecated. Please use status="error" and statusMessage',
```

src/GoogleAddressInput/index.d.ts
```diff
@@ -12,6 +14,8 @@ export interface GoogleAddressInputProps
   valuePrefix?: string;
   countryCode?: string;
   value?: string;
+  status?: GoogleAddressInputStatus;
```
mykas: Just an example of better pattern that I learned from @simast :
```
import { InputProps } from '../Input';
.
.
status?: InputProps['status'];
statusMessage?: InputProps['statusMessage'];
.
.

EgoziE: nice 
I didn't think of it

PR #5245, opened 2020-03-23
[wsr8] Remove redundand exports 

mykas:


src/index.js
```diff
@@ -76,22 +63,9 @@ export { default as DragDropContextProvider } from './DragDropContextProvider';
 export { default as SortableList } from './SortableList';
 export { default as NestableList } from './NestableList';
 export { default as FilePicker } from './FilePicker';
-export {
-  Container,
-  RawContainer,
```
mykas: Hey @shlomitc  from the storybook docs I don't see RawContainer examples, we can probably delete this, right?

src/index.js
```diff
@@ -76,22 +63,9 @@ export { default as DragDropContextProvider } from './DragDropContextProvider';
...
-  AutoAdjustedColumns,
-  AutoAdjustedColumns as AutoAdjustedRow,
-  Col,
-} from './Grid';
-
+export { Container, Row, Col, AutoAdjustedColumns } from './Grid';
 export { Layout, Cell } from './Layout';
-export {
-  default as Typography,
-  convertFromUxLangToCss,
```
mykas: Exports `convertFromUxLangToCss` and `convertFromCssToUxLang` looks a bit weird and I can't remember anybody asking for those

src/index.js
```diff
@@ -161,7 +135,6 @@ export { default as VerticalTabsItem } from './VerticalTabsItem';
 export {
   default as ListItemSection,
   listItemSectionBuilder,
-  TYPES as ListItemSectionTypes,
```
mykas: Hey @EgoziE I will remove this as we already have a different pattern for type exports and those are exported from d.ts files

src/index.js
```diff
@@ -76,22 +63,9 @@ export { default as DragDropContextProvider } from './DragDropContextProvider';
 export { default as SortableList } from './SortableList';
 export { default as NestableList } from './NestableList';
 export { default as FilePicker } from './FilePicker';
-export {
-  Container,
-  RawContainer,
-  Columns,
-  Columns as Row,
-  AutoAdjustedColumns,
```
mykas: Also these I have never seen anybody using it

src/index.d.ts
```diff
@@ -149,7 +149,7 @@ export * from './GoogleAddressInput';
 export { default as GooglePreview } from './GooglePreview';
 export * from './GooglePreview';
 
-export * from './Grid';
```
mykas: Hey @mosheKerbel  if we export like this: `export *` does that mean that user will get autocomplete for additional components exported from `Grid` even though we don't have them exported in the `index.js` ?

src/index.js
EgoziE: 👍

src/index.d.ts
mosheKerbel: If you export in the d.ts some class that was not exported in the js, so it will explode when the user tries to use it
but for types and interfaces it works.

if you remove here the `import * from ./Grid` then all the export types here won't be exported to the user nomore :(
https://github.com/wix/wix-style-react/blob/15cbb3f059e6c5220e16bb94749cb2e96576aa8e/src/Grid/index.d.ts#L9

src/index.js
mosheKerbel: How did you look it up?
https://github.com/wix-private/wix-vod/blob/af6cdfc74585d0c30d1d6656a74aafdcb3085f97/wix-vod/src/dashboard/views/channels/channel-info/cards/info/info.js#L4

there are many other examples
https://github.com/search?q=org%3Awix-private+RawContainer&type=Code

src/index.js
mosheKerbel: 👍 
for the convert function i don't see any usage inside the library also
so it can be removed and also the types
https://github.com/wix/wix-style-react/blob/7576d9960db85683f1dabc7f6e5165f887e600bc/src/Typography/Utils.js#L1

src/index.js
mosheKerbel: I replied about it below...
https://github.com/search?q=org%3Awix-private+RawContainer&type=Code

src/index.d.ts
```diff
@@ -212,7 +212,7 @@ export * from './MarketingLayout';
 export { default as MediaOverlay } from './MediaOverlay';
 export * from './MediaOverlay';
 
-export * from './MessageBox';
+export { MessageBoxMarketerialLayout, MessageBoxFunctionalLayout } from './MessageBox';
```
mosheKerbel: https://github.com/wix/wix-style-react/blob/15cbb3f059e6c5220e16bb94749cb2e96576aa8e/src/MessageBox/index.d.ts#L28

src/index.js
mykas: Oh right the types too

src/index.js
mykas: oh right, missed that. The old grid system is so complex..

src/index.d.ts
mykas: So if I understand correctly the interfaces would not be exported if you cherry pick classes?

src/index.d.ts
mykas: I've talked with Shlomi and HeaderLayout and FooterLayout should not be exposed

mykas: I have now added MessageBoxMarketrialLayoutProps and MessageBoxFunctionalLayoutProps to this export, I guess that should be enough.

src/index.d.ts
mykas: We need to take those ComponentProps and that would be ok, right?

mykas: Oh I see now. You have to export everything because there is this, right?
Screenshot 2020-03-24 at 09 30 03

src/MessageBox/index.d.ts
```diff
@@ -17,19 +11,10 @@ import MessageBoxMarketerialLayout, {
 
 export {
   MessageBoxMarketerialLayout,
-  MessageBoxMarketerialLayout as MessageBoxLayout1,
   MessageBoxMarketerialLayoutProps,
   MessageBoxMarketerialLayoutTheme,
   MessageBoxMarketerialLayoutThemePrimaryButtonTheme,
-  HeaderLayout,
```
mykas: Hey @mosheKerbel. I've deleted those here and left `export * from './MessageBox'` in entry index.d.ts . I think that should be good enough, right?

mosheKerbel: let me just pull it locally and verify it looks good.

mosheKerbel: change look solid
do you have a note for updating the migration guide accordingly? because I see no mention for these changes there.

src/index.js
```diff
@@ -85,13 +72,8 @@ export {
...
   Col,
 } from './Grid';
-
 export { Layout, Cell } from './Layout';
-export {
-  default as Typography,
-  convertFromUxLangToCss,
-  convertFromCssToUxLang,
-} from './Typography';
+export { default as Typography } from './Typography';
```
mosheKerbel: plz remove the 2 functions from here as well, otherwise, TS consumers will get them as autocomplete suggestions for impot statements.
https://github.com/wix/wix-style-react/blob/15cbb3f059e6c5220e16bb94749cb2e96576aa8e/src/Typography/index.d.ts#L4

src/MessageBox/index.d.ts
mykas: Right, I think I will  add those. Though we never actually advertised in storybook so some found those through types or maybe just legacy.

src/index.js
```diff
@@ -161,7 +143,6 @@ export { default as VerticalTabsItem } from './VerticalTabsItem';
 export {
   default as ListItemSection,
   listItemSectionBuilder,
-  TYPES as ListItemSectionTypes,
```
mosheKerbel: if it is indeed necessary to remove it
 you will also need to remove it from the d.ts, because it is a const, not a type:
https://github.com/wix/wix-style-react/blob/180020600474351c648ce32a3a8be8ac32181fed/src/ListItemSection/index.d.ts#L9

also from this statement, becuase there's no point
https://github.com/wix/wix-style-react/blob/4ec10fab6446ea8892589c50c0506292e12d31c5/src/ListItemSection/index.js#L1

and plz make sure to document it in the migration gude / take a note do it.

src/MessageBox/index.d.ts
mosheKerbel: I understand what you are saying
But, bottom line many users are using it, especially this weird `2` suffix components you removed, and we need to give them proper guidance so they won't get broken.
You know, probably most of this code owners are not necessarily the one who wrote these lines of code :)

src/index.d.ts
mosheKerbel: Exactly

PR #5251, opened 2020-03-23
WSR 7 - FINAL PR

EgoziE:
This PR will be the closing of wsr7!
It must include:

1. Make deployment domain permanent - https://wix-style-react-v7.now.sh/
   In order to enable deploying critical fixes to wsr7 in the future.

2. Remove caret (^) from all dependencies.
    In order not to break consumers in the future
    And help us maintain it better.

3. Anymore suggestions?

After merging this PR, don't forget to rebase branch version_7.x


shlomitc:
I think we can close this one.
Since we will support old versions anyway on the in `version_7.x`, we can decide to leave the dependencies as they are. the fix to `https://wix-style-react-v7.now.sh/` was done already there

PR #5252, opened 2020-03-23
(WSR 8) <Card/> - Remove withoutDivider prop

EgoziE:
Props:
- removed `withoutDivider` - use `<Card.Divider/>` when needed.


EgoziE:
@argshook I think a codemod can help here ⚡️

src/Stepper/docs/examples/card.js
```diff
@@ -1,4 +1,7 @@
 /* eslint-disable */
+import Card from "wix-style-react/*";
```
argshook: what's this `wix-style-react/*`?

EgoziE: webstorm's auto complete... 🤦‍♂ 
removing
thanks!

PR #5254, opened 2020-03-24
[wsr8 ]Migration.md - footerlayout,headerlayout,TYPES

mykas:


MIGRATION.md
```diff
@@ -676,6 +679,11 @@ Testkit:
 - removed `getTooltip` - use `isErrorIconShown` and `getTooltipErrorMessage` instead.
 - removed `isTooltipShown` - use `isErrorIconShown` and `getTooltipErrorMessage` instead.
 
+## \<ListItemSection/>
+
+General:
+- TYPES export no longer is supported. Import types definitions through: `ListItemSectionProps`.
```
mosheKerbel: you give here a solution for TS consumers only
what about JS consumers that used to import TYPES?
what's the motivation of removing it, to begin with?

MIGRATION.md
```diff
@@ -75,6 +75,8 @@ npx wix-ui-codemod wix-style-react/icons-common <path-to-your-project>
 
 4. Remove `<DataTable/>` - This component became _internal_ and should not be used directly - use `<Table/>` instead.
 
+5. Remove `<HeaderLayout>`, `<HeaderLayout1>`, `<FooterLayout>`, `<FooterLayout1>`, `<MessageBoxLayout2>`, `<MessageBoxLayout1>` - these components no longer available. Use `<MessageBoxMarketerialLayout>` or `<MessageBoxFunctionalLayout>` instead.
```
mosheKerbel: 👍

MIGRATION.md
mykas: Because this is not the pattern we support. We don't have any means to support JS types. That would be another initiative and I don't think we will go through this.

PR #5256, opened 2020-03-24
 `<Pagination/> - new component

jjaskulis:
### 🔦 Summary

...

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


src/Pagination/Pagination.st.css
```diff
@@ -0,0 +1,79 @@
...
+  color: value(D10);
+}
+
+.pagination::pageButton:hover,
+.pagination::navButton:hover {
+  color: value(D80);
+  background-color: value(B20);
+}
+
+.pagination::pageButton:focus, .pagination::navButton:focus {
```
mykas: I think thats duplicate to line https://github.com/wix/wix-style-react/pull/5256/files#diff-6b4411bd8fc948b3661321608f66a573R37 . No?

src/Pagination/Pagination.st.css
```diff
@@ -0,0 +1,79 @@
...
+}
+
+.pagination::currentPage {
+  -st-mixin: button-common-mixin;
+  background-color: value(B10);
+  color: value(D80);
+}
+
+.pagination::navButton {
+  -st-mixin: button-common-mixin;
```
mykas: Now I am wondering does this could be move to the `.pagination` instead? I don't think we need to repeat ourselves with this mixin, can you check?

src/Pagination/Pagination.st.css
jjaskulis: yep it's redundant - removed this.

src/Pagination/Pagination.st.css
jjaskulis: tried it, but that doesn't work.

mykas: Oh actually :: is not a state, but rather how it targets the classes. so it make sense to override all of those

mykas:
@argshook please check this branch for testkit autogenerator not working issue. To replicate:
 run storybook
 open pagination component testkit and see no api table

src/Pagination/Pagination.uni.driver.js
```diff
@@ -0,0 +1,16 @@
+import { paginationDriverFactory as corePaginationDriverFactory } from 'wix-ui-core/dist/src/components/pagination/Pagination.uni.driver';
+import { baseUniDriverFactory } from 'wix-ui-test-utils/unidriver';
+
+export const paginationDriverFactory = base => {
```
mykas: No need to export like this. Just do the whole file likes this:
```
export { paginationDriverFactory  } from 'wix-ui-core/dist/src/components/pagination/Pagination.uni.driver';
```

src/Pagination/Pagination.uni.driver.d.ts
```diff
@@ -0,0 +1,11 @@
+import { BaseUniDriver } from 'wix-ui-test-utils/unidriver';
```
mykas: Just re-export unidriver from wix-ui-core like this:
```
import { PaginationDriver } from 'wix-ui-core/drivers/unidriver';
interface PaginationUniDriver extends ReturnType<typeof PaginationDriver> {}
```

PR #5257, opened 2020-03-24
<DropdownBase/> - add testkit function - optionContentAt

mosheKerbel:


src/DropdownBase/DropdownBase.uni.driver.js
```diff
@@ -45,6 +45,25 @@ export const dropdownBaseDriverFactory = (base, body) => {
     optionsCount: async () =>
       (await createDropdownLayoutDriver()).optionsLength(),
 
+    optionContentAt: async id => {
```
EgoziE: When I talked to shlomi about a similar case he wanted to split this to 2 functions:
1. getContent: () => node
2. getContentText () => string

mosheKerbel: Reviewed it actually with Shlomi before I pushed :)
He said he doesn't mind if it's a one func or two, depends on the case...

EgoziE:
@TalZa1810 Once this is merged, you can delegate this function to `<PopoverMenu/>`

src/DropdownBase/DropdownBase.uni.driver.js
```diff
@@ -45,6 +45,25 @@ export const dropdownBaseDriverFactory = (base, body) => {
...
+      /*
+      Option content can be
+      1. node - <div>some text</div>
+      2. text - some text
+       */
+      const nodeContent = options[id].element().$$(':first-child');
+      const contentIsNode = (await nodeContent.count()) > 0;
+      if (contentIsNode) {
+        // eslint-disable-next-line no-restricted-properties
+        return await nodeContent.get(0).getNative();
```
TalZa1810: @mosheKerbel 
I know it was already merged, but why returning the native element and not the uni driver element?
The reason I'm asking this is because I want to implement such a driver function for the `<PopoverMenu/>`.
I need a function to check the text of the `<PopoverMenu.MenuItem/>`. Since I'm using the `dropdownbaseTestkit`
it returns the actual element and I need to manipulate the element which I prefer not to do: https://github.com/wix/wix-style-react/pull/5262

In my case, the user wants to get the text of the `<PopoverMenu.MenuItem/>` in order to check that it is translated correctly.

TalZa1810: Nevermind. I've managed to resolve this. I've added you as a reviewer so you can have a look :)

PR #5263, opened 2020-03-24
Added deprecation logs

EgoziE:
Preparation to WSR 8

Added deprecation logs for props:
1. error and errorMessage
2. help
3. theme

in relevant input family components.


EgoziE:
> I tested your changes with a Yoshi project
> It's enough that you use an empty `<Input/>` to get warnings,
> So all the other components in the family get it as well :(
> 
> Notice that some of the props have default values, for example here:
> https://github.com/wix/wix-style-react/blob/235ada1b77d967356b4d8f6f353e5f230ef6a975/src/Input/Input.js#L431
> 
> So the `hasOwnProperty('errorMessage')` always return true

This is bad, I'll remove default values of deprecated props

mosheKerbel:
> > I tested your changes with a Yoshi project
> > It's enough that you use an empty `<Input/>` to get warnings,
> > So all the other components in the family get it as well :(
> > Notice that some of the props have default values, for example here:
> > https://github.com/wix/wix-style-react/blob/235ada1b77d967356b4d8f6f353e5f230ef6a975/src/Input/Input.js#L431
> > 
> > So the `hasOwnProperty('errorMessage')` always return true
> 
> This is bad, I'll remove default values of deprecated props

Need to verify that no regressions can happen because of this
I mean the component **does* count on this default value. What would happen if both a deprecated prop and his new replacement will be undefined?

PR #5266, opened 2020-03-25
Added deprecation logs 2

EgoziE:
Preparation to WSR 8

Added deprecation logs for props in relevant components.


src/Loader/Loader.js
```diff
@@ -91,6 +92,16 @@ class Loader extends React.PureComponent {
     shouldLoadAsync: false,
   };
 
+  constructor(props) {
+    super(props);
+
+    if (props.hasOwnProperty('shouldLoadAsync')) {
```
mosheKerbel: same comment as my other PR - https://github.com/wix/wix-style-react/pull/5263#pullrequestreview-381051333

mosheKerbel: maybe there are other props like this in the PR, did not verify.

PR #5269, opened 2020-03-25
<ToggleButton/>: fix disabled prop and pass it also to <ButtonNext/>

Wo1v3r:
### 🔦 Summary

Latest changes of the component made it so when passing 'disable' it only styles it as disabled and noops `onClick`  callback, however, this should be handled by `ButtonNext` especially if we want to reuse its' drivers.
...

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [x] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [ ] Visual test


Wo1v3r:
@mykas 
No UX \ API change here, just implementation.

PR #5270, opened 2020-03-25
adding note for migrating tooltip with popover

TalZa1810:
Adding a disclaimer for migrating white `<Tooltip/>` to `<Popover/>`


MIGRATION.md
```diff
@@ -1402,6 +1402,10 @@ Tooltip no longer supports theme white, but white skin is available on Popover w
 > Warning #2: White Tooltip should be used only in cases when app background is dark - all other cases should have dark Tooltip.
 
 
+> Note:
+> - `content` prop should be replaced with `<Popover.Content/>`.
```
mykas: Thats a bit strange: Why `content` prop should be replaced with `<Popover.Content/>` ? Maybe would be more understandable `content` prop value should be used inside `<Popover.Content/>` or I did not get something here?

TalZa1810: yeah, that's what I meant :) I'll re-write this :)

TalZa1810: > Note:
> - `content` prop value should be used within `<Popover.Content/>`.
> - `children` prop value should be used within `<Popover.Element/>`.

Is it a better phrasing in your opinion? @mykas

mykas: yep :)

MIGRATION.md
```diff
@@ -1402,6 +1402,10 @@ Tooltip no longer supports theme white, but white skin is available on Popover w
 > Warning #2: White Tooltip should be used only in cases when app background is dark - all other cases should have dark Tooltip.
 
 
+> Note:
```
shlomitc: @TalZa1810 @mykas - maybe just add the "Before" example like in the other cases?

`<Tooltip skin="light"/>

PR #5272, opened 2020-03-25
(WSR 8) `<Tooltip/>` example  of old use

TalZa1810:
a better example of the old use of the Tooltip.


MIGRATION.md
```diff
@@ -1401,12 +1401,29 @@ Tooltip no longer supports theme white, but white skin is available on Popover w
 
 > Warning #2: White Tooltip should be used only in cases when app background is dark - all other cases should have dark Tooltip.
 
+Before:
+```jsx
+class WhiteTooltip extends React.PureComponent {
```
shlomitc: I think you can keep it simple
Before:
```jsx
<Tooltip content="Content">
  <div>Trigger</div>
</Tooltip>
```

And I think we can reduce the "white tooltip" example:

After:
```jsx
() => {
  const [shown, setShown] = React.useState(false);
  const toggle = () => setShown(!shown);
  return (
      <Popover
        shown={shown}
        onMouseEnter={toggle}
        onMouseLeave={toggle}
        placement="top"
        animate
        showArrow
      >
        <Popover.Element>Trigger</Popover.Element>
        <Popover.Content>
          <div style={{padding: '12px 24px'}}>Content</div>
        </Popover.Content>
      </Popover>
    );
}
```

TalZa1810:
I'm creating a new PR with this one. too many conflicts.

PR #5273, opened 2020-03-25
(WSR 8) <Input/> - remove updateControlledOnClear prop

EgoziE:
Props:
- removed `updateControlledOnClear`


src/MultiSelect/test/InputWithTags.spec.js
```diff
@@ -48,22 +48,6 @@ describe('InputWithTags', () => {
       expect(onClear).toHaveBeenCalledTimes(1);
     });
 
-    it('should call onChange', () => {
```
shlomitc: Technically this a change also in the `MultiSelect` component, but I don't think this it's that critical

MIGRATION.md
```diff
@@ -568,6 +568,7 @@ Props:
   For more information and examples go to [help prop](#help-prop) section.
 - removed sub component `<Input.Units/>` use `<Input.Affix/>` component instead
 - removed `magnifyingGlass` use `<Input.Affix/>` or `<Search/>` component instead. (see example below)
+- removed `updateControlledOnClear` - Clearing the value in a controlled component through onChange() is obsolete. Clearing the input will not trigger onClear().
```
shlomitc: I think consumers will find it hard to understand, the same as it was for us.

Let's add a title of "Behavior"
- `showClearButton` - previously, clicking on the clear button would have trigger `onChange()` **and** `onClear()`. Now, clicking on the clear button will trigger `onClear()` **only**

And i this line of "removed `updateControlledOnClear`" - refer to `showClearButton` change of behavior

PR #5275, opened 2020-03-25
(WSR 8) change sync-ts comparison branch

mosheKerbel:
@shlomitc what you showed me that failed  https://github.com/wix/wix-style-react/pull/5240 with no reason on types error.
I simply override the default branch for comparison.
(I accidentally thought this feature is buggy, but it was just the param order which should come before the name of the script)


mosheKerbel:
@EgoziE before we merge wsr8 to master, I will remove this

shlomitc:
For the future, I think that this should be done automatically by passing some kind of an environment variable from the CI, indicating the merge target branch. This will remove the need for a hard-coded value

mosheKerbel:
> For the future, I think that this should be done automatically by passing some kind of an environment variable from the CI, indicating the merge target branch. This will remove the need for a hard-coded value

how would you do it in such use case? it should behave like this only for `wsr8` branch, `master` should stay the same.

shlomitc:
You already have the feature for selective branch comparison. What I suggest is just to automatically identify which branch it should be compared to, instead of passing `wsr8` or any other branch name manually.
For example, in the `storybook-snapper` we have some configuration to identify which branch to compare with, that's how we make the Applitools visual tests to work between branches
https://github.com/wix-incubator/storybook-snapper/blob/master/config/applitools.config.js

PR #5285, opened 2020-03-26
<Search/> + <InputWithOptions/> - Clear value in controlled mode

EgoziE:
** Blocker for WSR 8 **
There's a bug that in a controlled `<Search/>` the value isn't really cleared when clicking the clear button.

How to reproduce:
1. Paste this code in the playground
```
class MyComponent extends React.Component {
  state={
    value: '',
  };

  render() {
    return (
    <Search
    updateControlledOnClear
      options={Array(26)
        .fill(0)
        .map((_, id) => ({
          id,
          value: `Option ${String.fromCharCode(97 + id)}`,
        }))}
      showOptionsIfEmptyInput={false}
      closeOnSelect={false}
      value={this.state.value}
      onChange={e => this.setState({value: e.target.value})}
      onSelect={option => this.setState({value: option.value})}
    />
    );
  }
}
```
2. Click/Focus on the search bar and type any letter, for example the letter `p`.
3. Click the clear button on the far right side of the component.

If you followed all steps correctly, the result should be that the value is cleared from the input, but the **dropdown is open** and the **letter `p` is still marked**.
Like this screenshot:
image

The wanted result, which you can achieve by removing the prop `updateControlledOnClear`, is once clicked the clear button, the **dropdown is closed** and the **letter `p` is not marked**.


src/InputWithOptions/InputWithOptions.js
```diff
@@ -96,6 +96,15 @@ class InputWithOptions extends Component {
     ) {
       this.showOptions();
     }
+
+    // Clear value in controlled mode
```
shlomitc: Please add a test in `inputWithOptions` spec file for this scenario

src/InputWithOptions/InputWithOptions.js
```diff
@@ -96,6 +96,15 @@ class InputWithOptions extends Component {
...
+    if (
+      prevProps.value !== '' &&
+      this.props.value === '' &&
+      this.state.inputValue !== ''
```
shlomitc: Isn't it be enough just to check if the new `value` prop is an empty string?
```suggestion
      prevProps.value !== this.props.value && this.props.value == ''
```

src/InputWithOptions/InputWithOptions.spec.js
```diff
@@ -9,6 +9,8 @@ import { makeControlled } from '../../test/utils';
 import InputWithOptions from './InputWithOptions';
 import inputWithOptionsDriverFactory from './InputWithOptions.driver';
 import { inputWithOptionsUniDriverFactory } from './InputWithOptions.private.uni.driver';
+import { mount } from 'enzyme';
+import { inputWithOptionsTestkitFactory } from '../../testkit/enzyme';
```
shlomitc: Please don’t import the testkit file. This is the library’s output and should not be consumed from inside. Initialize the driver with this instead:
```jsx
import {
  enzymeTestkitFactoryCreator,
} from 'wix-ui-test-utils/enzyme';
```

src/InputWithOptions/InputWithOptions.spec.js
```diff
@@ -150,6 +152,32 @@ describe('InputWithOptions', () => {
...
+          const driver = inputWithOptionsTestkitFactory({
+            wrapper,
+            dataHook,
+          });
+
+          expect(await driver.dropdownLayoutDriver.isShown()).toBe(false);
+          await driver.inputDriver.enterText('o');
+          expect(await driver.dropdownLayoutDriver.isShown()).toBe(true);
+          await driver.inputDriver.clickClear();
+          expect(await driver.dropdownLayoutDriver.isShown()).toBe(false);
```
shlomitc: I don’t think this is enough. It doesn’t indicate if the internal input value state was changed or not.

PR #5291, opened 2020-03-29
Pass RowDataType to TableColumn

splintor:
### 🔦 Summary

In https://github.com/wix/wix-style-react/pull/5190 that added type to the data displayed in `<Table>`, I forgot to pass the `RowData` to the `TableColumn` type used in the `columns` prop. This is needed to make the column's `render` method properly typed.  

### ✅ Checklist
- 🔬 Change is tested
  - [x] Component type

### DIFF
--- a/src/Table/index.d.ts
+++ b/src/Table/index.d.ts
@@ -12,9 +12,9 @@ export interface TableProps<RowData = RowDataDefaultType> extends UsedDataTableP
   selectionDisabled?: boolean;
   deselectRowsByDefault?: boolean;
   withWrapper?: boolean;
-  onSortClick?(colData: TableColumn, colNum: number): void;
+  onSortClick?(colData: TableColumn<RowData>, colNum: number): void;
   totalSelectableCount?: number;
-  columns: TableColumn[];
+  columns: TableColumn<RowData>[];
 }
 
 export default class Table<RowData = RowDataDefaultType> extends React.Component<TableProps<RowData>> {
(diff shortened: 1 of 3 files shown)


test/types/Table.tsx
```diff
@@ -83,7 +83,10 @@ function typedTable() {
       name: 'Joe',
       age: 42,
     }]}
-    columns={[]}
+    columns={[{
+      title: 'name',
+      render: (row: TableRowData) => row.name,
```
mosheKerbel, line 88: This is something you could do prior to your change
In order to prove that your change works, remove the type:
`render: (row) => row.name`

src/Table/index.d.ts
```diff
@@ -12,9 +12,9 @@ export interface TableProps<RowData = RowDataDefaultType> extends UsedDataTableP
   selectionDisabled?: boolean;
   deselectRowsByDefault?: boolean;
   withWrapper?: boolean;
-  onSortClick?(colData: TableColumn, colNum: number): void;
+  onSortClick?(colData: TableColumn<RowData>, colNum: number): void;
```
mosheKerbel, line 15: I'm not sure it will serve you as you wish
because if you look at the `TableColumn` type, the RowData is only part of the `render` function 
https://github.com/wix/wix-style-react/blob/902a3295c92b1a739933afda67b4c9c34993b052/src/DataTable/index.d.ts#L82

So how would your `onSortClick` callback would look? how would you use the actual data of the column to run comparisons?

splintor: Yes, this was done only for completeness. Ideally (if we didn't have to support existing code), I would make `RowData` generic parameter with no defaults, so users are required to set the type of their data rows, so I tried to make every usage of `TableColumn` be with the `RowDataType`, but here it really has no meaning, so I removed it per your request.

The main purpose of this PR it to pass `RowDataType` to `columns` to make their `render` method typed.

test/types/Table.tsx
splintor: This change also won't probe my change works, as it would compile even without my change (since then, render will accept a `row: any` parameter).
But you are right that the idea is that with my change, the `render` method parameter is now types, and should be auto-completed by the IDE, without specifying its type explicitly, so I updated the code.

PR #5284, opened 2020-03-26
`<Page/>`- remove warning messages from <Page/> story page

TalZa1810:
### 🔦 Summary
Fixing `<Page/>`  example to use the new PopoverMenu

PR #5295, opened 2020-03-29
`<Table/>` - removing warning messages from story page 

TalZa1810:
`<Table/>` - removing warning messages from story page


TalZa1810:
@EgoziE 
I get here the same warning messages as described [here](https://github.com/wix/wix-style-react/pull/5284) so I guess they are both related to the `TableActionCell` and the `Search` components, so same comments here as the PR #5284

PR #5297, opened 2020-03-30
`<Sidebar/>`-  Add shadow to sidebar for scrolling indication

TalZa1810:
### 🔦 Summary

1. Adding shadow to sidebar scroll. 
2. Fixing `Upgrade` Button example to have margin of : `24px 18p`

Jira: https://jira.wixpress.com/browse/DSM-266
Zeplin: https://zpl.io/a7Bry88

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


src/Sidebar/Sidebar.js
```diff
@@ -120,6 +121,15 @@ class Sidebar extends Component {
     this._setInnerMenus(props);
   }
 
+  componentDidMount() {
+    this._shouldAddGradient();
+  }
+
+  _shouldAddGradient() {
+    const { scrollHeight, clientHeight } = this.onScreenChildrenRef.current;
+    this.setState({ isScrollbarDisplayed: scrollHeight > clientHeight });
```
shlomitc: What happens if we just add it always, even without checking for the scroll? how will it look like?

TalZa1810:
fails on visual which is not related.

EgoziE:
Right side - master
Left side - yours

There's a gap of one item that is the gradient, is it intentional?

image

EgoziE:
What about gradient at the top to indicate scroll up?
The scrollbar fades with blur so I think it might be needed too (maybe ask the UX)

TalZa1810:
@EgoziE 
I've talked to Ayelet and currently, the design is to have the gradient only at the bottom of the Sidebar.

@shlomitc 
As for rendering the gradient permanently, it creates diffs with the original sidebar (check the diff Egozi added above). It means that there's a gap always at the bottom of the sidebar. Do we want to render it only when there's a scrollbar or rendering this all the time?

shlomitc:
@TalZa1810 - doing dom calculation has its cost, it might not that critical, but I would prefer to prevent doing it.
In these cases, I summon my CSS Guru - @giladsegal (which is probably busy these days but might have a better suggestion for us).

giladsegal:
Thou summon is heard.
There is actually a really cool trick to achieve using CSS only,
Check out [this link](http://lea.verou.me/2012/04/background-attachment-local/)

TalZa1810:
So visuals pass. Currently implemented with JS. Waiting for Ayelet for another review before merging.

PR #5298, opened 2020-03-30
<Input/> - Add test functions to status

EgoziE:
Added status functions to `<Input/>`

  // Status
  hasStatus: () => boolean;
  getStatus: () => InputStatus;
  hasStatusMessage: () => boolean;
  getStatusMessage: () => string;


src/Input/Input.driver.js
```diff
@@ -1,7 +1,8 @@
 import ReactTestUtils from 'react-dom/test-utils';
+import { tooltipDriverFactory } from 'wix-ui-core/dist/src/components/tooltip/Tooltip.driver';
 import styles from './Input.scss';
 
-const inputDriverFactory = ({ element }) => {
+const inputDriverFactory = ({ element, wrapper, eventTrigger }) => {
```
shlomitc: why is `wrapper` needed here?

src/Input/Input.driver.js
```diff
@@ -90,7 +91,7 @@ const inputDriverFactory = ({ element }) => {
     hasError: () => element.classList.contains(styles.hasError),
     hasWarning: () => element.classList.contains(styles.hasWarning),
     getTooltipElement: () => element,
-    hasLoader: () => element.querySelector(`.loaderContainer`),
+    hasLoader: () => !!element.querySelector(`.loaderContainer`),
```
shlomitc: don't forget to change the `.d.ts` type to boolean as well

src/Input/Input.driver.js
```diff
@@ -119,6 +120,31 @@ const inputDriverFactory = ({ element }) => {
...
+      element.classList.contains(styles.hasError) ||
+      element.classList.contains(styles.hasWarning) ||
+      !!element.querySelector(`.loaderContainer`),
+    /** If there's a status, returns its type */
+    getStatus: () =>
+      (element.classList.contains(styles.hasError) && 'error') ||
+      (element.classList.contains(styles.hasWarning) && 'warning') ||
+      (!!element.querySelector(`.loaderContainer`) && 'loading'),
+    /** Return true if there's a status message */
+    hasStatusMessage: () => true, // This is a stub for wsr7, will be fixed at wsr8.
```
shlomitc: maybe it should just call `getStatus()`?

src/Input/Input.uni.driver.js
```diff
@@ -119,6 +120,28 @@ export const testkit = base => {
...
+      (await base.hasClass('hasError')) ||
+      (await base.hasClass('hasWarning')) ||
+      (await base.$$(`.loaderContainer`).count()) > 0,
+    /** If there's a status, returns its type */
+    getStatus: async () =>
+      ((await base.hasClass('hasError')) && 'error') ||
+      ((await base.hasClass('hasWarning')) && 'warning') ||
+      ((await base.$$(`.loaderContainer`).count()) > 0 && 'loading'),
+    /** Return true if there's a status message */
+    hasStatusMessage: () => true, // This is a stub for wsr7, will be fixed at wsr8.
```
shlomitc: same comment as for the regular driver

src/Input/tests/Input.spec.js
```diff
@@ -273,17 +189,17 @@ describe('Input', () => {
     });
 
     describe('status attribute', () => {
-      it('should display a loader icon if status is loading', async () => {
```
shlomitc: maybe worth keeping these tests? since we didn't delete the methods, I think it's better not to remove the tests.
Adding the new ones is fine

src/NoBorderInput/NoBorderInput.driver.js
```diff
@@ -8,14 +8,14 @@ const noBorderInputDriverFactory = ({ element, wrapper }) => {
   });
 
   return {
+    ...inputDriver,
```
shlomitc: good catch, however now the question is why do we even need the specific `getStatusMessage` override and not use the one that comes from the `Input`?

src/Input/Input.driver.js
EgoziE: getStatus returns 'error', 'warning', or 'loading' strings, 
hasStatusMessage returns true only if a **message** is provided

For `<Input status="error">`
getStatus => 'error'
hasStatusMessage => **false**

For `<Input status="error" statusMessage="message">`
getStatus => 'error'
hasStatusMessage => **true**

src/Input/Input.driver.js
EgoziE: tooltipDriverFactory won't work without it

shlomitc: but I don't see `wrapper` in the core component, which is the driver you're initializing

src/Input/Input.driver.js
shlomitc: so maybe this function is not very useful?
since you're already providing the `getStatusMessage`, wouldn't that be enough?

EgoziE: removed for wsr7

src/NoBorderInput/NoBorderInput.driver.js
EgoziE: `<NoBorderInput/>` is built differently than `<Input/>`
The status icon is outside the input (and implemented with StatusIndicator)

src/Input/Input.driver.js
```diff
@@ -119,6 +120,28 @@ const inputDriverFactory = ({ element }) => {
     isCustomInput: () => {
       return input.getAttribute('data-hook') === 'wsr-custom-input';
     },
+
+    // Status
+    /** Return true if there's a status */
+    hasStatus: () =>
```
shlomitc: since the `status` prop is no longer a boolean (as oppose to `error`), I don't think that `hasStatus()` should just return boolean but rather boolean for some value.

for example - `hasStatus('error')` will return true if it is in error state or not.

I know it might be a bit of a hassle, but I think this is what users will expect in their tests, not just to see if ther's some status

PR #5300, opened 2020-03-30
[DO NOT MERGE][JUST A TEST] Release alpha test

mykas:


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "wix-style-react",
   "description": "wix-style-react",
-  "version": "7.53.1",
+  "version": "7.53.2",
```
shlomitc: not good, this will release a new version, you need to mention it as `-alpha`

PR #5308, opened 2020-03-31
[version_7.x] Release 7.54.0-alpha

mykas:
This is a test of our ci to check if we can release alpha.


shlomitc:
That's great.
However, I see an issue in the publishing mechanism
```bash
⇒  npm view wix-style-react | grep alpha
wix-style-react@7.54.0-alpha | MIT | deps: 39 | versions: 7962
.tarball: https://registry.npmjs.org/wix-style-react/-/wix-style-react-7.54.0-alpha.tgz
latest: 7.54.0-alpha
next: 5.0.3-alpha
```

As you can see, the `npm publish` task marked the version as the `latest`. That's not good because after v8 is released, once we release a patch for 7 - if someone will install `wix-style-react` for the first time, he will get the v7 instead of v8.

mykas:
You're right! I was confused about the `latest` tag too so I tested and tried to install locally and it did not catch the alpha version on version 7. What I did not thought of was that it could hurt our version 8 users.

mykas:
I just talked with @netanelgilad and it seems when we gonna have next major version 8 released - whenever our branch version_7.x will try to publish it will not use the `latest` tag. Yesterday it published that alpha version because we still have the major version 7 as latest.

PR #5310, opened 2020-03-31
wsr7: update sync-ts

mosheKerbel:
@mykas


mykas:
It seems there is some strange merge conflict. Check it in builds

PR #5324, opened 2020-04-05
Avatar Custom Indication

ayal:
Add Custom Indication property to Avatar component. Acts and styled like the Indication prop only is not wrapped in an IconButton.

### ✅ Checklist

- [x] 👨‍💻 API change is approved by the UI Infra Developers @nitayneeman 
- [x] 👨‍🎨 UX change is approved by the Design System UX @ayeletr 
- 📚 Change is documented
  - [x] Story
  - [x] API description
  - [x] Cheatsheet
- 🔬 Change is tested
  - [x] Component
  - [x] Visual test


src/Avatar/Avatar.js
```diff
@@ -87,7 +94,7 @@ class Avatar extends React.PureComponent {
           {...styles('avatarContainer', {
             shape,
             size,
-            indication,
+            indication: indication || customIndication,
```
nitayneeman, line 97: What would happen if both `indication` and `customIndication` are provided? I think `customIndication` should override

src/Avatar/Avatar.js
```diff
@@ -69,11 +70,17 @@ class Avatar extends React.PureComponent {
...
     const renderOnHover = !showIndicationOnHover || showIndication;
     const renderIndication =
-      indication && renderOnHover && sizeNumber > minIndicationRenderSize;
+      indication &&
+      renderOnHover &&
+      sizeNumber > minIndicationRenderSize &&
+      !customIndication;
+
+    const renderCustomIndication =
+      customIndication && renderOnHover && sizeNumber > minIndicationRenderSize;
```
nitayneeman, line 83: What do you think about:
```suggestion
    const indicationConstraints = renderOnHover && sizeNumber > minIndicationRenderSize;
    const renderIndication = indicationConstraints && !customIndication && indication;
    const renderCustomIndication = indicationConstraints && customIndication;
```

src/Avatar/Avatar.js
```diff
@@ -200,6 +216,8 @@ Avatar.propTypes = {
   presence: PropTypes.oneOf(['online', 'offline', 'busy']),
   /** A node to be rendered as Indication. */
   indication: PropTypes.node,
+  /** A node to be rendered as Custom Indication. Not wrapped by an IconButton */
```
nitayneeman, line 219: A suggestion:
```suggestion
  /** A node to be rendered as a custom indication, that is not wrapped by an IconButton. This node could be rendered in other shapes (such as square). */
```

src/Avatar/docs/index.story.js
```diff
@@ -48,6 +50,28 @@ const indicationExamples = [
   },
 ];
 
+const customIndicationExamples = [
+  {
+    label: 'Box',
```
nitayneeman, line 55: Let's be specific:
```suggestion
    label: 'Facebook Logo',
```

src/Avatar/docs/index.story.js
```diff
@@ -48,6 +50,28 @@ const indicationExamples = [
...
+    value: (
+      <Box
+        align="center"
+        verticalAlign="middle"
+        backgroundColor="#3b5998"
+        color="white"
+        minHeight="100%"
+        padding={0}
```
nitayneeman, line 63: This seems redundant (also in the rest of examples)

src/Avatar/docs/index.story.js
```diff
@@ -48,6 +50,28 @@ const indicationExamples = [
   },
 ];
 
+const customIndicationExamples = [
+  {
+    label: 'Box',
+    value: (
+      <Box
```
nitayneeman, line 57: How are you going to use it as a consumer?

If it's an SVG we can render the asset instead of simulating it using `Box`

nitayneeman:
@ayeletr could you do a UX review please?

nitayneeman:
Also I think we should align with this:
image

In contrast to the circle shape, in a square for size 60 - the height is supposed to be `26px` and not `24px`.

@ayeletr is this designed on purpose, right?

src/Avatar/docs/index.story.js
ayal: actually in our case it a div with a background image

src/Avatar/docs/index.story.js
ayal: if I remove the background color for the Box it will be transparent and you can't see it's a square. The other properties are there so the Box looks nicer (i.e 100% height of indication wrapper div, no padding, etc..).. padding might be redundant.

src/Avatar/Avatar.js
ayal: > Also I think we should align with this:
> image
> 
> In contrast to the circle shape, in a square for size 60 - the height is supposed to be `26px` and not `24px`.
> 
> @ayeletr is this designed on purpose? If so, is it relevant only for the `customIndicator` or also for the regular one?

Yeah right now I do the exact same thing as in regular indication - do we need to change the behaviour for custom indication?

src/Avatar/docs/index.story.js
nitayneeman: I meant only for the padding ;)

src/Avatar/docs/index.story.js
nitayneeman: It could be nice to reflect the correct usage at least in the examples page, so maybe a `Box` with `backgroundImage` might be better?

PR #5327, opened 2020-04-06
1.4 General Use Icons

TalZa1810:
### 🔦 Summary
This is an initial draft for this story. Should consider dynamic import for the icons. 

**Missing:**
1. Asking UX which icons to use for each example.
2. a UX review. 
3. Categorize the icons.


shlomitc:
While you’re Already doing some order in the mess - can your please check how we can categorize the components? Try talking with zilvinas about it

mosheKerbel:
It looks awesome, much more welcoming than the existing one.
A nice to have feature you can consider - a `search` to filter the icons.

TalZa1810:
Yeah, I'm on it already :) 
Thanks @mosheKerbel

src/new-icons/docs/index.story.js
```diff
@@ -0,0 +1,108 @@
...
+        ],
+      }),
+
+      tab({
+        title: 'Examples',
+        sections: [
+          description({
+            title: 'Important:',
+            text:
+              'In storybook, we import icons using `Icons` prefix. When using in your application, remember to **remove** this prefix.',
```
mosheKerbel, line 76: good note
I suggest we put some red alert here to make sure we got the user's attention, our Notification maybe? maybe UX can advise on the proper design.

src/new-icons/docs/index.story.js
```diff
@@ -0,0 +1,108 @@
...
+          title('Examples'),
+
+          example({
+            title: 'Custom Icon color',
+            description:
+              'The icon color can be customized using the `style` or `className` props.',
+            source: examples.colorsExample,
+          }),
+
+          example({
```
mosheKerbel, line 88: We have different icons for the different sizes
We should instruct users to use those
For example - Add - AddSmall, AlignRight - AlignRightSmall

src/new-icons/docs/AllIcons.js
```diff
@@ -0,0 +1,52 @@
...
+            }
+            options={this.iconsNames.map((name, id) => ({ id, value: name }))}
+            onSelect={({ value }) => this.setState({ searchQuery: value })}
+            onClear={() => this.setState({ searchTerm: '', searchQuery: '' })}
+          />
+        </Cell>
+        <Cell>
+          {searchQuery
+            ? renderIcon(searchQuery)
+            : this.iconsNames.map(renderIcon)}
```
mosheKerbel, line 45: A suggestion - could be nicer if the icons are filtered upon key pressing, and not only on the selection of a single one.
Again, just a feature suggestion.

src/new-icons/docs/AllIcons.js
```diff
@@ -0,0 +1,52 @@
...
+  render() {
+    const { searchTerm, searchQuery } = this.state;
+
+    return (
+      <Layout>
+        <Cell span={2}>
+          <Heading appearance="H2">All Icons:</Heading>
+        </Cell>
+        <Cell span={2}>
+          <Search
```
mosheKerbel, line 32: it's really small, for long icon names it's partly hidden
image

src/new-icons/docs/examples.js
```diff
@@ -0,0 +1,88 @@
+import { colors, sizes } from './constants';
+
+export const colorsExample = `<Layout>
+                        <Cell>
+                          <Layout justifyItems="center" cols="4">
+                              <Box direction="vertical" align="center">
+                                 <Icons.Duplicate style={{ color: '${colors.standard}' }} /><Text>Standard</Text><Text size="tiny">B10</Text>
```
shlomitc, line 7: since you're using `Box` as a parent, just pass `color="B10"` and ditch the inline styles

src/new-icons/docs/installation.md
```diff
@@ -0,0 +1,5 @@
+```bash
+npm i wix-ui-icons-common
+# or
+yarn add wix-ui-icons-common
```
shlomitc, line 4: you can remove it, it's just taking space and not adding too much

src/new-icons/docs/installation.md
```diff
@@ -0,0 +1,5 @@
+```bash
+npm i wix-ui-icons-common
```
shlomitc, line 2: `npm install wix-ui-icons-common` will look a little bit more official

src/new-icons/docs/index.story.js
```diff
@@ -0,0 +1,108 @@
...
+            title: 'Important:',
+            text:
+              'In storybook, we import icons using `Icons` prefix. When using in your application, remember to **remove** this prefix.',
+          }),
+
+          title('Examples'),
+
+          example({
+            title: 'Custom Icon color',
+            description:
```
shlomitc, line 83: I think that the more accurate description will be:
"The icon is an svg that inherits its parent's color or can directly be colored by passing a `className` or using the `style` prop.

src/new-icons/docs/index.story.js
shlomitc: That is correct.
For best results, we advise using the smaller size icons and in cases when such doesn't exist, then `size` prop can be used

src/new-icons/docs/storySettings.js
```diff
@@ -0,0 +1,6 @@
+import { Category } from '../../../stories/storiesHierarchy';
+
+export const storySettings = {
+  category: Category.FOUNDATION,
+  storyName: '1.4 General Use Icons',
```
shlomitc, line 5: In general, I think that this file is a bit obsolete, also it is a duplication of the `symbolsComponentsMapping` you created.

Any reason why we can't use the `symbolsComponentsMapping/symbols.js` here and in all other places in stories?

src/new-icons/docs/index.story.js
```diff
@@ -0,0 +1,108 @@
...
+            "import Duplicate from 'wix-ui-icons-common/Duplicate';\n",
+          ),
+
+          divider(),
+          <AllIcons />,
+        ],
+      }),
+
+      tab({
+        title: 'Examples',
```
shlomitc, line 71: I'd call it "Common Usages"

src/new-icons/docs/index.story.js
```diff
@@ -0,0 +1,108 @@
...
+const example = config => baseExample({ components: allComponents, ...config });
+
+export default {
+  category: storySettings.category,
+  storyName: storySettings.storyName,
+
+  sections: [
+    tabs([
+      tab({
+        title: 'Description',
```
shlomitc, line 40: let's just call it "Inventory"

src/new-icons/docs/AllIcons.js
```diff
@@ -0,0 +1,52 @@
...
+            value={searchQuery || searchTerm}
```
shlomitc, line 33: you should bypass all search when input becomes empty again.
1. type some icon name
2. click on it
3. delete the input content (not using the clear button)

**current behavior**: only the last icon appear
**expected**: all icons appear

shlomitc:
@TalZa1810 - for a better user experience, instead of using the `<Search/>` component, I would simply use an `<Input/>` and filter the list of the icons. Since you already see all icons and names, you don't need auto-completion.

src/new-icons/docs/examples.js
TalZa1810: The Box is for the alignment of the icon and color description.
The `Box` children is the `Icon` and two `Text` elements so coloring the `Box` would affect the look of the Text elements within. Basically, the example was to show the use of inline styling.

shlomitc: 1. It will not color the text because text elements' styles are more specific.
2. These are two separate concerns in this example. One is how to properly color icons in the design system and the other is the API of it. The example of the different colors in the system is good, and using `Box` is good enough.

if you wish to have a more technical example of the `className` and `style` props, they can be separated and don't need that many examples, just one for each prop.

shlomitc: anyway, holding the constant object is just duplicating code, not tested and not maintainable, so let's remove it

TalZa1810:
1. Fixed the option to mark the icon name.
2. Removed the search option. We'll re-add this after categorizing the icons (in `wix-ui-common` repository) in order to have a better search (when we have more metadata on each icon).
3. As for the `Icon` red alert (Moshe's note)- should consult with UX.
4.  As for copy-paste feature- should consult UX regarding this feature as well. 
5. Fixing the custom colors example. 

I'll talk to Zilvinas about the meta data for each icon and I'll work on the search and categorizing the icons in a different PR.

Let me know what you think :)

shlomitc:
## Icons are not vertically aligned next to the texts
image

## Overlapping icons on small screens
image
Icons don't look good on mobile screens. I think that using flex with max-width. would do the trick

## Search
I actually liked it, even if we're going to do the work on `wix-ui-icons-common`. Keep it really simple, just an input field that filters the icons list by their name. This will give great value with minimal effort

TalZa1810:
fails on types which is irrelevant

shlomitc:
@TalZa1810 - I reverted the PR.
You indexed the story wrongly and it caused the storybook to open on the icons page instead of on the main page
image

TalZa1810:
PR with the correct linking: https://github.com/wix/wix-style-react/pull/5361

PR #5332, opened 2020-04-07
allow text in size 24

ayal:
Allow text on size24 (when using name prop) but only one letter out of initials, depends on core Avatar allowing limitInitials=1
https://github.com/wix/wix-ui/pull/1829

### 🔦 Summary

...

### ✅ Checklist

- [x] 👨‍💻 API change is approved by the UI Infra Developers @nitayneeman 
- [x] 👨‍🎨 UX change is approved by the Design System UX @ayeletr 
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [x] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [x] Visual test


src/Avatar/Avatar.st.css
```diff
@@ -351,3 +352,7 @@
   visibility: hidden;
 }
 
+.avatarContainer:size(size24):name .avatar:contentType(text)::content  {
```
nitayneeman, line 355: According to the design spec, `font-size` should be exactly as size 36 and 30 - so I wouldn't change it particularly

src/Avatar/test/Avatar.spec.js
```diff
@@ -99,4 +99,15 @@ describe('Avatar', () => {
     expect(await driver.indicationExists()).toBe(false);
     expect(await driver.customIndicationExists()).toBe(true);
   });
+
+  it('For size < 30 text content should be 1 character', async () => {
```
nitayneeman, line 103: I would add another test for sizes that are larger than 30 (mainly to make sure we didn't break)

src/Avatar/Avatar.st.css
nitayneeman: Actually if we remove the font-size, we can adjust line 350 instead of this block:
```
.avatarContainer:size(size24):not(:name) .avatar:contentType(text)::content,
.avatarContainer:size(size18) .avatar:contentType(text)::content  {	
```

But the question is - why should it be conditioned by `name` at all (the bigger sizes currently aren't, what's the difference...)?

ayal: I added the name conditioning (name && !text) to make sure in case of text we don't show the text because then avatar core is not doing any initials calculation and we get ugly small text that exceeds the container

src/Avatar/test/Avatar.spec.js
ayal: ok, and do you think this test is ok? since it's dependant on a new version of ui-core with this feature, i.e currently it breaks.

src/Avatar/Avatar.st.css
ayal: you are right about the font-size, it was set explicitly for size 36 and 30 for text-tiny-bold so i missed it, adding another line to include size24 there

src/Avatar/test/Avatar.spec.js
nitayneeman: We will need to upgrade wix-ui-core anyway, in order to make the feature work - so I think we can add them

src/Avatar/Avatar.st.css
nitayneeman: Regarding the name conditioning - it makes sense, but I don't think this behavior was specified in the design spec (maybe you've spoken about that by heart?). I think it's good in general but also needs to be consistent with other sizes (at least 30).

@ayeletr what do you think?

src/Avatar/Avatar.st.css
```diff
@@ -336,18 +337,21 @@
...
 .avatarContainer:size(size36)[data-madefor="true"] .avatar:contentType(text)::content,
-.avatarContainer:size(size30)[data-madefor="true"] .avatar:contentType(text)::content  {
+.avatarContainer:size(size30)[data-madefor="true"] .avatar:contentType(text)::content,
+.avatarContainer:size(size24)[data-madefor="true"] .avatar:contentType(text)::content,
+.avatarContainer:size(size18)[data-madefor="true"] .avatar:ceverontentType(text)::content { 
   -st-mixin: text-tiny-bold;
 }
 
-.avatarContainer:size(size24) .avatar:contentType(text)::content,
+.avatarContainer:size(size24):text .avatar:contentType(text)::content,
```
ayal, line 354: better like this?

src/Avatar/Avatar.js
```diff
@@ -111,6 +112,7 @@ class Avatar extends React.PureComponent {
                 text,
                 name,
                 onClick,
+                initialsLimit: sizeNumber < 30 ? 1 : undefined,
```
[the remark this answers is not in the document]
ayal: well this introduces something unexpected - in case of size >= 30 it passes "false" not "undefined" and somehow this always renders only 1 character :/ basically core avatar props are typed and should only get 1 | 2 | 3 or undefined - but this is not fully typed. might need to fix core avatar a little bit more to make sure this doesn't ever happen accidentally

src/Avatar/Avatar.js
```diff
@@ -89,6 +89,7 @@ class Avatar extends React.PureComponent {
             presenceType: presence,
             clickable: !!onClick,
             fade: fadeIndication,
+            text: !!text,
```
nitayneeman, line 92: I would rename to the following to clarify that's a boolean and not the actual text:
```suggestion
            hasText: !!text,
```

src/Avatar/Avatar.st.css
nitayneeman: I think we should rename it to `hasText` to make it clearer

PR #5333, opened 2020-04-07
<BadgeSelect/> - expose popoverCommonProps

rostyslavP:
### 🔦 Summary
Passing props to Popover and fixing problem when BadgeSelect is wrapped by Tooltip 

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


mykas:
Hi, thanks for the PR. Though it needs some additional changes for it to make it work. Currently we have also `d.ts` files which needs to have this exposed too.

Also I suggest adding appendTo="window" as default value and only if user passes `popoverProps` it could override it.

src/BadgeSelect/BadgeSelect.js
```diff
@@ -186,6 +189,8 @@ BadgeSelect.propTypes = {
 
   /** Applied as data-hook HTML attribute that can be used to create driver in testing */
   dataHook: PropTypes.string,
+
+  popoverProps: PropTypes.shape(PopoverCommonProps),
```
mykas, line 193: This also is missing a comment: `common popover props`

src/BadgeSelect/index.d.ts
```diff
@@ -1,5 +1,6 @@
 import * as React from 'react';
-import {BadgeSize, BadgeType, BadgeSkin} from '../Badge';
+import { BadgeSize, BadgeType, BadgeSkin } from '../Badge';
+import { PopoverCommonProps } from '../common/PropTypes/PopoverCommon';
```
mykas, line 3: This is not the right place to get it. There is a `common.d.ts` file in `./src` folder

rostyslavP: but should it be updated here as well?

src/BadgeSelect/BadgeSelect.driver.js
```diff
@@ -26,7 +26,7 @@ const badgeSelectDriverFactory = ({ element, eventTrigger }) => {
 
       // Since the popover may be closed, we need to create the DropdownLayout driver every time
       const dropdownLayoutDriver = dropdownLayoutDriverFactory({
-        element: element.querySelector(
+        element: document.body.querySelector(
```
mykas, line 29: This is not good, because there could many content elements likes this in the dom with the same data-hook (believe me, users tend to do stuff like that). For it to work properly you should use popoverDriver to:
1. Open popover content.
2. get content element
3. query the badge datahook 
4. pass it to dropdownLauyoutDriver

EgoziE:
I'm splitting this PR
1. <BadgeSelect/> - expose popoverCommonProps (this PR)
2. <BadgeSelect/> - support appendTo="window' (next PR)

PR #5334, opened 2020-04-07
<CheckToggle/> - new component

EgoziE:
Avishai's component, review it carefully 😬


mosheKerbel:
@EgoziE 
Didn't really review it, just something that popped up
the `handleClick` is considered public in API docs, because it doesn't have `_` prefix (`_handleClick`)
is this the desired behaviour?

package.json
```diff
@@ -44,6 +44,7 @@
     "test:e2e": "npm run test:e2e-runtime && build-storybook -c ./.storybook/e2e && npm run test:e2e-only",
     "test:e2e-only": "NODE_ENV=production yoshi test --protractor",
     "test:e2e-runtime": "cd ./test-runtime/imports && NODE_ENV=production npm run test",
+    "pretest:visual": "npm run generate-autodocs-registry",
```
shlomitc, line 47: why is this needed?

src/CheckToggle/CheckToggle.js
```diff
@@ -0,0 +1,136 @@
...
+   * Except when disabled
+   */
+  handleClick = () => {
+    const { checked } = this.state;
+    const { disabled, onChange } = this.props;
+
+    if (disabled) return;
+
+    this.setState({ checked: !checked });
+    if (onChange) onChange(!checked);
```
shlomitc, line 33: have this as a callback function to `setState` to ensure it happens after the render

src/CheckToggle/CheckToggle.js
```diff
@@ -0,0 +1,136 @@
...
+  /**
+   * Renders the button itself
+   * @returns React.ReactNode
+   * @private
+   */
+  _renderButton = () => {
+    const { checked } = this.state;
+    const { size, skin, disabled } = this.props;
+    return (
+      <button
```
shlomitc, line 45: doesn't it make sense to use `input type="checkbox"` instead of a `button`?
this would also make it possible to use the component in an `uncontrolled` way

package.json
EgoziE: I'm using autodocs to get all components metadata to create permutations

shlomitc: Then you will need this also in `storybook:visual`.
Please note that this action adds 16s to the build step.
In general, I think that this task should be some **storybook addon**, so it will be configured part of the storybook and not part of the build step - @argshook WDYT?

src/CheckToggle/CheckToggle.js
```diff
@@ -0,0 +1,139 @@
...
+
+    if (disabled) return;
+
+    this.setState({ checked: !checked }, () => {
+      if (onChange) onChange(!checked);
+    });
+  };
+
+  /**
+   * Renders the button itself
```
shlomitc, line 38: a "toggle", not a button

src/CheckToggle/CheckToggle.st.css
```diff
@@ -0,0 +1,127 @@
...
+
+.input {
+  position: absolute;
+  opacity: 0;
+  cursor: pointer;
+  height: 0;
+  width: 0;
+}
+
+.button {
```
shlomitc, line 36: if it's no longer a button, but a toggle, let's change the terminology.

src/CheckToggle/CheckToggle.js
```diff
@@ -0,0 +1,139 @@
...
+   * Toggles checked state and triggers the onChange callback.
+   * Except when disabled
+   */
+  _handleClick = () => {
+    const { checked } = this.state;
+    const { disabled, onChange } = this.props;
+
+    if (disabled) return;
+
+    this.setState({ checked: !checked }, () => {
```
shlomitc, line 32: The existing beahvior mixes the controlled and uncontrolled pattern. 
I'd expect `<CheckToggle skin="standard" checked />` to not change the checked state when I click on it.

Take the native `input`:
`<input/>` - this will behave as an `uncontrolled`, meaning it will call `onChange` for every character type and update the internal state value.
`<input value=""/>` - this will behave as `controlled` and will call the `onChange`, but will **not** change the internal state unless explicitly passed as props.

so I think you should also check in this component if you're in controlled mode or not

src/CheckToggle/CheckToggle.st.css
```diff
@@ -0,0 +1,127 @@
...
+  -st-extends: Focusable;
+  -st-states: checked, disabled, size(enum(small, medium)),
+    skin(enum(standard, success)), size(enum(small, medium));
+
+  outline: none;
+  border-radius: 100%;
+}
+
+/* Input */
+.label {
```
shlomitc, line 21: any reason why we need an extra `label` and not merge it with the `root` div?

src/CheckToggle/docs/examples.js
```diff
@@ -0,0 +1,67 @@
+export const skin = `
+<Container>
```
shlomitc, line 2: I think you mixed a bit the `Grid` component with the `Layout` one.

If you use `Col`, you should have `Row` before. `Container` sets min-max width which isn't necessary here.

```jsx
<Layout>
  <Cell span={3}>
    <CheckToggle skin="standard" checked />
  </Cell>
  <Cell span={9}>
    <CheckToggle skin="standard" />
  </Cell>
  <Cell span={3}>
    <CheckToggle skin="success" checked />
  </Cell>
  <Cell span={9}>
    <CheckToggle skin="success" />
  </Cell>
</Layout>
```

src/CheckToggle/docs/examples.js
```diff
@@ -0,0 +1,67 @@
...
+  <Col span={1}>
+    <CheckToggle skin="success" checked />
+  </Col>
+  <Col span={1}>
+    <CheckToggle skin="success" />
+  </Col>
+</Container>
+`;
+
+export const size = `
```
shlomitc, line 24:
```jsx
<Layout>
  <Cell span={3}>
    <CheckToggle checked size="medium"/>
  </Cell>
  <Cell span={9}>
    <CheckToggle  size="medium"/>
  </Cell>
  <Cell span={3}>
    <CheckToggle checked  size="small"/>
  </Cell>
  <Cell span={9}>
    <CheckToggle  size="small"/>
  </Cell>
</Layout>
```

src/CheckToggle/docs/examples.js
```diff
@@ -0,0 +1,67 @@
...
+  <Col span={1}>
+    <CheckToggle size="medium" checked />
+  </Col>
+  <Col span={1}>
+    <CheckToggle size="medium" />
+  </Col>
+</Container>
+`;
+
+export const disabled = `
```
shlomitc, line 47: like the above

src/CheckToggle/docs/examples.js
```diff
@@ -0,0 +1,67 @@
...
+  <Col span={1}>
+    <CheckToggle disabled checked />
+  </Col>
+  <Col span={1}>
+    <CheckToggle disabled />
+  </Col>
+</Container>
+`;
+
+export const tooltip = `
```
shlomitc, line 58: like the above

src/CheckToggle/CheckToggle.js
```diff
@@ -0,0 +1,152 @@
...
+
+  /**
+   * Renders the toggle itself
+   * @returns React.ReactNode
+   * @private
+   */
+  _renderInput = () => {
+    const { size, disabled, onChange } = this.props;
+    let _onChange = null;
+    if (!disabled) {
```
shlomitc, line 54: Since you're rendering input that already handles all input props, why not pass the `disable` prop as well?

I find it a bit confusing that you're conditioning it on the `disabled` while you can actually condition this on the `isControlled`.

as I see it
```js
if(controlled) {
 inputProps = {onChange: this._handleChange, value: props.value, disabled: props.disabled}
} else {
 inputProps = {onChange: props.onChange, disabled: props.disabled}
}
```

src/CheckToggle/CheckToggle.js
```diff
@@ -0,0 +1,152 @@
...
+      disabled,
+      tooltipContent,
+      focusableOnFocus,
+      focusableOnBlur,
+    } = this.props;
+
+    return (
+      <label
+        {...styles('root', { checked, size, skin, disabled }, this.props)}
+        data-checked={checked}
```
shlomitc, line 104: you can just check the actual input instead of adding a custom data- attribute

src/CheckToggle/CheckToggle.uni.driver.js
```diff
@@ -0,0 +1,38 @@
...
+    /** Get the current count */
+    isChecked: async () =>
+      (await base.$(`[data-hook="${dataHooks.toggle}"]`).getNative()).checked, // eslint-disable-line no-restricted-properties
+
+    /** Click the check toggle */
+    click: async () => {
+      const input = await base.$(`[data-hook="${dataHooks.toggle}"]`);
+
+      // eslint-disable-next-line no-restricted-properties
+      const isDisabled = (await input.getNative()).disabled;
```
shlomitc, line 21: Using `getNative()` is not allowed, as it breaks the whole concept of Unidriver.
I'd want this to be natively in Unidriver or use ReactBase as a last resort

mosheKerbel: You have the `_prop` function in UniDriver
So you can 
`const isDisabled = (await input._prop('disabled'))`

PR #5339, opened 2020-04-12
Align status API

EgoziE:
Aligned all status API in all components + added tests

Unfortunately I already did this, I guess somehow it got run over during merges...
This time I added a lot more tests including when there's no status at all (no error is thrown)


src/VariableInput/test/VariableInput.spec.js
```diff
@@ -9,6 +9,7 @@ import privateDriverFactory from './VariableInput.private.uni.driver';
 import VariableInput from '../VariableInput';
 import { sizeTypes } from '../constants';
 import { selectionBehaviorPolyfill } from '../../../testkit/polyfills';
+import Input from '../../Input';
```
shlomitc, line 12: not sure this is needed

src/ColorInput/ColorInput.spec.js
```diff
@@ -6,6 +6,7 @@ import {
   cleanup,
 } from '../../test/utils/react/index';
 import { extractHex } from './hex-helpers';
+import Input from '../Input';
```
shlomitc, line 9: not needed

src/ImageViewer/ImageViewer.driver.d.ts
```diff
@@ -1,4 +1,5 @@
 import { BaseDriver } from 'wix-ui-test-utils/driver-factory';
+import { InputStatus } from '../Input';
```
shlomitc, line 2: Drivers should not be aware of the components' code (JSX). we never do it in tests and shouldn't do it in types as well. 
all such statuses should sit in a shared place, not in `Input`

PR #5346, opened 2020-04-14
Improve docs and readability

EgoziE:
Improved readability of our code + fix some documentation


src/AutoComplete/AutoComplete.js
```diff
@@ -40,4 +23,31 @@ class AutoComplete extends InputWithOptions {
   }
 }
 
+AutoComplete.displayName = 'AutoComplete';
+
+AutoComplete.propTypes = {
```
argshook, line 28: this kind of change i don't agree with. the question of if it's readable is debatable, but i can leave it aside.
what's not good with this type imo is that now instead of one `AutoComplete` in file, there will be three.

EgoziE: @argshook  Thanks for the comment,  I guess I owe you at least an explanation.
Bottom line, all I want is to have more order in our code, we have ~160 components and every one is written differently.

I found this way to be more readable for development so I took the time to arrange some components.
The reasons I think this is more readable is:
* Some component classes have many props and they get huge
* Some components have a list of constants definitions before them
* Moving the props/defaultProps list to the bottom of the file makes it easier to find.

I don't mind moving it inside the component as static fields if you think it's such a bad practice.

argshook: > we have ~160 components and every one is written differently.

And, unfortunately, it will continue to be so. Bottom line is if some
pattern cannot be sanely enforced by automation, then it is doomed to be
dismissed.

> I found this way to be more readable for development so I took the time to arrange some components.

That's cool, however the issue is that next developer will find something else
to be more readable and make a change to their liking. I remember not one occasion
there was a debate about

```diff
-import PropTypes from 'prop-types'
+import { bool } from 'prop-types'
```

Which is opposite to the change in this PR. The argument was that "you
pick only what you need, thus its easier to reason about".

To me it doesn't matter which way is the "norm", i care that things
work. Tomorrow there will be some new pattern anyway.

> Moving the props/defaultProps list to the bottom of the file makes it easier to find.

This is personal taste, i could say it's the opposite for me and we'd
never settle the debate. IDEs usually allow folding code blocks and have
tools to search for code, go to definitions and alike. It should be easy
enough to find things with good tools.

---

I'm not here to discourage changes, but want to point out that not all
change is necessary and unnecessary change can even be harmful.

For example, the surface area of this PR can likely inflict merge
conflicts for other devs, there could be a typo somewhere (which may or
may not be important), contributors will most definitely not know about
new pattern and future PR reviews might cause additional cognitive load
to review, whether `propTypes` are inside or outside the class (or
something similar).

It's kind of like when js didn't have code fixer (i mean
eslint/tslint/prettier) and everyone was like "but muh indentation". Now
no one cares, machine takes care of it, there's no debate.

---

Or maybe in just old, so scratch all that

EgoziE:
closing for now

PR #5289, opened 2020-03-28
Dord live session 3 - !!DO NOT MERGE!!

dorduch:
### 🔦 Summary

...

### ✅ Checklist

- [ ] 👨‍💻 API change is approved by the UI Infra Developers 
- [ ] 👨‍🎨 UX change is approved by the Design System UX 
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test

PR #5348, opened 2020-04-14
<ListItemEditable/> - new component

EgoziE:
Finishing https://github.com/wix/wix-style-react/pull/5289


milkyfruit:
I updated the specs: https://zpl.io/VO8ozEQ

src/ListItemEditable/ListItemEditable.uni.driver.js
```diff
@@ -0,0 +1,76 @@
...
+    enterText: text => inputDriver.enterText(text),
+
+    /** Simulates clicking the approve button */
+    clickApprove: () => approveButtonDriver.click(),
+
+    /** Simulates clicking the cancel button */
+    clickCancel: () => cancelButtonDriver.click(),
+
+    /** Returns true if cancel button has a visible tooltip */
+    isCancelButtonTooltipExists: () =>
```
shlomitc, line 47: looks redundant

src/ListItemEditable/ListItemEditable.uni.driver.js
```diff
@@ -0,0 +1,76 @@
...
+      cancelButtonTooltipDriver.tooltipExists(),
+
+    /** Returns cancel button tooltip text */
+    cancelButtonTooltipText: () => cancelButtonTooltipDriver.getTooltipText(),
```
shlomitc, line 51: I think we can add a `get` prefix for better readability, wdyt?

src/ListItemEditable/ListItemEditable.uni.driver.js
```diff
@@ -0,0 +1,76 @@
...
+
+    /** Hovers the cancel button */
+    hoverCancelButton: () => cancelButtonTooltipDriver.mouseEnter(),
```
shlomitc, line 54: redundant. it's an internal part of getting the tooltip related content

src/ListItemEditable/ListItemEditable.uni.driver.js
```diff
@@ -0,0 +1,76 @@
...
+
+    /** Returns true if approve button has a visible tooltip */
+    isApproveButtonTooltipExists: () =>
```
shlomitc, line 57: redundant

src/ListItemEditable/ListItemEditable.uni.driver.js
```diff
@@ -0,0 +1,76 @@
...
+      approveButtonTooltipDriver.tooltipExists(),
+
+    /** Returns approve button tooltip text */
+    approveButtonTooltipText: () => approveButtonTooltipDriver.getTooltipText(),
```
shlomitc, line 61: `get`

src/ListItemEditable/ListItemEditable.uni.driver.js
```diff
@@ -0,0 +1,76 @@
...
+
+    /** Hovers the approve button */
+    hoverApproveButton: () => approveButtonTooltipDriver.mouseEnter(),
```
shlomitc, line 64: `redundant. it's an internal part of getting the tooltip related content`

src/ListItemEditable/ListItemEditable.uni.driver.js
```diff
@@ -0,0 +1,76 @@
...
+
+    /** Return true if the component size is equal to the given size */
+    isInputOfSize: size => inputDriver.isOfSize(size),
```
shlomitc, line 67: either not have it at all because it's visual property, or ditch the `input` from the name

src/ListItemEditable/constants.js
```diff
@@ -0,0 +1,7 @@
+export const dataHooks = {
+  editableListInput: 'editable-list-input',
```
shlomitc, line 2: want to change the names here as well to `listItemEditableXXX`

src/ListItemEditable/docs/index.story.js
```diff
@@ -0,0 +1,106 @@
...
+            title: 'Description',
+            text:
+              'ListItemEditable is an internal editable component which is used to build dropdown or menu like components. Usually this item should not be used by consumers, though custom options builder is exposed for usage with DropdownBase.',
+          }),
+
+          importExample(`
+// Use directly
+import { ListItemEditable } from 'wix-style-react';
+// Or use a builder
+import { listItemEditableBuilder } from 'wix-style-react/ListItemEditable';
```
shlomitc, line 71: should be picked from the root `wix-style-react`

src/ListItemEditable/docs/index.story.js
```diff
@@ -0,0 +1,106 @@
...
+        title: 'Description',
+        sections: [
+          description({
+            title: 'Description',
+            text:
+              'ListItemEditable is an internal editable component which is used to build dropdown or menu like components. Usually this item should not be used by consumers, though custom options builder is exposed for usage with DropdownBase.',
+          }),
+
+          importExample(`
+// Use directly
```
shlomitc, line 68: I think we can create a common file for all import types
- simple
- with item builder
- compound components
etc...

We have something similar in baseui which works really nicely
https://github.com/wix-private/wix-base-ui/blob/master/stories/storySectionHelpers.js#L9-L49

src/ListItemEditable/ListItemEditable.js
```diff
@@ -0,0 +1,199 @@
...
+  cancelButtonTooltip: PropTypes.string,
+
+  /** Input status - use to display an status indication for the user */
+  status: PropTypes.oneOf(['error', 'warning', 'loading']),
+
+  /** Specifies the size of the input */
+  size: PropTypes.oneOf(['small', 'medium']),
+
+  /** Approve button tooltip text */
+  approveButtonTooltip: PropTypes.string,
```
shlomitc, line 189: is this working well with the work done to consolidate passing tooltip props to components?

src/ListItemEditable/ListItemEditable.js
```diff
@@ -0,0 +1,199 @@
...
+  cancelButtonTooltip,
+  approveButtonTooltip,
+  status,
+  statusMessage,
+}) => ({
+  id,
+  disabled: true,
+  overrideStyle: true,
+  value: props => (
+    <Box margin="3px 24px">
```
shlomitc, line 134: is this only for the builder or also for the regular component?

EgoziE: Builder only:
image

shlomitc: interesting. I think it's the first time we see such a thing.
in the previous list items it wasn't needed, do you know why not it is?

EgoziE: @milkyfruit instructed me to do this. My personal guess is that other list items just don't have any outline like the input has and just start with an already padded text.

shlomitc: Let’s figure out why, because the builder purpose is just to attaching to the dropdown, not introduce new styles
As I see it - all list items styles should be in the component, builder is just a helper

milkyfruit: I aligned the specs to match the examples. It was shown as a stand alone and as in the list.
Previous didn't specify how it should be be layouted in the list: https://zpl.io/2yZEJ19

For what I know this kind of component can be used in dropdown lists and EditableSelector:
https://wix-style-react.now.sh/?path=/story/components-api-components--editableselector

src/ListItemEditable/ListItemEditable.uni.driver.js
EgoziE: removed

src/ListItemEditable/docs/index.story.js
EgoziE: I think it's should be part of  https://jira.wixpress.com/browse/DSM-258
Mentioned your comment there

src/ListItemEditable/ListItemEditable.js
EgoziE: Changed to:

cancelButtonTooltipContent
cancelButtonTooltipProps

approveButtonTooltipContent
approveButtonTooltipProps

src/ListItemEditable/ListItemEditable.js
shlomitc: I think this should be an explicit feature prop of the list item.

By default, it will apply the padding (because it's a list item), and when set to false, it will remove it, for other use cases. And now, the builder is clean.
Another option is to do the opposite, and the builder will just turn on the flag. both options fine by me.

WDYT @milkyfruit @EgoziE

EgoziE: More props means more maintenance. (and sometimes mistakes by users)
But let's say we'll add a prop. As I see it, making the builder turn the flag is changing styles exactly like wrapping it with a div does, the builder is not "clean" of styles.

That makes me wonder, why should a builder be clean of styles?
By passing `overrideStyle: true` the builder already says to the layout, let me handle this.

shlomitc: well, we still didn't finalize the concept of builders, but the original purpose was to make it easy to integrate with our dropdowns since it's API is super complicated. If I would rewrite the dropdown component again, it would probably be something like:
```jsx
<Dropdown>
  { (getItemProps) => 
    <ListItemSelect {...getItemProps({id:1})} /> 
    <ListItemSelect {...getItemProps({id:2})} /> 
    <ListItemEditable enableTheAmazingPadding {} {...getItemProps({id:3})}/> 
  }
</Dropdown>
```

this would make the builders redundant, and we would just use the render props from the dropdown to achieve this beautiful markup.
in that case, turning on/off the padding flag would be necessary, so I think this to be a feature of the component and not the builder.

Issue #5277, opened 2020-03-26
ellipsis inside dropdownBase don't work


marik22312:
# 🐛 Bug Report

### 🏗 Relevant Components

`<Dropdown/>`

### 😯 Current Behavior

when having components with ellipsis (Text, FormField) as a trigger for a dropdown, seems that the ellipsis don't work at all

### 🤔 Expected Behavior
The text should be hidden with ellipsis, a tooltip should be displayed on hover.

### 👣 Steps to Reproduce
```javascript
  <DropdownBase
    dynamicWidth
    appendTo="window"
    zIndex={5500}
    options={this.renderFieldTagsOptions()}
    onSelect={tag => this.handleSelectTagClick(tag.id)}
    open={this.state.isOpen}
    onClickOutside={() => this.setState({ isOpen: false })}
   >
    <TextButton
       placement="top"
       skin="dark"
    >
      <FormField
		label="very long test text"
      />
     </TextButton>
   </DropdownBase>
```

  - **WSR version**: `7.48`
  - **Yoshi version**: `4.1`

### 👀 Severity

- Critical - it's blocking a feature release.

mosheKerbel:
To refine the problem - textButton has no ellipsis support. Putting inside of it `formField` or `text` still won't help.
@mykas as i understand there's a plan to add support for `ellipsis` to `TextButton`?
Anyways, here's an exact example to reproduce in the playground:
```jsx
<Box width="150px" border="1px solid black">
  <DropdownBase
    data-hook="story-dropdown-base-uncontrolled-click"
    options={[
      { id: 0, value: 'First option' },
      { id: 1, value: 'Second option' },
      { id: 2, value: 'Third option' },
      { id: 3, value: 'Fourth option' },
      { id: 4, value: 'Fifth option' },
      { id: 5, value: 'Sixth option' },
    ]}
  >
    {({ toggle, selectedOption = {} }) => {
      return (
        <TextButton
          skin="dark"
          suffixIcon={<Icons.ChevronDown />}
          onClick={toggle}
          dataHook={'drop-down-opener'}
        >
          {selectedOption.value || 'Please choose an option'}
        </TextButton>
      );
    }}
  </DropdownBase>
</Box>;
```

mykas:
Hi, currently we have an issue with this kind of approach in getting the ellipsis working deep in nested wrappers when only the parent component defines the width. For this to work we need to develop our ellipsis solution in different way. 

[...]

Issue #4612, opened 2019-11-28
<TextButton/> - Add ellipsis


EgoziE:
# ✨ Feature Request

### 🔦 Scope
Add ellipsis to `<TextButton/>` using the HOC.

**NOTE!**
When done, please remove the hard-codded ellipsis in `<ListItemSection/>` (src/ListItemSection/ListItemSection.st.css)

### 👀 Severity
Low

PR #5349, opened 2020-04-15
`<DropdownBase/>, <TextButton/>` - introduce fluid working & examples

mykas:
This PR is about long awaited feature where users would like to have ellipsis working for Button,TextButton and other places. By introducing fluid prop in core for Popover trigger element now user can tell `<DropdownBase/>` to flow with the parent container width instead of going inline based on children content.

Example like this now works:
```jsx
<Box width="150px">
  <TextButton fluid>
    <Text ellipsis={true}>TextButton that gets shrinked</Text>
  </TextButton>
</Box>;
```
Screenshot 2020-04-15 at 15 06 17

Which now also can be used together with DropdownBase:
```jsx
<Box align="center">
  <div style={{ width: '100px' }}>
    <DropdownBase
      fluid
      options={[
        { id: 0, value: 'First option' },
        { id: 1, value: 'Second option' },
        { id: 2, value: 'Third option' },
        { id: 3, value: 'Fourth option' },
        { id: 4, value: 'Fifth option' },
        { id: 5, value: 'Sixth option' },
      ]}
    >
      {({ toggle, selectedOption = {} }) => {
        return (
          <TextButton
            skin="dark"
            fluid
            suffixIcon={<Icons.ChevronDown />}
            onClick={toggle}
            dataHook={'drop-down-opener'}
          >
            <Text ellipsis={true}>
              {selectedOption.value || 'Please choose'}
            </Text>
          </TextButton>
        );
      }}
    </DropdownBase>
  </div>
</Box>;
```
Screenshot 2020-04-15 at 15 07 11

fixes https://github.com/wix/wix-style-react/issues/5277 resolves https://github.com/wix/wix-style-react/issues/4612


mosheKerbel:
Nice one! a very elegant solution.

A reminder - Please update tsx tests when modifying d.ts props :)
`TextButton.tsx` `DropdownBase.tsx`

mykas:
we might need another sync-ts for those tests :D I always forget

PR #5354, opened 2020-04-16
release: 7.56.0

nitayneeman:


nitayneeman:
Merging this PR.

The broken visuals were excluded purposefully (there are Chrome version conflicts and sort of a technical gap with Applitools - but both aren't directly related to the minor feature we're trying to release).

PR #5356, opened 2020-04-16
`style` added to Page.* components.

daliusd:


daliusd:
> Let's focus only on `Page.Sticky`, not all other components use `style`

Done.

PR #5357, opened 2020-04-16
Update components_guidelines.md

mosheKerbel:
@shlomitc


docs/contribution/COMPONENT_GUIDELINES.md
```diff
@@ -29,6 +29,13 @@ Every component folder should be a single component.
 | *xxxProps*        | If the component is wrapping another component (or native DOM element) and its props is needed - pass it with the convention of xxxProps. for example inputProps tooltipProps, etc..
 | `PropTypes.node` instead of `PropTypes.string` | a frequent case for props that accept string is to use a component which evaluates to string. For example, component that handles translation or experiment, or some kind of provider. For those cases `PropTypes.string` would give a warning, therefore almost all `PropTypes.string` cases should actually be `PropTypes.node` |
 
+## Function names
```
shlomitc, line 32: Component instance methods

docs/contribution/COMPONENT_GUIDELINES.md
```diff
@@ -29,6 +29,13 @@ Every component folder should be a single component.
...
+
+The functions of the component should be prefixed with `_`.  For example - `_handleOnClick`. <br/>This convention implies that the function is private and should not be tiggered externaly by the component's consumers.
```
shlomitc, line 34: I'd separate here into two parts - private methods and public methods, explaining the obvious case first.


Every method that is set on a React class component is publically available on the component `ref`. Methods like `focus()` or `blur()` are common public API for input elements and should be exposed to the consumer. Other methods, for example, `_handleButtonClick()` are usually private and should not be exposed to the consumer.

The library's convention is to add an `_` prefix for any private method. This convention is also handled as part of the `wix-storybook-utils` documentation tool, dropping it from the public API documentation.

PR #5360, opened 2020-04-16
fix types for PopoverMenu.MenuItem text prop

brumeregan:
component `<PopoverMenu.MenuItem />` should accept prop `text` with type string and Node


EgoziE:
As far as the design indicates, text should accept only text.
Need to add an Icon? please use prefixIcon prop.

If your design is completely different, let's discuss in the channel with both your UX and ours.
I'm closing this PR for now.

mykas:
@EgoziE you forgot about translation strings which are node based

EgoziE:
@brumeregan would you please provide an example?
Are your translations not text based?

EgoziE:
Ok I got this, will reopen and merge.

brumeregan:
We are implementing Roles&Permissions feature and it was designed like that
Screen Shot 2020-04-21 at 1 58 58 PM

PR #5369, opened 2020-04-19
<Avatar> added loading prop

adirou:
### 🔦 Summary

...

### ✅ Checklist

- 📚 Change is documented
  - [x] Story
  - [x] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [x] Visual test


src/Avatar/Avatar.st.css
```diff
@@ -242,30 +242,43 @@
...
   height: 72px;
   width: 72px;
   border-radius: 36px;
 }
-
-.avatarContainer:size(size60) .avatar {
+.size60-mixin{
   height: 60px;
   width: 60px;
   border-radius: 36px;
```
TalZa1810, line 259: the mixins should be without the `border-radius` since we have two shapes- `circle` and `square`.

src/Avatar/test/Avatar.visual.js
```diff
@@ -156,3 +156,7 @@ storiesOf(`Avatar/Props`, module).add('Placeholder', () => (
     {allSizesAvatarWithProps({ shape: 'circle', placeholder: <div /> })}
   </div>
 ));
+
+storiesOf(`Avatar/Props`, module).add('Loading', () => (
+  <div>{allSizesAvatarWithProps({ loading: true })}</div>
```
TalZa1810, line 161: I think that here a test with shape `square` should be added too.

src/Avatar/Avatar.js
```diff
@@ -214,6 +220,9 @@ Avatar.propTypes = {
   onIndicationClick: PropTypes.func,
   /** Show indication on hover. */
   showIndicationOnHover: PropTypes.bool,
+  /** Show indication only when hover */
```
TalZa1810, line 223:
```suggestion
  /** Avatar in loading state when set to true */
```

src/Avatar/Avatar.js
```diff
@@ -214,6 +220,9 @@ Avatar.propTypes = {
...
+  loading: PropTypes.bool,
+  /** Avatar in loading state. */
```
TalZa1810, line 225: prop description should be above the prop itself so that it generates the prop description correctly.
Currently, it generated "Show indication only when hover" for the `loading` prop. 
https://wix-wix-style-react-pr-5369.surge.sh/?activeTab=API&path=%2Fstory%2Fcomponents-api-components--avatar

shlomitc:
1. This is not yet described in the Zeplin https://zpl.io/V11PxE4, @ayeletr - is this planned to be added?
2. I don't think this should be a `loading` prop, but rather an internal behavior while an image is loaded. @adirou - can you explain how this is going to be used by projects?

adirou:
@shlomitc 

1. this is the us
/https://www.google.com/url?q=https://zpl.io/VOGWE8Y&sa=D&source=hangouts&ust=1587372945719000&usg=AFQjCNF9u3FoHOJx1pqs7lTgUWE7yaUsNg

2.main scenario:
2.a. avatar shows placeholder or other image
2.b. The user clicks the avatar or indication.
2.c avatar moving to loading state.
2.d. choosing an image from the file system
2.e. image file uploading.
2.f. image exit loading state and get new img prop.
2.g. image shows new image.

i think this is the simple solution

shlomitc:
@adirou 
So if I understand correctly, the loader should appear while the image is loaded, which means it needs to know when the image started its loading to show the loader, and remove it when finished.

These events are not part of the component's props, thus it should probably be an internal logic of the component.

adirou:
@shlomitc 
and how the component will know this behavior internaly?
would you prefer a promise to some img prop instead?  doesn't it become a little bit complex.

ayeletr:
The loader it self shouldn't have opacity, according to spec

adirou:
@ayeletr fixed

src/Avatar/Avatar.st.css
```diff
@@ -242,24 +242,27 @@
 .avatar:contentType(text).colorA6 { background-color: value(A6); }
 
 /* Sizes */
+
+.avatarContainer:size(size90) .loaderContainer,
 .avatarContainer:size(size90) .avatar {
   height: 90px;
   width: 90px;
-  border-radius: 45px;
+  border-radius: 45px;
```
shlomitc, line 250: redundant space

src/Avatar/Avatar.uni.driver.js
```diff
@@ -15,6 +18,7 @@ export default base => {
     clickIndication: () => getIndication().click(),
     clickCustomIndication: () => getCustomIndication().click(),
     indicationExists: () => getIndication().exists(),
+    loaderExists: () => getLoader().exists(),
```
shlomitc, line 21: since the prop name is `loading`, then the equivalent in the driver should be  `isLoading`

adirou:
@shlomitc updated

src/Avatar/Avatar.uni.driver.d.ts
```diff
@@ -4,5 +4,6 @@ export default interface AvatarUniDriver extends AvatarDriverCore {
   clickIndication: () => Promise<void>;
   clickCustomIndication: () => Promise<void>;
   indicationExists: () => Promise<boolean>;
+  loaderExists: () => Promise<boolean>;
```
shlomitc, line 7: remove

shlomitc: actually change with `isLoading`

adirou:
types has been fixed

PR #5370, opened 2020-04-19
<AutoCompleteWithLabel /> - fix hasError in testkit 

saarkuriel:
#pr


EgoziE:
The problem was deeper than it seem.
`<Input/>` used the internal `<StatusIndicator/>` to determine if there's a status.
`<AutoCompleteWithLabel/>` used the prop hideStatusSuffix that made this check obsolete.

Therefore changed `<Input/>` to check a data-attribute instead.

saarkuriel:
@EgoziE thanks for taking care of this!

Issue #5353, opened 2020-04-15
<AddItem /> add missing prop "tooltipContent" 


miroshSa:
# 🐛 Bug Report

### 🏗 Relevant Components

<AddItem />

...

### 😯 Current Behavior

prop "tooltipContent" is missing in version 8

...

### 🤔 Expected Behavior

prop "tooltipContent" should exist for this component
...

### 👣 Steps to Reproduce

  - **WSR version**: `8.1.0`
  - **Yoshi version**: `4.56.1`

### 👀 Severity

- Critical

EgoziE:
PR - https://github.com/wix/wix-style-react/pull/5371

PR #5371, opened 2020-04-19
<AddItem/> & <FillButton/> - Restore tooltipContent prop

EgoziE:
Restoring `tooltipContent` prop
We removed `tooltipContent` prop in order to use `tooltipProps` prop, but then we decided to exclude `content` from it. 

I made sure both are supported in order not to break anyone!
TS users will have to use `tooltipContent={...}`
JS users can use both `tooltipContent={...}` and `tooltipProps={{ content: ... }}`


EgoziE:
Resolves https://github.com/wix/wix-style-react/issues/5353

PR #5378, opened 2020-04-20
Modals - move story docs out of WIP

mosheKerbel:
Since I moved it out of WIP, should it replace the old modal layouts?
Should I add notifications in the old modals layouts referencing to the new one?
@shlomitc @itaytay


shlomitc:
@mosheKerbel - The `CustomModalLayout` is the only one approved by UX to roll-out.
The two other are still a WIP since we need to change the font heading to something that doesn't exist in the library yes

mosheKerbel:
go it, will change it then.

PR #5380, opened 2020-04-20
<DateInput/> - remove public story for inactive comopnent

mosheKerbel:
@shlomitc please review


shlomitc:
If the component is not used - I think we can remove it completely. Wdyt?

mosheKerbel:
I just didn't want to loose a code we might need to work on in the future
If you say it's irrelevant then I will remove.

mosheKerbel:
> If the component is not used - I think we can remove it completely. Wdyt?

it's actually being used inside the `DatePicker`
so I will nest it in it's folder and remove the story.

mosheKerbel:
1. Removed the story
2. Removed from cheatsheet
3. Moved under DatePicker which still uses DateInput

mosheKerbel:
@argshook Autodocs is still trying to parse DateInput
I removed the story and nested the component dir under `DatePicker` dir
can u take a look?
http://pullrequest-tc.dev.wixpress.com/viewLog.html?buildId=7459193&buildTypeId=FedInfra_WixStyleReac_WixStyleReactNew_Parallel_Tests_Test2Visual&tab=buildLog&_focus=659#_state=659
image

argshook:
@mosheKerbel 
https://github.com/wix/wix-style-react/blob/dateInput-remove-story/src/DatePicker/DateInput/DateInput.meta.js#L7

because files were moved, this here has a wrong path, should be changed:
this should have a change:

```diff
-path: 'src/DateInput/DateInput.js',
+path: 'src/DatePicker/DateInput/DateInput.js',
```

mosheKerbel:
> @mosheKerbel
> https://github.com/wix/wix-style-react/blob/dateInput-remove-story/src/DatePicker/DateInput/DateInput.meta.js#L7
> 
> because files were moved, this here has a wrong path, should be changed:
> 
> ```diff
> -path: 'src/DateInput/DateInput.js',
> +path: 'src/DatePicker/DateInput/DateInput.js',
> ```

that was silly :(
thanks!

PR #5385, opened 2020-04-20
<LabelledElement> - move out of WIP

mosheKerbel:
Move to a new category under `Components API` -> `Internal`
@shlomitc


shlomitc:
I wonder if it's actually a component anyone would like to use. it looks to me more like an internal component

mosheKerbel:
> I wonder if it's actually a component anyone would like to use. it looks to me more like an internal component

that's why I put it under Internal instead of removing the story
this way we can easily play with it for dev purposes.
I can add a doc in the category describing it.

PR #5387, opened 2020-04-20
<Timeline/> - new component

adimorwix:


src/Timeline/index.d.ts
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
+
+export interface TimelineItem {
+  id: string | number;
+  label: string;
+  labelAction?: React.ReactNode;
+  customPrefix?: React.ReactNode;
+  suffix?: React.ReactNode | string;
```
EgoziE, line 8: string is covered inside React.ReactNode, can be removed

src/Timeline/Timeline.uni.driver.js
```diff
@@ -0,0 +1,42 @@
+import { baseUniDriverFactory } from 'wix-ui-test-utils/base-driver';
+import { dataHooks } from './constants';
+
+export const timelineDriverFactory = (base, body) => {
+  return {
+    ...baseUniDriverFactory(base, body),
+
+    /** Is the timeline exists */
+    isTimelineExists: () =>
```
shlomitc, line 9: `baseUniDriverFactory` already has `exists()` so I think you can remove it

src/Timeline/Timeline.uni.driver.js
```diff
@@ -0,0 +1,42 @@
...
+      base.$(`[data-hook="${dataHooks.timelineList}"]`).exists(),
+
+    /** Get the label text by item id */
+    getLabelText: async itemsId =>
```
shlomitc, line 13: rename to `getLabelTextAt(idx)`
where `idx` is the position of it, instead of an id

src/Timeline/Timeline.uni.driver.js
```diff
@@ -0,0 +1,42 @@
...
+      await base
+        .$(`[data-hook="${dataHooks.timelineLabel}-${itemsId}"]`)
+        .text(),
+
+    /** Get the suffix text by item id */
+    getSuffixText: async itemsId =>
```
shlomitc, line 19: same comment as `getLabelText`

src/Timeline/Timeline.uni.driver.js
```diff
@@ -0,0 +1,42 @@
...
+      base.$(`[data-hook="${dataHooks.timelineLabelAction}-${itemsId}"]`),
+
+    /** Is default prefix wrapper by item id */
+    isDefaultPrefixExists: async itemsId =>
+      await base
+        .$(`[data-hook="${dataHooks.timelineDefaultPrefix}-${itemsId}"]`)
+        .exists(),
+
+    /** label action wrapper by item id */
+    getCustomPrefixElement: itemsId =>
```
shlomitc, line 39: same comment as `getLabelActionElement`

src/Timeline/Timeline.js
```diff
@@ -0,0 +1,117 @@
...
+  /** Applied as data-hook HTML attribute that can be used in the tests */
+  dataHook: PropTypes.string,
+
+  /** A css class to be applied to the component's root element */
+  className: PropTypes.string,
+
+  /** timeline events items */
+  items: PropTypes.arrayOf(
+    PropTypes.shape({
+      id: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
```
shlomitc, line 106: I think an `id` is not needed, we can simply use the index of the item

src/Timeline/Timeline.js
```diff
@@ -0,0 +1,117 @@
...
+import styles from './Timeline.st.css';
+import { dataHooks } from './constants';
+
+/** A timeline is a display of a list of events */
+class Timeline extends React.PureComponent {
+  state = {};
+
+  _isString = a => typeof a === 'string';
+
+  _getListItem(item) {
```
shlomitc, line 14: can you please split this to a `<TimeLineItem/>` component?
When a method on a component becomes that big, I prefer it separated

src/Timeline/Timeline.js
```diff
@@ -0,0 +1,117 @@
...
+            <div
+              data-hook={`${dataHooks.timelineCustomPrefix}-${item.id}`}
+              className={styles.prefixCustom}
+            >
+              {item.customPrefix}
+            </div>
+          ) : (
+            <div
+              data-hook={`${dataHooks.timelineDefaultPrefix}-${item.id}`}
+              className={styles.prefixDefault}
```
shlomitc, line 32: let's call it `bulletIndicator` instead of prefixDefault

src/Timeline/Timeline.js
```diff
@@ -0,0 +1,117 @@
...
+        </div>
+        <div className={styles.label}>
+          <Text
+            dataHook={`${dataHooks.timelineLabel}-${item.id}`}
+            weight="normal"
+            size="small"
+          >
+            {item.label}
+          </Text>
+          <span> </span>
```
shlomitc, line 44: I think we can remove this `span` and just use some margin on the `labelAction` class

src/Timeline/docs/index.story.js
```diff
@@ -0,0 +1,206 @@
...
+              '    labelAction: (\n' +
+              '      <TextButton weight="normal" size="small">\n' +
+              '        View Order\n' +
+              '      </TextButton>\n' +
+              '    ),\n' +
+              '  },\n' +
+              ']}/>',
+          }),
+
+          code({
```
shlomitc, line 142: you can delete this one

src/Timeline/Timeline.js
adimorwix: If only the action down a line we have unnecessary space with a margin

.wuf/components.json
```diff
@@ -424,5 +424,8 @@
   },
   "VerticalTabsItem": {
     "path": "src/VerticalTabsItem"
+  },
+  "Timeline": {
```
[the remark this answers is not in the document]
EgoziE: @shlomitc https://github.com/wix/wix-style-react/pull/5390

src/Timeline/Timeline.js
shlomitc: so we can add `margin-right` to the text, so it will break one word earlier

src/Timeline/Timeline.js
```diff
@@ -0,0 +1,118 @@
...
+  /** Applied as data-hook HTML attribute that can be used in the tests */
+  dataHook: PropTypes.string,
+
+  /** A css class to be applied to the component's root element */
+  className: PropTypes.string,
+
+  /** timeline events items */
+  items: PropTypes.arrayOf(
+    PropTypes.shape({
+      /** event text */
```
shlomitc, line 106: timeline item textual description

src/Timeline/index.d.ts
adimorwix: In labelAction too for example?
I think type make it more clear

src/Timeline/Timeline.st.css
```diff
@@ -0,0 +1,82 @@
+:import {
+  -st-from: "../Foundation/stylable/colors.st.css";
+  -st-named: D30,D80;
+}
+
+.root {
+  text-align: center;
```
shlomitc, line 7: not needed

src/Timeline/Timeline.st.css
```diff
@@ -0,0 +1,82 @@
...
+  background-color: value(D80);
```
shlomitc, line 8: not needed.
Usually, this component and similar widgets will appear on top of a Card component. We can add another example in the playground that uses `<Card/>` and `<Card.Content/>`

src/Timeline/Timeline.st.css
```diff
@@ -0,0 +1,82 @@
...
+  padding-left: 23px;
```
shlomitc, line 9: not needed

src/Timeline/Timeline.st.css
```diff
@@ -0,0 +1,82 @@
...
+  padding-right: 30px;
```
shlomitc, line 10: not needed

src/Timeline/Timeline.st.css
```diff
@@ -0,0 +1,82 @@
...
+  padding-top: 28px;
```
shlomitc, line 11: not needed

src/Timeline/Timeline.st.css
```diff
@@ -0,0 +1,82 @@
...
+}
+
+.timeline {
```
shlomitc, line 14: all of these are not needed besides the `list-style: none;`
Maybe we can just merge this with `root` and remove the `div` from the markup?

src/Timeline/Timeline.st.css
```diff
@@ -0,0 +1,82 @@
...
+  list-style: none;
+  text-align: left;
+  padding: 0;
+  margin: 0;
+}
+
+.timeline .event {
+  display: list-item;
+  position: relative;
+  padding-bottom: 30px;
```
shlomitc, line 26: This should appear for all besides the last child

src/Timeline/Timeline.st.css
```diff
@@ -0,0 +1,82 @@
...
+  border-left: 1px solid value(D30);
```
shlomitc, line 27: it's an interesting solution, but this means that the prefix should cover it. I guess it will cause some small issues in CSS so let's see if we can overcome them in an elegant way

src/Timeline/Timeline.st.css
```diff
@@ -0,0 +1,82 @@
...
+  display: block;
+  position: relative;
+  list-style: none;
+  text-align: left;
+  padding: 0;
+  margin: 0;
+}
+
+.timeline .event {
+  display: list-item;
```
shlomitc, line 24: what's the point of using list-item if we don't actually display it? maybe it should be semantically used as `ul` and `li` in the markup, this will lead to better a11y

src/Timeline/Timeline.st.css
```diff
@@ -0,0 +1,82 @@
...
+  padding-bottom: 30px;
+  border-left: 1px solid value(D30);
+}
+
+.timeline .event:last-child {
+  border-left: none;
+}
+
+.timeline .event .prefix{
+  width: 3%;
```
shlomitc, line 35: I don't think we use percentages in the values of items. it's either fixed pixel size or "flex" containers, please find the right solution following this description
image

src/Timeline/Timeline.st.css
```diff
@@ -0,0 +1,82 @@
...
+  display: inline-block;
+}
+
+.timeline .event .prefix .prefixDefault{
+  position: absolute;
+  display: block;
+  background: value(D30);
```
shlomitc, line 42: background should be transparent

src/Timeline/Timeline.st.css
```diff
@@ -0,0 +1,82 @@
...
+.timeline .event:last-child {
+  border-left: none;
+}
+
+.timeline .event .prefix{
+  width: 3%;
+  display: inline-block;
+}
+
+.timeline .event .prefix .prefixDefault{
```
shlomitc, line 39: let's split this, there should be a standalone `bulletIndicator` which is the component styles (width, height, border-radius, ...) and there should be 

`.prefix` that takes care of the positioning relative to the parent node. both classes should be applied to the HTML node.

src/Timeline/Timeline.st.css
```diff
@@ -0,0 +1,82 @@
...
+  position: absolute;
+  display: block;
```
shlomitc, line 41: isn't it block by default?

src/Timeline/Timeline.st.css
```diff
@@ -0,0 +1,82 @@
...
+  left: -4px;
+}
+
+.timeline .event .prefix .prefixCustom:first-child{
+  position: absolute;
+  display: block;
+  border-radius: 50%;
+  content: "";
+  top:-1px;
+  box-shadow: 0 0 0 1px value(D80);
```
shlomitc, line 58: looks strange, why is this needed?

src/Timeline/Timeline.st.css
```diff
@@ -0,0 +1,82 @@
...
+  content: "";
+  top: 4.5px;
+  box-shadow: 0 0 0 5px value(D80);
+  left: -4px;
+}
+
+.timeline .event .prefix .prefixCustom:first-child{
+  position: absolute;
+  display: block;
+  border-radius: 50%;
```
shlomitc, line 55: I don't think this should be, this should be the responsibility of the custom prefix node which is rendered by the consumer

src/Timeline/Timeline.st.css
```diff
@@ -0,0 +1,82 @@
...
+  content: "";
+  top:-1px;
+  box-shadow: 0 0 0 1px value(D80);
+  background-color: value(D80);
```
shlomitc, line 59: not needed

src/Timeline/Timeline.st.css
```diff
@@ -0,0 +1,82 @@
...
+  left: -12px;
+}
+
+.timeline .event .label{
+  width: 67%;
```
shlomitc, line 64: fix according to spec constraints

src/Timeline/Timeline.st.css
adimorwix: only the padding-bottom and border-left should not exist in  last-child

src/Timeline/Timeline.st.css
adimorwix: we can't use transparent color. the color cover the border

src/Timeline/Timeline.st.css
adimorwix: the color cover the border

src/Timeline/Timeline.st.css
adimorwix: We have to override the defaults

src/Timeline/TimelineItem.st.css
```diff
@@ -0,0 +1,67 @@
...
+.event .suffix{
+  width: 25%;
+  display: inline-block;
+  text-align: right;
+  vertical-align: top;
+  padding-top: 3px;
+}
+
+.event .labelAction{
+  display: inline-block;
```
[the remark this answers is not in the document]
adimorwix: the action should be in the same line

ayeletr:
Some fixes needed:
Suffix - align to content top, and adjust color.
Remove for now the different prefix, we will define them when we’ll have specific requests.

milkyfruit:
Why `label` is string and `suffix` is a node?
Can't it all be made as a node?

Also, I think the first example should be more simple showing the main structure. e.g.:
```
<Timeline items={[
  {label:'Item 1',},
  {label:'Item 2',}
  ]}
/>
```

shlomitc:
@milkyfruit - The current requirements were for a textual label, it is also important to keep a line-height of 18px string in order to align the "bullet" and make the line stretch.
If we decide on having an open node, this will change the component structure, so I prefer not doing so until we have an actual need.

In any case, let's open a JIRA task related to documentation fixes, I think the code is ready to merge and I don't want to block it for this

PR #5388, opened 2020-04-20
`<ModalPreviewLayout/>`- fix test

TalZa1810:
### 🔦 Summary
Fix flaky test


TalZa1810:
ok, I see now that it still hasn't resolved the issue. There's still a tiny gap which I cannot understand what causes this change:


I think that maybe we should skip this test, add an issue about this and investigate this later. wdyt?

shlomitc:
opening an issue means just sits in the backlog and no one will ever touch it until I close it for inactivity.
So either investigate a bit deeper to see where this issue is coming from, or delete this test and forever hope it will never break

mykas:
Hey @shlomitc , lets remove this test. I don't see the benefit of this except for checking the styling of tooltip. We can already check in unit tests if tooltip appears on hover.

TalZa1810:
ok, I've decided to remove the e2e file and moved the navigation button tests to the visual.js file.

PR #5390, opened 2020-04-21
Makes sure that new generated components are ordered alphabetically

EgoziE:
.wuf/components.json is safe to remove


package.json
```diff
@@ -29,7 +29,7 @@
     "storybook:build": "npm run prestorybook && build-storybook -s ./src/assets",
     "------new components generators": "",
     "generate": "wuf generate --output src --codemods .wuf/generator/codemods --templates .wuf/generator/templates",
-    "postgenerate": "npm run update-components",
+    "postgenerate": "rm ./.wuf/components.json && npm run update-components",
```
argshook, line 32: should be part of `.wuf/update-components.sh`

EgoziE: good idea, i'll change it

EgoziE: changed and verified it works

.wuf/update-components.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+rm ./.wuf/components.json
```
argshook, line 3: sorry for nitpicking a bit more :)

```suggestion
rm .wuf/components.json || true
```

1. the `./` is not necessary at the beginning of path
2. the `|| true` part will allow `rm` to fail but continue executing the rest of the script.
   that's important in the case if `.wuf/components.json` doesn't exists for example, freshly cloned repo.
   `rm` would throw and the rest of `update-components.sh` would be prevented from executing

EgoziE: cool thanks

Issue #5393, opened 2020-04-21
babel-plugin-dynamic-import-node - version fixation because of bug


mykas:
Related to this issue: airbnb/babel-plugin-dynamic-import-node#88

mykas:
@EgoziE I assign you this issue because if we are gone to Thunderbolt - maybe you can handle this and unpin the version when they fix it?

EgoziE:
Sure

PR #5398, opened 2020-04-22
use latest babel-plugin-dynamic-import-node again

shlomitc:
check if this issue is resolved https://github.com/wix/wix-style-react/issues/5393


mykas:
Nice, it works!

PR #5401, opened 2020-04-22
Madefor font story page

EgoziE:
People starts asking how to use the new Madefor font!
So I created a small story page to answer all questions.

For now this is all the information needed, will expand this story if more questions pop.
image


shlomitc:
@EgoziE - the experiment doesn't relate to the library, so I don't want it to leak here. let's keep it generic - explaining consumers to load the font.
in the business manager case, it already loads the Helvetica font, so it won't be special.
We can create an internal doc explaining each one of our consumer should apply it in the relevant platform (business manager, editor, etc...)

PR #5406, opened 2020-04-26
`<FontUpgrade/>`- make changes scoped

TalZa1810:
### 🔦 Summary
Using `FontUpgrade` context in the following components:
- Avatar
- Breadcrumbs
- Button
- EmptyState
- SectionHelper
- SidebarSectionItem

Deleting `utils` file for font upgrade and moving the relevant functions to the `FontUpgrade` component itself.


src/FontUpgrade/FontUpgrade.js
```diff
@@ -2,12 +2,25 @@ import React from 'react';
...
+  _setMadefor(value) {
+    if (this._isWindowDefined()) {
+      if (value) {
+        window['useMadeforFont'] = value;
+      } else {
+        delete window['useMadeforFont'];
+      }
+    }
+  }
+
```
EgoziE, line 20: no need for window['useMadeforFont'] anymore, can be deleted

PR #5412, opened 2020-04-26
Fix flaky visual tests

EgoziE:
pre-load images and wait for animation


src/MediaOverlay/test/MediaOverlay.visual.js
```diff
@@ -11,14 +11,21 @@ import {
...
+import { snap, story, visualize } from 'storybook-snapper';
 
 const createDriver = () =>
   mediaOverlayTestkitFactory({
     wrapper: document.body,
     dataHook: storySettings.dataHook,
   });
 
-const hover = async () => await createDriver().hover();
+const wait = timeToDelay =>
```
shlomitc, line 22: I think we can put it once and for all in a common location as it is shared across many tests

src/MediaOverlay/test/MediaOverlay.visual.js
```diff
@@ -173,19 +180,21 @@ const MediaOverlayWrapper = ({ componentDidMount, rtl, ...props }) => {
...
-    );
+visualize('MediaOverlay', () => {
+  tests.forEach(({ describe, its }) => {
+    its.forEach(({ it, props, componentDidMount, rtl }) => {
+      story(describe, () => {
+        snap(it, done => (
+          <MediaOverlayWrapper
+            {...commonProps}
+            {...props}
+            done={done}
```
shlomitc, line 191: instead of passing a new `don` prop, simply pass `componentDidMount={() => componentDidMount(done)}

src/Carousel/test/Carousel.visual.js
```diff
@@ -81,15 +80,19 @@ const tests = [
...
-      ),
-    );
+visualize('Carousel', () => {
+  tests.forEach(({ describe, its }) => {
+    its.forEach(({ it, props }) => {
+      story(describe, () => {
+        snap(it, done => (
+          <div style={{ maxWidth: '550px' }}>
+            {/* Load the first image separately and snap when it loads */}
+            <img src={sampleImages[0]} alt="" onLoad={() => done()} hidden />
```
shlomitc, line 90: You can simply use the `isLoading` method from the driver, waiting for it to become `false`

src/Carousel/test/Carousel.visual.js
```diff
@@ -1,7 +1,9 @@
-import React from 'react';
-import { storiesOf } from '@storybook/react';
-
+import React, { useEffect } from 'react';
+import { visualize, story, snap } from 'storybook-snapper';
 import Carousel from '..';
+import { carouselTestkitFactory } from '../../../testkit';
```
shlomitc, line 4: Please don't use the testkit. They are public API for consumers and not internally in our tests.
 We aim for this to be auto-generated file during the build, and not something to rely on in our tests (same as unit-tests don't use it)

src/Carousel/test/Carousel.visual.js
```diff
@@ -81,15 +83,44 @@ const tests = [
...
-    );
+const createDriver = () =>
+  carouselTestkitFactory({
+    wrapper: document.body,
+    dataHook: storySettings.dataHook,
+  });
+
+const checkIsLoading = done => {
+  const driver = createDriver();
+  const interval = setInterval(async () => {
```
shlomitc, line 94: you can `eventually` instead

PR #5415, opened 2020-04-27
Remove testkit generated files

EgoziE:
@argshook anymore files you think it's better to .ignore?


EgoziE:
🗒️ Note: `npm run generate` won't work well without running `npm run build` first
so maybe it's best not to ignore components.json

EgoziE:
CI fails - https://pullrequest-tc.dev.wixpress.com/viewLog.html?buildId=7587386&buildTypeId=FedPerformance_PerferAutomaticExecution_NoVcsRoot&tab=buildLog

So maybe i'll bring components.json back

argshook:
> CI fails - https://pullrequest-tc.dev.wixpress.com/viewLog.html?buildId=7587386&buildTypeId=FedPerformance_PerferAutomaticExecution_NoVcsRoot&tab=buildLog

hmm, that's because `components.json` is used in perfer check which requires `components.json` here https://github.com/wix/wix-style-react/blob/master/.perfer/yoshi.config.js#L6

so perhaps yeah, let's keep `components.json` for now, as it seems it's not as trivial to remove, it has become part of the build

other than that, ignoring testkit export files seems ok to me

Issue #4976, opened 2020-02-11
<Slider/> - Global styles are overridden


simast:
# 🐛 Bug Report

### 🏗 Relevant Components

`<Slider/>`

### 😯 Current Behavior

Slider does not display correctly in some projects because the global class names are overridden by external styles:

![slider-bug](https://user-images.githubusercontent.com/433527/74221006-bf474080-4cb9-11ea-8411-3b7d1c0194b7.png)

The screenshot above was provided by @drasius

### 🤔 Expected Behavior

Slider is displayed correctly. Ideally we should try not to use global class names or somehow isolate them to wix-style-react scope only.

### 👣 Steps to Reproduce

I am not able to reproduce this myself in production. It seems to happen for some users only (maybe it matters how you navigate to an app in business manager).  Currently it is not clear where the class name `.rc-slider-handle` came from (it is a default style for slider handle from `rc-slider` lib, but it is not clear who imported it since we do not use/import default `rc-slider` styles in wix-style-react ourselves).

We can emulate this in our Storybook by temporarily adding the following stylesheet to `.storybook/preview-head.html` file:

```html
<link
  rel="stylesheet"
  href="https://cdn.jsdelivr.net/npm/rc-slider@8.7.1/assets/index.css"
/>
```

  - **WSR version**: `7.42.0`
  - **Yoshi version**: `4.32.0`

### 👀 Severity

- Major

simast:
I did some experiments with `rc-slider` lib and it seems it supports an undocumented prop `prefixCls` that can be used to change the prefix for global styles. So this would be one way to fix this issue - to provide our own custom class name prefix so we do not conflict with other `rc-slider` users.

TalZa1810:
I've talked to Drąsius and it seems like the issue is not a new one. It happened also with the old `Slider` style for some users from time to time (randomly).

drasius:
@TalZa1810 @shlomitc @simast 
What is status for this one? Are you planning to fix?
This is what I see now. 
![image (10)](https://user-images.githubusercontent.com/46714829/76313116-033f5c80-62dd-11ea-9120-eda7d082ebfc.png)

shlomitc:
@drasius - can you please ask one of your developers to help us reproduce this in your project?
Otherwise, it is really hard to understand what should be the right fix for it

drasius:
[...]

PR #5418, opened 2020-04-28
`<Slider/>`-fixing global styles override

TalZa1810:
### 🔦 Summary
Resolves https://github.com/wix/wix-style-react/issues/4976

Resolving this by using `prefixCls` undocumented prop and setting the `rc-slider` version to `"9.2.4"` to prevent future issues regarding this change.

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [x] Visual test


src/Slider/Slider.js
```diff
@@ -110,6 +110,7 @@ export default class Slider extends Component {
     return Array.isArray(value) && value.length > 1 ? (
       <Slide.Range
         {...rest}
+        prefixCls="wsr"
```
simast, line 113: I would maybe use a bit longer prefix - `wsr-slider`? Because `.wsr` is quite a short class name and it could clash with some other global stuff in the future.

src/Slider/Slider.scss
```diff
@@ -9,15 +9,15 @@ $sliderMarkLine: 6px;
 $sliderMarkTextMarginTop: 12px;
 $sliderMarkTextSize: 18px;
 
-:global(.wix-style-react-slider) :global(.rc-slider) {
+:global(.wix-style-react-slider) :global(.wsr) {
```
simast, line 12: Is `.wix-style-react-slider` global class still required? I think it was maybe used to increase the selector complexity at some point and to solve a similar issue. But since we now have a totally custom prefix this should no longer be required. I would check git blame and try to ask why this class name was added.

TalZa1810: The PR is too big. I wasn't sure why we added this. I think the custom prefix resolves that issue as well.
https://github.com/wix/wix-style-react/pull/3062

The tests pass on my local host so I think it's fine.

mykas: Hey, @simast is correct. This increases the specificity. The problem occurs only in production so I suggest running production storybook and making sure everything works

PR #5431, opened 2020-05-03
<Carousel/> - Fix indicators color and margin-top

bamai:
### 🔦 Summary
Fix indicators color and margin-top: https://jira.wixpress.com/browse/DSM-132
...

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


src/Carousel/Carousel.js
```diff
@@ -146,11 +146,12 @@ class Carousel extends React.Component {
         aspectRatio={Proportion.PREDEFINED_RATIOS.landscape}
       >
         <div
+          style={{ height: '100%' }}
           data-hook={dataHooks.imagesContainer}
           data-is-loading={this._isLoading()}
         >
           <img
-            src={image.src}
+            {...image}
```
TalZa1810, line 154: Can you clarify what's the purpose of the change?
Do we test this change?

TalZa1810: @EgoziE do you think that we should add a test for this change? and if so- a visual test or unit test? WDYT?

EgoziE: Right now this image variable is just an object containing `src`.
It should be of type `ImgHTMLAttributes` and be spread over the `` element.

1. Need to change the d.ts accordingly (I'm not sure what is the right way to write it, please consult Moshe/Guilad/Google)
2. No need to test this specific change, it's just for adding native props like `alt` etc.

mosheKerbel: @maia-b 
```jsx
images?: CarouselImage[];
export type CarouselImage = Partial<HTMLImageElement> & { src: string };
```
What happens here?
You are telling TS that the item **must** contain an `src` prop, but it **can** also contain, part of the props from `HTMLImageElement`
`Partial<x>` makes all props from x to be optional.
Make sure to test it in the `carousel.tsx` types test.

`HTMLImageElement` exists out of the box, meaning you get it automatically from the typescript environment, you do not have to import it in your d.ts.

PR #5432, opened 2020-05-03
`<Table/>`- Add TableActionCell support for a table with horizontal scroll

TalZa1810:
### 🔦 Summary
Adding TableActionCell support for a table with a horizontal scroll

### ✅ Checklist

- 📚 Change is documented
  - [x] Story
  - [x] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


src/PopoverMenu/PopoverMenu.js
```diff
@@ -305,10 +309,12 @@ class PopoverMenu extends React.PureComponent {
         moveBy={moveBy}
         maxHeight={maxHeight}
         zIndex={zIndex}
+        onMouseLeave={closeOnMouseLeave ? this.close : null}
```
TalZa1810, line 312: I need to expose this prop in order to close the popoverMenu when hovering on a different table row.
exposing a prop named `closeOnMouseLeave` is a good approach in this use case? wdyt?
@EgoziE

TalZa1810: it seems to me like that it should have similar functionality as the `ClosablePopover` internal component.

PR #5438, opened 2020-05-06
<Skeleton/> - add uni.driver + migrate to stylable

EgoziE:
Adding uni.driver and migrating to stylable.

This PR was created to help estimating the effort of removing all scss files and creating uni.drivers for all components.

My estimation: small effort - big scope
If we split the work we can finish this in no time.


src/Skeleton/Skeleton.js
```diff
@@ -1,28 +1,33 @@
...
+      >
+        {content.map(({ type, size }, key) => (
           <div
             key={key}
-            data-hook="placeholder-line"
-            className={classnames(styles.placeholderLine, styles[spacing], {
-              [styles.middle]: alignment === 'middle',
-            })}
+            data-hook={dataHooks.line}
+            {...styles('line', { alignment, spacing }, this.props)}
```
TalZa1810, line 24: why not using `styles.line`? you can get the `alignment` & `spacing` props from the root.

src/Skeleton/Skeleton.js
```diff
@@ -1,28 +1,33 @@
...
           >
             <div
-              data-hook="placeholder-chunk"
-              className={classnames(styles.chunk, styles[item.size])}
+              data-hook={dataHooks.chunk}
+              data-type={type}
+              data-size={size}
+              {...styles('chunk', { size }, this.props)}
```
TalZa1810, line 30: same comment here. just pass the `size` prop to the root div

src/Skeleton/Skeleton.spec.js
```diff
@@ -11,32 +16,44 @@ const content = [
...
-      expect(driver.hasSpacing('medium')).toBe(true);
-    });
+describe('[async]', () => {
+  runTests(createRendererWithUniDriver(skeletonUniDriverFactory));
+});
 
-    it('should have lines with expected sizes', () => {
-      expect(driver.hasSizes(content.map(({ size }) => size))).toBe(true);
+function runTests(render) {
+  afterEach(() => cleanup());
```
TalZa1810, line 28:
```suggestion
  afterEach(cleanup);
```

src/Skeleton/Skeleton.js
EgoziE: Alignment and spacing is a property relevant to the line only, it has no effect it the root scope.

In the css itself
on root:
`.root:alignmet(sheker) .line { ... }`
on line:
`.line:alignmet(sheker){ ... }`

PR #5445, opened 2020-05-07
upgrade stylable to v3

argshook:
This is a PR tackling the stylable v3 migration

---

* [x] `test1:unit`
* [x] `test2:visual`
* [x] `test3:e2e`
* [x] `test4:e2e`
* [x] `test5:types`
* [x] `test6:sanity`
* [x] 0 stylable warnings
      for example
    ```
    WARNING in Stylable: src/Tooltip/Tooltip.st.css:162:83: unknown pseudo element "arrow"
    > 162 | .root::popover[data-placement*="bottom"]::arrow {
    ```

depends on:

* [X] https://github.com/wix/wix-ui/pull/1859
* [x] https://github.com/wix/wix-ui/pull/1864


mykas:
I wish you force of patience and greenish builds with no or little effort!

scripts/transpile/tasks/replace-stylable-imports.js
```diff
@@ -31,10 +31,11 @@ const WRONG_PATHS = [
     regexp: /wix-ui-core\/dist\/src\/hocs\/.*\/[A-Za-z]*\.st\.css/,
     correct: 'hocs.st.css',
   },
-  {
+  /** temp ignore */
+  /** {
```
argshook, line 35: note to self: this needs resolution, it's possible this fix is no longer required.

need to test project with ssr

src/Popover/Popover.st.css
```diff
@@ -51,7 +51,7 @@
 }
 
 /* Overriding arrow definitions */
-.root::popover::arrow {
+.root::popover .arrow {
```
shlomitc, line 54: why not just `.root::arrow`

argshook: most of selectors targeting `.arrow` in this file also use attribute selector, for example `.root::popover[data-placement*="right"].withArrow`

`.root::arrow` would work bet it might be confusing, why is this one `.root::arrow` but all the rest `.root::something .arrow`?

`.root::popover::arrow` is no longer a valid selector

src/Tooltip/index.d.ts
```diff
@@ -8,6 +8,7 @@ export type TooltipSize = 'small' | 'medium';
 
 export interface TooltipProps {
   dataHook?: string;
+  className?: string;
```
EgoziE, line 11: Add to tsx

src/DatePicker/DatePicker.js
```diff
@@ -215,7 +215,7 @@ export default class DatePicker extends React.PureComponent {
 
     return (
       <Popover
-        {...styles('root', {}, this.props)}
+        className={st(classes.root)}
```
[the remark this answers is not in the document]
argshook: correct, it is set on `Calendar.propTypes`, but the `className` should likely be set on `Calendar` and not root popover. wdyt?

src/InputWithOptions/InputWithOptions.js
```diff
@@ -225,7 +228,7 @@ class InputWithOptions extends Component {
     const body = popoverProps.appendTo === 'window';
     return !native ? (
       <Popover
-        {...styles('root', {}, this.props)}
+        className={st(classes.root)}
```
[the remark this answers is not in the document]
argshook: same as with `DatePicker` and `Calendar`. in here `className` should probably not be set on root popover but on `input`, which is already the case.

.wuf/generator/templates/$ComponentName/$ComponentName.js
```diff
@@ -25,18 +25,18 @@ class <%= ComponentName %> extends React.PureComponent {
 
     return (
       <div
-        {...styles('root', { even: isEven, odd: !isEven }, this.props)}
+        className={st(classes.root, { even: isEven, odd: !isEven })}
```
EgoziE, line 28: className is defined here too 🐰

src/ImageViewer/ImageViewer.st.css
```diff
@@ -9,10 +9,13 @@
...
 
 :import {
-  -st-from: "wix-ui-core/hocs.st.css";
-  -st-named: Focusable;
+  -st-from: "wix-ui-core/dist/src/hocs/Focusable/Focusable.st.css";
+  -st-default: Focusable;
 }
 
+:vars {
+  overlayColor: rgba(32, 69, 94, 0.8);
```
EgoziE, line 17: This was changed to be D10-66

src/Accordion/AccordionItem/AccordionItem.js
```diff
@@ -109,35 +110,45 @@ class AccordionItem extends React.PureComponent {
...
+        className={st(classes.root, {
+          disabled,
+          hover,
+          open,
+          skin,
+          hideShadow,
+        })}
+=======
+        className={st(classes.root, { disabled, hover, open, skin }, className)}
+>>>>>>> 7c3811c86... chore(*): accept className prop in components that extend Focusable
```
EgoziE, line 133: 🍦

src/Popover/Popover.st.css
shlomitc: So it looks to me like we didn't use stylable correctly for these customizations...

in `wix-ui-core` - we have this public API:
`root` - the root div
`popoverElement` - the wrapping div for the trigger element
`popover` - the wrapping div of the content and the arrow.
`arrow` - the arrow div

`withArrow`- an additional class, next to the wrapping `div` of the content - displayed only when arrow should be displayed. this could actually be a state.
`popoverContent` - a class that wraps the content div only if an arrow is not displayed **OR** added as an additional class next to `popover` if an arrow is displayed

and in addition, there is the `data-placement` which is something that is required by `react-popper` to indicate the direction of the popover.

So sounds to me like we could convert the `withArrow` in the `wix-ui-core` to state and this will make the code cleaner.

For now, let's make it work, but let's open a task to make it use best practices

src/DatePicker/DatePicker.js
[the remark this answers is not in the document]
EgoziE: I see your point, can be marked as resolved

src/ImageViewer/ImageViewer.st.css
argshook: ah, the joys of rebasing. thanks for checking!

src/Popover/Popover.st.css
argshook: moved to jira DSM-594

PR #5460, opened 2020-05-13
chore(package.json): pin tslib version to 1.11.2

argshook:
hopefully temporary fix to avoid errors like the following:

```
Cannot set property avatarDriverFactory of [object Object] which has only a getter
```

related https://github.com/microsoft/tslib/issues/103


argshook:
not enough of a fix, closing :(

PR #5473, opened 2020-05-17
docs: revise the Madefor migration guide

shlomitc:
### 🔦 Summary

This makes the migration guide a bit more self-explanatory.
@ayeletr - do you already know of common pitfalls in the migration to Madefor? If so, we could add these [here](https://wix-wix-style-react-pr-5473.surge.sh/?path=/story/introduction-getting-started--madefor-font).


EgoziE:
I know that a common pitfall might be a linebreak because overall the font is wider.
In this case the best solutions is to let UX change spacing or translations.

shlomitc:
@ayeletr - do you have an example like @EgoziE mentioned? we will add it to the doc

Issue #5474, opened 2020-05-18
`<SectionHelper />` - without title, close button position is wrong


frontned:
# 🐛 Bug Report

### 🏗 Relevant Components

`<SectionHelper />`

https://wix-style-react.now.sh/?activeTab=Playground&path=%2Fstory%2Fcomponents-api-components--sectionhelper

### 😯 Current Behavior

When the title is removed, the close button jumps to the center of the container.

### 🤔 Expected Behavior

Close button should stay at the top.

### 👣 Steps to Reproduce

https://wix-style-react.now.sh/?activeTab=Playground&path=%2Fstory%2Fcomponents-api-components--sectionhelper

Remove the title in the playground with close button toggled on.

PR #5475, opened 2020-05-18
#5474 fix close button position in SectionHelper without a title

frontned:
### 🔦 Summary

Fix close button position in SectionHelper without a title


EgoziE:
Thanks for the fix 😎

PR #5480, opened 2020-05-19
Improve Playground experience

EgoziE:
1. Hide unnecessary props like 'dataHook' and 'className'
2. Callback Props - make interactive
3. Added - commonTooltipPropsExample & commonPopoverPropsExample


.wuf/generator/templates/$ComponentName/docs/index.story.js
```diff
@@ -38,6 +38,8 @@ export default {
     // https://github.com/wix/wix-ui/blob/master/packages/wix-storybook-utils/docs/usage.md#using-list
   },
 
+  hiddenProps: ['dataHook', 'className'],
```
shlomitc, line 41: Maybe this should be done in the Wix-storybook-Utils, otherwise it will be duplicated all over

EgoziE: If we do it in Wix-storybook-Utils, Won't it look like a bug if someone ever tries to use these prop for some reason?

EgoziE: I think it's better to duplicate this line and encourage us/consumers to use `hiddenProps` more.
Luckily it doesn't affect the bundle size :)

shlomitc: Then I think this should be some global configuration in the `.storybook` that is shared between all stories

argshook: `wix-storybook-utils` can be updated to read "global" `hiddenProps` from `.storybook/webpack.config.js`

not really sure we need to, but if we decide to have this setting by default, then imo it's better to do it in wsr only and not force on all other projects (like, maybe wix-ui-tpa wants to support `className`)

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5480/

PR #5485, opened 2020-05-20
Madefor - added two Q&A

EgoziE:


EgoziE:
image

EgoziE:
image

shlomitc:
What is Madefor regular? weight="normal"?
We shouldn't really care if it's Helvetica 55, we care for the appropriate prop name and value
I think we should also map it to the correct Text/Heading prop

ayeletr:
Yes, ok we can use the SB text name/prop.  
I created these examples for UX issues, that can happen after the change, following your request to have "common UX pitfalls". 
I suggest we'll have another section (title) under Q&A, named "Common UX Pitfalls",
And put the UX example there. 
What do you think? Tell me if you have other idea

shlomitc:
@ayeletr - sounds good, let's add a subsection for UX related issues

PR #5487, opened 2020-05-21
infra: fix broken puppeteer testkit that can't be required in node

shlomitc:
### 🔦 Summary

Puppeteer testkit should be required in node env, meaning it shouldn't import `scss` files or components that import a scss file.
The fix is simple, just remove it from the driver :) The hard part is to find it because logs are not accurate.
The bug is here: https://github.com/wix/wix-style-react/blob/e56e5e895101cabb7c7a0aa814fd9b7a95552930/src/AutoCompleteWithLabel/AutoCompleteWithLabel.uni.driver.js#L6
by importing `Input`, you also import `scss` files.

Anyway, I added a runtime-e2e test to prevent this from happening


test-runtime/imports/test/e2e/puppeteer-testkit.e2e.js
```diff
@@ -27,4 +28,15 @@ describe('React application to be interacted with puppeteer testkit', () => {
 
     expect(await approvalTextDriver.getValue()).toContain('It was clicked!');
   });
+
+  // This test makes sure that no css files or similar are imported by the drivers. kind of an SSR test
+  it('should be imported in node-env', async () => {
+    let error;
+    try {
+      await axios.get('http://localhost:3105/puppeteer-testkit-require');
```
EgoziE, line 36: Can this be replaced by fetch API?
(instead of having another dependency)

shlomitc: it's running in node, so you will need either a low-level HTTP request or some other dependency like `node-fetch`.
`axios` is really small and very common, so I think that for testing purposes it shouldn't be an issue

PR #5497, opened 2020-05-26
<Page/> - allowing content to grow horizontally with scroll

Jefoss:
### 🔦 Summary
Allowing the page to have content that doesn't have a fixed width or limit, with a horizontal scrollbar.

image


src/Page/Page.js
```diff
@@ -512,6 +516,8 @@ Page.propTypes = {
   maxWidth: PropTypes.number,
   /** Sets the min width of the content (Both in header and body) NOT including the page padding */
   minWidth: PropTypes.number,
+  /** Expands the content container to the content's width */
+  horizontalScroll: PropTypes.bool,
```
shlomitc, line 520: You should also add this prop to the typescript types:
1. `index.d.ts`
2. `test/Page.tsx`

src/Page/Page.js
```diff
@@ -267,7 +268,10 @@ class Page extends WixComponent {
     });
 
     const pageDimensionsStyle = this._getPageDimensionsStyle();
-    const style = contentFullScreen ? null : pageDimensionsStyle;
+    let style = contentFullScreen ? null : pageDimensionsStyle;
+    if (horizontalScroll) {
+      style = { ...style, display: 'flex', flexDirection: 'column' };
```
shlomitc, line 273: it's static so it should probably be in the scss file and used as a class name in `mergedClassNames`

src/Page/Page.js
```diff
@@ -512,6 +516,8 @@ Page.propTypes = {
   maxWidth: PropTypes.number,
   /** Sets the min width of the content (Both in header and body) NOT including the page padding */
   minWidth: PropTypes.number,
+  /** Expands the content container to the content's width */
```
shlomitc, line 519: allow the page to scroll horizontally for large width content

src/Page/docs/index.story.js
```diff
@@ -214,6 +214,40 @@ export default {
...
               compact: true,
             },
+            {
+              title: 'Horizontal Scroll',
+              description: 'Allows for horizontal scrolling without set width',
+              source: `
+                <Page height="40vh" horizontalScroll>
+                  <Page.Header title="Page Header" actionsBar={<Button>Click me</Button>}/>
+                  <Page.Content>
+                    <div style={{width: '3000px', height: '100px', backgroundColor: 'white', padding: '20px'}}>
```
shlomitc, line 224: How does it actually work currently in your local code?
Do you set a specific size or a max-width?

Because in this case, it will always cause an overflow and a showing the scrollbar, even if the content doesn't need to exceed the limits.

shlomitc:
@Jefoss - feature-wise, this is not working as expected. when you scroll it, I'd expect only the content to scroll but now it also scrolls the header:
image

ayeletr:
Hi, Iooked at the example in Page component,
As Shlomi mentioned:
1. The header should be sticky and not scroll
2. The scrolling and the scroll bar should appear only if the content gets cut.

src/Page/docs/index.story.js
Jefoss: we don't have any specific sizes in that area, the content just grows horizontally

shlomitc: so why do you need adding `width:3000px` here?

shlomitc: anyway, API-wise, I think that in order to create a simple API, we need to allow setting the content width as a prop (if it isn't available already). So let's change it to this:
`<Page.Content horizontalScroll width="3000px"/>`

let me know if there are any issues

milkyfruit:
@Jefoss @shlomitc could you specify which product needs this feature?

shlomitc:
> @Jefoss @shlomitc could you specify which product needs this feature?

Workflows. It’s already implemented there locally and finally is contributed to the library

src/Page/docs/index.story.js
Jefoss: i just wanted to simulate content wider than your maximum width to see that the content scrolls horizontally, the current api has max/minWidth props, but anyway i dont know my content's width, it could be 500px and could be 1500px depending on the specific site.
i dont think introducing the width prop helps with our problem

Jefoss:
> It might be browser's bug, but if content overflows both vertically and horizontally, on Chrome scrollbar hides underneath the content:
> image
> 
> on Safari it works fine.

@shlomitc not sure how we want to address this, is it a blocker? we can add right margin on the container so the vertical scrollbar will show, wdyt?

src/Page/Page.scss
```diff
@@ -165,6 +165,11 @@ $minimizedHeaderContainerZIndex: $stickyZIndex + $zIndexLayerStep;
   padding-right: 0;
 }
 
+.horizontalScroll {
+  overflow-x: auto;
+  margin-right: 15px;
```
shlomitc, line 170: why is the magic number 15 works?
In any way, let's make it a variable and add a comment

src/Page/docs/index.story.js
```diff
@@ -214,6 +214,41 @@ export default {
...
             },
+            {
+              title: 'Horizontal Scroll',
+              description: 'Allows for horizontal scrolling without set width',
+              source: `
+                <Page height="40vh" horizontalScroll>
+                  <Page.Header title="Page Header" actionsBar={<Button>Click me</Button>}/>
+                  <Page.Content>
+                    <Box directon="horizontal">
+                      <Box backgroundColor="lightyellow" marginRight="12px" padding="12px" minWidth="200px" height="1000px">
```
shlomitc, line 225: just do a simple loop, just to make the code simper

src/Page/test/examples/SomeContentComponent.js
```diff
@@ -82,9 +83,13 @@ export default class SomeContentComponent extends React.Component {
         style={{
           backgroundColor: 'white',
           minHeight: this.props.stretchVertically ? 'inherit' : undefined,
+          width: this.props.wideContent ? 'max-content' : undefined,
```
shlomitc, line 86: I don't understand why it's needed. You managed to do the same without it in the storybook example, why does this extra `max-content` needed

shlomitc:
image

I colored the playground background to show some issues
1. by hacking the scrollbar 15px to the right, it moved also the button, so If this is the way to hack it, we should reduce some padding from the header.
2. another thing to notice is that when you scroll all the way to the right, the right padding doesn't appear... I think this one should be fixed as well

src/Page/Page.scss
Jefoss: we can calculate the exact width of the scrollbar but this is something that works in majority of cases, ill add a comment

src/Page/test/examples/SomeContentComponent.js
Jefoss: in the story book i had a bunch squares with specific width next to each other, in this example just a single div with lorem ipsum, and we dont want to content to break to fit the screen, we want to simulate some content wider than the screen, and in this example `max-content` allows this containing div to expand to its content's width. (i could do a similar thing like in the story book if you want, i just enhanced the existing code)

shlomitc:
@milkyfruit - what do you think about the side padding?

milkyfruit:
@Jefoss @shlomitc I think that pushed CTA button to the left is not critical, though it would be nice if it could stay in the same place.

Regarding 15px, I think it should be changed to 18px, because we use 6px grid and other OS might have larger scrollbars: https://codepen.io/sambible/post/browser-scrollbar-widths

PR #5498, opened 2020-05-26
<FontUpgrade/> - Fix Modal, Popover, Tooltip content

EgoziE:


shlomitc:
How does it fix it?

EgoziE:
> How does it fix it?

when it's undefined the default is true

EgoziE:
I checked it on a clean yoshi project.
Managed to reproduce and test the fix.

shlomitc:
So maybe the default should be false?

EgoziE:
Yeah it should have, when I added this prop it was only was only for the Storybook add-on..
Changing it now means it will be breaking.

EgoziE:
I can delete this prop and replace it with a disabled instead

PR #5499, opened 2020-05-27
<InputWithOptions/> - fix use of excludeClass prop

EgoziE:
**THIS IS A TEMPORARY FIX UNTIL WE UPGRADE wix-ui-core**

Opening a popover and clicking outside should close it.
Currently if the outside click target is another popover, the first one doesn't close.

Which means popover's prop excludeClass value should be **unique for each instance**.
A real fix should be developed for the core popover in wix-ui-core, but for now until we can upgrade to it, this is a workaround.

A proper fix should be one of:
1. Adding a unique class generator to core popover
2. Changing ClickOutside to work with context (need to investigate this option)

A follow up Jira ticket - https://jira.wixpress.com/browse/DSM-583


EgoziE:
To reproduce:
https://wix-style-react.now.sh/?path=/story/introduction-playground--playground
1. open one dropdown
2. open the second dropdown

expected:
the first one closes when the second opens

actual:
the first one remains open

PR #5506, opened 2020-05-31
Fixed react-modal to 3.10.1 to enable tabs in modals

marik22312:
### 🔦 Summary

when using inputs inside models, the tab key doesn't seem to work at all,
after some investigation, @shlomitc were able to confirm that this issue only reproduces in the modal.
we were able to trace back the issue to `react-modal` starting from version 3.11.1.

We have tested this locally and seems to work perfectly with `react-modal@3.10.1`

can be reproduced with the following code:

```javascript
class SimpleExample extends React.Component {
  state = {
    isModalOpened: false,
  };

  openModal = () => this.setState({ isModalOpened: true });

  closeModal = () => this.setState({ isModalOpened: false });

  render() {
    const { isModalOpened } = this.state;
    return (
      <Box>
        <Button onClick={this.openModal}>Open Modal</Button>
        <Modal isOpen={isModalOpened} onRequestClose={this.closeModal}>
          <Card>
            <Card.Content>
              <Input />
              <Input />
            </Card.Content>
          </Card>
        </Modal>
      </Box>
    );
  }
}
```
https://wix-style-react.now.sh/?path=/story/introduction-playground--playground&snippet=3bce4c3b-865e-48f7-b620-0e3019a2f944

relevant issue in `react-modal`
https://github.com/reactjs/react-modal/issues/782

### ✅ Checklist

- 🔬 Change is tested
  - [X] Component
  - [X] Visual test


EgoziE:
Thanks for this great PR and investigation
Unit tests fail in CI (and locally)
```
TypeError: firstInput.pressKey is not a function
```

marik22312:
Talked with @EgoziE privately, basically, we had issues with loading the driver properly...
i left a skipped unit test as a reference, this test should be removed before merging

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5506/

PR #5508, opened 2020-05-31
`<Notification/>` - exposing ellipsis prop

TalZa1810:
### 🔦 Summary
Exposing ellipsis prop for notification label

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [x] Visual test


TalZa1810:
Hi @milkyfruit , 
Do we want to add the margin to the `Label` or the `Button`? Currrently, this is how it looks on a narrow viewport:
narrow
If we add it to the label, it will be added also when there's no button. However, if we add it to the button ( I mean margin-top for the button) it will be added also when viewing the notification on a wide viewport.
WDYT?

src/Notification/TextLabel/index.js
```diff
@@ -2,17 +2,28 @@ import React from 'react';
...
+    ellipsis={ellipsis}
+    light
+    dataHook={dataHooks.notificationLabel}
+  >
     {children}
   </Text>
 );
 
 TextLabel.propTypes = {
+  ellipsis: PropTypes.bool,
```
TalZa1810, line 19: set to true in order not to break existing users.

src/Notification/TextLabel/index.js
```diff
@@ -2,17 +2,28 @@ import React from 'react';
...
   children: PropTypes.node,
 };
 
+TextLabel.defaultProps = {
+  ellipsis: true,
```
TalZa1810, line 24: set to true in order not to break existing users.

src/Notification/Notification.js
```diff
@@ -147,19 +146,13 @@ class Notification extends React.PureComponent {
           aria-labelledby="notification-label"
           aria-live="polite"
         >
-          {themeIcon[theme]}
-          <div
-            id="notification-label"
```
TalZa1810, line 152: Was not used in our drivers. Therefore, was removed.

TalZa1810:
Fails on visuals which are not related to the change. 
As for types- the types were updated but the compound components do not have on their own `d.ts` files, so it seems fine.

PR #5519, opened 2020-06-04
DSM-538/Using scrollable container in modal layouts

lironch:
### 🔦 Summary
Using ScrollableContainer in the Content LayoutBlock to provide the dynamic appearance of top/bottom dividers according to scroll position
...
### ✅ Checklist

- [V] 👨‍🎨 UX change is approved by the Design System UX @ayeletr 
- 📚 Change is documented
  - [V] API description
- 🔬 Change is tested
  - [-] Visual test - **Currently missing, will be added later**


src/common/ScrollableContainer/test/ScrollableContainer.visual.js
```diff
@@ -0,0 +1,139 @@
...
+    ).add(
+      it,
+      () => {
+        useEffect(() => {
+          componentDidMount && componentDidMount();
+        }, []);
+        return (
+          <ScrollableContainerWrapper
+            componentDidMount={componentDidMount}
+            expectedDataHook={expectedDataHook}
```
jonathanadler, line 129: I think you're not using the `expectedDataHook` so I guess it's redundant

lironch: Yes, that's correct.
I'll modify the test to use it as an `expected` value with a conditional `data-test-read="true"`, it makes the test easier to understand as well.
Thanks for the feedback! 💪

PR #5548, opened 2020-06-14
Remove WixComponent - Part 2

EgoziE:
Part 2


src/ColorPicker/ColorPickerConverterHex.js
```diff
@@ -1,32 +1,26 @@
...
-  constructor(props) {
-    super(props);
-    this.change = this.change.bind(this);
-    this.state = {
-      hex: getHexOrEmpty(props.current),
-      inFocus: false,
-    };
-  }
+  state = {
+    hex: getHexOrEmpty(this.props.current),
```
argshook, line 16: @EgoziE does this really work? isn't `this.props.current` undefined at the time `getHexOrEmpty` is called?

argshook: ok, it works, react initializes `props` and fills in `current`, urgent to bcc :)

PR #5562, opened 2020-06-15
DSM-538/AnnouncementModalLayout without borders 

lironch:
### 🔦 Summary
Adding the option to render the AnnouncementModalLayout without borders and background-color.
This is done so it will be able to use it as a replacement for the deprecated EndorseContentLayout.
### ✅ Checklist

- 📚 Change is documented
  - [V] API description
- 🔬 Change is tested
  - [V] Visual test


lironch:
After discussing with @ayeletr we decided to not modify AnnouncementModalLayout to replace EndoresedModalLayout.
This means that for now we don't have a solution for migrating users and the only option they have is to use the library atoms to implement this simple layout in their products.
FYI @shlomitc

PR #5566, opened 2020-06-16
<InputArea/> counter not updated when value prop is changed

Roeia1:
### 🔦 Summary

<InputArea/> counter not updated when value prop is changed
currently, it's only being updated by interactions with the input area

### ✅ Checklist

- 🔬 Change is tested
  - [ ] Component


EgoziE:
Looks legit, would you please add a simple test for this so it won't break in the future?

Roeia1:
@EgoziE in order to test this, I need to add to both drivers the ability to change the value prop
(and not by using the enter text methods) meaning exploring the `ReactBase` util for your uni driver etc'...
I don't have the time to invest in this one, unfortunately.

EgoziE:
I see, well I meant something really really basic like this:
```
      it('should change counter when value change', async () => {
        const { driver, rerender } = render(
          <InputAreaForTesting hasCounter maxLength={30} value={'abc'} />,
        );
        expect(await driver.getCounterValue()).toEqual('3/30');
        rerender(<InputAreaForTesting hasCounter maxLength={30} value={'abcd'} />);
        expect(await driver.getCounterValue()).toEqual('4/30');
      });
```

Don't worry, I'll add it.
Thanks for this PR :)

Roeia1:
@EgoziE well didn't notice you have a rerender function 🤦 😆 
thanks!

PR #5581, opened 2020-06-17
Fixing the badge component height when it has `ellipsis` enabled.

lironch:
### 🔦 Summary

Fixing the badge component height when it has `ellipsis` enabled.
To do this i had to modify the `ellipsedTooltipHOC` to accept a className and position it on the internal `EllipsedTooltip` element, this way the badge component is able to extend this element and set it's line-height back to the right value.


shlomitc:
1. can you share the "before and after" screenshots to understand what was broken
2. add a test that expects this class name to be provided

lironch:
@shlomitc 
You can see the problem here on the bottom `Ellipsis` example
Screen Shot 2020-06-18 at 10 48 06

lironch:
@shlomitc regardless of the above, I talked to Egozi and he suggested a much more elegant solution that will make this one redundant. i'll try it and update later.

lironch:
Closing this PR since we decided to work on a new EllipsisHOC that will take issues like that into account

PR #5582, opened 2020-06-18
<Accordion/> - Adding support for `initiallyOpen` prop for <AccordionItem/>

lironch:
### 🔦 Summary
<Accordion/> - Adding support for `initiallyOpen` prop in <AccordionItem/>, this way the `open` prop is for controlled items and `initiallyOpen` for uncontrolled ones.

### ✅ Checklist
- 📚 Change is documented
  - [V] Story
  - [V] API description
- 🔬 Change is tested
  - [V] Component
  - [V] Visual test


src/Accordion/Accordion.js
```diff
@@ -43,16 +44,18 @@ class Accordion extends React.Component {
     hideShadow: false,
   };
 
-  _findOpenIndexes = items =>
+  _findOpenIndexes = (items, initial) =>
     items
-      .map((item, index) => (item.open ? index : null))
+      .map((item, index) =>
+        (initial && item.initiallyOpen) || item.open ? index : null,
```
EgoziE, line 50: Why not:
```suggestion
        (item.initiallyOpen || item.open) ? index : null,
```

lironch:
@EgoziE 
Since this method is invoked both on the component `constructor` and `componentDidUpdate` your suggested change will make `initiallyOpen` a controlled prop as well.
I want it to be taken into account only one time when constructor calls it.

PR #5591, opened 2020-06-21
Testing docs

EgoziE:
New testing docs


EgoziE:
Thanks for your comment, let me explain better what went though my mind.

First of all, the obvious change, how to import and use testkits.
Second, I was aiming to reduce the required knowledge in testing of the reader.
Therefore there's an abstract overview which is enough for most readers (with previous knowledge in testing) and a **full** advanced example for each platform with all the small differenced between them all.

BTW the example is a bit different between (React testing library & Enzyme) and (Puppeteer & Protractor).

If you found any mistake I'll fix it, I checked all tests myself, they run and pass.
I'll rename testing_overview.md

argshook:
> If you found any mistake I'll fix it, I checked all tests myself, they run and pass.

pardon, the mistake i thought is not a mistake. In some examples you display app code and test code as if it were in one file, at first i thought there were copy-paste errors

docs/usage/testing_overview.md
```diff
@@ -0,0 +1,120 @@
...
+      </div>
+```
+
+#### Importing a testkit
+
+Wix-style-react currently supports 4 testing platforms.
+Choose one and import the relevant component from it.
+
+```js
+// ReactTestUtils
```
shlomitc, line 70: Not sure it's actually for ReactTestUtils, it's simply dom if I remember correctly

docs/usage/testing_overview.md
```diff
@@ -0,0 +1,120 @@
...
+// Puppeteer
+import { InputTestkit } from 'wix-style-react/dist/testkit/puppeteer';
+```
+
+_Note: A testkit's name is always the name of the component following the word Testkit._
+
+#### Initializing a testkit
+
+Once we imported the relevant testkit from the desired testing platform, two parameters are needed in order to initialize it:
+1. `wrapper` - A platform specific React node containing the component we want to test.
```
shlomitc, line 85: that's not necessarily a react node. in puppeteer, for example, it should be an element

docs/usage/testing_overview.md
```diff
@@ -0,0 +1,120 @@
...
+2. `dataHook` - A string that matches the same `dataHook` of component we want to test.
+
+In this example we use `InputTestkit` from `ReactTestUtils`:
```
shlomitc, line 88: testing-library

docs/usage/testing_overview.md
```diff
@@ -0,0 +1,120 @@
...
+
+```jsx
+import { InputTestkit } from 'wix-style-react/dist/testkit';
+import { render } from '@testing-library/react';
+
+import App from './App';
+
+const inputDriver = InputTestkit({
```
shlomitc, line 96: in the puppeteer example, it should be `await InputTestkit({...})`.
I would actually be happy if we could make this the same in all platforms

docs/usage/testing_overview.md
```diff
@@ -0,0 +1,120 @@
...
+```jsx
+describe('My App', () => {
+  it('should get the correct value', async () => {
+    expect(await inputDriver.getValue()).toEqual('hello world');
+  });
+});
+```
+
+### Next steps
+For further reading visit our best practices page and advanced examples.
```
shlomitc, line 120: any relevant link here?

stories/Introduction/index.js
```diff
@@ -36,9 +41,39 @@ storiesOf(Category.CHEATSHEET, module).add('Components Cheatsheet', () => (
...
 
-storiesOf(Category.TESTING, module).add('Testing', () => (
+storiesOf(Category.TESTING, module).add('Overview', () => (
   <div style={{ margin: '0 48px', width: '100%', maxWidth: 1161 }}>
-    <Markdown source={TestingReadme} />
+    <Markdown source={TestingOverview} />
+  </div>
+));
+
+storiesOf(Category.TESTING, module).add('React tesing library', () => (
```
shlomitc, line 50: maintaining (and showing in our docs) so many examples seem to be an overkill and harder maintenance.
Also, it won't compile, meaning we would not be able to validate it actually works if someone complains.

In popular libraries, it's common to create an example folder which you can simply do `npm install` and `npm test` and it will actually simulate the cases as real projects. For instance, see [storybook's examples](https://github.com/storybookjs/storybook/tree/next/examples).

Once stylable@3 is merged and we have clear examples of how to use our project, we can simply use this configuration to create these test projects.

PR #5595, opened 2020-06-22
`<ModalPreviewLayout/>`- exposing  nextButtonProps & prevButtonProps props

TalZa1810:
### 🔦 Summary

1. deprecating `nextButtonTooltipText` and `prevButtonTooltipText` (adding deprecation log and deprecation notation in the documentation.
2. Add new props instead of the previous ones:
```
nextButtonProps: {
      onClick: PropTypes.func,
      tooltipText: PropTypes.node,
}
```

```
prevButtonProps: {
      onClick: PropTypes.func,
      tooltipText: PropTypes.node,
}
```

3. Have onClick for each navigation button separately. 
4. Remove conditionally tooltip rendering (using the disable prop for that purpose).
5. Update example with the new props.
6. Update **private** driver to return the tooltip driver for both navigation buttons and close button.
7. Update tests with the new API change. 

This change still supports the previous API for backwards competability. 

I've decided to leave `closeButtonTooltipText` as is since we have also `onClose` prop which is required. Maybe in a different PR should be changed. I didn't want to mix these. 

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [x] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [x] Visual test


src/ModalPreviewLayout/ModalPreviewLayout.js
```diff
@@ -31,31 +33,48 @@ class ModalPreviewLayout extends React.PureComponent {
     prevButtonTooltipText: PropTypes.string,
     /** Tooltip next button text */
     nextButtonTooltipText: PropTypes.string,
+    /** A NavigationButton onClick callback
+     * ##### Signature:
+     * function( direction: ARROWS_DIRECTIONS, childIndexDisplayed: number) => void
+     * * `direction`: the direction of the navigation button. Can be either `right-arrow` or `left-arrow`
+     * * `childIndexDisplayed`: represent the index of the child node currently displayed
+     */
+    onNavigationButtonClick: PropTypes.func,
```
shlomitc, line 42: don't you think it would be easier just to expose a prop per-button?
you anyway pass ones for the tooltip text, so this will only make it less consistent.
this means that every consumer will need to do an `if` statement, instead of simply connecting the right callback function

EgoziE: Honestly, I don't like this components per-button API.
I think 2 more props will be a mess...

I suggest to change it to:
```
nextButtonProps: {
    tooltipText: PropTypes.string,
    onClick: PropTypes.func,
}
```

```
prevButtonProps: {
    tooltipText: PropTypes.string,
    onClick: PropTypes.func,
}
```

and deprecate `nextButtonTooltipText`, `prevButtonTooltipText`, and perhaps `closeButtonTooltipText` too.

TalZa1810: I agree. I've checked `wix-private` and saw that it is used only once. I'll contact the developer to change it to a new API with the next version?  WDYT? 
@EgoziE @shlomitc

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5595/

src/ModalPreviewLayout/ModalPreviewLayout.js
```diff
@@ -26,60 +27,105 @@ class ModalPreviewLayout extends React.PureComponent {
...
   };
 
-  state = { childIndexDisplayed: 0 };
+  constructor(props) {
+    super(props);
+    this.state = { childIndexDisplayed: 0 };
 
-  shouldClose(id) {
+    deprecationLog(
+      'nextButtonTooltipText and prevButtonTooltipText props are deprecated and will be removed as part of the next major version.',
```
EgoziE, line 70:
```suggestion
      'nextButtonTooltipText and prevButtonTooltipText props are deprecated and will be removed as part of the next major version, please use nextButtonProps and prevButtonProps.',
```

src/ModalPreviewLayout/ModalPreviewLayout.js
```diff
@@ -26,60 +27,105 @@ class ModalPreviewLayout extends React.PureComponent {
...
 
-  _onArrowClick(direction) {
+  _onRightNavigationClick() {
+    const { childIndexDisplayed } = this.state;
+    const { onClick } = this.props.nextButtonProps;
+    const newChildIndexDisplayed = childIndexDisplayed + 1;
+
+    onClick && onClick(newChildIndexDisplayed);
+
+    this.setState({ childIndexDisplayed: newChildIndexDisplayed });
```
EgoziE, line 88:
```suggestion
    this.setState({ childIndexDisplayed: newChildIndexDisplayed }, onClick && () => onClick(newChildIndexDisplayed));
```

EgoziE: I think it's better to call the callback after the action has finished.

src/ModalPreviewLayout/ModalPreviewLayout.js
```diff
@@ -26,60 +27,105 @@ class ModalPreviewLayout extends React.PureComponent {
...
     const { childIndexDisplayed } = this.state;
-    direction === arrowsDirection.rightArrow
-      ? this.setState({ childIndexDisplayed: childIndexDisplayed + 1 })
-      : this.setState({ childIndexDisplayed: childIndexDisplayed - 1 });
+    const { onClick } = this.props.prevButtonProps;
+    const newChildIndexDisplayed = childIndexDisplayed - 1;
+
+    onClick && onClick(newChildIndexDisplayed);
+
+    this.setState({ childIndexDisplayed: newChildIndexDisplayed });
```
EgoziE, line 98:
```suggestion
    this.setState({ childIndexDisplayed: newChildIndexDisplayed }, onClick && () => onClick(newChildIndexDisplayed));
```

EgoziE: I think it's better to call the callback after the action has finished.

src/ModalPreviewLayout/ModalPreviewLayout.js
```diff
@@ -26,60 +27,105 @@ class ModalPreviewLayout extends React.PureComponent {
...
           />
         )}
         {hasRight && (
           <NavigationButton
-            tooltipText={nextButtonTooltipText}
-            dataHook={dataHooks.modalPreviewRightArrow}
+            tooltipText={nextButtonProps.tooltipText || nextButtonTooltipText}
             direction={arrowsDirection.rightArrow}
-            onClick={() => this._onArrowClick(arrowsDirection.rightArrow)}
+            onClick={() => this._onRightNavigationClick()}
```
EgoziE, line 128:
```suggestion
            onClick={this._onRightNavigationClick}
```

src/ModalPreviewLayout/ModalPreviewLayout.js
```diff
@@ -26,60 +27,105 @@ class ModalPreviewLayout extends React.PureComponent {
...
     return (
       <React.Fragment>
         {hasLeft && (
           <NavigationButton
-            tooltipText={prevButtonTooltipText}
-            dataHook={dataHooks.modalPreviewLeftArrow}
+            tooltipText={prevButtonProps.tooltipText || prevButtonTooltipText}
             direction={arrowsDirection.leftArrow}
-            onClick={() => this._onArrowClick(arrowsDirection.leftArrow)}
+            onClick={() => this._onLeftNavigationClick()}
```
EgoziE, line 121:
```suggestion
            onClick={this._onLeftNavigationClick}
```

src/ModalPreviewLayout/NavigationButton/NavigationButton.js
```diff
@@ -28,32 +23,30 @@ const tooltipProps = {
...
+const iconButtonProps = {
+  [arrowsDirection.leftArrow]: {
+    as: 'button',
+    skin: 'transparent',
+    dataHook: dataHooks.modalPreviewLeftArrow,
+    children: <ChevronLeft />,
+  },
+  [arrowsDirection.rightArrow]: {
+    as: 'button',
+    skin: 'transparent',
```
EgoziE, line 35: Shared code, can be on IconButton directly

src/ModalPreviewLayout/index.d.ts
```diff
@@ -10,8 +10,20 @@ export interface ModalPreviewLayoutProps {
   closeButtonTooltipText?: string;
   prevButtonTooltipText?: string;
   nextButtonTooltipText?: string;
+  prevButtonProps?: PrevButtonProps;
+  nextButtonProps?: NextButtonProps;
 }
 
+export type PrevButtonProps = {
+  onClick?: (childIndexDisplayed: number) => {};
+  tooltipText:React.ReactNode;
```
EgoziE, line 19:
```suggestion
  tooltipText?: React.ReactNode;
```

src/ModalPreviewLayout/index.d.ts
```diff
@@ -10,8 +10,20 @@ export interface ModalPreviewLayoutProps {
...
+};
+
+export type NextButtonProps = {
+  onClick?: (childIndexDisplayed: number) => {};
+  tooltipText:React.ReactNode;
```
EgoziE, line 24:
```suggestion
  tooltipText?: React.ReactNode;
```

PR #5610, opened 2020-06-25
<InputArea/> - Tests noise fixes

GuyPerkal:
### 🔦 Summary

Fixes 2 separate bugs, both occur when testing `InputArea` with the.`autoGrow` property and both cause a lot of noise.

The first is this error - `Error: Uncaught [TypeError: Cannot read property 'style' of null]`
* It is caused by a debounced function getting called after the `InputArea` has been unmounted, and is fixed by cancelling the debounced function on unmount.

The second is this warning - `Warning: Received NaN for the 'rows' attribute. If this is expected, cast the value to a string.`
* It is caused by the `InputArea` not having a font-size in its computed style when running tests, and is fixed by returning `MIN_ROWS` as its row count in that case.

### ✅ Checklist

- 🔬 Change is tested
  - [x] Component

### DIFF
--- a/src/InputArea/InputArea.js
+++ b/src/InputArea/InputArea.js
@@ -40,7 +40,9 @@ class InputArea extends React.PureComponent {
       this._calculateComputedRows();
     }
     if (this.props.hasCounter && prevProps.value !== this.props.value) {
-      this.setState({ counter: (this.props.value || this.props.defaultValue || '').length })
+      this.setState({
+        counter: (this.props.value || this.props.defaultValue || '').length,
+      });
     }
   }
 
--- a/src/InputArea/InputArea.js
+++ b/src/InputArea/InputArea.js
@@ -243,9 +243,18 @@ class InputArea extends React.PureComponent {
       computedStyle.getPropertyValue('line-height'),
       10,
     );
-    const lineHeightValue = isNaN(lineHeight)
-      ? this._getDefaultLineHeight() * fontSize
-      : lineHeight;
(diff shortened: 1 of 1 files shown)


GuyPerkal:
There are not tests for the autoGrow property, BTW.

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5610/

src/InputArea/InputArea.js
```diff
@@ -237,6 +243,11 @@ class InputArea extends React.PureComponent {
...
       10,
     );
+
+    if (isNaN(fontSize)) {
+      return InputArea.MIN_ROWS;
+    }
+
     const lineHeightValue = isNaN(lineHeight)
       ? this._getDefaultLineHeight() * fontSize
       : lineHeight;
```
lironch, line 253: From what i see here `fontSize` will be used only when the `isNaN(lineHeight)` is true, hence we shouldn't return  the `MIN_ROWS` value so early here.
I would go with a more defensive approach:
```suggestion
    let lineHeightValue;
    if (isNaN(lineHeight)) {
      if (isNaN(fontSize)) {
        return InputArea.MIN_ROWS;
      } else {
        lineHeightValue = this._getDefaultLineHeight() * fontSize;
      }
    } else {
      lineHeightValue = lineHeight;
    }
```

GuyPerkal: Done 👍

PR #5626, opened 2020-06-30
Address input poc

lironch:
AddressInput POC
Defining a suggested api for the future component and ways we can compose with it new products.
The POC is written like a blog post so you can read the comments in between to understand the thinking process led to writing each one.

### DIFF
--- a/src/AddressInput/AddressInput.part1.poc.tsx
+++ b/src/AddressInput/AddressInput.part1.poc.tsx
@@ -117,7 +117,7 @@ interface Suggestion {} // This is the raw "suggestion" object returned from the
  *
  * Therefore, we will need an adapter to make the two talk to each other.
  * Clarification, when speaking about adapters i will refer to function components that receive one type of api signature
- * and return an different type of api-signature by applying transformations to the received props.
+ * and return a different type of api-signature by applying transformations to the received props.
  * These adapters should be pure, hence, shouldn't hold any state or perform side-effects.
  */
 
@@ -134,7 +134,7 @@ interface AdapterProps extends ServiceProviderProps {
   children?(props: AddressInputContentProps): JSX.Element;
 }
 /** Note that the `AddressInputAppearanceProps` are not part of the renderChildren signature, hence this is not the
- * responsibility of the adapter, we should be provided them to the component independently. */
+ * responsibility of the adapter, we should be providing them to the component independently. */
 
 /*---------------------------------------------------------------------------------------------------------------------*/
 
--- a/src/AddressInput/AddressInput.part1.poc.tsx
+++ b/src/AddressInput/AddressInput.part1.poc.tsx
@@ -110,7 +110,7 @@ interface Suggestion {} // This is the raw "suggestion" object returned from the
 /**
  * So far we manged to define two building blocks required for our solution, one is the UI-Layer component <AddressInput/>
  * and the second is the ServiceProvider component <GooglePlacesServiceProvider/>.
- * It is easy to see that these components don't "speaks" the same language, the provider is not familiar with the
+ * It is easy to see that these components don't "speak" the same language, the provider is not familiar with the
  * UI-Layer API and vise-versa.
  *
  * GooglePlacesService  -/->  AddressInput (service-api !== component-props)
@@ -139,8 +139,8 @@ interface AdapterProps extends ServiceProviderProps {
 /*---------------------------------------------------------------------------------------------------------------------*/
 
 /** <GooglePlacesAddressInput/>
- * Let's put everything we defined so far together,
- * this component should fulfil our first milestone in the end product, an <AddressInput/> component that is
+ * Let's put everything we defined so far together.
+ * This component should fulfil our first milestone in the end product, an <AddressInput/> component that is
  * powered by Google Places. */
 interface GooglePlacesAddressInputProps
   extends AddressInputBaseProps,
@@ -150,7 +150,6 @@ const GooglePlacesAddressInput: React.FunctionComponent<GooglePlacesAddressInput
   initialValue,
   ...restAddressInputProps
 }) => (
-  // @ts-ignore
   <GooglePlacesServiceProvider initialValue={initialValue}>
     {serviceProviderProps => (
       <GooglePlacesServiceAdapter {...serviceProviderProps}>
@@ -171,7 +170,7 @@ const GooglePlacesAddressInput: React.FunctionComponent<GooglePlacesAddressInput
  * from outside, we will see later how we can still provide "extending" capabilities that can modify these props
  * for additional features.
  * Now, i can see why the code above can be eye pleasing to some while mind bending to others, it is just for
- * demonstration purposes to shows the separation of concerns between all of our building blocks and the props
+ * demonstration purposes to show the separation of concerns between all of our building blocks and the props
  * composition.
  * We can always consolidate the Provider/Adapter into a single component <GooglePlacesProvider/> resulting a code that
  * would look like...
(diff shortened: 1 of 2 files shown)


src/AddressInput/AddressInput.part1.poc.tsx
```diff
@@ -0,0 +1,197 @@
...
+    data: Suggestion[];
+    status: 'OK' | 'NOT_FOUND' | 'ZERO_RESULTS';
+  };
+  clearSuggestions?(): void;
+}
+interface Suggestion {} // This is the raw "suggestion" object returned from the google places service, untyped to reduce noise
+
+/*---------------------------------------------------------------------------------------------------------------------*/
+
+/**
```
shlomitc, line 110: great bullet

src/AddressInput/AddressInput.part1.poc.tsx
```diff
@@ -0,0 +1,197 @@
...
+ * GooglePlacesService  -/->  AddressInput (service-api !== component-props)
+ *
+ * Therefore, we will need an adapter to make the two talk to each other.
+ * Clarification, when speaking about adapters i will refer to function components that receive one type of api signature
+ * and return an different type of api-signature by applying transformations to the received props.
+ * These adapters should be pure, hence, shouldn't hold any state or perform side-effects.
+ */
+
+/**
+ * <GooglePlacesServiceAdapter/>
```
shlomitc, line 125: Why not simply combine the Adapter and the Provide?
The adapter can be initialized as a simple instance and passed to the adapter. It's one layer less, and I think the Adapter doesn't have to be a component

src/AddressInput/AddressInput.part1.poc.tsx
```diff
@@ -0,0 +1,197 @@
...
+          />
+        )}
+      </GooglePlacesServiceAdapter>
+    )}
+  </GooglePlacesServiceProvider>
+);
+/**
+ * This component is in fact taking the responsibility to implement the full <AddressInput/> interface.
+ * `AddressInputBaseProps` & `AddressInputAppearanceProps` are being exposed outside for the consumer to provide while
+ * the `AddressInputContentProps` is being served internally by our Provider/Adapter and is not open for modification
```
shlomitc, line 169: that's not completely true. in some case you do want to hook into some `onChange` or `onSelect` events, for example, to send a BI event.

src/AddressInput/AddressInput.part1.poc.tsx
```diff
@@ -0,0 +1,197 @@
...
+ * from outside, we will see later how we can still provide "extending" capabilities that can modify these props
+ * for additional features.
+ * Now, i can see why the code above can be eye pleasing to some while mind bending to others, it is just for
+ * demonstration purposes to show the separation of concerns between all of our building blocks and the props
+ * composition.
+ * We can always consolidate the Provider/Adapter into a single component <GooglePlacesProvider/> resulting a code that
```
shlomitc, line 175: oh, now I see you do suggest to mix them
So I wonder if it's not too complicated that you explain first the entire thinking process. it is interesting, but it takes some time to read and understand it and eventually you don't use it...

src/AddressInput/AddressInput.part1.poc.tsx
```diff
@@ -0,0 +1,197 @@
+import { DropdownLayoutOption, DropdownLayoutProps } from '../DropdownLayout';
```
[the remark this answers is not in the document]
lironch: Great point, i will add more information about what we currently have and what are the new requirements looking forward so the reader will have more context

src/AddressInput/AddressInput.part1.poc.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+/**
+ * <AddressInput/>
+ * This is our "marionette" component, it's responsibility is to follow the product visual definitions and behaviour
+ * guidelines, this way any consumer that wishes to build it's own address lookup solution will be able to reuse it and
+ * only take care of his specific use cases.
+ *
+ * In the following suggested api, we can see our props divided into 3 different groups, we will see later how this
+ * can provide us clarity when composing with this component... */
+class AddressInput extends React.PureComponent<AddressInputProps> {} // This can be a Function Component as well using hooks
```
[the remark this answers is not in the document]
lironch: I'm not sure i understand why, the options array is provided from outside and controlled by the user and not the component, i don't manage it in state or anything else similar

src/AddressInput/AddressInput.part1.poc.tsx
```diff
@@ -0,0 +1,197 @@
...
+interface AddressInputContentProps {
+  value?: string;
+  onChange?: React.ChangeEventHandler<HTMLInputElement>;
+  onClear?: () => void;
+  options?: DropdownLayoutOption[];
+  status?: 'loading' | 'error' | 'warning';
+}
+// Appearance related props, defining the component appearance & layout
+interface AddressInputAppearanceProps {
+  size?: 'small' | 'medium' | 'large';
```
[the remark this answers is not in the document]
lironch: I wanted to have everything as close to reality so this document can serve as a baseline for development, when i'll work on converting it to a blog post i will definitely reduce some props to make easier to the reader

src/AddressInput/AddressInput.part1.poc.tsx
```diff
@@ -0,0 +1,197 @@
...
+ * The component abstraction must stand by itself.
+ * */
+
+/*---------------------------------------------------------------------------------------------------------------------*/
+
+/**
+ * So we used the spec we got to define our first building block, however, this is not enough in order to build the
+ * product we were requested.
+ * Luckily, we don't need a detailed spec to realize that this component will need to get it's info from some kind of
+ * location service provider, Google Places is a good use case since it is what we currently use in our old component
```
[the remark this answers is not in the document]
lironch: Done 👍

src/AddressInput/AddressInput.part1.poc.tsx
```diff
@@ -0,0 +1,197 @@
...
+interface ServiceProviderProps {
+  value?: string;
+  setValue?(value: string): void;
+  suggestions?: {
+    data: Suggestion[];
+    status: 'OK' | 'NOT_FOUND' | 'ZERO_RESULTS';
+  };
+  clearSuggestions?(): void;
+}
+interface Suggestion {} // This is the raw "suggestion" object returned from the google places service, untyped to reduce noise
```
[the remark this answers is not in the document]
lironch: I added a better description, let me know if it helps

src/AddressInput/AddressInput.part1.poc.tsx
lironch: 🎉

src/AddressInput/AddressInput.part1.poc.tsx
lironch: As you noticed bellow i am indeed suggesting to combine the two, i will address it more on other comment.
Regarding the "adapter not being a component", that is true to some extent, since this content is already technical i wanted to stick to familiar elements that won't add friction.

src/AddressInput/AddressInput.part1.poc.tsx
lironch: You are correct, having it like that is intended, i am laying down here the basics just to get a minimal working product, in part2 i will show how we can add more features and capabilities to this solution and here is exactly where we will need to "open up" our solution for additional "plugins".

src/AddressInput/AddressInput.part1.poc.tsx
lironch: Like i mentioned in the above comment, most of part2 will deal with extending the capabilities of what we got so far, since the UI-Layer and ServiceProvider are pretty much defined and closed for additions, most of the changes will happen in the Adapter area, this is why i think it is important to understand what was it's original purpose and how we suggest to touch when adding more code.

shlomitc:
@lironch - what's the future of this PR?

PR #5637, opened 2020-07-02
<Badge> - use <Caption/> internally

EgoziE:
Instead of a span with ellipsisHOC


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5637/

src/Badge/Badge.st.css
```diff
@@ -78,13 +73,7 @@
 }
 
 .root .text {
-  -st-mixin: text-medium-bold;
-  font-size: 10px;
-  line-height: 12px;
-  letter-spacing: 1px;
   color: currentColor;
```
shlomitc, line 76: You should -st-extend the caption component

src/Text/Caption.st.css
```diff
@@ -0,0 +1,14 @@
...
+  -st-from: "./Text.st.css";
+  -st-default: Text;
+}
+
+.caption {
+  -st-extends: Text;
+}
+
+.c1 {
+  font-size: 10px !important;
```
shlomitc, line 11: I think you won’t need the important if you target the state you’re overriding (which is more specific)

PR #5457, opened 2020-05-12
Inputs & selection families- aligning large size to 42px

TalZa1810:
### 🔦 Summary

Inputs & selection families- aligning large size to 42px
Work in progress

### ✅ Checklist

- [ ] 👨‍💻 API change is approved by the UI Infra Developers 
- [x] 👨‍🎨 UX change is approved by the Design System UX 
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [x] Visual test

PR #5645, opened 2020-07-05
<LabelledElement /> Height should be !important

OriMarron:
### 🔦 Summary
This is a bug fix.

`LabelledElement`'s `inputContainer`'s height should be marked as `!important`.

the bug was introduced in #5457, when the previous implementation which was `!important` was replaced with current one.

This causes inputs to not have correct height **in production**. 

Note:
The bug is not always reproduced. There are conflicting css rules and order of cascade changes from environment to environment (works fine in storybook), and if compiled in production or dev mode (works in dev)

expected:
image
actual:
image


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5645/

TalZa1810:
There was an issue with this component when we tried to align the height of all inputs. 
Currently, since it seems like this is internal change which affect only the `<InputWithLabel/>` I'll merge this.

Issue #5596, opened 2020-06-22
Add scroll listener to Page component


hitpopdimestop:
# ✨ Feature Request

### 🔦 Scope
`<Page />`
...

### 💁 Explanation
Right now it is pretty hard to track scroll events on a Page component, you have to make pretty ugly workarounds. What would be nice to have is these optional properties on a Page component:

`onPageScroll: ({top, bottom, left, right, reachedTop, reachedBottom, reachedLeft, reachedRight}) => {...}, // event returns offsets for all of the sides of the viewport and booleans representing if some of the edges are reached`

`scrollDebounce: 200 // we are able to set the debounce for the scroll event in ms`

`bypassDebounceOnEndReach: true // if we would like to receive onPageScroll event immediately when the user scrolls to one of the edges of the page`
...

### 🐾 Possible solution 
Listening to the event in an app that uses WSR is possible by adding event listener to this element

`const scrollableContent = document.querySelector(
      '[data-hook="page-scrollable-content"]',
    );`

in WSR it is possible to add a ref to that element as well as an onScroll event listener
...

### 👀 Severity

-  Low

lironch:
@hitpopdimestop 
Thank you for the submitting this Feature Request.
In case you manage to find some time to work on it, it will be great if you can reach out to one of our team members for some initial guidance with the api definitions and implementation guidelines.
We have an internal component that can save you most of the work in this use case so it will be a shame if you will write it from scratch.

PR #5669, opened 2020-07-10
feat(Page): add onScroll callbacks

hitpopdimestop:
This PR adds scroll listeners to Page component and resolves this feature request:
[#5596 ](https://github.com/wix/wix-style-react/issues/5596)

### 🔦 Summary

...

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [x] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5669/

src/Page/Page.js
```diff
@@ -95,6 +99,9 @@ class Page extends React.PureComponent {
     // Maybe there is a transition
     const ARBITRARY_SHORT_DURATION_MS = 100;
     setTimeout(this._calculateComponentsHeights, ARBITRARY_SHORT_DURATION_MS);
+
+    this.props.scrollableContentRef &&
+      this.props.scrollableContentRef(this.scrollableContainerRef.current);
```
lironch, line 104: Consider adding a comment above these lines to make it clear it is not part of the "hack" above.
Another option is to restore the `_setScrollContainer` method, rename it to `_publishScrollableContainer` and call it here.

lironch:
@shlomitc
Regarding the issue you pointed out with delaying the header from minimising according to the `scrollThrottleWait` value.
We need to remember that the situation before was much worst, we were handling the scroll event in an `active` way and blocking by that the rendering thread, now at least we provide a `passive` handling to this event using the throttle mechanism which should provide a much smoother experience to the user as long as the `scrollThrottleWait` is not a big value.

Here are the options i think we have moving forward:
1. Declare an internal value for `scrollThrottleWait` in the `Page` component without exposing it as a prop to the users for modifications. We can make sure the value is small enough to provide us with `passive` handling of the event while visually there is no noticeable change to the component behaviour. 
In terms of cost-effectiveness, this one is very "cheap" and should suffice for the time being which makes it my favorite option, if we will get requests for configurable `scrollThrottleWait` value in the future we will handle it then.
2. Pass a `ref` from `Page` to `ScrollableContainer` and register the `Page` component to the `scroll` events on that `ref`.
This means that `Page` will have full control on the event handling for it's internal usages while the event handling for the `Page` users is done by `ScrollableContainer` according to the provided `scrollThrottleWait` value they set. 
Cost-effectiveness wise, this one is more expensive since we will need to write and test more code in the `Page` component   which is already written and tested at `ScrollableContainer` so i this this can be done when option 1 won't be enough.

Regarding adding visual/e2e tests, i looked on the current suite we have and it felt "good enough" to cover this new addition, every new test i thought off felt to me like something that tests `ScrollableContainer` and not the integration, any ideas for uncovered flows?

shlomitc:
Regarding the solution - if it doesn't need to be configurable, just remove the prop and leave some comment to explain that this value is hardcoded because it affects the page header animation.

I must say I don't really understand the difference between `onScrollChanged` and `onScrollPositionChanged`. Since there are no tests or examples, it's really hard to understand how to use the. the comments explain what they do, but not when or how should I use it. This is why I think tests can be important here

lironch:
I created a [new PR](https://github.com/wix/wix-style-react/pull/5766) that addresses the issues in this one
FYI @hitpopdimestop @shlomitc

PR #5702, opened 2020-07-15
Tooltip: Explain that children's "disabled" prop can affect the Tooltip

splintor:
### 🔦 Summary

Update `<Tooltip>`'s `disabled` prop documentaiton to explain that if it is not specified, its value is taken from the `children`'s `disabled` prop.

### ✅ Checklist

- 📚 Change is documented
  - [X] API description


src/Tooltip/Tooltip.js
```diff
@@ -52,7 +52,7 @@ class Tooltip extends React.PureComponent {
     onHide: PropTypes.func,
     /** tooltip content placement in relation to target element */
     placement: PropTypes.string,
-    /** disables tooltip element trigger behaviour */
+    /** disables tooltip element trigger behaviour. If not specified, `disabled` prop of the `children` element will be used. */
```
shlomitc, line 55: I think that it will be best to add this to the documentation page instead since it's not directly related to this prop.

EgoziE:
Hey @splintor thanks for this PR, please make sure your commits are signed for the next one.

src/Tooltip/Tooltip.js
EgoziE: I missed your comment because my page wasn't refreshed... I'll add it.

EgoziE:
I added this example:
image

PR #5711, opened 2020-07-16
<CloseButton/> - fix incorrect style extending

shlomitc:
### 🔦 Summary
I was playing around with extending some styles and for some reason, I couldn't override `IconButton::content`.
So I took a look inside the component and saw that we're using both `...styles(***)` and passing `className` which looks like a mistake.


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5711/

shlomitc:
> That was messed up.. 😵
> 
> I'd also change:
> 
> ```
> 
> {...styles('root', { skin, size }, this.props)}
> 
> ```
> 
> 
> 
> to 
> 
> ```
> 
> {...styles('root', { skin, size }, { className })}
> 
> ```

I'm not sure that's a good change. The idea is to pass all props so also the data-* attributes will be applied. The stylable runtime extracts them from the props object

PR #5718, opened 2020-07-19
<Text/> - Align non textual direct children of text to top

EgoziE:
There's a bug that children of `<Text/>` is not aligned to middle.
The problem is that the the current (and default) vertical alignment is to "baseline" which means that the child's top is aligned to **text** (determined by the font-size) instead of the whole Text element top (determined by the line-height)

* Before:
image

* After:
image


EgoziE:
image
image

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5718/

shlomitc:
looks like an important fix but it breaks some visual tests. need to see why it happens and if it has the potential to break something in other usages

EgoziE:
@shlomitc I conducted a deeper investigation, the **only** case where this solution breaks is when line-height is set manually from outside.
image

As far as I know, according to our design guidelines, there is no valid use case where line-height needs to be set manually by a consumer. @milkyfruit please correct me if I'm wrong.

(The **only** visual test that fails is Breadcrumbs which is using line-height incorrectly)

milkyfruit:
Yup, you're right @EgoziE, end users shouldn't care about line height and we don't allow customizing it.
> As far as I know, according to our design guidelines, there is no valid use case where line-height needs to be set manually by a consumer. @milkyfruit please correct me if I'm wrong.

PR #5721, opened 2020-07-19
`<Animate/>` - new component

TalZa1810:
### 🔦 Summary
Currently we have only one type `bounce`. In the future, we may have more types, so we can add `type` to the stylable states. 

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [ ] Visual test


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5721/

TalZa1810:
@ZilvinasMaciulis 
I'm still in the process of understanding how to test this component. In the meantime, you can have a look in the examples I've created and tell me what you think :)

src/Animate/docs/index.story.js
```diff
@@ -0,0 +1,103 @@
...
+  componentProps: {},
+
+  exampleProps: {
+    // Put here presets of props, for more info:
+    // https://github.com/wix/wix-ui/blob/master/packages/wix-storybook-utils/docs/usage.md#using-list
+  },
+
+  sections: [
+    header({
+      sourceUrl: `https://github.com/wix/wix-style-react/tree/master/src/${Animate.displayName}/`,
```
argshook, line 41: if i remember well, `sourceUrl` in wsr is automated, no need to set it manually

perfer.config.js
```diff
@@ -140,6 +140,7 @@ const javascriptFiles = [
   ['VerticalTabs.bundle.min.js', 38],
   ['VerticalTabsItem.bundle.min.js', 38],
   ['WarningIndicator.bundle.min.js', 40],
+  ['Animate.bundle.min.js', 10],
```
EgoziE, line 143: sort and decrease to 5

stories/index.js
```diff
@@ -189,3 +189,6 @@ require('./PopoverMenu/index.story'); // 7.3 PopoverMenu
 // require('../src/CustomModalLayout/docs/design.story'); // 9.3 Custom Modal (WIP)
 require('./ModalMobileLayout/index.story'); // 9.5 Mobile Modal
 require('../src/MessageBox/docs');
+
+// TODO: move to correct position
+require('../src/Animate/docs/index.story');
```
EgoziE, line 194: move

src/Animate/test/Animate.visual.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
```
EgoziE, line 1: I guess we can remove visual tests completely, it's not relevant for animations and prone to be flaky when it comes to animations.

src/Animate/Animate.js
```diff
@@ -0,0 +1,131 @@
...
+  dataHook: PropTypes.string,
+
+  /** The component which we would like to animate. When given multiple components- only the first child node will be animated. */
+  children: PropTypes.node.isRequired,
+
+  /** Triggers the animation transition */
+  triggerAnimation: PropTypes.bool.isRequired,
+
+  /** A callback fired immediately after the transition enters.*/
+  onEnter: PropTypes.func,
```
argshook, line 117: i would name this `onStart`

src/Animate/Animate.js
```diff
@@ -0,0 +1,131 @@
...
+
+  /** A callback fired immediately after the transition finishes.*/
+  onExited: PropTypes.func,
```
argshook, line 120: i would name this `onEnd`

src/Animate/index.d.ts
```diff
@@ -0,0 +1,13 @@
+import * as React from 'react';
+
+export interface AnimateProps {
+  dataHook?: string;
+  children?: React.ReactNode;
+  triggerAnimation?: boolean;
+  onEnter?(): void;
+  onExited?(): void;
+  loop?: boolean;
+  delay?: boolean;
```
EgoziE, line 10: string

src/Animate/index.d.ts
```diff
@@ -0,0 +1,13 @@
+import * as React from 'react';
+
+export interface AnimateProps {
+  dataHook?: string;
+  children?: React.ReactNode;
+  triggerAnimation?: boolean;
```
EgoziE, line 6: required

src/Animate/index.d.ts
```diff
@@ -0,0 +1,13 @@
+import * as React from 'react';
+
+export interface AnimateProps {
+  dataHook?: string;
+  children?: React.ReactNode;
```
EgoziE, line 5: required

src/Animate/docs/index.story.js
```diff
@@ -0,0 +1,103 @@
...
+    }),
+
+    tabs([
+      tab({
+        title: 'Description',
+        sections: [
+          description({
+            title: 'Description',
+            text:
+              'This line here should briefly describe component in just a sentence or two. It should be short and easy to read.',
```
EgoziE, line 52: write a description

src/Animate/Animate.js
```diff
@@ -0,0 +1,131 @@
...
+
+  /** when set to true, the child component animate repetitively until stopped by other event*/
+  loop: PropTypes.bool,
+
+  /** set a delay before the animation execution */
+  delay: PropTypes.string,
```
argshook, line 126: this should support a number also imo. number can be treated as `ms`
```suggestion
  delay: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
```

src/Animate/Animate.js
```diff
@@ -0,0 +1,131 @@
...
+  };
+
+  render() {
+    const { classNames } = this.state;
+    const { dataHook, delay } = this.props;
+
+    return (
+      <div
+        data-hook={dataHook}
+        {...classNames}
```
EgoziE, line 87: This is a bit complicated to understand.
I'd change it to something like:

```
{ ...styles('root', { loop, size }), { className: someBoolean && styles.animation} }
```

or:

```
{ ...styles('root', { active, loop, size })}
```

src/Animate/Animate.js
```diff
@@ -0,0 +1,131 @@
...
+
+Animate.propTypes = {
+  /** Applied as data-hook HTML attribute that can be used in the tests */
+  dataHook: PropTypes.string,
+
+  /** The component which we would like to animate. When given multiple components- only the first child node will be animated. */
+  children: PropTypes.node.isRequired,
+
+  /** Triggers the animation transition */
+  triggerAnimation: PropTypes.bool.isRequired,
```
argshook, line 114: i'd probably go with `play` or just `trigger`

src/Animate/Animate.js
```diff
@@ -0,0 +1,131 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styles from './Animate.st.css';
+import { dataHooks, childSize, childWidthRange } from './constants';
+
+// return true if in range, otherwise false
+const inRange = (x, min, max) => (x - min) * (x - max) <= 0;
```
argshook, line 7:
```suggestion
const inRange = (x, min, max) => x >= min && x <= max;
```

src/Animate/Animate.js
```diff
@@ -0,0 +1,131 @@
...
+    );
+  };
+
+  _toggleAnimation = () => {
+    const size = this._getChildComponentSize();
+    const { classNames } = this.state;
+    const { loop } = this.props;
+
+    this.setState({
+      classNames: classNames ? '' : { ...styles('animation', { loop, size }) },
```
argshook, line 63: if `this.state.classNames` is truthy, it will be set to `string` which
during render will be spread. it works but doesn't feel right.

moreover, `...styles()` returns an object, like `{ classNames: '...' }`

thus, i think `this.state.classNames` should always be a string and not
change type.

apply to `div` as `className={this.state.classNames}` without the spread

src/Animate/Animate.js
```diff
@@ -0,0 +1,131 @@
...
+        {...classNames}
+        style={delay ? { animationDelay: delay } : null}
+        onAnimationStart={this._onAnimationStart}
+        onAnimationEnd={this._onAnimationEnd}
+      >
+        <div
+          data-hook={dataHooks.animateContent}
+          className={styles.animateContent}
+        >
+          <div className={styles.observer} ref={this.childObserverRef} />
```
argshook, line 96: could this `ref` be taken from parent div? in a way, to avoid additional div just for getting width.

src/Animate/Animate.js
```diff
@@ -0,0 +1,131 @@
...
+          {this._renderFirstChild()}
```
argshook, line 97: is it a requirement to disallow more than one child to be rendered?

src/Animate/index.d.ts
argshook: is it tho? maybe default behaviour is when no `triggerAnimation` set, then `Animate` would start animation as soon as it renders

TalZa1810: it's more of a UX decision no? @ZilvinasMaciulis wdyt?

src/Animate/Animate.js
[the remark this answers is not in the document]
TalZa1810: I've set it eventually to the root `div`.  In the first render, the ref is set to `null` and than it fails to get the width.  In order to prevent this, I've set the width to `0` in case the `ref` equals null.

src/Animate/Animate.js
```diff
@@ -0,0 +1,142 @@
...
+
+Animate.propTypes = {
+  /** Applied as data-hook HTML attribute that can be used in the tests */
+  dataHook: PropTypes.string,
+
+  /** The component which we would like to animate. When given multiple components- only the first child node will be animated. */
+  children: PropTypes.node.isRequired,
+
+  /** Triggers the animation transition */
+  trigger: PropTypes.bool,
```
shlomitc, line 125: "trigger" is a bit confusing in my opinion.
I'd expect using `trigger()` as action.
or pass some component as a "trigger".

I think naming it "active" could be better

src/Animate/Animate.js
```diff
@@ -0,0 +1,142 @@
...
+      case 'string':
+        return { animationDelay: delay };
+      case 'number':
+        return { animationDelay: `${delay}ms` };
+      default:
+        return null;
+    }
+  };
+
+  _getDataAttributes = ({ active, loop, size }) =>
```
shlomitc, line 83: I think that this is redundant...
sorry, but this is not testing logic, but rather testing UI and for UI testing we have visual tests.
In this case, animations are hard to test so we just drop it.

adding the `data-*` tags in old components was necessary because we used to expose test methods that queried the class names and couldn't find a better alternative. in this case, I don't think we should introduce it

TalZa1810: The `size` and `active` stylable props are doing some logic. The `size` is being calculated based on the child component width and the `active` is changed based on the component state ( part of the `onTransitionEnd` and `onTransitionStart`). I will continue research of a different way to test this kind of component.

shlomitc: I understand that, but AFAIK JSDom does not implement any style or animation API, meaning you can't actually test this logic without proper browser level APIs.
I think it doesn't worth the effort

PR #5723, opened 2020-07-20
fixing imports with the new syntax 

TalZa1810:
### 🔦 Summary
Running the codemode to fix the imports with the new syntax


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5723/

src/TableToolbar/docs/MainExample.js
```diff
@@ -28,36 +24,30 @@ export default class MainExample extends React.Component {
...
-            </Item>
-          </ItemGroup>
-          <ItemGroup position="end">
-            <Item>
+          <TableToolbar.ItemGroup position="start">
+            <TableToolbar.Item>
+              <TableToolbar.Title>My Table</TableToolbar.Title>
+            </TableToolbar.Item>
+            <TableToolbar.Item>
+              <FormField label="Collection" labelPlacement="left">
```
TalZa1810, line 32: I saw that there's a `<TableToolbar.Label/>` component and it has a span with inline styles. Here we prefer to use `<FormField/>` instead, right? Also, the `<TableToolbar.Label/>` component is not documented.

argshook: for transparency: i reverted this change for now. although it makes sense to use `FormField` instead of poorly documented `TableToolbar.Label`, it breaks existing visual tests. Thus plainly using `FormField` is not enough, more tweaks are needed.

because of this, it's better to do this change as a separate PR

argshook:
There are 2 visual diffs that are neglible, unrelated to this PR and seemingly because of newer chrome version. Other than that, it's good to merge!

PR #5729, opened 2020-07-21
chore(Focusable): remove unused E2E test

argshook:
I want to eradicate the usage of `AutoExampleDriver` and in wix-style-react library this is the only place it is used.

or, not really used, these tests are not executed at all, thus no need to carry them around.

In any way, testing focusable visibility is much better done with visual test.

There were also documentation mentions about merging private and public drivers, which was discussed many times to be ill-advised.


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5729/

PR #5730, opened 2020-07-21
docs(Tooltip): add example with wrapped tooltip

argshook:
1. fix one description which has `<button>` in text which was actually
   rendered as native html
2. add example with Tooltip wrapped with `<Box inline>`.

point 2. is a recurring issue that people are asking for from time to
time, thus it's a good idea to have example covering it.

Ideally issue should not be present at all, but until we get there, an
example is better than nothing.


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5730/

PR #5733, opened 2020-07-22
feat(filterObject): add utility function and refactor Object.fromEntries

argshook:
this PR refactors existing `Object.fromEntries` usages with internal utility function `filterObject`

it is needed to prevent runtime errors for users who don't have polyfills big enough to include one for `Object.fromEntries`


src/utils/extractAttributes.js
```diff
@@ -1,12 +1,16 @@
...
  * @param prefix
  * @return {any}
  */
 export const extractAttributes = (props, prefix = '') =>
-  Object.fromEntries(
-    Object.entries(props).filter(entry => entry[0].indexOf(prefix) === 0),
+  filterObject(
+    props,
+    key =>
+      key.length > prefix.length && key.substring(0, prefix.length) === prefix,
```
lironch, line 13: This changes the condition logic to exclude keys that are strictly equal to the prefix where before they were included, WDYT?

src/utils/filterObject.js
```diff
@@ -0,0 +1,9 @@
+export const filterObject = (object, filter = () => true) => {
```
lironch, line 1: I would add here some documentation about the method signature and it's purpose to provide an alternative to `Object.fromEntries`.
This way in case a developer will search for usages of `Object.fromEntries` in the code in the future he/she will end up discovering this method thanks to that.

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5733/

src/utils/extractAttributes.js
argshook: fixed to work as before. also added test case to check for that

src/utils/filterObject.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * filterObject is a utility function for excluding desired properties from an object. It works similar to Array.prototype.filter.
+ * This function should be used instead of Object.fromEntries(Object.entries(object).filter(...))
+ *
+ * @param {object} object - original object to be filtered
+ * @param {function} filter
```
argshook, line 6:
```suggestion
 * @param {function} filter - function of signature (key, value) => boolean. executed for each property in `object`
```

EgoziE:
I approve this PR with great sadness 😢

Issue #5511, opened 2020-06-02
<BarChart /> label not vertically aligned


MSiliunas:
# 🐛 Bug Report

### 🏗 Relevant Components

`<BarChart />`
`<Heading />`

### 😯 Current Behavior

![image](https://user-images.githubusercontent.com/6944305/83518793-94d8f780-a4e3-11ea-8a20-408713b6e194.png)

### 🤔 Expected Behavior

I expect that the label would be centered vertically.

### 👣 Steps to Reproduce

I don't have exact steps to reproduce, only happens in production. I've found out that BarChart's style `margin: auto` lacks specificity as it gets overriden by Heading's margin (see attached screenshot).
You can see the issue in Wix Email Marketing statistics.

  - **WSR version**: `8.13.1`
  - **Yoshi version**: `4.84.0`

### 👀 Severity

- Critical

MSiliunas:
@mykas is familiar with the problem, had a few words with him beforehand :)

drasius:
@mykas @EgoziE @milkyfruit 
Hi, what is the status of this one? It looks really bad and we face it in production, can you please prioritize this bug?

EgoziE:
@drasius Thanks for your concern, we prioritized it as "low" so there's no update currently.
Due to thunderbolt effort we have very limited capacity.
Would you like to contribute this fix?

PR #5734, opened 2020-07-22
fix(barchat): css specificity issues

mastertheblaster:
This PR fixes the issue #5511

It happens due to CSS specificity clash.

```css
.class {}
.class > * {}
```

Have the same specificity and the override is being controlled by order of declaration.
Current fix eliminates the possibility of the clash.


src/BarChart/BarChart.st.css
```diff
@@ -28,17 +28,14 @@
...
+  padding: 0 24px;
   height: value(height);
   display: flex;
+  align-items: center;
+  justify-content: center;
   margin-bottom: 12px;
 }
 
-.value > * {
-  padding: 0 12px;
```
mastertheblaster, line 38: The horizontal padding was set for all children, however it's horizontal padding so i just doubled the padding of the parent component instead.

src/BarChart/BarChart.st.css
```diff
@@ -28,17 +28,14 @@
...
-  margin: auto;
```
mastertheblaster, line 39: Instead of margin for children i used `flex` props to align items in the center. The parent element was already using `flex` display.

argshook:
Let's wait for visual tests to run, once green i will merge

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5734/

src/BarChart/BarChart.st.css
```diff
@@ -28,15 +28,16 @@
...
+  padding: 0 24px;
   height: value(height);
   display: flex;
+  align-items: center;
+  justify-content: center;
   margin-bottom: 12px;
 }
 
-.value > * {
+.value:empty {
```
mastertheblaster, line 39: A special usecase for empty elements to save the horizontal space.

mastertheblaster: https://developer.mozilla.org/en-US/docs/Web/CSS/:empty

mastertheblaster: Screenshot 2020-07-22 at 15 10 40

argshook: just fyi, we can achieve this with stylable state, without the worry of css support. but as the support is well, i don't mind it without stylable to

PR #5743, opened 2020-07-22
<TagList/> - add size prop

gendeld:
### 🔦 Summary

Adds size prop to TagList to control Tag and ActionButton sizes.

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5743/

src/TagList/TagList.js
```diff
@@ -46,6 +62,11 @@ TagListAction.propTypes = {
   dataHook: PropTypes.string,
   children: PropTypes.node,
   onClick: PropTypes.func,
+  size: PropTypes.oneOf(['tiny', 'small', 'medium']),
```
EgoziE, line 65: Remove size prop from here, it's not a public API

src/TagList/index.d.ts
```diff
@@ -2,16 +2,17 @@ import * as React from 'react';
 import { TagProps } from '../Tag';
 import { ButtonProps } from '../Button';
 
+export type TagListSize = 'small' | 'medium' | 'large';
+
 export type TagListProps = {
   dataHook?: string;
   tags: Array<TagProps>;
```
EgoziE, line 9: Not all `<Tag/>` props are public API in this case, please pick only those we expose.

src/TagList/TagList.js
EgoziE: Also add `id` and `theme` and maybe more, check wha't relevant from `<Tag/>` component in this or another PR

src/TagList/index.d.ts
```diff
@@ -2,16 +2,17 @@ import * as React from 'react';
...
+  size?: TagListSize;
```
EgoziE, line 10: And also add it to the type test `src/TagList/test/TagList.tsx` as well

src/TagList/TagList.js
```diff
@@ -57,11 +78,18 @@ TagList.propTypes = {
   /** List of tags props to be rendered */
   tags: PropTypes.arrayOf(PropTypes.object),
 
+  /** The size of each individual `<Tag />` */
+  size: PropTypes.oneOf(['small', 'medium', 'large']),
```
EgoziE, line 82: Please add a visual test for this prop

src/TagList/index.d.ts
gendeld: The code as written accepts any TagProp and works properly, shouldn't the types reflect that?

src/TagList/TagList.js
gendeld: Can you explain what we add to propTypes for private components?

src/TagList/index.d.ts
```diff
@@ -2,16 +2,17 @@ import * as React from 'react';
 import { TagProps } from '../Tag';
 import { ButtonProps } from '../Button';
 
+export type TagListSize = 'small' | 'medium' | 'large';
+
 export type TagListProps = {
   dataHook?: string;
-  tags: Array<TagProps>;
+  tags: Array<Omit<TagProps, 'size'>>;
```
splintor, line 9: @EgoziE This use of `Omit<>` which [was added in TypeScript 3.5](https://www.infoq.com/news/2019/06/typescript-3-5-release/) broke [our build](http://tc.dev.wixpress.com/viewLog.html?buildId=118814619&buildTypeId=mono_repo_wixos_ping_settings_PingSettingsClient_O&tab=buildLog&_focus=169#_state=467), as we were still using `"typescript": "~3.0.1"`.

You might want to add a pre-req TypeScript version in https://github.com/wix/wix-style-react/blob/master/docs/FAQ/TYPES.MD

PR #5747, opened 2020-07-23
TableActionCell: Fix showing tooltip for visible disabled actions

splintor:
### 🔦 Summary
https://github.com/wix/wix-style-react/pull/4661 tried to make visible secondary items show tooltip when they are disabled, but since `<Tooltip>` [uses its children's `disabled` prop](https://github.com/wix/wix-style-react/pull/5702), this won't work, and the tooltip is not displayed.

This PR changes it, so the tooltip's `disabled` prop is set to `false`, unless the internal `<IconButton>` is disabled, and `disabledDescription` is explicitly set to an empty string.

It is difficult to test this change, as `<Tooltip>` testkit doesn't provide an easy way to check its `disabled` prop.

### ✅ Checklist
- 📚 Change is documented
  - [X] Story


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5747/

PR #5750, opened 2020-07-26
<Ellipsis/> - strengthen ellipsis rules

EgoziE:
An element has ellipsis if:
1. The parent element is shorter than the element
2. The scroll width of the element is bigger than its offset width

There's a very rare bug related to **pixel rounding** that makes the element width be rounded differently than its parent, therefore in very rare cases there's an ellipsis when there shouldn't be.
The fix is to set ellipsis only when the difference between the element width to its parent is larger than 1px.


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5750/

PR #5755, opened 2020-07-27
<TagList/> - add onTagRemove prop

gendeld:
### 🔦 Summary

...

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


src/TagList/index.d.ts
```diff
@@ -5,13 +5,12 @@ import { ButtonProps } from '../Button';
 export type TagListProps = {
   dataHook?: string;
   tags: Array<TagProps>;
+  onTagRemove?: (id: string) => void;
```
EgoziE, line 8: Add this to type test `src/TagList/test/TagList.tsx`

src/TagList/TagList.js
```diff
@@ -57,6 +58,9 @@ TagList.propTypes = {
   /** List of tags props to be rendered */
   tags: PropTypes.arrayOf(PropTypes.object),
 
+  /** Callback function that passes `id` property as parameter when removing a Tag  */
+  onTagRemove: PropTypes.func,
```
EgoziE, line 62: Please add a unit test for this prop

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5755/

src/TagList/test/TagList.spec.js
```diff
@@ -3,6 +3,7 @@ import { createRendererWithUniDriver, cleanup } from '../../../test/utils/unit';
 
 import TagList from '../TagList';
 import { tagListPrivateDriverFactory } from './TagList.private.uni.driver';
+import { element } from 'prop-types';
```
EgoziE, line 6: Irrelevant here

src/TagList/TagList.uni.driver.js
```diff
@@ -14,5 +14,7 @@ export const tagListDriverFactory = (base, body) => {
     clickActionButton: async () => (await getTagListActionTestkit()).click(),
     /** Get if action button exists */
     actionButtonExists: async () => (await getTagListActionTestkit()).exists(),
+    /** Get Tag by data-hook */
+    getTagDriver: dataHook => base.$(`[data-hook=${dataHook}]`),
```
EgoziE, line 18: We don't like a driver returning a driver.
Anyway this function can be deleted, take a look at my comment in the spec file

src/TagList/test/TagList.spec.js
```diff
@@ -77,4 +78,23 @@ describe(TagList.displayName, () => {
...
+            children: 'Some Tag',
+            dataHook: tagDataHook,
+          },
+        ]}
+      />,
+    );
+    const tagDriver = await driver.getTagDriver(tagDataHook);
+    const tagElement = await tagDriver.element();
+    expect(tagElement.onClick).toBe(onTagRemoveFn);
+  });
```
EgoziE, line 99:
```suggestion
  it('should trigger onRemove when removing a tag', async () => {
    const onRemove = jest.fn();
    const id = '1';
    const { driver } = render(
      <TagList
        tags={[
          {
            id,
            children: 'Some Tag',
            onRemove,
          },
        ]}
      />,
    );
    await driver.removeTag(id);
    expect(onRemove).toHaveBeenCalledWith(id);
  });
```

EgoziE: A valid driver function should be:
removeTag(id) => void

gendeld: ^ This does not test the onRemoveTag prop, it tests <Tag/>'s onRemove, which is already being tested somewhere else

EgoziE: apologies
```
  it('should trigger onRemove when removing a tag', async () => {
    const onTagRemove = jest.fn();
    const id = '1';
    const { driver } = render(
      <TagList
        onTagRemove={onTagRemove}
        tags={[
          {
            id,
            children: 'Some Tag',
          },
        ]}
      />,
    );
    await driver.removeTag(id);
    expect(onTagRemove).toHaveBeenCalledWith(id);
  });
```

EgoziE: The bottom line is, don't let a driver function return a driver.
When it comes to a callback function prop, we test it by checking it's called (with params if needed).

Testing equality between the callback function and the element's onClick is not a good solution because:
1. You're not testing the signature.
2. In some cases we create our own onClick that calls the callback function (therefore they won't be equal)

src/TagList/TagList.uni.driver.d.ts
```diff
@@ -1,7 +1,9 @@
 import { BaseUniDriver } from 'wix-ui-test-utils/unidriver';
+import { TagUniDriver } from '../Tag/Tag.uni.driver';
 
 export interface TagListUniDriver extends BaseUniDriver {
   actionButtonLabel: () => Promise<string>;
   clickActionButton: () => Promise<void>;
   actionButtonExists: () => Promise<boolean>;
+  getTagDriver: (dataHook: string) => Promise<TagUniDriver>;
```
EgoziE, line 8:
```suggestion
  removeTag: (id: string) => Promise<void>;
```

PR #5760, opened 2020-07-27
TableActionCell: add placeholder to primary and only visible secondary

assafganor:
### 🔦 Summary
TableActionCell in most cases has a placeholder which causes the user hovering the table row.
Placeholder types:
1) When only primaryAction without secondaryActions -> ChevronRight
2) When secondaryActions has hiddenActions -> More
3) When no primaryAction and all secondaryActions are visible there is no placeholder.
4) When primaryAction and all secondaryActions are visible there is no placeholder.

I changed the behaviour of 4 to display ChevronRight.

I think you should think about 3 too. How does the user know he should hover the row if there is no placeholder?

### ✅ Checklist

- 📚 Change is documented
  - [x] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [x] Visual test


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5760/

src/TableActionCell/TableActionCell.js
```diff
@@ -143,15 +143,16 @@ const TableActionCell = props => {
...
-        <HoverSlot
-          display="notOnHover"
-          className={style.placeholderIcon}
-          data-hook={dataHooks.tableActionCellPlaceholder}
-        >
-          {renderPlaceholder()}
-        </HoverSlot>
-      )}
+      {primaryAction &&
+        (secondaryActions || []).length === numOfVisibleSecondaryActions && (
```
lironch, line 147: * You can remove the array fallback here since `secondaryActions` has a default value of `[]`.
* This change cause a regression when the ActionCell has only a primary action, now it won't show a placeholder for this case.
* I want you to support the 3rd case you described in the PR summary as well (no primary, all secondaries are visible on hover)

milkyfruit:
@assafganor 
It might be there was a miscommunication. When discussed with Nataly, we agreed on a solution to show the the secondary action on default when they are, but none of them hidden under 3 dots:
image

So, instead of an arrow there should be secondary action exposed.

src/TableActionCell/test/TableActionCell.spec.js
```diff
@@ -16,7 +16,15 @@ const primaryActionProps = (actionTrigger = () => {}, disabled = false) => ({
   },
 });
 
-const secondaryActionsProps = ({ actionTriggers, actionDataHooks } = {}) => {
+const secondaryActionsProps = ({
```
lironch, line 19: I think it will be better to expose `numOfSecondaryActions` & `numOfVisibleSecondaryActions` as optional parameters with default values, then just alter these values when needed

src/TableActionCell/TableActionCell.js
assafganor: *  I'll remove the fallback
* no regression, the test for only primary passed too. when no `secondaryActions`: `numOfVisibleSecondaryActions===secondaryActions.length===0` 
* I'll check @milkyfruit comments

lironch: This is the snippet you shared on slack before your code change:
Screen Shot 2020-07-29 at 11 53 29
And this is how it looks like with your code:
Screen Shot 2020-07-29 at 11 52 35
You can see that after your change the first line that only has a primary button now lack the placeholder, i don't think that was intended right?

lironch: You can see that it fails the e2e tests as well.

assafganor:
@milkyfruit 
- with your suggestion, we need to think about another scenario:
1 primary with 2 visibleSecondary and 0 hiddenSecondary, which placeholder to show?
- I think it's not logical to display the secondaryAction as the placeholder when we have a primaryAction.
In your example, I see the printer button, and if I don't want to print I'll not hover the row. if I'll see arrow I'll hover to check what is there.

src/TableActionCell/test/TableActionCell.spec.js
assafganor: changing

src/TableActionCell/TableActionCell.js
assafganor: fixing

assafganor:
@lironch I changed and fixed

milkyfruit:
> @milkyfruit
> 
> * with your suggestion, we need to think about another scenario:
>   1 primary with 2 visibleSecondary and 0 hiddenSecondary, which placeholder to show?
> * I think it's not logical to display the secondaryAction as the placeholder when we have a primaryAction.
>   In your example, I see the printer button, and if I don't want to print I'll not hover the row. if I'll see arrow I'll hover to check what is there.

@assafganor Ok, I spoke with Nataly again and we agreed for now to run what you developed, and later to make a proper fix which would involve repositioning main cta and secondary action positions. It will involve some design process first.

assafganor:
@lironch thanks for approval.
When running tests locally I get errors  on different components, so it's hard for me to understand if there is really a problem.
Does the failed tests are only eyes scroller position? is it ok?

TalZa1810:
@milkyfruit 
This is a behaviour change. We're not sure how our users are currently using this component. If we merge this PR, it will affect our other users as well. Therefore, we should decide whether this is the proper usage or not in order not to break existing users.

TalZa1810:
One visual test fails which is not related (`MessageModalLayout`).
I had to check 38 visual diffs of this specific change (`TableActionCell`) manually. It seems fine. I hope that I haven't missed anything.

PR #5778, opened 2020-07-30
`<MultiselectCheckbox/>`- using both builders and object

TalZa1810:
### 🔦 Summary
Supporting using builders in `<MultiselectCheckbox/>`. Supporting also the current component API for backwards compatibility.


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5778/

src/ListItemSelect/ListItemSelect.js
```diff
@@ -200,24 +200,28 @@ export const listItemSelectBuilder = ({
...
-      size={size}
-      ellipsis={ellipsis}
-      selected={selected}
-      highlighted={props.hovered}
-      {...props}
-    />
-  ),
-});
+}) => {
+  debugger;
```
lironch, line 204: oh no! 😱

TalZa1810:
Few things I wasn't sure about:
1. API description
2. Story description.
3. Types changes (builders)

src/MultiSelectCheckbox/MultiSelectCheckbox.st.css
```diff
@@ -0,0 +1,4 @@
+.showPointer,
```
shlomitc, line 1: please revert this. we decided to not do any further scss->st migration until v9

src/DropdownLayout/DropdownLayout.js
```diff
@@ -412,7 +412,7 @@ class DropdownLayout extends WixComponent {
           selected: selected && selectedHighlight,
           hovered,
           disabled,
-          title,
+          title: title && !overrideStyle,
```
shlomitc, line 415: Why is this needed?
I think it's because you're misusing the `title` prop.
`DropdownLayout` expects `title` to define an item which is visual but isn't selectable.
`ListItemSelect` expects `title` to define the main title.

In that case, you need to change the builder so `title` will not be mapped to the DropdownLayout object

src/MultiSelectCheckbox/test/MultiSelectCheckbox.visual.js
```diff
@@ -76,6 +76,32 @@ const interactiveTests = [
           valueParser: option => (option.label ? option.label : option.value),
         },
       },
+      {
+        it: 'Builders items',
```
TalZa1810, line 80: I wasn't sure whether adding this visual test is necessary

src/MultiSelectCheckbox/MultiSelectCheckbox.js
```diff
@@ -139,13 +150,17 @@ MultiSelectCheckbox.propTypes = {
 
   /** delimiter between the selected options that will be displayed in the input. */
   delimiter: PropTypes.string,
+
+  /** Function that receives an option, and should return the value to be displayed. By default returns `option.value` in case the option is an object and `option.optionTitle` in case it's a builder */
+  valueParser: PropTypes.func,
```
shlomitc, line 155: 1. I don't think that the component should not declare anything regarding the builder, so I'd remove it from the comment.

2. Regarding the `optionTitle` - now that's a bit more gentle case. It is an interesting idea to add it, but this becomes a public API of an option. The case where it becomes public is whenever the `option` is being sent as a callback to the consumer (like in `valueParser`)

I can suggest two options:
1. don't do this assistance. whoever decides uses `listItemBuilder` should supply it's own `valueParser`.
Since this `title` is basically in the hands of the consumer, he can simply take `id` which is already part of the required API, and combine it with the right title himself.
`<MultiSelectCheckbox valueParser(option => myOptionList[option.id].title)/>`

2. Define this API as part of a `DropdownLayout` option.
The `options` API of `DropdownLayout` which expects an array of `option`s will allow getting a new `label` property, same as it expects a `value` or `overrideStyles` here https://github.com/wix/wix-style-react/blob/9ed9ce86027fcd5c21ffe4e66330ce4ea9a80e13/src/DropdownLayout/DropdownLayout.js#L498
so basically `label` should always be the string applied in case of selection, unless `value` is a string` and then it can be used

TalZa1810: We saw that a lot of the consumers use anyways `label` in their own `valueParser` so maybe exposing `label` prop in the builder (instead of `optionTitle`) and have a default valueParser of `option => option.label || option.value` can be a solution (and updating the optionPropTypes to have a `label` as well of course).

src/MultiSelectCheckbox/test/MultiSelectCheckbox.visual.js
shlomitc: I don't think you need them

src/DropdownLayout/DropdownLayout.js
```diff
@@ -501,6 +501,7 @@ const optionPropTypes = PropTypes.shape({
     .isRequired,
   disabled: PropTypes.bool,
   overrideStyle: PropTypes.bool,
+  label: PropTypes.string,
```
shlomitc, line 504: please add a test for it

src/InputWithOptions/InputWithOptions.js
```diff
@@ -14,7 +14,7 @@ import uniqueId from 'lodash/uniqueId';
 
 import Popover from '../Popover';
 
-export const DEFAULT_VALUE_PARSER = option => option.value;
+export const DEFAULT_VALUE_PARSER = option => option.label || option.value;
```
shlomitc, line 17: and add a test for this

TalZa1810: This is used within the `MuiltiSelectCheckbox` itself and not in `InputWithOptions`. Moreover, it is used in `Dropdown` (since it extends `InputWithOption`.  Therefore, I've made a different `valueParser` specifically for `MultiSelectCheckbox` in order to to affect `Dropdown`.

src/MultiSelectCheckbox/docs/index.story.js
```diff
@@ -77,9 +76,16 @@ export default {
...
-          code({
+          example({
+            title: 'Simple usage',
+            text: 'Simple usage of object options.',
             source: examples.simple,
-            compact: false,
+          }),
+
+          example({
+            title: 'Advanced usage',
```
shlomitc, line 86: That's not an "Advanced Usage" but simple as it is - "Using builders".
And the description can be - "Use the Multi-selection dropdown behavior using the predefined list items.

In the future, when we create a well-documented page for the list items, this could be a good place to point to that doc

src/MultiSelectCheckbox/test/MultiSelectCheckbox.spec.js
```diff
@@ -150,20 +150,82 @@ describe('multiSelectCheckbox', () => {
       );
     });
 
-    it('should use provided valueParser that will enable handling option with a component in value', async () => {
-      const specialOption = options.find(x => typeof x.value !== 'string');
-      const selectedOptions = [specialOption.id];
-      const valueParser = option =>
-        typeof option.value === 'string' ? option.value : option.label;
+    describe('valueParser', () => {
+      it('should use provided valueParser when given', async () => {
```
shlomitc, line 154: Let's split this test into two:
1. it('should use the default valueParser') - this test will make sure that `value` is parsed.
2. it('should use the default valueParser for options contains `label` property) - this will make sure that `label` is parsed

src/MultiSelectCheckbox/test/MultiSelectCheckbox.spec.js
```diff
@@ -150,20 +150,82 @@ describe('multiSelectCheckbox', () => {
...
+
+        const { driver, inputDriver } = createDriver(
+          <MultiSelectCheckbox
+            valueParser={valueParser}
+            options={options}
+            selectedOptions={selectedOptions}
+          />,
+        );
+        expect(await driver.getLabelAt(0)).toBe(specialOption.title);
+        expect(await inputDriver.getValue()).toBe(specialOption.title);
```
shlomitc, line 173: Why is both `getLabelAt(0)` and `getValue()` is needed? looks to me like one of them is redundant. probably the second one

src/MultiSelectCheckbox/test/MultiSelectCheckbox.spec.js
```diff
@@ -150,20 +150,82 @@ describe('multiSelectCheckbox', () => {
...
+        expect(await driver.getLabelAt(0)).toBe(options[0].label);
+        expect(await driver.getLabelAt(1)).toBe(options[1].value);
+
+        expect(await inputDriver.getValue()).toBe(
+          `${options[0].label}, ${options[1].value}`,
+        );
+      });
+    });
+
+    describe('Builder Options', () => {
```
shlomitc, line 205: I wouldn't test it against a builder but rather against the new protocol, meaning `value` which is a function.

This makes sure the implementation is independent of any builder and supports the render method of an option

it should be very simple. It doesn't even have to have all the rendering options of the ListItemSelect.

Think that this test could have been written before any builder was created

src/Input/Input.js
```diff
@@ -516,7 +516,7 @@ Input.propTypes = {
   /** Placement of status tooltips */
   tooltipPlacement: PropTypes.string,
 
-  /** Specifies the type of <input> element to display.default is text. */
+  /** Specifies the type of `<input/>` element to display.default is text. */
```
shlomitc, line 519: I don't think it needs to be mentioned as a component, it's referring an input field

src/DropdownLayout/DropdownLayout.js
shlomitc: still missing a test here

src/MultiSelectCheckbox/test/MultiSelectCheckbox.spec.js
TalZa1810: The second one is referring to the value parser because it takes the value from the input. so I think that the first one is the one redundant.

src/Input/Input.js
TalZa1810: It renders an input in the documentation:
https://www.wix.com/pages/wix-style-react/?activeTab=API&path=%2Fstory%2Fcomponents-api-components--input
I thought it was by mistake.

src/MultiSelectCheckbox/test/MultiSelectCheckbox.spec.js
shlomitc: From what I see in the driver - the first one encapsulates the second

src/DropdownLayout/DropdownLayout.js
TalZa1810: I've added an if statement to the `optionValidator`. is this what you were referring?

src/MultiSelectCheckbox/test/MultiSelectCheckbox.spec.js
```diff
@@ -150,20 +149,84 @@ describe('multiSelectCheckbox', () => {
...
+            valueParser={valueParser}
+            options={options}
+            selectedOptions={selectedOptions}
+          />,
+        );
+        expect(await driver.getLabelAt(0)).toBe(specialOption.title);
+      });
+    });
+
+    it('should allow using functions as options', async () => {
```
TalZa1810, line 211: it this what you were referring?

src/DropdownLayout/DropdownLayout.js
shlomitc: 1. it's missing in the prop types: https://github.com/wix/wix-style-react/blob/9ed9ce86027fcd5c21ffe4e66330ce4ea9a80e13/src/DropdownLayout/DropdownLayout.js#L498
2. add a test for the `DropdownLayout` that gets and item with a label and when clicking, the value is received in the callback https://github.com/wix/wix-style-react/blob/9ed9ce86027fcd5c21ffe4e66330ce4ea9a80e13/src/DropdownLayout/DropdownLayout.js#L420

src/MultiSelectCheckbox/docs/index.story.js
shlomitc: again, it's not "advanced"

src/MultiSelectCheckbox/test/MultiSelectCheckbox.spec.js
shlomitc: yes but the expectation should be testing the selection logic, meaning that you should have multiple actions clicked and each one's render function should be called with the "selected" being true.
https://github.com/wix/wix-style-react/pull/5778/files/eb368bc421f18f1a4c9f2ffaa5f0cdf52ce9ec1b#diff-fa476392e31f6bf3c05eb58029ab1267R19

src/MultiSelectCheckbox/docs/index.story.js
TalZa1810: I've changed this already: 
https://github.com/wix/wix-style-react/blob/eb368bc421f18f1a4c9f2ffaa5f0cdf52ce9ec1b/src/MultiSelectCheckbox/docs/index.story.js#L86

Or you refer also to the "Simple usage"?

shlomitc: I don't see the changes reflected in this PR. maybe you didn't push it yet?

src/MultiSelectCheckbox/test/MultiSelectCheckbox.spec.js
shlomitc: I see you're still using builders here - let me know if I need to reexplain myself here

src/MultiSelectCheckbox/docs/index.story.js
TalZa1810: Check out here: https://github.com/wix/wix-style-react/pull/5778/files

src/MultiSelectCheckbox/test/MultiSelectCheckbox.spec.js
TalZa1810: I've refactored this: 
https://github.com/wix/wix-style-react/blob/07b4809da663ab0f827402c9ce52bddeb69a4240/src/MultiSelectCheckbox/test/MultiSelectCheckbox.spec.js#L211

PR #5782, opened 2020-08-02
<TestimonialList/> - new component

sivanharel:
### 🔦 Summary

...

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5782/

shlomitc:
@milkyfruit sounds good from my side. I would also suggest splitting this from the Zeplin spec to a separate `IC`.
I would also add examples of what happens when some properties are missing.

@sivanharel - regarding the API, I think that `authorName` is too opinionated, maybe it should be more of a generic kind of name?

sivanharel:
It sounds like a good idea to change the name of this component.
I know that we talked about it also in our meeting and especially if we will make it a generic component  I agree with Shlomi that it will better to split it into different Zeplin specs. 
And yes if it's a generic footer so I have to change the 'authorName' prop maybe it could be  'name'.

Regarding the other comments, I will change it after the Crash course :)

milkyfruit:
Great! Then let's rename it to `TestimonialList`.
`name` is perfect, because it matches with `<Avatar />`.

Meanwhile I'll update the specs.

sivanharel:
@milkyfruit - I updated the PR.
Do you want me to keep this example as well or to keep just the Columns and Customization examples? (Structure example looks the same as the Columns example but it has a very informative description.)

**Structure**
Component lists horizontally any number of testimonials. Each testimonial has text, author name, and avatar. Avatar should be 60px size.

src/TestimonialList/TestimonialList.js
```diff
@@ -0,0 +1,100 @@
...
+            />
+          );
+        })}
+      </div>
+    );
+  }
+}
+
+const TestimonialItem = ({ index, avatar, text, authorName }) => (
+  <div className={classes.testimonialItem} data-hook="testimonial">
```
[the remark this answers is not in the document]
sivanharel: 1. I use this 'dataHook ' in order to calculate the number of testimonials that were rendered and to check it.
const getTestimonials = () =>
    base.$$(`[data-hook="${dataHooks.testimonial}"]`);

2. I agree and changed it.

src/TestimonialList/TestimonialList.js
```diff
@@ -0,0 +1,100 @@
...
+/** TestimonialList is a group of layouts that display avatar, description and name. It's used in a footer of a marketing page layout. */
+class TestimonialList extends React.PureComponent {
+  render() {
+    const { className, dataHook, testimonials, cols } = this.props;
+
+    return (
+      <div
+        className={st(classes.root, className)}
+        data-hook={dataHook}
+        style={{ gridTemplateColumns: `repeat(${cols}, 1fr)` }}
```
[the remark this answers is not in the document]
sivanharel: Good to know :)
I chose to use the Layout here but also learned how to use CSS variables during runtime :)

src/TestimonialList/TestimonialList.js
```diff
@@ -0,0 +1,101 @@
...
+
+    return (
+      <Layout
+        className={st(classes.root, className)}
+        data-hook={dataHook}
+        cols={cols}
+      >
+        {testimonials.map((testimonialItem, index) => {
+          return (
+            <TestimonialItem
```
shlomitc, line 23: if you're using `Layout`, each item should be wrapped with a `Cell`

src/TestimonialList/TestimonialList.js
```diff
@@ -0,0 +1,101 @@
...
+import PropTypes from 'prop-types';
+
+import Text from '../Text';
+import { Layout } from '../Layout';
+
+import { st, classes } from './TestimonialList.st.css';
+import { dataHooks } from './constants';
+
+/** TestimonialList is a group of layouts that display avatar, description and name. It's used in a footer of a marketing page layout. */
+class TestimonialList extends React.PureComponent {
```
shlomitc, line 11: I think it's no longer a `PureComponent` if you have object which are props

src/TestimonialList/TestimonialList.js
```diff
@@ -0,0 +1,101 @@
...
+/** TestimonialList is a group of layouts that display avatar, description and name. It's used in a footer of a marketing page layout. */
+class TestimonialList extends React.Component {
+  render() {
+    const { className, dataHook, testimonials, cols } = this.props;
+
+    return (
+      <div
+        className={st(classes.root, className)}
+        data-hook={dataHook}
+        // Using CSS variable
```
shlomitc, line 18: you can remove this comment

Issue #5781, opened 2020-08-02
<PopoverMenu.Divider /> changes background on hover


marik22312:
# 🐛 Bug Report

### 🏗 Relevant Components

`<PopoverMenu.Divider />`

### 😯 Current Behavior

When hovering the divider, it's background color is changed
### 🤔 Expected Behavior

Dividers should be interactable and shouldn't change background color 

### 👣 Steps to Reproduce

https://www.wix.com/pages/wix-style-react/?path=/story/introduction-playground--playground&snippet=7db731b6-2324-412d-a2b7-074e81670474

  - **WSR version**: `8.35`

### 👀 Severity

- Low

EgoziE:
There's also an accessibility issue here, can't use arrows to go over items.

PR #5810, opened 2020-08-09
<PopoverMenu/> - change divider to ListItemSection internally

EgoziE:
Fix accessibility issues with dividers inside PopoverMenu


EgoziE:
Resolves https://github.com/wix/wix-style-react/issues/5781

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5810/

shlomitc:
It works better now but it's still focusing on the element when navigating with the keyboard.
In `DropdownLayout`, the way to point that an item is not selectable (or navigatable) is one of the following:
1. It is defined explicitly as a [DropdownLayout Divider](https://github.com/wix/wix-style-react/blob/master/src/DropdownLayout/DropdownLayout.js#L334-L336)
2. it is a [disabled item](https://github.com/wix/wix-style-react/blob/master/src/DropdownLayout/DropdownLayout.js#L494)
3. it has a [`title` property](https://github.com/wix/wix-style-react/blob/master/src/DropdownLayout/DropdownLayout.js#L494) on the object

The first option is probably not good in this case, but both the second and third will work. `title` was created specifically for this use-case if I remember correctly


Also - I think that we shouldn't a divider to each one of the examples, it's better just to create a separate example that showcases the divider scenario

EgoziE:
Yes, I took that in mind, it's a global problem in all dropdown family components, I think I'll dedicate a different PR for this issue.

Issue #5811, opened 2020-08-09
<RadioGroup/> puppeteer testkit exceptoion


alexandraseri:
# 🐛 Bug Report

### 🏗 Relevant Components

`<RadioGroup />` puppeteer testkit

...

### 😯 Current Behavior

When trying to use `<RadioGroup />` puppeteer testkit, an exception is thrown:

```
import { radioGroupTestkitFactory } from 'wix-style-react/dist/testkit/puppeteer';
​
//in test
const radioGroupDriver = await radioGroupTestkitFactory({ dataHook, page });
​
​
/// causes error
TypeError: driverFactory is not a function
```

### 🤔 Expected Behavior

`<RadioGroup />` testkit works just like all others

### 👣 Steps to Reproduce

create puppeteer Box testkit in e2e

- WSR version: 8.39.0
- Yoshi version: 4.139.0

### 👀 Severity

Low

shlomitc:
I think this is related to the fact that we export multiple functions exported from the driver. [This function export](https://github.com/wix/wix-style-react/blob/739b5d7f7569779923d482aa37aee3d7245c4f28/src/RadioGroup/RadioGroup.uni.driver.js#L5) is causing the pain
Our automated tools generating the `/dist/testkit/puppeteer'` gets confused and exports both methods instead of a single one

I console-logged this line and got `load(require('../src/RadioGroup/RadioButton/RadioButton.uni.driver'))` 

```js
{
  createRadioButtonsGetter: [Function: createRadioButtonsGetter],
  radioGroupUniDriverFactory: [Function: radioGroupUniDriverFactory]
}
```

After removing the exported `createRadioButtonsGetter`, this is the output
```js
Promise {
  {
    exists: [Function: exists],
    element: [Function: element],
    click: [Function: click],
    selectByValue: [Function: selectByValue],
    selectByIndex: [Function: selectByIndex],
    getRadioValueAt: [Function: getRadioValueAt],
    getRadioAtIndex: [Function: getRadioAtIndex],
    getSelectedValue: [Function: getSelectedValue],
    isRadioDisabled: [Function: isRadioDisabled],
    getNumberOfRadios: [Function: getNumberOfRadios]
  }
}
```

I think that simply extracting this method to some shared file will solve it without any massive refactors.
I'll try doing a quick fix and post here the results

shlomitc:
fixed in 8.41.0

PR #5814, opened 2020-08-09
<RadioGroup/> - fix incorrectly exported puppeteer uni-driver

shlomitc:
This PR comes to solve the following issue - https://github.com/wix/wix-style-react/issues/5811
tl;dr - in some cases when a driver exports more than one method, the automatic driver loader fails to define which file is the right one.

@argshook - this obviously solves the case but I think that at least having some sort of a meaningful error message in the `load` method could help us understand there are issues that need to be fixed. In this case, we didn't have any indication that the driver doesn't work


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5814/

argshook:
thanks @shlomitc, let's keep track of it here https://github.com/wix/wix-ui/issues/1898

PR #5823, opened 2020-08-11
`<MultiSelectCheckbox/>`- use builders instead of manually using `<ListItemSelect/>` and `<ListItemSection/>`

TalZa1810:
### 🔦 Summary
Jira: https://jira.wixpress.com/browse/DSM-744

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [ ] Visual test


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5823/

src/MultiSelectCheckbox/MultiSelectCheckbox.js
```diff
@@ -3,51 +3,47 @@ import PropTypes from 'prop-types';
...
+            checkbox: true,
+            title: option.value,
+          });
+          return this._renderOption(builder);
         }
       }
     });
   }
 
+  _renderOption(option) {
```
TalZa1810, line 34: I'm not sure about the naming, since a divider is also sort of an option. Let me know if you can think of a better naming.

src/MultiSelectCheckbox/test/MultiSelectCheckbox.spec.js
```diff
@@ -24,7 +24,7 @@ describe('multiSelectCheckbox', () => {
     const options = [
       { value: 'Alabama', id: 'Alabama-1' },
       { value: 'Alaska', id: 'Alaska' },
-      { value: <div>Arkansas</div>, id: 'Arkansas', label: 'Arkansan Label' },
+      { value: <div>Arkansas</div>, id: 'Arkansas' },
```
TalZa1810, line 27: This should have been removed with the previous PR, since we're using now `label` for the `listItemSelectBuilder`. https://github.com/wix/wix-style-react/pull/5778

src/DropdownLayout/DropdownLayout.js
```diff
@@ -498,7 +498,7 @@ const optionPropTypes = PropTypes.shape({
     .isRequired,
   disabled: PropTypes.bool,
   overrideStyle: PropTypes.bool,
-  label: PropTypes.string,
+  label: PropTypes.node,
```
EgoziE, line 501: Add to types & tests

src/MultiSelectCheckbox/MultiSelectCheckbox.js
```diff
@@ -3,51 +3,47 @@ import PropTypes from 'prop-types';
...
   createOptions(options) {
     return options.map(option => {
       if (typeof option.value === 'function') {
-        const value = option.value;
-        return {
-          ...option,
-          value: props =>
-            value({ ...props, selected: this.isSelectedId(option.id) }),
-        };
+        return this._renderOption(option);
```
EgoziE, line 15: rearrange maybe

src/DropdownLayout/DropdownLayout.js
shlomitc, line 501: why it became a node? isn't the purpose of the label is to simply be a textual representation?
eventually, this is what will appear as the input's value in Dropdowns and similars

argshook: `node` is also a `string`, so it's ok. `node` is also good when users want to use a component that eventually renders just text, but in the middle it also translates or sets some feature toggle

shlomitc: In this case it should be strictly a string, I don’t see a case when an object is passed. The translations function should also return a string here. An object for example will not work. It will be stringified to [object Object] in the input field

TalZa1810: I've changed this to node because `title` can be a node and we return the `title` when we use the builder:
https://github.com/wix/wix-style-react/blob/master/src/ListItemSelect/ListItemSelect.js#L207 
https://github.com/wix/wix-style-react/blob/master/src/ListItemSelect/ListItemSelect.js#L34

When I changed this to used the builder, there was a test that was failing because of this change.

TalZa1810: A value can be a [node](https://github.com/wix/wix-style-react/blob/bfd3f3ba7268ae459b6ebe21e3354f106f5b1753/src/MultiSelectCheckbox/test/MultiSelectCheckbox.spec.js#L27), and when we use it within the builder [here](https://github.com/wix/wix-style-react/blob/bfd3f3ba7268ae459b6ebe21e3354f106f5b1753/src/MultiSelectCheckbox/MultiSelectCheckbox.js#L30) it causes problems. If `title` should be strictly `string`, we should update the tests and PropTypes.

src/MultiSelectCheckbox/docs/examples.js
```diff
@@ -75,15 +75,17 @@ class BuildersExample extends React.Component {
             checkbox: true,
             id: 'Logan Chandler',
             title: 'Logan Chandler',
+            label: 'Logan Chandler Label',
         }),
-       { value: 'Paul Simon', id: 'Paul Simon' },
+       { value: <div>Paul Simon</div>, id: 'Paul Simon', label: 'Paul Simon Label' },
```
shlomitc, line 80: `div` is not needed here

src/MultiSelectCheckbox/MultiSelectCheckbox.js
```diff
@@ -3,51 +3,54 @@ import PropTypes from 'prop-types';
...
-                suffix={option.suffix}
-                ellipsis={option.ellipsis}
-                onClick={e => e.preventDefault()} // This is prevented because there's an event listener wrapping the option
-              />
-            ),
-          };
+            checkbox: true,
+            /* the node displayed within the DropdownLayout*/
+            title: option.value,
+            /* the string displayed within the input when the option is selected */
```
shlomitc, line 32: This is the API of an `option` in DropdownLayout`, so these comments should be set there

src/MultiSelectCheckbox/MultiSelectCheckbox.js
```diff
@@ -3,51 +3,54 @@ import PropTypes from 'prop-types';
...
 import styles from './MultiSelectCheckbox.scss';
-import ListItemSelect from '../ListItemSelect';
-import ListItemSection from '../ListItemSection';
+import { listItemSelectBuilder } from '../ListItemSelect';
+import { listItemSectionBuilder } from '../ListItemSection';
 
 const OPEN_DROPDOWN_CHARS = ['Enter', 'ArrowDown', 'Space', ' '];
 
 class MultiSelectCheckbox extends InputWithOptions {
+  /** An option value can be either a function or an object:
```
shlomitc, line 12: I think you can remove these comments by changing it to a more declarative code.

the comment regarding the "select mechanism" should go down to the `_renderFunctionOption`, explaining there why it is done

src/MultiSelectCheckbox/MultiSelectCheckbox.js
```diff
@@ -3,51 +3,54 @@ import PropTypes from 'prop-types';
...
       if (typeof option.value === 'function') {
-        const value = option.value;
-        return {
-          ...option,
-          value: props =>
-            value({ ...props, selected: this.isSelectedId(option.id) }),
-        };
+        return this._renderFunctionOption(option);
       } else {
         if (option.value === '-') {
```
shlomitc, line 21:
```suggestion
      } else if (option.value === '-') {
```

src/MultiSelectCheckbox/MultiSelectCheckbox.js
```diff
@@ -3,51 +3,54 @@ import PropTypes from 'prop-types';
...
 const OPEN_DROPDOWN_CHARS = ['Enter', 'ArrowDown', 'Space', ' '];
 
 class MultiSelectCheckbox extends InputWithOptions {
+  /** An option value can be either a function or an object:
+   * If it's a function, there's a need to override it in order to support the select mechanism within the MultiSelectCheckbox itself.
+   * If it's a string, we check which string in order to know which builder to use. We use builders instead of manually override the styles of the DropdownLayout.
+   * */
   createOptions(options) {
     return options.map(option => {
       if (typeof option.value === 'function') {
```
shlomitc, line 18: I think that instead of adding the comments above, you can make it a bit more declarative:
```js
if(isUsingCustomRenderFunction(option))  {
 ...
} else if(isDivider(option)) {
 ...
} else {
 ...
}
```

src/MultiSelectCheckbox/MultiSelectCheckbox.js
```diff
@@ -3,51 +3,54 @@ import PropTypes from 'prop-types';
...
+            /* the string displayed within the input when the option is selected */
+            label: option.label,
+          });
+          return this._renderFunctionOption(builder);
         }
       }
     });
   }
 
+  _renderFunctionOption(option) {
```
shlomitc, line 41:
```suggestion
  _patchOptionWithSelectionMechanism(option) {
```

src/MultiSelectCheckbox/docs/examples.js
TalZa1810: I just wanted to demonstrate that we can have a node as a value, but I guess that this can be removed.

TalZa1810:
It's a flaky test with the `MessageModalLayout`. Not related to the change. 
I'm merging this.

PR #5831, opened 2020-08-13
<ModalSelectorLayout/> - use new <CustomModalLayout/>

bamai:
### 🔦 Summary
https://jira.wixpress.com/browse/DSM-613
...

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5831/

src/ModalSelectorLayout/ModalSelectorLayout.js
```diff
@@ -205,16 +211,32 @@ export default class ModalSelectorLayout extends React.PureComponent {
...
       shouldShowNoResultsFoundState,
+      selectedItems,
     } = this.state;
 
+    const enabledItems = this._getEnabledItems(selectedItems);
+
     return (
-      <div
+      <CustomModalLayout
         data-hook={dataHook}
```
shlomitc, line 221: `dataHook` probably

src/ModalSelectorLayout/ModalSelectorLayout.js
```diff
@@ -205,16 +211,32 @@ export default class ModalSelectorLayout extends React.PureComponent {
...
       searchValue,
       shouldShowNoResultsFoundState,
+      selectedItems,
     } = this.state;
 
+    const enabledItems = this._getEnabledItems(selectedItems);
+
     return (
-      <div
+      <CustomModalLayout
```
shlomitc, line 220: Looking in the cheatsheet, previously the close button appeared
image

While now it isn't
image

can you check this out, please?

shlomitc, line 220: The initial rendering of the modal is broken
image

in oppose to the current one
image

lironch: To show the close button in `CustomModalLayout` you need to pass a `onCloseButtonClicked` handler

src/ModalSelectorLayout/ModalSelectorLayout.js
```diff
@@ -205,16 +211,32 @@ export default class ModalSelectorLayout extends React.PureComponent {
...
         data-hook={dataHook}
-        className={css.modalContent}
-        style={{ height, maxHeight }}
+        showHeaderDivider
```
lironch, line 222: You probably don't need this unless it was requested specifically by Ayelet.
The header/footer dividers will show automatically according to your scroll position.

src/ModalSelectorLayout/ModalSelectorLayout.js
```diff
@@ -205,16 +211,32 @@ export default class ModalSelectorLayout extends React.PureComponent {
...
+        style={{ height, maxHeight, width: '600px' }}
```
lironch, line 223: Please avoid adding styles this way for now, when we will understand this is a must we will rethink what is the best way to do so.

src/ModalSelectorLayout/ModalSelectorLayout.js
```diff
@@ -205,16 +211,32 @@ export default class ModalSelectorLayout extends React.PureComponent {
...
+        title={title}
+        onCloseButtonClick={onClose}
```
lironch, line 225: I'm not sure why this component need to expose both `onClose` and `onCancel` which represent the exact same use case for the component user.
@shlomitc i would consider deprecating one of them as part of this change

src/ModalSelectorLayout/ModalSelectorLayout.js
```diff
@@ -205,16 +211,32 @@ export default class ModalSelectorLayout extends React.PureComponent {
...
+        primaryButtonOnClick={() =>
+          onOk(multiple ? enabledItems : enabledItems[0])
+        }
+        secondaryButtonText={cancelButtonText}
+        primaryButtonText={okButtonText}
+        primaryButtonProps={{
+          disabled: disableConfirmation || !selectedItems.length,
+        }}
+        sideActions={
+          sideActions ? sideActions : multiple && this._renderFooterSelector()
```
lironch, line 236: Do we really have a good product justification to expose `sideActions` to the users of this component?
It means that in case they choose to use it you will probably lose the `select-all` behaviour of this component

src/ModalSelectorLayout/ModalSelectorLayout.js
```diff
@@ -33,7 +32,7 @@ export default class ModalSelectorLayout extends React.PureComponent {
     title: PropTypes.node,
 
     /** Fixed text displayed above the list */
-    subtitle: PropTypes.node,
+    subtitle: PropTypes.string,
```
shlomitc, line 35: are you sure this is a required change? we usually allow passing nodes because of translation functions that return a node

src/ModalSelectorLayout/ModalSelectorLayout.js
shlomitc: @bamai - at least in the storybook preview - the cheatsheet is still not displaying the x button, please check it
https://www.wix.com/pages/wix-style-react.pr_5831/?path=/story/introduction-cheatsheet--components-cheatsheet

src/ModalSelectorLayout/ModalSelectorLayout.js
```diff
@@ -205,82 +211,87 @@ export default class ModalSelectorLayout extends React.PureComponent {
...
-                <Search
-                  dataHook={dataHooks.search}
-                  placeholder={searchPlaceholder}
-                  onChange={this._onSearchChange}
-                  onClear={this._onClear}
-                  debounceMs={searchDebounceMs}
-                  value={searchValue}
-                />
-              </div>
+              <Search
```
shlomitc, line 246: The search should stay sticky once scrolling the modal content, see how it cuts the content:
image

After your change, the Search is scrolled with the content which isn't good

src/ModalSelectorLayout/ModalSelectorLayout.js
shlomitc: @bamai - what's the decision here?

src/ModalSelectorLayout/ModalSelectorLayout.js
shlomitc: You're probably right, but this widget is quite old and some wrong decisions were taken along the way.
Anyway, that's not urgent, so let's skip this for now

src/ModalSelectorLayout/ModalSelectorLayout.js
shlomitc: same as before - let's not change the current product, it's an old component...

shlomitc (changes requested):

src/ModalSelectorLayout/ModalSelectorLayout.js
shlomitc: currently, we decided to not apply any borders due to search sticky position limitations.
We will need to extend the ScrollContainer to support an additional `fixedHeader` in which the `Search` component will be placed. It will not be part of the scroll, but the divider will appear above it. We'll do it as a separate task

src/ModalSelectorLayout/ModalSelectorLayout.scss
```diff
@@ -58,6 +47,7 @@
   align-items: center;
 }
 
-.subtitleWrapper + .searchWrapper {
-  margin-top: 18px;
+.noResultsFoundStateWrapper {
+  width: 100%;
+  height: calc(100% - #{$search-height});
```
shlomitc, line 52: not sure if this should actually be done with `flex` instead of `calc`

shlomitc (approved):

PR #5834, opened 2020-08-13
Update Readme

EgoziE:


README.md
```diff
@@ -35,11 +35,14 @@ yarn add wix-style-react
 
 ## 🔨 Setup
 
-If 🐉 yoshi build runner is used for your application then all required loaders are already defined. If you'd rather set up your project manually, take a look at our [webpack guide](https://github.com/wix/wix-style-react/blob/master/docs/usage-without-yoshi.md).
+We recommend using a **Stylable compatible** project template in order to save some configurations.
 
-Minimal Yoshi version to be used with wix-style-react:
+- [Yoshi](https://github.com/wix/wix-style-react/blob/master/docs/usage/usage-with-yoshi.md) - A Galaxy of toolkits to develop applications at Wix.
```
argshook, line 40: yoshi is going private, i don't think we should advertise it any longer.
in near future we will drop using it ourselves

docs/usage/usage-with-yoshi.md
```diff
@@ -0,0 +1,9 @@
+# Usage with Yoshi
+
+Yoshi already holds all needed configurations in order to use wix-style-react.
```
argshook, line 3: same as previous comment, i think we no longer need such document. It should be just usage in general, without any info about yoshi

README.md
EgoziE: I don't mind deleting this, @shlomitc what do you think?

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5834/

README.md
shlomitc: yes, we can probably remove it

README.md
```diff
@@ -35,11 +35,14 @@ yarn add wix-style-react
...
+- [Create Stylable App](https://github.com/wix/wix-style-react/blob/master/docs/usage/usage-with-create-stylable-app.md) - Create a new Stylable project from a boilerplate.
 
-`Version 4 - ^4.1.0`
+If you decide not to use a Stylable compatible project template, please read [here](https://stylable.io/docs/getting-started/install-configure) how to add Stylable loader manually.
```
shlomitc, line 43: I think we can simply merge the two bullets into a single one, saying something like:
"`wix-style-react` is powered by `Stylable`, follow the installation setup instructions to see how to integrate it".

Inside the "usage-with-create-stylable-app" add a first paragraph that says that we recommend using the `create-stylable-app` when using `create-react-app` but you can use any custom build tool following the stylable installation guide

docs/usage/usage-with-create-stylable-app.md
```diff
@@ -0,0 +1,94 @@
...
+```bash
+npm install wix-style-react --save
+```
+
+3. Install necessary loaders:
+```bash
+npm install url-loader style-loader css-loader node-sass resolve-url-loader sass-loader --save-dev
+```
+
+4. Use loaders in `webpack.config.js` file:
```
shlomitc, line 21: Do I really need to do this even when using `create-stylable-app`? that's really strange, I'd expect integration to be seamless.

EgoziE: Unfortunately yes, scss is not recognized at all, I also talked with Stylable team about it, they objected with the explanation that projects that use Stylable don't need to use scss as well.

shlomitc: So is the create-stylable-app is needed? Why not simply use create-react-app?

EgoziE: Because then you have to eject it and it gets the project messy and the guide complicated.
And if you must eject the project it makes you ask, why using create-react-app at all?

Bottom line, create-stylable-app was the easiest and quickest way I found to get a working project with wsr.

PR #5839, opened 2020-08-16
<Table/> - use TagList in Table SubToolbar example

gendeld:
### 🔦 Summary

...

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5839/

TalZa1810:
I think that we should make the `TagList` interactive. I mean removing and adding tags should filter the table data. Also, it seems like there's no need for both `TagsList` and `Dropdown` for filtering data. I would consult a UX to see if this is how they want this to look like or they want this to be a separate example: https://zpl.io/bo8Kl7v

shlomitc:
@TalZa1810 - The Table examples are quite complex already so I think that we need to aim to keep them as minimal as possible.
We can create a `2.3 Table Layout` story page for common use cases and real-life examples. @milkyfruit - wdyt about defining such a page with common examples?

Anyway, I think we can merge this as-it-is, but let's fix the description
image

It should be something more like:
A Table can contain a sticky sub-toolbar, an area for additional actions such as displaying a tag list of filtered items

PR #5855, opened 2020-08-18
<TextButton/> - fix Text child component not inheriting button's color

gendeld:
When a `<Text/>` component is passed as `<TextButtons/>`'s child, it does not inherit the text color.

As can be seen in the Ellipsis example here:

https://www.wix.com/pages/wix-style-react/?path=/story/components-api-components--textbutton

Before:
image
After:
image


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5855/

shlomitc:
I don't really like this solution. it makes that TextButton intimately knows how about Text...
If this is a common use case - I think that `Text` should get a prop that explicitly states it doesn't want the default colors but rather inherit it

PR #5885, opened 2020-08-25
`<Ellipsis/>` - adding "maxLines" prop

TalZa1810:
### 🔦 Summary
Adding `maxLines` prop to `<Ellipsis/>`. 

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [x] Visual test


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5885/

src/common/Ellipsis/Ellipsis.js
```diff
@@ -48,23 +51,38 @@ class Ellipsis extends React.PureComponent {
...
   _updateEllipsisState = () => {
     const { ellipsis } = this.props;
     const { isActive } = this.state;
-    const { current: textElement } = this.ref;
     const shouldBeActive =
-      ellipsis &&
-      textElement &&
-      (textElement.scrollWidth - textElement.parentNode.offsetWidth > 1 ||
-        textElement.offsetWidth < textElement.scrollWidth);
+      ellipsis && (this._widthCheck() || this._HeightCheck());
```
shlomitc, line 61: is it really necessary to check both for all cases?

src/Text/Text.js
```diff
@@ -12,11 +12,12 @@ const TextWithEllipsis = ({ className, ...props }) => {
         size: props.size,
         weight: props.weight,
       })}
-      render={({ ref, ellipsisClasses }) => (
+      render={({ ref, ellipsisClasses, style }) => (
         <RawText
           {...componentProps}
           ref={ref}
           className={ellipsisClasses(className)}
+          style={style}
```
shlomitc, line 20: why is this needed?

src/common/Ellipsis/Ellipsis.js
```diff
@@ -48,23 +51,38 @@ class Ellipsis extends React.PureComponent {
...
 
     if (shouldBeActive !== isActive)
       this.setState({ isActive: shouldBeActive });
   };
 
+  _widthCheck = () => {
```
shlomitc, line 67: give this function a name that describes what it does

src/common/Ellipsis/Ellipsis.js
```diff
@@ -77,12 +95,19 @@ class Ellipsis extends React.PureComponent {
...
   _renderText = () => {
-    const { ellipsis, render } = this.props;
+    const { ellipsis, render, lineClamp } = this.props;
 
     return render({
       ref: this.ref,
+      ...(lineClamp && { style: { WebkitLineClamp: lineClamp } }),
       ellipsisClasses: (...classNames) =>
-        [ellipsis && classes.text, ...classNames].filter(Boolean).join(' '),
+        [
```
shlomitc, line 104: That's not very readable, don't you think?

src/common/Ellipsis/Ellipsis.st.css
```diff
@@ -13,6 +13,12 @@
   vertical-align: bottom !important;
 }
 
+.lineClamp {
+  white-space: normal !important;
+  display: -webkit-box !important;
+  -webkit-box-orient: vertical;
```
shlomitc, line 19: is this supported by all browsers?

shlomitc (changes requested):

src/common/Ellipsis/Ellipsis.st.css
TalZa1810: It is [not supported ](https://www.w3.org/TR/css-overflow-3/#propdef--webkit-line-clamp) in all browsers.  I've read [this article](https://medium.com/mofed/css-line-clamp-the-good-the-bad-and-the-straight-up-broken-865413f16e5) regarding the `line-clamp` css property and saw [this solution](https://codepen.io/nilsynils/pen/KgaRBo?editors=1100
).  
WDYT about this solution for unsupported browsers? 
There is another [suggested solution](https://codepen.io/nilsynils/pen/zKNpkm?editors=1111).

shlomitc: What they suggest in the article is hiding some content in case line-clamp is not supported.
image

I guess that this can be considered a good enough fallback as long as the tooltip will appear

src/Text/Text.js
TalZa1810: It is needed since we don't know in advance how many lines the user wants to have. Therefore, it has to be inline styles.

src/common/Ellipsis/Ellipsis.js
```diff
@@ -77,15 +101,27 @@ class Ellipsis extends React.PureComponent {
...
+    const inlineStyle = {
+      style: { WebkitLineClamp: maxLines, MozLineClamp: maxLines },
+    };
 
     return render({
       ref: this.ref,
-      ellipsisClasses: (...classNames) =>
-        [ellipsis && classes.text, ...classNames].filter(Boolean).join(' '),
+      ellipsisClasses: this._getEllipsisClasses(ellipsis, maxLines),
+      ...(ellipsis && maxLines && inlineStyle),
```
shlomitc, line 112: 1. from this, it is unclear that you are passing the `style` entry. I need to go to `inlineStyle` only to understand there that you create the `style` entry.

2. In order to prefer confusion, I think it's better to give it a more indicative name. same as we have `ellipsisClasses`, use `ellipsisInilneStyle`. then the consumer of it understand what he attaches to his component

3. instead of inline styles, you can simply use css class + css variable. this means that the `style` object will contain the variable of `--max-lines` and you will use it in the `.maxLines` class

src/Text/Text.js
shlomitc: so doesn't `RawText` should now open the `style` prop?

shlomitc (changes requested):

src/Text/index.d.ts
```diff
@@ -36,6 +36,7 @@ export type TextPropsBase = EllipsisCommonProps & {
   dataHook?: string;
   tagName?: string;
   className?: string;
+  style?: React.CSSProperties;
```
shlomitc, line 39: I'm actually not sure this is the right place since this is exported out as part of the `<Text/>` component, while your change is kind of internal as it stays in `RawText`.

src/common/Ellipsis/Ellipsis.js
```diff
@@ -77,15 +100,24 @@ class Ellipsis extends React.PureComponent {
...
     return render({
       ref: this.ref,
-      ellipsisClasses: (...classNames) =>
-        [ellipsis && classes.text, ...classNames].filter(Boolean).join(' '),
+      ellipsisClasses: this._getEllipsisClasses(ellipsis, maxLines),
+      ellipsisInlineStyle: { [vars.maxLines]: maxLines },
     });
   };
 
+  _getEllipsisClasses = (ellipsis, maxLines) => (...classNames) =>
```
shlomitc, line 112: I really think using `classNames()` will make this much more readable and declerative

src/common/Ellipsis/Ellipsis.st.css
```diff
@@ -13,6 +13,17 @@
   vertical-align: bottom !important;
 }
 
+.maxLines {
```
shlomitc, line 16: to keep it declerative - `.multiLine`

src/common/Ellipsis/Ellipsis.js
```diff
@@ -77,15 +100,24 @@ class Ellipsis extends React.PureComponent {
...
+    [
+      ellipsis && classes.text,
+      maxLines > 1 && ellipsis && classes.maxLines,
```
shlomitc, line 115: does it make sense to use both `text` and `maxLine`? aren't they overriding each other?

shlomitc (changes requested):

src/common/Ellipsis/Ellipsis.js
TalZa1810: It just overrides `white-space` & `display`. I've removed the default prop of `maxLines = 1` because it broke some of the examples. That's why I check maxLines is bigger then 1.

src/Text/index.d.ts
TalZa1810: yep, I was confused because `RawText` doesn't have a specific `d.ts` file.

src/common/Ellipsis/Ellipsis.js
```diff
@@ -77,15 +100,27 @@ class Ellipsis extends React.PureComponent {
...
     return render({
       ref: this.ref,
-      ellipsisClasses: (...classNames) =>
-        [ellipsis && classes.text, ...classNames].filter(Boolean).join(' '),
+      ellipsisClasses: this._classNames(),
+      ellipsisInlineStyle: { [vars.maxLines]: maxLines },
     });
   };
 
+  _classNames = () => (...classNames) => {
```
shlomitc, line 112: declarative function name, please :)

src/common/Ellipsis/Ellipsis.js
shlomitc: so let's do a small refactor.
one class that sets the common ellipsis styling (overflowing and such), another class that does single line ellipsis, and one last the does multiple lines

TalZa1810: I've used stylable for this. not sure about the name of the main class name though.

src/common/Ellipsis/Ellipsis.js
```diff
@@ -77,15 +100,30 @@ class Ellipsis extends React.PureComponent {
...
+
+    return className =>
+      st(
+        classes.ellipsis,
+        {
+          text: ellipsis,
+          singleLine: ellipsis,
+          multiline: maxLines > 1 && ellipsis,
+        },
+        className,
```
TalZa1810, line 123: It seems that all components that use ellipsis: `Text`, `Caption` and `Heading`, it gets `className` and not array of classNames so I used stylable for this case and rename the variable.

src/common/Ellipsis/Ellipsis.js
```diff
@@ -77,15 +100,30 @@ class Ellipsis extends React.PureComponent {
...
+  _getEllipsisClasses() {
+    const { ellipsis, maxLines } = this.props;
+
+    return className =>
+      st(
+        classes.ellipsis,
+        {
+          text: ellipsis,
+          singleLine: ellipsis,
+          multiline: maxLines > 1 && ellipsis,
```
shlomitc, line 121: this should be a single state since they can not live together

shlomitc (approved):
👍

PR #5914, opened 2020-09-01
<Button/> - add ellipsis prop

gendeld:
### 🔦 Summary

...

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


src/Button/Button.js
```diff
@@ -75,12 +87,21 @@ class Button extends PureComponent {
...
             )}
             data-hook={dataHook}
           >
-            {children}
+            <Ellipsis
+              ellipsis={ellipsis}
+              showTooltip={showTooltip}
+              {...tooltipProps}
+              render={({ ref, ellipsisClasses }) => (
+                <div ref={ref} className={ellipsisClasses()}>
```
shlomitc, line 100: instead of adding an extra div, why not simply make `Ellipsis` wrap `<ButtonNext/>`? then `ButtonNext` can accept a `ref` and classname

src/Button/Button.js
```diff
@@ -75,12 +87,21 @@ class Button extends PureComponent {
             {...generateDataAttr(this.props, ['skin', 'size', 'priority'])}
             className={st(
               classes.root,
-              { fluid: fullWidth, skin, priority, size },
+              { fluid: fullWidth, skin, priority, size, ellipsis },
```
shlomitc, line 90: can `fluid` and `ellipsis` live together?

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5914/

shlomitc (commented):
nice! that's actually a good demonstration of how easy it is to integrate the `Ellipsis` component.

Wrote a few comments + might want to add a visual test for it.

It will obviously increase the bundle size of many components so we should just be aware of it

src/Button/Button.js
gendeld: Yes!
https://recordit.co/XO1alx24GL

src/Button/Button.js
gendeld: I think we want Ellipsis to measure the Button content instead of the button itself,
otherwise we don't account for padding, suffix, prefix, ex cetera.

gendeld:
"Expected bundle .perfer/dist/statics/Button.bundle.min.js gzipped size to be less than 10kb. Actual gzipped size: 34.864kb"
@shlomitc Is this acceptable?

src/Button/Button.js
shlomitc: You're right, the measuring should be done according to the content.
I remember that we did something in the past that checked if the child element is `ref`able or not and then decided if we wrap it or not with an extra div. So for most cases, we simply clone the children and don't need to add it.

can you investigate it a little further? if I'll find the example I'll add it here

gendeld: Searched all implementations using <Ellipsis/> and didn't find what you're describing.
Maybe you're remembering something from <EllipsisHOC/> days?
I've seen using React.Children to check if a child is a component of a certain type (we want to support any child),
or to see if element is valid element and if it has props, but that won't tell us if it can receive ref and className.

Do you have an idea? I currently don't.

gendeld: We can add `contentClassName` and `contentRef` props to ButtonNext.

shlomitc (approved):

PR #5935, opened 2020-09-07
<StatisticsWidget/> - show dash when no value given

yogevitz:
### 🔦 Summary

added `emptyValues` prop - when set to true, show "-" instead of all values

### ✅ Checklist

- 📚 Change is documented
  - [x] Story
  - [x] API description
- 🔬 Change is tested
  - [x] Component
  - [x] Visual test


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5935/

src/StatisticsWidget/StatisticsWidget.js
```diff
@@ -12,6 +12,8 @@ class StatisticsWidget extends React.PureComponent {
     dataHook: PropTypes.string,
     /** Displayed value size (default: large) */
     size: PropTypes.oneOf(['tiny', 'large']),
+    /** When set to true show "-" (dashes) instead of values */
+    emptyValues: PropTypes.bool,
```
shlomitc, line 16: maybe instead of a new prop, this should be the default behavior when no value is passed.
based on your experience - what is the state and flow of your project? are you fetching the data and getting empty results?

yogevitz: In our case, we want to display the empty state when there is no data - in order for the user to see which data can be presented when he will have this data.
We wanted to use the `emptyValues` prop in order to see "dashed" in **all** of the items.
Your suggestion is a good idea, and it will be useful for cases in which some of the data exists and some doesn't.
I will make the change :)

shlomitc: 👍

yogevitz:
@shlomitc made the change

src/StatisticsWidget/StatisticsItem/StatisticsItem.js
```diff
@@ -52,7 +52,7 @@ class StatisticsItem extends React.PureComponent {
 
   _renderValue = (value, valueInShort, size) => (
     <AdaptiveHeading
-      text={value}
+      text={value || '-'}
```
shlomitc, line 55: I think we should add a really small unit-test for it because it's quite a hidden feature

yogevitz: added 👍

PR #5954, opened 2020-09-09
<PageHeader /> BackButton support - RTL

Shaniko:
### 🔦 Summary

A small css update - to support RTL:
Support the back-button to be aligned correctly.
Add small space between action buttons

### ✅ Checklist

- 📚 Change is documented
  - [x] Story

### DIFF
--- a/src/PageHeader/PageHeader.scss
+++ b/src/PageHeader/PageHeader.scss
@@ -90,7 +90,6 @@ See: https://css-tricks.com/flexbox-truncated-text/#article-header-id-3 */
     height: 30px;
     margin-top: 6px;
     margin-left: auto;
-    button {margin-left: 3px}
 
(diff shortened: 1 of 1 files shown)


Shaniko:
A small css update - to support RTL:
Support the back-button to be aligned correctly.
Add small space between action buttons

TalZa1810 (changes requested):
Please add visual tests. Also note that 20 visual tests were failed. 
I would use `dir=rtl` instead of adding the same class. We have multiple examples for rtl within the library that you can use as a reference.

Shaniko:
Hey.
Thanks for your comments and for your help.
Here is the problem this PR is trying to resolve:

image

TalZa1810:
Duplication of - https://github.com/wix/wix-style-react/pull/5956

PR #5965, opened 2020-09-13
<ColorPicker/> - refactor story page

tatiPodolskaya:
### 🔦 Summary

...

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5965/

src/ColorPicker/docs/index.story.js
```diff
@@ -1,43 +1,110 @@
...
+          }),
+
+          importExample(),
+
+          divider(),
+
+          title('Examples'),
+
+          example({
+            title: 'Simple usage',
```
sivanharel, line 77: Please change the first letter of each word to be a capital letter.
'Simple usage' ==> 'Simple Usage'

src/ColorPicker/docs/index.story.js
```diff
@@ -1,43 +1,110 @@
...
+            text:
+              'A simple example of color picker with `HEX`/`RGB`/`HSB` converter and input.',
+            source: examples.withConverter,
+          }),
+          example({
+            title: 'With history',
+            text:
+              'Previous selected color and current one can be shown below the picker area.',
```
sivanharel, line 85: I think that this could be a better description but let's talk with Assaf Zinger.
"Whether to display the previously selected color. Used for comparing the current and previous selected colors."

src/ColorPicker/docs/index.story.js
```diff
@@ -1,43 +1,110 @@
...
+  componentProps: {
+    value: '',
+    showConverter: true,
+    showInput: true,
+    showHistory: false,
+    allowEmpty: false,
+    onChange: () => {},
+    onCancel: () => {},
+    onConfirm: () => {},
+    onAdd: value => value,
```
sivanharel, line 43: Since the "onAdd" is not a required prop it shouldn't be here.
You can keep it in the example prop.

src/ColorPicker/docs/index.story.js
```diff
@@ -1,43 +1,110 @@
...
-          />
-        ),
-      },
-    ],
+    value: 'lightblue',
+    onChange: value => ({ value: value.alpha() === 0 ? '' : value.hex() }),
+    onCancel: () => {},
+    onConfirm: () => {},
+    onAdd: value => value,
+    addTooltipContent: 'Add your color',
```
sivanharel, line 52: It should be an array of objects with label and value properties: {label:"", value: ""}
So in this case it should be:

addTooltipContent: [
      {
        label: 'With add tooltip content',
        value: 'Add your color',
      },
    ],

src/ColorPicker/docs/index.story.js
```diff
@@ -1,43 +1,110 @@
...
-          <Swatches
-            colors={['red', 'green', 'blue']}
-            onClick={_color => {
-              changeColor(_color);
-            }}
-          />
-        ),
-      },
-    ],
+    value: 'lightblue',
```
sivanharel, line 47: Same as I wrote on below on "addTooltipContent".
The playground is broken as a result of the syntax error here.
You should change it to the correct form :)
It could be: 
value: [{label: 'Example Value', value: 'lightblue'}],

sivanharel (commented):
Great job!
There are some comments here.
Let's talk with Asaf Zinger about the examples.

sivanharel (changes requested):
Great job!
There are some comments here.
Let's talk with Asaf Zinger about the examples.

src/ColorPicker/docs/examples.js
```diff
@@ -0,0 +1,168 @@
...
+  }
+}
+`;
+
+export const withHistory = `
+class MyComponent extends React.Component {
+
+  state= {
+    current: '#c479ed',
+    previous: '#34b7eb',
```
sivanharel, line 75: Since the component gets only the current value (in the value prop) and in the initial state it sets the previous color to be the current I would say that it is confusing to write here 2 different values.
I suggest setting both of them to the same color also in your state.

sivanharel (approved):

PR #5976, opened 2020-09-14
Themes - Create an EditorX theme

shlomitc:
This is an initial PR to create an EditorX theme on top of the WSR components.
It intends to provide the flavor of the design system for the modal layout components, not yet create a full design system.

This PR includes the following:
1. Introduce a new theme - EditorX (tokens, theme provider function)
2. A new story section and page for the theme
3. Theme the `CustomModalLayout` component
4. Reuse visual tests per-theme.
5. Change in existing components to support straightforward stylable theming
6. Change in existing component to support theming icons assets
7. Introduce `ThemeProviderContext` and `ThemeProviderConsumerBackwardCompatible` to propagate the themed assets

Regarding the discussion that we had on whether to use CSS Vars or stylable approach - currently let's stick, with stylable but with css vars "in-mind". already in the css I tried to give names to each variable and understand what it will mean torefactor it to vars


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5976/

src/ThemeProvider/ThemeProviderConsumerBackwardCompatible.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import { ThemeProviderContext } from './ThemeProviderContext.js';
+
+/**
+ * This is a utility for every component that uses an icon to declare its defaults but also replace it with the theme context
+ * Usage Example:
+ *
+ * <ThemeProviderConsumerBackwardCompatible
```
argshook, line 8: i don't really understand why this is needed. can't icons be passed in `ThemeProvider.Consumer` without this wrapper?

src/ThemeProvider/ThemeProvider.js
```diff
@@ -1,24 +1,32 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import kebabCase from 'lodash/kebabCase';
+import { ThemeProviderContext } from './ThemeProviderContext';
```
argshook, line 4: i would name it `ThemeContext`, so that later it's `ThemeContext.Provider` and not `ThemeProviderContext.Provider`

src/Themes/editorX/editorX.js
```diff
@@ -0,0 +1,16 @@
+import { st, classes } from './editorX.st.css';
+import { Close, Help } from './icons';
+
+export default () => ({
+  className: st(classes.root),
```
argshook, line 5:
```suggestion
  className: classes.root,
```

src/BaseModalLayout/BaseModalLayout.js
```diff
@@ -59,28 +60,42 @@ class BaseModalLayout extends React.PureComponent {
...
-                dataHook={dataHooks.helpButton}
-                className={classes.controlButton}
-                onClick={onHelpButtonClick}
-                size="large"
-                skin="dark"
-              >
-                <Help className={classes.helpIcon} />
-              </CloseButton>
+          <ThemeProviderConsumerBackwardCompatible
+            defaultIcons={{
```
argshook, line 64: ideally default items should not be needed to be passed for every component.

this can be done by setting checking if `theme.icons` is set in `ThemeProviderContext.Provider`. If it's not, then use a default set of icons.

https://github.com/wix/wix-style-react/pull/5976/files#diff-e0fe1adecda63c78e1b36a94a5cdc2f9R27

src/CloseButton/CloseButton.js
```diff
@@ -72,13 +73,27 @@ class CloseButton extends PureComponent {
...
+        ) : (
+          <ThemeProviderConsumerBackwardCompatible
+            defaultIcons={{
+              CloseButton: {
+                small: Close,
+                medium: CloseLarge,
+                large: CloseLarge,
+              },
+            }}
+          >
```
argshook, line 91:
```suggestion
          <ThemeContext.Consumer>
```

i think this is how ideally it should look like to use the theme context. Default values should be set in `React.createContext({ icons: { CloseButton: ... } })`

src/Themes/editorX/editorX.js
shlomitc: 👍

src/CloseButton/CloseButton.js
shlomitc: If I understand you correctly, this means to put all icons in a single file, and then you simply use it from the context.

I tried that, but the problem is that it will immediately increase the bundle size of each component since you will bundle all the icons, even if you're not using them.

You can see [this commit](https://github.com/wix/wix-style-react/pull/5976/commits/965c6810196545479d4fee0ed4ce3a5970dd7436). I created a `defaultIcons` file and used it [here in the ThemeProviderConsumerBackwardCompatible](https://github.com/wix/wix-style-react/pull/5976/commits/965c6810196545479d4fee0ed4ce3a5970dd7436#diff-02e304bd6251cb0ff409ad88bc44e39aR13)
But then I had to ditch it and define the default icon per component.

src/ThemeProvider/ThemeProviderConsumerBackwardCompatible.js
shlomitc: AFAIK, you can initialize the `value` only in the `Provider`.
So if I don't use this provider, I can't "merge" the default ones with the overridden ones.
As you can see in the implementation, I'm wrapping it again with another `Provider`

src/ThemeProvider/ThemeProvider.js
shlomitc: 👍

src/BaseModalLayout/BaseModalLayout.js
shlomitc: see my answer in your next comment.
Also worth mentioning that you pass the defaultIcons only for your component.
This means the `BaseModalLayout` need to pass only the icons it uses explicitly (the help icon), but the default close button icons aren't required

shlomitc:
closing in favor of a clean PR - https://github.com/wix/wix-style-react/pull/6038

PR #5982, opened 2020-09-15
infra: cancel files deletion in the build task

shlomitc:
This was introduced when we were moving from WSR7 to WSR8 and we had a lot of icons in the root that weren't needed anymore.

This is no longer needed and also makes lives bad when it deletes stuff you didn't want to delete


shlomitc:
actually we can delete this entire task, I'll update

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5982/

shlomitc:
@EgoziE ping?

PR #5985, opened 2020-09-15
Remove wix-animations

EgoziE:


argshook:
i understand this is a copy paste of wix-animation, no changes done?

most likely we can reduce the size of whole animator util, cause we don't need all functionalities. for example debug mode, not sure if it's really needed.

probably should be done as a separate task

EgoziE:
Copy paste only, didn't touch it at all.
I can open a task removing unwanted content

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5985/

PR #5988, opened 2020-09-16
<Text/> - change ellipsis example

EgoziE:
change ellipsis example
image


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://www.wix.com/pages/wix-style-react.pr_5988/

shlomitc:
shouldn't there be a simple example of a one-line ellipsis before the multi-line appears?

EgoziE:
@milkyfruit

milkyfruit:
It seemed the most optimal to make this way, otherwise we would have to show 3 examples.

Now, first example shows text without any props and text with two props. So it should be relatively easy to understand that maxLines doesn't work without ellipsis, but ellipsis work alone. 
Second example starts with text with ellipsis only, which is simple example.

shlomitc:
it's just that you don't need the `maxLines` in order to create a single-line ellipsis. so this is kind of redundant.
It's not critical though

EgoziE:
@milkyfruit Maybe I can change the description to be:

Text component can wrap the text, show ellipsis or do both. Setting `maxLines` prop truncates the text at a specific number of lines (use this prop only when you need to show 2 lines or more).

milkyfruit:
@EgoziE Sounds great, please do it.

PR #6005, opened 2020-09-21
VariableInput: add readOnly prop

splintor:
### 🔦 Summary
Add `readOnly` prop to `<VariableInput>`, similar to what we have for `<Input>` ([Slack discussion](https://wix.slack.com/archives/C39FTGUBZ/p1600670672020700)).

### ✅ Checklist

- [X] 👨‍💻 API change is approved by the UI Infra Developers (@EgoziE)
- [X] 👨‍🎨 UX change is approved by the Design System UX (@milkyfruit)
- 📚 Change is documented
  - [X] Story
  - [X] API description
- 🔬 Change is tested
  - [X] Component
  - [X] Visual test


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6005

milkyfruit (changes requested):
@splintor The hover style and behaviour is not according to input specs.

- [ ] Allow to focus and select the text
- [ ] Add the hover style
- [ ] Change border color

Please check the input and input specs:
https://www.wix-style-react.com/?path=/story/components-api-components--input
https://zpl.io/2ERjBxM

splintor:
@milkyfruit 
1. Allow to focus and select the text - seems to work from what I checked. Do you see something else?
2. I don't have permission to open zpl.io/2ERjBxMSh. I'll try to work with what I see in `<Input>` code, but would be nice if you can provide me access.

milkyfruit:
1. I can select the text, but the input doesn't show focus outline and other styles:

VariableInput:
image

Input:
image

2. My back, I pasted the wrong link. This should work: https://zpl.io/2ERjBxM

splintor:
I pushed a fix that fixes the border and background colors on hover and focus.

Since the draft.js control sets `contenteditable` to `false` when `readOnly`, it doesn't have focus by default. To make it focusable so we will have focus outline, [we need to set its `tabIndex`](https://web.dev/control-focus-with-tabindex/). To achieve this, I added `tabIndex` prop to the `<VariableInput>` component, and set its default to `0` if `readOnly` is set. @EgoziE, please make sure it is OK with you.

milkyfruit (approved):

src/VariableInput/VariableInput.js
```diff
@@ -32,11 +32,14 @@ class VariableInput extends React.PureComponent {
       rows,
       size,
       disabled,
+      readOnly,
+      tabIndex,
```
shlomitc, line 36: I think this prop is not needed. the `tabIndex` hack internally doesn't relate to it.
if `tabIndex` is need for some reason, it should be applied to the root element.

src/VariableInput/VariableInput.js
```diff
@@ -32,11 +32,14 @@ class VariableInput extends React.PureComponent {
...
       placeholder,
       status,
       statusMessage,
       className,
     } = this.props;
+    const tabIndexToUse = tabIndex || (readOnly ? 0 : undefined);
```
shlomitc, line 42: this logic can be removed and inlined `tabIndex={readlOnly ? 0 : undefined }`

splintor: Removed - please review.

src/VariableInput/VariableInput.js
splintor: Removed.

EgoziE (approved):

PR #6023, opened 2020-09-23
<PageHeader /> - convert scss to stylable

yardenez:
_Note:_
stylable requires **additional changes** - when `showBackButton` &  `hasBackgroundImage` are enabled, back button should be rendered with white color and not blue, as currently happening.


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6023

src/PageHeader/PageHeader.js
```diff
@@ -125,19 +124,15 @@ export default class PageHeader extends React.PureComponent {
...
+      <div className={st(classes.root, {}, className)} data-hook={dataHook}>
+        <div className={classes.header}>
           <div>
             {this._animateComponent(
               breadcrumbsExists || minimized,
               !breadcrumbsExists,
               <div
-                className={classNames(s.breadcrumbsContainer, {
-                  [s.absolute]: !breadcrumbsExists,
-                  [s.minimized]: minimized,
```
mykas, line 140: Was there a reason to remove `minimized` here?

src/PageHeader/PageHeader.st.css
```diff
@@ -0,0 +1,125 @@
...
+}
+
+:import {
+  -st-from: "../Foundation/stylable/mixins/calc_rgba.js";
+  -st-default: calc_rgba;
+}
+
+/* Allow flex-child to shrink when there is a truncated descendant
+See: https://css-tricks.com/flexbox-truncated-text/#article-header-id-3 */
+.truncatedFlexChild {
```
mykas, line 13: I think no need to have this as mixin. We have only two places that uses this, but please leave the comment above those cases

src/PageHeader/PageHeader.st.css
```diff
@@ -0,0 +1,125 @@
...
+  margin-left: -10px;
+}
+
+:global([dir='rtl']) .breadcrumbsContainer {
+  margin-left: 0;
+  margin-right: -10px;
+}
+
+.breadcrumbsContainer:absolute {
+  position: absolute;
```
mykas, line 40: Lets give this state a better name. Can you check and explain me what this `absolute` does for the PageHeader container?

mykas: I see that this is applied when breacrumbExists, is this right?

src/PageHeader/PageHeader.st.css
```diff
@@ -0,0 +1,125 @@
...
+}
+
+.actionsBar {
+  -st-states: withBreadcrumbs, minimized;
+  height: 30px;
+  margin-top: 6px;
+  margin-left: auto;
+}
+
+.actionsBar:withBreadcrumbs {
```
mykas, line 119: Is this `widthBreadcrumbs` the same as above where `breadcrumbExists` applies absolute value?

src/PageHeader/PageHeader.js
daugerdas: `minimized` was not used as a selector for styling `.breadcrumbsContainer`, and does not seem to be used for anything else on this specific `div`

src/PageHeader/PageHeader.st.css
daugerdas: was trying to find a specific case for this, but could not yet. It is applied when `breadcrumbs` does not exist (`absolute: !breadcrumbsExists`), so renamed it now to `withoutBreadcrumbs`

Found that it can be seen on `storybook:visual` in `Page: EmptyState: should render with empty state` (when header is minimized), but the applied css rule for `position: absolute` does not change anything in that case

src/PageHeader/PageHeader.st.css
daugerdas: this one checks if `breadcrumbs` does exist (`widthBreadcrumbs: breadcrumbsExists`)

src/PageHeader/PageHeader.js
mykas: what do you mean it was not used? From what I see the classname: `minimized` is applied if `minimized` prop is true

src/PageHeader/PageHeader.st.css
mykas: Position: absolute removes the component from regular flow. So maybe you don't see it, but it does have an effect

src/PageHeader/PageHeader.st.css
```diff
@@ -0,0 +1,123 @@
...
+  min-width: 0;
+  flex: 1;
+}
+
+.breadcrumbsContainer {
+  -st-states: withoutBreadcrumbs;
+  margin-left: -10px;
+}
+
+:global([dir='rtl']) .breadcrumbsContainer {
```
mykas, line 30: lets move these `global([dir='rlt')` to the bottom

src/PageHeader/PageHeader.js
daugerdas: meant to say that `breadcrumbsContainer` class in `st.css` does not have `minimized` state/pseudo-class, so no additional css styling is applied when `minimized` is added

mykas: Yeah but I meant that it was being applied in sass file. The classname if applied had some styles there, no?

daugerdas: as far as I checked, did not found a specific style for that in `sass` file
Only this:
```css
.breadcrumbsContainer {
      margin-left: -10px;

      &.absolute {
        position: absolute;
      }
    }
```

mykas (approved):

PR #6024, opened 2020-09-24
<Ellipsis/> - tooltip value is not updated when changing the text content

gendeld:
When updating the Ellipsis value without rerendering the component the label value is changing but the tooltip text is not updating.

In [this example](https://www.wix.com/pages/wix-style-react/?path=/story/introduction-playground--playground&snippet=5e11a8f8-26b7-43f1-a0de-14d66b318d78), we are changing the <BarChart /> items every time the current tab id is changed and we can see that the tooltip text of the 'description' does not change correctly.

[Fixed example](https://preview.wix-style-react.com/pr_6024/?path=/story/introduction-playground--playground&snippet=5e11a8f8-26b7-43f1-a0de-14d66b318d78).

The solution here is to add a textContent field to state,
and trigger an update of tooltip activity & text content only once text is rendered.

When window size or ellipsis props have changed but hadn't triggered text rendering (memoized),
only check whether tooltip should be active.


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6024

src/common/Ellipsis/Ellipsis.js
```diff
@@ -45,24 +72,58 @@ class Ellipsis extends React.PureComponent {
...
+
+    const shouldBeActive = this._checkShouldBeActive();
+    if (shouldBeActive !== isActive) {
+      newState.isActive = shouldBeActive;
+    }
+
+    const stateChanged = Object.keys(newState).length > 0;
+    if (stateChanged) {
+      this.setState(newState);
+    }
```
EgoziE, line 103: Variable is not used elsewhere:
My suggestion:
```suggestion
    if (Object.keys(newState).length > 0) {
      this.setState(newState);
    }
```

src/common/Ellipsis/Ellipsis.js
```diff
@@ -92,29 +153,21 @@ class Ellipsis extends React.PureComponent {
...
-    });
-  };
-
-  _getEllipsisClasses() {
-    const { ellipsis, maxLines } = this.props;
-    const ellipsisLines = maxLines > 1 ? 'multiline' : 'singleLine';
-
-    return className =>
-      ellipsis ? st(classes.text, { ellipsisLines }, className) : className;
+  _renderText() {
```
EgoziE, line 162: Just out of curiosity, why did you changed the arrow function to a class function?

EgoziE (approved):
Nice job, I understand now why you wanted to use hooks...
I promise someday you'll be able to.

src/common/Ellipsis/Ellipsis.js
```diff
@@ -7,6 +7,33 @@ import Tooltip from '../../Tooltip';
...
+    return className =>
+      ellipsis ? st(classes.text, { ellipsisLines }, className) : className;
+  };
+
+  componentDidMount() {
+    this.props.textDidUpdate();
+  }
+
+  componentDidUpdate() {
+    this.props.textDidUpdate();
```
shlomitc, line 24: better name this prop as `onTextChange`

shlomitc (commented):
don't forget to add a test

src/common/Ellipsis/Ellipsis.js
gendeld: @EgoziE 
After the text render's there is a state change that would cause a rerender in the previous implementation leading to an infinite loop in some cases.

This utilizes the memoization of React.PureComponent to only update text once relevant props update,
making sure we only measure and change state when necessary.

src/common/Ellipsis/Ellipsis.js
gendeld: Don't you think it's more readable this way?
I know we don't need the variable.

EgoziE: It's a simple condition in my opinion, I find it pretty readable.

PR #6042, opened 2020-09-29
DatePicker - remove the interactive example from doc and generator

shlomitc:
basically, that's what you see when you click on the dropdown.

image

The reason is that the code example is setting an `overflow: auto`.

Since the overflow is good for the big playground, I suggest we use only the compact examples, so I removed it from the generator as well
image


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6042

PR #6047, opened 2020-09-30
<DropdownBase/> - fix flaky test

shlomitc:
Well, this one bugged me for a while now. In some cases, the eyes test will catch this:
image

And in other cases, it will be this:
image

no, the images are not the same, the difference is subtle. Here's why:

## Why is this happening:
After the component is mounted and opens the dropdown, the internal `Popover` component sets the position of the dropdown in the middle to the `TextButton` (which is the trigger element).
This is ok, but once fonts are loaded, the size of the `TextButton` get smaller, meaning that if some re-render occurs (or initial rendering with the right fonts), then the Popover will have a different position, and will cause a flaky test.

## How to reproduce before this PR:
open the visual storybook and this specific story page with a clean cache (Cmd+Shift+r). Then open the dev-tools (which will trigger a rerender) - you will see the popover changes the location

## The solution
simply use a native HTML button which isn't affected by the font swapping.


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6047

PR #6058, opened 2020-10-06
test(protractor): balance protractor tests

argshook:
This PR readjusts e2e tests in order to minimize cumulative time it
takes to run all parallel builds

1. `npm run test:e2e-runtime` is moved to `npm run test6:sanity`

    previously `npm run test:e2e-runtime` were run twice before both `test3`
    and `test4`. It is enough to run it once. Also, `npm run sanity` name
    fits the purpose of `e2e-runtime`.

2. the configs of `npm run test3:e2e` and `npm run test4:e2e` were
   changed to explicitly add files with `.e2e.js` extension

    `protractor1.conf.js` and `protractor2.conf.js` were changed so that
    both would take a similar amount of time

as a result, the protractor e2e tests durations is more balanced and no
longer takes the longest time:

| build          | before  | after   |
| ---            | ---     | ---     |
| `test3:e2e`    | 11m:49s | 16m:54s |
| `test4:e2e`    | 22m:09s | 15m:58s |
| `test6:sanity` | 8m:27s  | 7m:38s  |

strangely adding `npm run test:e2e-runtime` into `test6:sanity` did not
have impact on time

---

in future we want to move away from protractor entirely, but
we're not there yet. `npm run test2:visual` currently can't easily do
things like hover or scroll. it's possible but quite hackish


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6058

package.json
```diff
@@ -41,7 +41,7 @@
     "test:unit": "jest",
     "test:watch": "jest --watch",
     "pretest:e2e": "./scripts/ensure-dist.sh",
-    "test:e2e": "npm run test:e2e-runtime && build-storybook -c ./.storybook/e2e && npm run test:e2e-only",
+    "test:e2e": "build-storybook -c ./.storybook/e2e && npm run test:e2e-only",
```
shlomitc, line 44: since you removed this from the `test:e2e`, it also removed it from the `test` task.
in that case, you should explicitly add it there so all tests could be run in a single build.

package.json
```diff
@@ -71,7 +71,7 @@
     "test3:e2e": "YOSHI_PROTRACTOR_CONFIG=./protractor1.conf.js npm run test:e2e",
     "test4:e2e": "YOSHI_PROTRACTOR_CONFIG=./protractor2.conf.js npm run test:e2e",
     "test5:types": "npm run test:types && ./scripts/sync-ts.sh",
-    "test6:sanity": "npm run lint && npm run sanity",
+    "test6:sanity": "npm run lint && npm run sanity && npm run test:e2e-runtime",
```
shlomitc, line 74: The sanity task was supposed to give some quick indication for failing structure instead of waiting for long runs, but it seems to become some kind of a "common" task.

lint - runs on src code and should fail fast for such issues.
sanity - runs on src code for the autotools. that's kind of "smoke" tests, but it runs on a bundle, so it must have a build phase.
e2e-runtime - runs on the real project output.

I think the right thing to do will be the e2e-runtime to a separate task

protractor1.conf.js
```diff
@@ -2,14 +2,26 @@ const {
   baseProtractorConfig,
 } = require('wix-ui-test-utils/dist/src/protractor/protractor.conf');
 
+// explicitly listing components to test, because:
+// 1. we want to prevent adding new .e2e.js files
+// 2. to balance the time taken to run tests between parallel e2e builds
+const components = [
```
shlomitc, line 8: I think it's better to have both lists in the same file, that way it's easy to see the entire picture.

protractor2.conf.js
```diff
@@ -2,9 +2,28 @@ const {
   baseProtractorConfig,
 } = require('wix-ui-test-utils/dist/src/protractor/protractor.conf');
 
+// explicitly listing components to test, because:
+// 1. we want to prevent adding new .e2e.js files
+// 2. to balance the time taken to run tests between parallel e2e builds
+const components = [
```
shlomitc, line 8: what happens if I forget to add here a file? will it be missing from the tests?
Maybe it's better that only one will rune a specific set of tests while the other will run "all-but-the-specific-ones"

shlomitc (changes requested):

package.json
argshook: `npm run test` has `npm run sanity`, which, with my change, also has `e2e-runtime`. Thus, running `npm run test` will also run `e2e-runtime`

shlomitc: it runs `sanity`, not `test6:sanity`.
`sanity` contains only the autotools, not the e2e-runtime.

protractor2.conf.js
argshook: > what happens if I forget to add here a file? will it be missing from the tests?

yes, test will not be run with protractor. I think this is what we want
and having a list of specific entries is what discourages of manually
adding new `e2e.js`. It's also not part of generator for a long time.

these lists are not "take all tests, split in half", but:

* take all e2e tests
* sort by the length of time they take in CI
* split into two lists, while keeping the length of time in both lists similar.

it is to solve a problem of `test3:e2e` taking 22 mins, while
`test4:e2e` takes 11, and make them both ~15 mins. This essentially wins
us 7 mins of CI waiting time for a complete build

| protractor1.conf.js time | component       | protractor2.conf.js time | component                 |
| ---                      | ---             | ---                      | ---                       |
| 257                      | TableActionCell | 196                      | Page                      |
| 90                       | InputArea       | 104                      | DatePicker                |
| 52                       | MultiSelect     | 60                       | Range                     |
| 41                       | Stepper         | 43                       | Search                    |
| 30                       | RadioGroup      | 41                       | StatisticsWidget          |
| 28                       | Slider          | 29                       | MultiSelectCheckbox       |
| 23                       | DropdownLayout  | 23                       | MessageBox                |
| 14                       | Table           | 20                       | RichTextInputArea         |
| 13                       | FilePicker      | 20                       | PageHeader                |
| 12                       | Tabs            | 13                       | DataTable                 |
| 11                       | AudioPlayer     | 12                       | CardGalleryItem           |
|                          |                 | 1                        | InputWithOptions          |
|                          |                 | 1                        | MultiSelectFocusBehaviour |
| 571                      | total time (s)  | 563                      | total time (s)            |

> Maybe it's better that only one will rune a specific set of tests while the other will run "all-but-the-specific-ones"

i can think of something more sophisticated but i'm not sure it's worth
the effort, especially considering that in future we want to use
different methods for running e2e (maybe eyes-storybook with better
hover/scroll/etc support, maybe sled, maybe puppeteer etc)

package.json
argshook: ah, you're right, i'll add `e2e-runtime` also to `test`

although it's only relevant for local, not sure anyone ever runs full test suite locally

protractor1.conf.js
argshook: do you mean to export them from `protractor.conf.js` and reuse in
`protractor1.conf.js` and `protractor2.conf.js`?

shlomitc: whatever name you think of.
just to have the list of all tests (by component names) in the same file

package.json
argshook: `sanity` in my mind can include the current `e2e-runtime` as its pretty lightweight.

adding dedicated `e2e-runtime` is a right thing to do, but i'd like to
have it as a separate task, because it requires ci people involvement.

`e2e-runtime` should also do more thorough testing than it does right
now, but it should be separate task.

goal here is to just reduce the maximum time it takes for all e2es in CI

protractor2.conf.js
shlomitc: I thought of something like this:

```
const specificComponents = ['InputArea', 'MultiSelect',,,,,,,];

//1 - specific ones
specs: components.map(component => `src/${component}/**/*.e2e.js`)

//2 - all but specific ones
specs: `src/![${specificComponents.join(', ')]}/**/*.e2e.js`,
```

Regarding not adding new "e2e"s. yeah, but you know - if someone needs to add it, it's easier if it simply works instead of figuring out that it needs to be added to some file (as it's not documented, new contributors are not familiar with this constrain, even not all team members are aware of it...)

package.json
shlomitc: probably not that critical

scripts/components-for-e2e-tests.js
```diff
@@ -0,0 +1,35 @@
+/*
```
shlomitc, line 1: Not sure that “scripts” is the right place for it, but I guess that’s not a biggie

shlomitc (approved):

scripts/components-for-e2e-tests.js
argshook: yep, i moved it to `test` folder, it's a more appropriate location

PR #6076, opened 2020-10-11
<Ellipsis/> - fix endless render loop

gendeld:
When text component inside ellipsis is rendered it calculated whether to activate the tooltip,
if tooltip state is active
text content is rendered inside <Tooltip />
else
text content is rendered outside of it
React does not know these components are the same,
causing a remount when tooltip activity status changes.
In marginal cases (where the difference in having ellipsis or not is a few pixels)
this can cause an infinite render loop.

To recreate this error - open these examples and resize the window until it breaks.

Before (Broken):
https://www.wix-style-react.com/?path=/story/introduction-playground--playground&snippet=58be2f3e-cc3b-419b-bba4-075f1b286bee

After (Fixed):
https://preview.wix-style-react.com/pr_6076/?path=/story/introduction-playground--playground&snippet=58be2f3e-cc3b-419b-bba4-075f1b286bee

### DIFF
--- a/src/Text/Text.js
+++ b/src/Text/Text.js
@@ -5,11 +5,12 @@ import { st, classes } from './Text.st.css';
 import Ellipsis, { extractEllipsisProps } from '../common/Ellipsis';
 
 const TextWithEllipsis = ({ className, ...props }) => {
+  const { ellipsisProps, componentProps } = extractEllipsisProps(props);
+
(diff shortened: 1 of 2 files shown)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6076

PR #6079, opened 2020-10-12
<AddItem/> - Support customising icons with theming mechanism

sivanharel:
### 🔦 Summary

...

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6079

src/AddItem/AddItem.js
```diff
@@ -79,9 +81,33 @@ class AddItem extends Component {
...
+
+  _isChildStringWrapByBox = () => {
+    const { children } = this.props;
+
+    const isBoxChild =
+      children && children.type && children.type.displayName === 'Box';
+    const isStringInnerChild =
+      children.props && isString(children.props.children);
+
+    return isBoxChild && isStringInnerChild;
```
shlomitc, line 110: consider a more generic solution using a render function
```jsx
<AddItem>
  {() => <div>this will not apply text styles</div>}
</AddItem>
```

src/AddItem/AddItem.st.css
```diff
@@ -96,12 +96,12 @@
   cursor: not-allowed;
   background-color: value(D10-05);
   color: value(D10-30);
-  border: dashed 1px value(D10-30);
+  border-color: value(D10-30);
```
shlomitc, line 99: are you sure `dashed` isn't needed?

shlomitc (changes requested):

src/AddItem/AddItem.js
sivanharel: sounds like a good idea.

src/AddItem/AddItem.st.css
sivanharel: It is not needed since it defined at the ".root:theme(dashes)".
The only change that is needed is the color.

src/AddItem/AddItem.js
```diff
@@ -11,24 +12,16 @@ import Add from 'wix-ui-icons-common/Add';
...
-  tiny: <Add width="26" height="26" style={{ flexShrink: 0 }} />,
-  custom: <AddMedia width="31" height="31" />,
-};
-
 class AddItem extends Component {
   static displayName = 'AddItem';
   static propTypes = {
-    /** any renderable node */
-    children: PropTypes.node,
+    /** any renderable node or a render function */
```
shlomitc, line 23: worth mentioning that in the case of a render function, text styles will not be applied. you should also back this with a visual test if possible

src/AddItem/index.d.ts
```diff
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { TooltipCommonProps } from '../common';
 
 export interface AddItemProps {
+  children?: React.ReactNode | Function;
```
shlomitc, line 5:
```suggestion
  children?: React.ReactNode | () => React.ReactNode;
```

src/AddItem/AddItem.js
```diff
@@ -75,13 +68,46 @@ class AddItem extends Component {
     removePadding: false,
   };
 
+  _renderAddItemIcon(Icon, size, isImageIcon) {
+    let AddItemIcon;
+
+    if (isImageIcon) {
+      AddItemIcon = <Icon width="31" height="31" />;
```
shlomitc, line 75: I think that it could be better if it's set only on the default icon since it's a very specific fit of this icon to the component's size constraints. WDYT?

src/AddItem/AddItem.js
```diff
@@ -75,13 +68,46 @@ class AddItem extends Component {
...
+    } else if (size === SIZE.tiny) {
+      AddItemIcon = <Icon width="26" height="26" style={{ flexShrink: 0 }} />;
```
shlomitc, line 77: same

src/AddItem/AddItem.js
sivanharel: I agree.. But the style={{ flexShrink: 0 }} is must be added in the tiny icon since it should be display vertical in this case. But I could add it to the css.

src/AddItem/AddItem.js
```diff
@@ -78,10 +71,26 @@ class AddItem extends Component {
...
+    return (
+      <ThemeProviderConsumerBackwardCompatible
+        defaultIcons={{
+          AddItemButton: {
+            tiny: () => <Add width="26" height="26" />,
+            small: AddItemSmall,
+            medium: AddItemMedium,
+            large: AddItemLarge,
+            image: () => <AddMedia width="31" height="31" />,
+          },
```
shlomitc, line 85: consider refactoring it out to a static variable to avoid re-renders of the function component.

const 
```jsx
const AddItemButtonIcons = {
  tiny: () => <Add width="26" height="26" />,
  small: AddItemSmall,
  medium: AddItemMedium,
  large: AddItemLarge,
  image: () => <AddMedia width="31" height="31" />,
};
```

```suggestion
          AddItemButton: AddItemButtonIcons
```

src/AddItem/AddItem.st.css
```diff
@@ -195,6 +195,10 @@
   flex-direction: row;
 }
 
+.content:size(tiny) svg {
```
shlomitc, line 198: instead of a global selector, you can target it as a general `icon` class
```css
.content:size(tiny) .icon {
  flex-shrink: 0;
}
```
each icon will be rendered with this class name:
```jsx
return <Icon className={classes.icon}/>;
```

and each icon is responsible to pass this class name
```jsx
tiny: ({className}) => <Add className={className} width="26" height="26" />
```

shlomitc (commented):
very last comments, I think we can merge afterwards

shlomitc (approved):

PR #6085, opened 2020-10-13
<Ellipsis/> - fix endless render loop

gendeld:
Try to fix Ellipsis endless rerender in marginal cases (where the difference in having ellipsis or not is a few pixels)
using debouncing and activating when text size matches container size


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6085

EgoziE (approved):

src/common/Ellipsis/Ellipsis.js
```diff
@@ -17,11 +17,15 @@ class TextComponent extends React.PureComponent {
...
+    if (!this.props.textRendered) {
+      this.props.onTextRendered();
+    }
   }
 
   componentDidUpdate() {
-    this.props.onTextChange();
+    if (!this.props.textRendered) {
+      this.props.onTextRendered();
+    }
```
ronenst, line 28: @EgoziE @gendeld Shouldn't these be `!this.state.textRendered` instead of `props`?

gendeld: `TextComponent` gets `textRendered` from props passed by `Ellipsis`

gendeld: So no, basically.
Why do you ask, have you been getting some errors relating to ellipsis?

ronenst: yeah, sorry, I think we're seeing an instance of this happening in the wild:
https://wix.slack.com/archives/CC0TSQ96H/p1604320081185200

But it's probably due to a stale version.

ronenst: https://sentry.wixpress.com/sentry/experts-experience/issues/29926695/?query=is%3Aunresolved

ronenst: actually according to the source map it's the latest version...

gendeld: If you can find a way to reproduce it reliably it would help to try and solve it.
According to the slack channel all occurrences happen to one user(?)
Do we know if this happened to anyone else in recent days?

ronenst: We don't really understand it either, we'll keep and eye I guess.

ronenst: thanks!

zivl: @gendeld we have another alert incident here:
https://sentry.wixpress.com/sentry/experts-experience/issues/29926695/?query=is%3Aunresolved%20exceeded

PR #6088, opened 2020-10-13
<DropdownLayout/>- using builder for selectable option

TalZa1810:
### 🔦 Summary

This PR includes:
1. Changing the selectable option to use builder. The idea is that the option is the one responsible to its design.
- If the option is a regular object (no overrideStyle)- the `listItemSelectBuilder` is used.
- If the option is a regular object (with overrideStyle)- creating a render function with the node provided as is. 
- If the option is already a render function- use this as is. 

We support having an object with `overrideStyle:true` (backwards compatibility) but would like to avoid using this. 
As for our builders- we can remove the `overrideStyle` because it is no longer used (can be in a different PR). 

2. Fixing some drivers functions.

**Custom elements**
1. _Previous version-_
In order to have a custom element, you needed to add `overrideStyle` property to the custom render function:
```
() => {
const setColor = ({ selected, hovered, disabled }) =>
  (selected && 'red') || (hovered && 'green') || (disabled && 'grey');

const customBuilderFunction = ({ id, disabled }) => ({
  id,
  disabled,
  overrideStyle:true,
  value: props => <div style={{ color: setColor(props) }}>value</div>})

  const options = [
    customBuilderFunction({ id: 1, title: 'Custom Option 1' }),
    customBuilderFunction({ id: 2, title: 'Custom Option 2', disabled: true }),
    listItemActionBuilder({
      id: 3,
      prefixIcon: <Icons.Edit />,
      title: 'Builder Option 3',
    }),
    ,
    { id: 4, value: 'Option 4' },
  { id: 5, value: 'Option 4', overrideStyle:true },
  ];

  return <DropdownLayout inContainer visible options={options} />;
};
```

2. _Current version-_ 
no need to add `overrideStyle` property:
```
() => {
const setColor = ({ selected, hovered, disabled }) =>
  (selected && 'red') || (hovered && 'green') || (disabled && 'grey');

const customBuilderFunction = ({ id, disabled }) => ({
  id,
  disabled,
  value: props => <div style={{ color: setColor(props) }}>value</div>})

  const options = [
    customBuilderFunction({ id: 1, title: 'Custom Option 1' }),
    customBuilderFunction({ id: 2, title: 'Custom Option 2', disabled: true }),
    listItemActionBuilder({
      id: 3,
      prefixIcon: <Icons.Edit />,
      title: 'Builder Option 3',
    }),
    ,
    { id: 4, value: 'Option 4' },
   { id: 5, value: 'Option 4', overrideStyle:true },
  ];

  return <DropdownLayout inContainer visible options={options} />;
};
```

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [x] Visual test


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6088

TalZa1810:
Removing the `overrideStyle` from the builders and the documentation will be in a different PR. Also, there's a fix that needs to be done with the `ListItemSelect` driver (which is why I skipped a test with `the DropdownBase`) and styles (when item is selected and hovered). This will be in a dedicated PR.  
`isHoveredWithGlobalClassName` and `isSelectedWithGlobalClassName` driver functions were used only within the library.

EgoziE (approved):

TalZa1810:
Fails on a visual test which is not related (FloatingHelper).

Issue #4041, opened 2019-09-02
<Multiselect/> - tags should change color on hover


b3llash:
# 🐛 Bug Report

### 🏗 Relevant Components
`<Multiselect>` and `<Tag>`

### 😯 Current Behavior
When hovering tags in a Multiselect, their background color doesn't change.
Instead, the input itself gets into hover state, which is misleading.
https://i.gyazo.com/508b154e942ff4ca513e7fc28f1b312d.gif

### 🤔 Expected Behavior
Tags should respond to cursor (get into hover state, show tooltip or whatever behavior attached to them).
The input itself should get into hover state only when hovered (and not when the tags in it are hovered).

### 👀 Severity
Low

shlomitc:
@milkyfruit what do you think?

milkyfruit:
@b3llash @shlomitc tags themselves have hover styles: https://zpl.io/bzq6pJE
It definitely makes sense to show it when they're in the input too.

shlomitc:
alright. added to our backlog

argshook:
closed by https://github.com/wix/wix-style-react/pull/6114

PR #6114, opened 2020-10-20
<Multiselect/> - tags change color on hover #4041

Zivilekin:
### 🔦 Summary

Before:
When hovering tags in Multiselect (or other relevant places - f.e. Tag Input), their background color didn't change.
Instead, only the input itself got into hover state.

With this fix:
Tags change color when hovered, also when they are used in Input. They don't change color only when are disabled.


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6114

src/Tag/Tag.private.driver.js
```diff
@@ -26,6 +26,7 @@ const tagPrivateDriverFactory = ({ element }) => {
     getTextSize: () => getTextDriver(element).getSize(),
     getTextWeight: () => getTextDriver(element).getWeight(),
     isClickable: () => isClassExists(element, 'clickable'),
+    isHoverable: () => isClassExists(element, 'hoverable'),
```
argshook, line 29: @Zivilekin we're slowly moving away from `.driver.js` towards `.uni.driver.js`. The goal is to have only `.uni.driver` or `.private.uni.driver` for
each component. This means we don't need to introduce new api to non-uni drivers.

no need to change anything, just fyi

argshook (approved):
great work!

PR #6120, opened 2020-10-21
fix(DropdownBase): add `className` to typings

argshook:
this PR adds `className` type definition to `<DropdownBase/>`.

`className` is passed to `<Popover>` and then applied to portaled element. it can be useful when, for example, one needs to target a component by
class name to exclude it from click-outside mechanism

it's not a new prop, just a typing


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6120

PR #6129, opened 2020-10-26
<DatePicker/> - RTL support

sivanharel:
### 🔦 Summary

...

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


shlomitc (commented):
is RTL really needed as a js prop? can't it be achieved only by css?

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6129

sivanharel:
> is RTL really needed as a js prop? can't it be achieved only by css?

In case of RTL, we have to make some CSS adjustments, and also for the a11y keyboard navigation, we have to pass the prop dir="rtl" to the DayPicker component (react-day-picker). So it's not only CSS changes.
When passing the dir="rtl", the keyboard navigation will be changed means pressing on the right arrow will navigate to the previous day, and pressing on the left arrow will navigate to the next day.

shlomitc:
can't you simply change the order of the buttons with flex-direction?

sivanharel:
> can't you simply change the order of the buttons with flex-direction?

Visually it changes the buttons orders without the rtl prop but if we are not passing the prop dir="rtl" then when pressing on the 'right arrow' and on the 'left arrow' it behaves the opposite than expected.

EgoziE:
@jonathanadler Just another good reason to let go of the buggy old DatePicker 🙄

argshook:
this is a recurring problem with `DatePicker` but it's possible for this
to happen in other components too. New and buggless `DatePicker` does
not guarantee to solve this problem.

It all boils down to the RTL detection mechanism. We currently check for
css selector and use a prop like `rtl`. This is not good, because users
need to do two things in order for RTL to work, that's way too much
work.

possible solution:

starting from component root dom element, climb upwards checking each
   dom element if it has `dir="rtl"`. if yes, apply rtl behaviour
  * pros
    * no more `rtl` prop
  * cons
    * maybe possibly negative performance impact

---

ultimate solution 9000

use react context and require users to wrap theirs apps with
`<WSRProvider rtl={isRtlEnabled}/>`.

It could also serve for theming or things like font upgrade:
`<WSRProvider rtl={isRtlEnabled} theme={backoffice()} font="madefor"/>`

* pros
  * no performance impact
  * no more `rtl` prop
* cons
  * users need to know about and use `<WSRProvider/>`

sivanharel:
@argshook -  I totally agree that it is really bad that the users
need to do two things in order for RTL to work.
But I don't think that DOM traversing in order to find an element that has dir="rtl" is better.

I think that the context solution could be great. This is how it is implemented in TPA, we are getting from the context indication if it's mobile mode and if it's rtl.
But yes I think we should open another task since it related also to some more WSR components.

jonathanadler:
Using `dir="..."` is pretty much a standard in web applications, and it's necessary for `rtl` support.
So I don't agree that it's really considered as "work" for a developer.
With Yoshi, developers get this free (and I bet the BM handles this as well - if it even supports RTL).

In wix-ui-tpa we try to have all components support RTL natively with flex.
Meaning, we rely on the existence of the `dir` attribute, so that flex items will flip, and work correctly in `rtl`.

In addition like @sivanharel wrote, we use React context to pass any app-level data to our components (`mobile` & `rtl` at the moment)

That way, for specific cases in which our components need to know the direction for the javascript logic, they get it "for free" without requiring an addition prop to the component.

This also makes the use of `rtl` an internal implementation issue.
So that we won't break any api by adding/removing the `rtl` prop.

argshook:
> So I don't agree that it's really considered as "work" for a developer.

just to clarify what i meant

`dir="rtl"` is ok
`rtl` prop is not ok, that's the "too much work" part.

there should be only one switch to enable rtl, maybe

`<WSRProvider dir="rtl"/>`

jonathanadler:
> > So I don't agree that it's really considered as "work" for a developer.
> 
> just to clarify what i meant
> 
> `dir="rtl"` is ok
> `rtl` prop is not ok, that's the "too much work" part.
> 
> there should be only one switch to enable rtl, maybe
> 
> `<WSRProvider dir="rtl"/>`

Oh, I  see, I misunderstood.

So yeah, totally agree 😄

PR #6132, opened 2020-10-26
<CustomModalLayout/> -  Remove 3px content padding

Zivilekin:
### 🔦 Summary 

As there were complaints about https://github.com/wix/wix-style-react/pull/5932 this change, those changes were reverted. 

With reverted change (removed content padding) the problem occurs - when Input field is used inside CustomModalLayout, the outline of that Input field (top and bottom) is hidden.

That's why a solution was added to storybook examples - which is to add Box component with 3px of height before and after Input field.


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6132

argshook (approved):
:+1: great work!

PR #6135, opened 2020-10-26
<Layout/> - Add rowHeight prop

EgoziE:
Added a `rowHeight` prop in order to set all rows height.
This prop will support the case when we want all rows to be the same height as the highest row. 

image


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6135

src/Layout/Layout/index.js
```diff
@@ -44,10 +45,14 @@ Layout.propTypes = {
 
   /** is used to aligns the grid items along the column axis */
   alignItems: PropTypes.string,
+
+  /** If true, all rows will be the same height according to the highest one */
+  rowHeight: PropTypes.string,
```
shlomitc, line 50: wouldn't it be more consistent to call it `gridAutoRows`? That way every new CSS Grid property that you need to expose, the naming will be straightforward.

src/Layout/Layout/index.js
```diff
@@ -44,10 +45,14 @@ Layout.propTypes = {
 
   /** is used to aligns the grid items along the column axis */
   alignItems: PropTypes.string,
+
+  /** If true, all rows will be the same height according to the highest one */
```
shlomitc, line 49: It's not a boolean, so it won't be "true"

src/Layout/Layout/index.js
EgoziE: Well, I did think about it!
But as a user of wsr that _might not_ speak fluent css, will you understand what `gridAutoRows` means without reading a lot?

src/Layout/Layout/index.js
EgoziE: Oh right, it started as a boolen in my PR and then I changed it to rowHeigt, thanks!

src/Layout/Layout/index.js
shlomitc: Good point. Is there another CSS grid property that related to a row's height? just so it won't collide.
Also, I see that the prop expects to receive the css property of a grid (e.g. 1fr). I think this is the first prop that actually means the component uses CSS grid inside to achieve some features, so I think it's a good time to mention it in it's docs

milkyfruit (approved):
Looks great, this is exactly what Promote team requested.

src/Layout/Layout/index.js
EgoziE: Of course, there are few that we don't use here and I don't think they're relevant.

`grid-template-rows` - specifies the number and heights of the rows in a grid layout, not relevant for us because it requires to know in advance the exact number of rows

`grid-auto-flow: column` - makes the grid column oriented instead of rows and then "cols" props becomes "rows" prop. We don't use it or need it currently.

I'll add some description explaining `Layout` is grid based.

milkyfruit:
@EgoziE I slightly adjusted your examples and added few more explaining horizontal and vertical alignment.

- [ ] **Layout presentation**
_What changed_: Added an additional sentence explaining about cols and gap

By default the layout is divided into 12 columns and the cell is spanned across all 12 columns.
Set the maximum number of columns with `cols` prop and adjust the space between cell with `gap` prop.
Build your own layout by setting a span to cells like in this example.


- [ ] **Row Height**
_What changed_: Changed to example with with long text, because it covers a real usecase

```
<Layout rowHeight="1fr">
  <Cell span={4}>
    <Box height="100%" backgroundColor="lightblue" />
  </Cell>
  <Cell span={4}>
    <Box height="100%" backgroundColor="lightblue" />
  </Cell>
  <Cell span={4}>
    <Box height="100%" backgroundColor="lightblue" />
  </Cell>

  <Cell span={4}>
    <Box height="100%" backgroundColor="lightblue" />
  </Cell>

  <Cell span={4}>
    <Box height="100%" backgroundColor="lightgreen">
      The Life and Strange Surprizing Adventures of Robinson Crusoe, Of York,
      Mariner: Who lived Eight and Twenty Years, all alone in an un-inhabited
      Island on the Coast of America, near the Mouth of the Great River of
      Oroonoque; Having been cast on Shore by Shipwreck, wherein all the Men
      perished but himself.With An Account how he was at last as strangely
      deliver'd by Pyrates.
    </Box>
  </Cell>
  <Cell span={4}>
    <Box height="100%" backgroundColor="lightblue" />
  </Cell>
</Layout>;
```

- [ ] **Vertical**
Cell content can be aligned in a middle by enabling the prop `vertical`
```
<Layout>
  <Cell span={4}>
    <Box background="gold" height="40px" />
  </Cell>
  <Cell span={4} vertical>
    <Box background="deeppink" height="20px" />
  </Cell>
  <Cell span={4} vertical>
    <Box background="deepskyblue" height="10px" />
  </Cell>
</Layout>;
```

- [ ] **Align Items**
Property `alignItems` is used to align cells along column axis. It supports all major CSS values, read more here: https://developer.mozilla.org/en-US/docs/Web/CSS/align-items
```
<Box height="150px">
  <Layout alignItems="flex-end">
    <Cell span={4}>
      <Box background="gold" height="100%" children="one" />
    </Cell>
    <Cell span={4}>
      <Box background="deeppink" height="100%" children="two" />
    </Cell>
    <Cell span={4}>
      <Box background="deepskyblue" height="100%">
        Three has way more text that doesn't fit in a single line and now it
        becomes importnat to find a way to align all three items the way we
        want.
      </Box>
    </Cell>
  </Layout>
</Box>;
```

- [ ] Justify Items
Property `justifyItems` is used to align cells along row axis. It supports all major CSS values, read more here:
https://developer.mozilla.org/en-US/docs/Web/CSS/justify-items
```
<Layout justifyItems="end">
  <Cell span={4}>
    <Box background="gold" children="one" />
  </Cell>
  <Cell span={4}>
    <Box background="deeppink" children="two" />
  </Cell>
  <Cell span={4}>
    <Box background="deepskyblue" children="three" />
  </Cell>
</Layout>;
```

PR #6140, opened 2020-10-27
chore(Dropdown): use imported classname

argshook:
this PR

1. removes the css style test. let's go away with it for now to not
   spend too much time as is not critical
2. use imported class name in stylable file. This way it's not an html
   `input` that's targeted, but an classname


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6140

daugerdas (approved):
Great, thank you!

PR #6149, opened 2020-10-28
<CustomModalLayout/> - fix top and bottom dividers

gendeld:
showHeaderDivider and showFooterDivider show persistent dividers that are usually only showed when CustomModalLayout is scrolled.

When CustomModalLayout is scrolled it also shows the Content dividers, resulting in doubled dividers:
image
(With showHeaderDivider you can see top divider is twice the height of bottom divider)

This PR fixes the doubled divider problem,
and adds vertical padding of 24px when persistent dividers are added.

Visual changes approved by Ayelet.


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6149

src/CustomModalLayout/index.d.ts
```diff
@@ -20,7 +20,7 @@ export interface CustomModalLayoutProps extends BaseModalLayoutProps {
   removeContentPadding?: boolean;
   showHeaderDivider?: boolean;
   showFooterDivider?: boolean;
-  contentHideDividers?: boolean;
+  hideContentDividers?: boolean;
```
lironch, line 23: I love this refactor but it needs to be done in the `content` implementations as well (it is still `contentHideDividers` there)

src/BaseModalLayout/LayoutBlocks/Content/Content.js
```diff
@@ -14,7 +14,9 @@ export const Content = ({
   dataHook,
   className,
   children,
-  contentHideDividers,
+  hideContentDividers,
+  showHeaderDivider,
+  showFooterDivider,
```
lironch, line 19: These props belong to `Header` & `Footer`, `Content` shouldn't be aware of them.
It creates coupling between the layout blocks that might prevent you from composing with them other use cases later.
I guess that you and Ayelet tried to get the best of all worlds with this solution but that's not how i would achieve it.
I originally thought that `contentHideDividers` (or `hideContentDividers` as you suggested) would suffice to allow users opting-out from the scroll monitoring capabilities of this component when using `showHeaderDivider` or `showFooterDivider`.
In case we want to have a finer grained solution for that (which i think is a bit of an overkill) i prefer that we will have a dedicated props for `Content` that hides dividers according to the user needs, `hideContentTopDivider` & `hideContentBottomDivider` for that matter.

src/BaseModalLayout/test/BaseModalLayout.tsx
```diff
@@ -22,16 +22,17 @@ function baseModalLayoutWithAllProps() {
       <BaseModalLayout.Header
         className={'cls-header'}
         dataHook={'dh-header'}
-        showHeaderDivider={false}
```
lironch, line 25: is `showHeaderDivider` no longer a prop of `Header`? 
I don't see that you removed it from the component definition

src/BaseModalLayout/test/BaseModalLayout.tsx
```diff
@@ -48,8 +49,8 @@ function baseModalLayoutWithAllProps() {
         secondaryButtonOnClick={() => {}}
         actionsSize={'large'}
         sideActions={<div />}
-        showFooterDivider={false}
```
lironch, line 51: is `showFooterDivider ` no longer a prop of `Footer`? 
I don't see that you removed it from the component definition

lironch (approved):

PR #6163, opened 2020-11-02
<DropdownLayout/>- convert option to ListItemSelect

TalZa1810:
### 🔦 Summary

We want all options to be converted to builders. Therefore, the solution is:
1. If we get an option as a render/builder function- we render this as is.
2. If we get a divider or title- we use the `listItemSectionBuilder`.
3. If we get a custom option- we convert this to a render function.
4. If we get a regular object option- we convert this to `listItemSelectBuilder`. 

This PR includes:
1.  Converting all `DropdownLayout` options to builders (as explained above).
2. Adding `DATA_HOOKS.OPTION` for backwards compatible of `DropdownBase` driver method `optionContentAt`. 
3. Adding unit tests to make sure that we get the relevant content for each of the scenarios. 

Here's a [snippet](https://preview.wix-style-react.com/pr_6163/?path=/story/introduction-playground--playground&snippet=38f09e6b-a229-4c88-b871-a420cef3b386) of all mapped scenarios of DDL options. 

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [x] Component
  - [x] Visual test


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6163

src/DropdownBase/DropdownBase.uni.driver.js
```diff
@@ -45,20 +47,22 @@ export const dropdownBaseDriverFactory = (base, body) => {
 
     optionContentAt: async id => {
```
TalZa1810, line 48: This method doesn't work as expected **(even before merging this PR)**.
You would expect to get option content **by id**, however it gets content by **position**.
This is an example for the test scenario I checked on master: 
```
it('should return the requested textual option', async () => {
    const options = [
      { id: 1, value: 'First option' },
      { id: 2, value: 'Second option', overrideStyle: true },
      { id: 3, value: 'Third option', overrideOptionStyle: true },
    ];
    const props = {
      onMouseLeave: jest.fn(),
      options,
    };
    const targetDataHook = 'myOpenButton';
    const driver = createDriver(
      <DropdownBase {...props}>
        {({ open }) => {
          return (
            <IconButton
              skin="inverted"
              dataHook={targetDataHook}
              onClick={open}
            >
              <ChevronDown />
            </IconButton>
          );
        }}
      </DropdownBase>,
    );
    await driver.clickTargetElement(targetDataHook);
    expect(await driver.optionContentAt(1)).toEqual(options[0].value);
  });
```
Expected: "First option"
Received: "Second option"

I think that we should use the DropdownLayout `optionContentAt` instead of implementing a different one for `DropdownBase` which is incorrect anyways. Also, this will help up to prevent maintaining the same method in 2 different places. I can try finding a different solution but it feels very hacky to make something work which doesn't work as expected anyways. 

@shlomitc  @EgoziE WDYT?

shlomitc: I don't understand.
the `optionContentAt()` expects a position, meaning an index and not id. it's an old (and bad) implementation, but let's just keep supporting it. don't change it to `id`

`expect(await driver.optionContentAt(1)).toEqual(options[1].value);`

TalZa1810: The `optionContentAt()` of the `DropdownBase` expects **id** (which you can tell by the variable name) and the `optionContentAt()` of the `DropdownLayout` expects a **position**.
The `optionContentAt()` of the `DropdownBase` returns a **node** (getNative) in case of a node whereas the `optionContentAt()` of the `DropdownLayout` returns just the **node content**.
The `DropdownBase` is basically using the `DropdownLayout` so I don't think that the `optionContentAt()`  method should be implemented different than how it is implemented in `DropdownLayout`. 
Moreover, sometimes the option rendered as `ListItemSelect` but when receiving `overrideStyle` or `overrideOptionStyle` it does not render `ListItemSelect` so I don't think that checking whether it renders `ListItemSelect` or not is the right approach for implementing the DropdownBase `optionContentAt()`  method. 
I hope it's more clear now. We can have a quick talk so I can elaborate more if it's unclear.

shlomitc (approved):
looks good. I think that since we have a lot of code in this file, we can start splitting it.
Deprecated methods, additional features (like the infinite scroll), utilities (_getDataAttributes), option creator (option <=> builders), etc...

We can finish this PR but also start refactoring so it will make more sense in the future

Issue #5818, opened 2020-08-10
<Search/> - Value not changing on blur


AtomicAlienZ:
# 🐛 Bug Report

### 🏗 Relevant Components

`<Search/>`

### 😯 Current Behavior

Situation: a pre-filled controlled search, which should revert to its original value when the user blurs from the field. Error occurs when the user clears the field and blurs. The result:
![image](https://user-images.githubusercontent.com/1439457/89793516-a0bde700-db2e-11ea-98e0-093f6dff1421.png)

If the original value is an empty string, the search becomes unresponsive: the user can't focus it anymore.

### 🤔 Expected Behavior

`<Search/>` should properly work when its value is changed on blur

### 👣 Steps to Reproduce
```typescript
type State = {override: null | string};

export class FiltersV2 extends React.Component<{}, State> {
  state: State = {
    override: null,
  };

  get value() {
    return this.state.override !== null ? this.state.override : 'Something Default';
  }

  onBlur = () =>
    this.setState({
      override: null,
    });

  onChange = (event: React.ChangeEvent<HTMLInputElement>) =>
    this.setState({
      override: event.target.value,
    });

  render() {
    return (
      <>
        {this.value}
        <br />
        <Search
          expandable={true}
          expandWidth={240}
          value={this.value}
          onBlur={this.onBlur}
          onChange={this.onChange}
          onClear={console.log}
        />
      </>
    );
  }
}
```
1. Focus on search
2. Type something
3. Delete everything
4. Blur from the field

  - **WSR version**: `8.41.1`
  - **Yoshi version**: `4.125.0`

### 👀 Severity

- Critical

EgoziE:
To reproduce:
https://www.wix-style-react.com/?path=/story/introduction-playground--playground&snippet=0794c0a2-e88a-4809-8b0d-6f8ced8a5ae4

```
() => {
  const initialState = {
    value: 'Unfocus me',
  };

  const [state, setState] = React.useReducer(
    (currentState, newState) => ({ ...currentState, ...newState }),
    initialState,
  );

  return (
    <>
      {state.value}
      <br />
      <Search
        expandable={true}
        expandWidth={240}
        value={state.value}
        onBlur={event => setState({ value: null })}
        onChange={event => setState({ value: event.target.value })}
        onClear={console.log}
      />
    </>
  );
};
```

PR #6169, opened 2020-11-03
<Search/> - fix value on blur

daugerdas:
### 🔦 Summary

If a custom `value` is set in `onBlur()`, expandable controlled `<Search/>` does not change the `value` correctly when element loses focus 

[DSM-939](https://jira.wixpress.com/browse/DSM-939), Github issue #5818 

Issues:
- If `value` is set to `""` in `onBlur()` and `<Search/>` is expandable, the input becomes unresponsive after deleting/clearing text and element losing focus
- If `value` is set to a custom string in `onBlur()` and `<Search/>` is expandable, the input collapses and shows `onClear` button when element loses focus and there is no text in the input

This fix:
- Checks collapse logic in `_onBlur()` only after the state is updated when element loses focus

At first was thinking that the issue is if `value` is set to `null` in `onBlur()`, then the text in the input box does not change when element loses focus. So made a unit test and made a check if `null` is being assigned to `value` to add empty string instead (89c805a). But probably `null` should not be used for `value` in a controlled component, so reverted it (a890493)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6169

daugerdas:
Added commit 7b79e8d with a fix and unit test for an additional issue:
- If `value` is set to custom in `onBlur()` and `<Search/>` is controlled/expandable, when clicked on `clearButton`, `<Search/>` collapses and `clearButton` is displayed even though there is a default value in the input

image

This commit: Checks in `_onClear()` whether input is empty before collapsing it

src/Search/test/Search.spec.js
```diff
@@ -198,6 +198,54 @@ describe('Search', () => {
         );
       });
 
+      it('should collapse when out of focus if value in onBlur() is set to empty', async () => {
+        const { driver, inputDriver } = createDriver(
+          <ControlledSearch
+            options={options}
+            onBlur={function(event) {
+              this.setState({ value: '' });
```
argshook, line 206: this is real magic coming from the `makeControlled` utility, i'd suggest not to use it, cause it's not entirely clear what `this.setState` really does.

perhaps instead make a local reusable controlled version of Search and
reuse that, something like:

```
const MockSearch = () => {
  const [value, setValue] = React.useState('');

  return (
    <Search
      expandable
      value={value}
      onBlur={() => setValue('')}
    />
  );
};
```

src/Search/test/Search.spec.js
```diff
@@ -198,6 +198,54 @@ describe('Search', () => {
...
+            }}
+            expandable
+          />,
+        );
+        await inputDriver.click();
+        await inputDriver.enterText('fox');
+        await inputDriver.blur();
+        expect(await driver.isCollapsed()).toBe(true);
```
argshook, line 214: here, along other tests, it would be worthwhile to also check value in input with `inputDriver.getText()`

argshook (changes requested):

argshook:
> But probably null should not be used for value in a controlled component

correct, react throws a warning when `<input value={null}/>`:

```
Warning: `value` prop on `input` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components.
```

src/Search/Search.js
```diff
@@ -153,8 +153,9 @@ class Search extends Component {
     onFocus && onFocus(event);
   };
 
-  _onBlur = event => {
+  _onBlur = async event => {
     const { onBlur } = this.props;
+    onBlur && (await onBlur(event));
```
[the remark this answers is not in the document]
daugerdas: currently was not able to come up with a better solution to this. Ideas that I had:
- to write `Promise` instead of async/await
- to handle collapse logic inside `componentDidUpdate()`, but then there should be a boolean to check if element is not in focus/might be not expected UI behaviour
- inside of `onBlur()` to check if `this._getIsControlled()` and then `setTimeout` to `0` for checking collapse logic

Was thinking that the best option would be to have a callback function in `setState()`, but in this case `setState()` would be defined inside `onBlur()`, so it is not accessible

src/Search/test/Search.spec.js
daugerdas: A question about reusable controlled version of Search - would be better to have a custom `onChange()` there (as in eda6b89)? I reverted it as I thought that may affect `test5:types` and it does not have an affect on current unit tests

daugerdas:
`test5:types` test is failing with the following error:
`error: Your local changes to the following files would be overwritten by checkout: src/Search/Search.js`
Error still exists after reverting last commit, a commit before that (9cd6b57) passed `test5:types`

src/Search/test/Search.spec.js
argshook: changes in `.spec.js` files have no impact on `test5:types`, there's likely some other issue.

regarding if it's better to have `onChange` or not, it's up to you, neither is better or worse.

argshook: please try to pull latest `master` and rebase this branch on it. This should fix the `test5:types` issue

argshook (approved):
great effort!

PR #6172, opened 2020-11-03
Create an example of how to create standalone lib

barak007:
### 🔦 Summary
Created an example for standalone library.
Read the README.md inside `.lib`

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test

### DIFF
--- a/.lib/README.md
+++ b/.lib/README.md
@@ -1,0 +1,29 @@
+This folder contains examples for how to create a lib bundle that can be consumed stand alone
+
+### Two modes
+
+- single chunk - all code in one file
+- dynamic - all code is spited via import()
+
+for the dynamic mode there is a react library context (`lib.dynamic.context.js`) that can be use to preload the components
+
+### Notes:
+
+- no yoshi!
+- each example is built to a separate folder
+- just open the html file after you build the example
+- the dynamic example dose not handle errors
+- the dependencies of the webpack config are not in the package.json there are there because of yoshi. (I used yarn to install the project npm may dedupe differently)
+- I did not give all the components the right props so there are some errors related to that
+- I did not generated source maps (but it can be easily added)
+- Not all component are included (only what was easy to do with regex) we need to complete the list (they are commented out) 
+
+### How to build:
+
+At the repo root just
+
+`yarn webpack --config ".lib/webpack.lib.config.js"`
+
+or
+
+`yarn webpack --config ".lib/webpack.dynamic.config.js"`
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ npm-debug.log
 lib/
 test-kit-lib/
 dist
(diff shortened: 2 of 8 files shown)


.lib/README.md
```diff
@@ -0,0 +1,29 @@
+This folder contains examples for how to create a lib bundle that can be consumed stand alone
+
+### Two modes
+
+- single chunk - all code in one file
+- dynamic - all code is spited via import()
+
+for the dynamic mode there is a react library context (`lib.dynamic.context.js`) that can be use to preload the components
```
tomrav, line 8:
```suggestion
for the dynamic mode there is a react library context (`lib.dynamic.context.js`) that can be used to preload the components
```

tomrav (commented):
a couple of language fixes

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6172

shlomitc:
This looks like a good direction for a standalone library, especially the dynamic loading which I think is the right way to go.
The single bundle which doesn't dynamically import will probably not be used so I'm not sure we should even promote it.

1. Currently, the output is generated in `.lib/dist-dynamic` can we make the output files to be created in `dist/standalone-dynamic`? This will assure all the distributed output is in the same folder.
2. In order to prevent version collisions - can we prefix the hashes to contain the version number of wsr? this will verify the uniqueness.
3. We'll probably need to add a `.d.ts` file for the dynamic index but that's not urgent anyway
4. I'd like to make the `lib.dynamic.context` concept a bit more approachable for consumers. I imagined something like this from a consumer side:

```jsx
import React from 'react;
import {DynamicComponentsLoader, Components} from 'wix-style-react/dist/standalone-dynamic'

const SomePageInMyApp = () => {
  return (
    <DynamicComponentsLoader.Consumer>
      {({Button, Slider}) => {
        <>
           <Button>Click Me</Button>
           <Slider/>
        </>
      }
    </DynamicComponentsLoader.Consumer>
  );
}

const App () => {
  return (
    <DynamicComponentsLoader.Provider components={[Components.Button, Components.Slider]}>
       <SomePageInMyApp/>
    </DynamicComponentsLoader.Provider>
  )
}
```

The idea is pretty much similar, but one that requires any "globals" in the system. everything should be encapsulated under the imports from the library (even if they look for some global variable).
I guess that someone who knows hooks can make it look better, I'm just an old-fashioned render-props fan.

5. I guess that all the content of `lib.dynamic.context` can be moved to the `src` folder, no reason for it to be in the `.lib` folder. any blockers you can think of?



Besides that, I think this is a really good direction. Thanks a lot!

PR #6179, opened 2020-11-04
Business dashboard theme - circular progress bar adjustments

dimafeldman:
### 🔦 Summary

...

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


src/Themes/businessDashboard/components/CircularProgressBar.st.css
```diff
@@ -33,15 +54,22 @@
...
-  top: calc(50% - 6px);
+}
+
+.root::progressBar:size(medium) {
+  height: 56px;
+  width: 56px;
+}
+
+.root::progressBar:size(medium)::arcsContainer {
+  height: 56px !important;
```
dimafeldman, line 65: I couldn't find a "cleaner" way to override the progress bar size :(

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6179

src/Themes/businessDashboard/components/CircularProgressBar.st.css
shlomitc: Yes, that's quite nasty.
The reason behind it is that the `wix-ui-core` component uses inline style to apply a `size`.
To make it css only will require a change both in the core component and in the WSR component that uses it.

You can keep it that way but let's just add a comment here like:
`/* TODO - requires a change in wix-ui-core to remove inline sizes */ `

src/Themes/businessDashboard/docs/components/CircularProgressBar.js
```diff
@@ -5,7 +5,8 @@ export const playground = `
     <CircularProgressBar showProgressIndication size="small" value={50} />
```
shlomitc, line 5: IIRC, I think that you only need the medium size, so maybe we can just remove this one from the examples and from the CSS. Can you please verify that only the medium size is needed?

dimafeldman: yes, we will use only the medium size. will remove it

PR #6183, opened 2020-11-05
chore(package.json): unpin `yoshi-style-dependencies` version

argshook:
This PR unpins `yoshi-style-dependencies` `devDependency` version.

As a result `yoshi-style-dependencies` used will be latest minor
version, it's currently `4.228.0`.

It was pinned so that we use the last publicly available version. It
means contributors outside of Wix could install it.

Nonetheless, as discussed, we're moving forward with continuing to use
yoshi tools even after they went closed source. This is one step towards
it and is required to use `yoshi-flow-library`


argshook:
Unfortunately it's not as trivial as just unpinning the version, because of internally changed dependencies.

This change can happen together with integration with `yoshi-flow-library`.

closing

PR #6206, opened 2020-11-09
BusinessDashboard Theme: Fix FontUpgrade not matching theme active state

ronenst:
### 🔦 Summary

Wrapping in `<ThemeProvider/>` with `active: false` still causes the `FontUpgrade` to apply.

Maybe a safer fix would be?
```ts
   componentWrapper: ({ children }) => active ? <FontUpgrade>{children}</FontUpgrade> : <>{children}</>, 
```


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6206

adicicurel:
@ronenst There was a specific request to always apply FontUpgrade on BusinessDashboard theme by UX, 
Initially i did what you are writing in this PR.

shlomitc (approved):
LGTM

ronenst:
@adicicurel To clarify, having FontUpgrade active even if the theme is not set to active is intentional?

ronenst:
I can only add the `ThemeProvider` when the FT is enabled, but the current behaviour is very surprising. Once expects that setting `active` to `false` will mean no changes are made to the existing styling even if it's a new wrapper.

shlomitc:
I agree - this is a necessary change

adicicurel:
But now you need to pass two active states, one for the font upgrade and one for our theme, since those are two different specs.
Or remove the fontUpgrade from the theme and wrap it when you wrap with ThemeProvider

adicicurel:
According to UX (Ayelet Livne), when we launch, font upgrade would be a must for the dashboard, therefor we need it even if turning the theme off.

shlomitc:
@adicicurel - I think that these are two unrelated features.

PR #6213, opened 2020-11-10
LinearProgressBar - remove flaky test

shlomitc:
it constantly fails on the Business Dashboard theme and has the potential to fail also in the regular theme due to tooltip positioning which is known as problematic.

I think it would simply be better without it, it's not that critical anyway and covered in unit-tests


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6213

PR #6218, opened 2020-11-10
Preload Madefor in visual tests

EgoziE:
This should do the trick.
The font file is downloaded before the font definitions and before the page even loads.
image


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6218

shlomitc (approved):
Looks fine for now.

In general, we can think of a more generic solution because now you need to manually specify the hash, like `9b7def89`. This will need to be changed once we update the Madefor version.
image

An idea:
1. create a json manifest in `wix-fonts` and expose it in the `npm module`.
```json
{
  "normal": "media/WixMadeforTextVF_W_Wght.9b7def89.woff2",
  "italic": "media/WixMadeforTextItalicVF_W_Wght.d6461449.woff2"
}
```
and then create a `<link/>` tag dynamically for each url

EgoziE:
@shlomitc I see your point.
Fortunately a future update will not be automatic, so it won't just suddenly break.
Moreover I think we can try to ditch the hash, these files are unique in their unique folder anyway.

shlomitc:
> these files are unique in their unique folder anyway.
Right. need to check if yoshi allows it

PR #6224, opened 2020-11-10
<EditableSelector/> - storybook updates

vidmantasa:
Storybook updates for `EditableSelector`


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6224

src/EditableSelector/docs/examples.js
```diff
@@ -0,0 +1,167 @@
...
+            </Card>
+          </Col>
+        </Row>
+      </Container>
+    );
+  }
+}
+`;
+
+export const popoverWithEditableSelector = `
```
EgoziE, line 82: The second example is at the bottom of the page, so when you click the button, the popup opens towards down, which means it's cut and the user need to scroll to interact with it.
I suggest 2 solutions:
1. Change `<Popover/>` placement to be `right-end`, which will open the popup towards up.
2. Wrap the button with a `<Box/>` with enough height.

I don't mind which solution you choose, or pick your own idea if you have one.

EgoziE (changes requested):
Wow, this page got a real "makeover".
Awesome work!

PR #6233, opened 2020-11-11
<NumberInput /> - add hideStepper property

Oripi:
### 🔦 Summary

This adds a property `hideStepper` that hides the stepper (up/down arrows) when the property is set to `true` and a `suffix` or `prefix` element is provided.
The idea is to use the component for price inputs, to get all `Number` functionality without the irrelevant stepper behaviour.
slack discussion here: https://wix.slack.com/archives/CR7P4A1UP/p1604846942007600

### ✅ Checklist

- [] 👨‍💻 API change is approved by the UI Infra Developers  
- [x] 👨‍🎨 UX change is approved by the Design System UX @assafzi 
- 📚 Change is documented
  - [x] Story
  - [x] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [x] Visual test


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6233

src/NumberInput/NumberInput.js
```diff
@@ -122,16 +131,19 @@ class NumberInput extends React.PureComponent {
...
           <Input.Group>
             {suffix}
-            <Input.Ticker
-              onUp={this._increment}
-              onDown={this._decrement}
-              dataHook="number-input-ticker"
-              upDisabled={strict && max === value}
-              downDisabled={strict && min === value}
-            />
+            {!(hideStepper && (suffix || prefix)) && (
```
EgoziE, line 138: This logic is just a UX guideline.
We cannot have one prop to affect another's functionality, only `hideStepper={true}` should hide the stepper regardless suffix / prefix.

src/NumberInput/NumberInput.js
```diff
@@ -145,10 +157,13 @@ NumberInput.propTypes = {
   defaultValue: PropTypes.number,
   /** If set to true - typing values beyond `min`/`max` values will round to nearest range  */
   strict: PropTypes.bool,
+  /** If set to true - hides the stepper. A `suffix` or a `prefix` must be provided to hide the stepper */
```
EgoziE, line 160: Also change the description

EgoziE (changes requested):

src/NumberInput/NumberInput.js
Oripi: that's not what @assafzi said. according to him something similar happens in `ColorPicker`

Oripi: generally i prefer it like this too (why force something like this...?) but he needs to approve it

src/NumberInput/NumberInput.js
Oripi: if we change the above i'll change this too

src/NumberInput/NumberInput.js
EgoziE: I spoke with @milkyfruit about it, not everything that is right for UX is right for dev

assafzi: Spoke with @milkyfruit and Ayelet. Let's keep it as @EgoziE said, but please add an example of this with something like "When stepper is hidden you should display other numerical visual clues like affix/suffix".
That's acceptable @EgoziE ?

EgoziE: @assafzi Let's not leave documentation for contributors.
Can you please write the full example with description? then Ori will add it to the PR

assafzi: Sure!

### Title
Hide Stepper

### Description
In order to keep the component's functionality, another visual clue should be added, such as prefix or suffix.

### Example
`<NumberInput hideStepper prefix={<Input.Affix>$</Input.Affix>} placeholder="0" />`


@EgoziE is it OK?

EgoziE (approved):

PR #6249, opened 2020-11-15
<SidebarSectionItem/> - adjust and limit suffix size

dht:


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6249

shlomitc (changes requested):
The visual tests fail because due to your changes which is somewhat fine, however, I’m not sure that it’s completely fine

Take a look at this one:
https://wixeyes.applitools.com/app/test-results/00000251796833133573/00000251796831635429/steps/1?accountId=mE99eP8CUn9UTSXSv98L3OSaJ0NZmAcY6Wndx6WSHODdY110&mode=step-editor

1 - it seems like there is no margin between the text and the icon. according to the specs there should be 6 px separating between them:
image

2 - the suffix area seems to shrink when the text is long but I'm not sure it should be. due to that, the badge will display an ellipsis

dht:
@shlomitc 
Yes, AFAICS this is due to the `max-width:100%` on the badge's container which causes the width of the badge to be calculated after its parent's width. This makes the badge non-eager to take up space which results in the `min-width` "winning" rather than `max-width`.
How would you suggest to advance?
I can suggest two options:
1. stay with the one line change which is already implemented and working
2. add a "noEllipsis" property to the badge which disables the shortening mechanism. On the `bizmgr` side we can then send something like:

```jsx
<Badge noEllipsis={text.length <= 4}>{text}</Badge>
```

dht:
Tnx Shlomi, you're the best! 🤴

PR #6252, opened 2020-11-15
Split visual tests

EgoziE:


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6252

shlomitc:
I liked the separation but there are few issues:

1. The testing times are not slower, unfortunately. You can see that `test2:visual` still takes the same 30+ minutes. that's because we have a concurrency limit by applitools for all Wix and not just per run. I can suggest increasing the `concurrency` field in the `applitools.config.js` to 60 which is the maximum we can get at the moment.
2. debugging the visual tests now becomes a bit more annoying. Previously you just run `npm run storybook:visual`, but now you will need to specify the env variable which will require at least having some command in the package.json. I can suggest adding another `if` statement in the `config.js` that if no env variable is set - it will display all stories

EgoziE:
1. Done - Increased to 60
2. Done:
 - test:visual - run tests without a parameter (All tests)
 - test2:visual - run `test:visual` with `base` param (Just base components)
 - test8:chromatic - run `test:visual` with `themes` param (Just themes)

EgoziE:
@shlomitc Update:
I don't see any improvement in runtime, but I'd like to merge this change regardless.

shlomitc:
So let’s merge but no need to split to test8. Simply chain them in the same test2 command

EgoziE:
Can you help me with that?
How can I make sure to chain it so they will run in parallel?

Because chain them with && will just run them one after the other:
```
"test2:visual": "STORYBOOK_VISUAL=base npm run test:visual && STORYBOOK_VISUAL=themes npm run test:visual"
```

EgoziE:
Ok nevermind, I see that using only a single & will do the trick 🤦🏼

shlomitc:
wait - if the tests are running anyway in the same command, let's just simply run them in the main task, let's drop STORYBOOK_VISUAL

PR #6260, opened 2020-11-16
infra: remove sync-ts

shlomitc:
It's been in the project for a while now and it didn't assist in detecting required changes.
We mostly encounter false-positives (files that don't require a .d.ts change) or failing builds for incorrect API.

I think that we're disciplined enough to verify that types are added and eventually we will start migrating to typescript.


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6260

PR #6263, opened 2020-11-17
Themes | Business Dashboard - added box-shadow theming

dimafeldman:
### 🔦 Summary

...

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test

### DIFF
--- a/src/Foundation/stylable/shadows.st.css
+++ b/src/Foundation/stylable/shadows.st.css
@@ -1,3 +1,5 @@
+@st-global-custom-property --shadow10, --shadow20, --shadow30, --shadow40;
+
 :vars {
   s1: rgba(22, 45, 61, 0.12);
   s2: rgba(22, 45, 61, 0.48);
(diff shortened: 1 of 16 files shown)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6263

src/Foundation/stylable/shadows.st.css
```diff
@@ -6,9 +8,9 @@
...
   s6: rgba(22, 45, 61, 0.3);
   s7: rgba(22, 45, 61, 0.36);
-
-  shadow10: 0 2px 1px 0 value(s2), 0 0 3px 0 value(s1);
-  shadow20: 0 2px 4px 0 value(s1), 0 0 6px 0 value(s1);
-  shadow30: 0 6px 6px 0 value(s3), 0 0 18px 0 value(s1);
-  shadow40: 0 8px 8px 0 value(s1), 0 3px 24px 0 value(s4);
 }
+
+.shadow10 { box-shadow: var(--shadow10, 0 2px 1px 0 value(s2), 0 0 3px 0 value(s1)); }
```
EgoziE, line 13: We prefer that design constants will not be mixins if the css is not complex.
Please change is to:
```
:vars {
  shadow10: var(--shadow10, 0 2px 1px 0 value(s2), 0 0 3px 0 value(s1));
}
```

and use it like before:

```
box-shadow: value(shadow10);
```

EgoziE (changes requested):

src/Foundation/stylable/shadows.st.css
dimafeldman: updated. thanks!

src/Themes/businessDashboard/BusinessDashboard.st.css
```diff
@@ -1,3 +1,10 @@
+@st-global-custom-property --shadow10, --shadow20, --shadow30, --shadow40;
```
EgoziE, line 1: Oh sorry just now I noticed that it's global.
This means you can accidentally override those from an external css source.
I suggest to remove `@st-global-custom-property` and import those variables just like you import any other variable.

EgoziE (changes requested):

src/Themes/businessDashboard/BusinessDashboard.st.css
EgoziE: Ok so I reached a conclusion it's not possible use scoped css variables as a stylable variable value.
Bottom line, let's keep those variables as global, BUT in order to reduce any possible overrides, let's change the name from `--shadow-10` to `--wsr-shadow-10`

dimafeldman: updated, thanks!

EgoziE (approved):

PR #6319, opened 2020-11-26
<Tag/> - add lightOutlined theme

splintor:
### 🔦 Summary
1. Add `lightOutlined` theme to `<Tag>` component, according to [this UX](https://zpl.io/bzq6pJE). This is done to implement **Channels indication** ([UX](https://share.goabstract.com/5877c8a2-7d85-4a0d-bdb5-db11d2bb928d?collectionId=b72f94b0-3cda-4d53-b677-4cfa55457dd7&collectionLayerId=69e79831-655a-4a28-b139-97e579773538&mode=build&selected=2892704313-8C555CB5-F234-4758-AC6C-99BDCE7F09DB&sha=bc3dc76d50c1143a0a798310077c9271157a66de)).
1. After I noticed that setting `size` to my `Tag`s in `TagList` had no effect, I realized I need to set the `size` of the `TagList`, so I updated the description of the `<Tag>`'s `size` prop (and also `onRemove`), so others will be aware of it.
1. Fix a bug where `Tag`s with no `onClick` handler are hoverable. They should not be hoverable, as nothing happens when the mouse hover over them.

Note: Because the `lightOutlined` theme adds `1px` border to the `Tag`, I reduced the padding in `1px` to keep the `Tag` in the same dimensions. However, for some sizes and dimensions, the padding was `0`, so I left it as is, which means it might make `Tag`s with `lightOutlined` theme slightly bigger. Let me know if you want me to handle it in another way.

### ✅ Checklist

- [X] 👨‍💻 API change is approved by the UI Infra Developers (@EgoziE)
- [X] 👨‍🎨 UX change is approved by the Design System UX (@milkyfruit)
- 📚 Change is documented
  - [X] Story
  - [X] API description
  - 🔬 Change is tested
  - [X] Component
  - [X] Visual test

### DIFF
--- a/src/Tag/Tag.js
+++ b/src/Tag/Tag.js
@@ -128,13 +128,13 @@ Tag.propTypes = {
    * and mouse event as second parameter when clicking on Tag */
   onClick: PropTypes.func,
 
-  /** Callback function that pass `id` property as parameter when removing the Tag  */
+  /** Callback function that pass `id` property as parameter when removing the Tag. If the Tag is rendered inside of a `TagList`, its `onRemove` handler is set by the TagList's `onTagRemove` prop  */
   onRemove: PropTypes.func,
 
   /** If the Tag is removable then it will contain a small clickable X */
   removable: PropTypes.bool,
 
-  /** The height of the Tag */
+  /** The height of the Tag. If the Tag is rendered inside of a `TagList`, its `size` is set by the TagList's `size` prop */
   size: PropTypes.oneOf(['tiny', 'small', 'medium', 'large']),
 
   /** theme of the Tag */
--- a/src/Tag/Tag.js
+++ b/src/Tag/Tag.js
@@ -87,7 +87,7 @@ class Tag extends React.PureComponent {
         [styles.withThumb]: thumb,
         [styles.disabled]: disabled,
         [styles.clickable]: onClick !== noop,
-        [styles.hoverable]: !disabled,
+        [styles.hoverable]: !disabled && onClick !== noop,
       },
     );
   }
--- a/src/Tag/Tag.js
+++ b/src/Tag/Tag.js
@@ -145,6 +145,7 @@ Tag.propTypes = {
     'dark',
     'neutral',
     'light',
+    'lightOutlined',
   ]),
 
   /** An optional thumb to display as part of the Tag */
--- a/src/Tag/Tag.js
+++ b/src/Tag/Tag.js
@@ -128,13 +128,13 @@ Tag.propTypes = {
    * and mouse event as second parameter when clicking on Tag */
   onClick: PropTypes.func,
 
-  /** Callback function that pass `id` property as parameter when removing the Tag. If the Tag is rendered inside of a `TagList`, its `onRemove` handler is set by the TagList's `onTagRemove` prop  */
+  /** Callback function that pass `id` property as parameter when removing the Tag  */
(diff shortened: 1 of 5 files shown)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6319

milkyfruit (changes requested):
@splintor everything looks great except what you mentioned about a height for `size="tiny"` — it should be 18px height, when it's set to `lightOutlined` it becomes 20px high. Adding `max-height: 18px` fixes the issue. Since tag component doesn't allow multiple lines, we can add max height to all sizes:

- tiny — max-height: 18px
- small — max-height: 24px
- medium — max-height: 30px
- large — max-height: 36px

splintor:
@milkyfruit max-heights added.

src/Tag/Tag.js
```diff
@@ -128,13 +128,13 @@ Tag.propTypes = {
    * and mouse event as second parameter when clicking on Tag */
   onClick: PropTypes.func,
 
-  /** Callback function that pass `id` property as parameter when removing the Tag  */
+  /** Callback function that pass `id` property as parameter when removing the Tag. If the Tag is rendered inside of a `TagList`, its `onRemove` handler is set by the TagList's `onTagRemove` prop  */
```
EgoziE, line 131: Tag is a standalone component that should not be aware of the existence of TagList.
If you want to add this comment, I recommend to move it to TagList.

EgoziE (approved):

milkyfruit (approved):
@EgoziE @splintor looks great, we can merge it now.

src/Tag/Tag.js
splintor: I understand your concern about `Tag` being a standalone component, so I removed these comments. I just want to say that when I tried to change the size of the tags in my `TagList`, my initial intuition was to change their `size` prop, and when it didn't make a change, I looked up `Tag` documentation, not `TagList`. When this didn't help, I turned to the WSR code, and realized this is set by `TagList`. So I'm not sure documenting it in `TagList` props description would help.

Actually, I really think that if `TagList`'s `tags` specify a `size` instead of `TagList`, it should be used in rendering the `Tag`s in the `TagList` ([Principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_surprise)). This would also help me in rendering `tiny` sized tags in a `TagList` without [fighting with TypeScript](https://wix.slack.com/archives/G01FL265WMQ/p1606400091008200?thread_ts=1606375933.000100&cid=G01FL265WMQ), but I understand there might be backward compatibility issues as the `size` prop of the `tags` in `TagList` is currently being ignored. What do you think?

EgoziE: First of all I agree.
In order to do that officially:
- Check usage of `TagList` in different places to make sure there's no specific reason why not making this prop more global.
- Add the prop to `TagList` directly (+ docs)
- Add deprecation message to adding size to `Tag` directly.

Can be done as a separate task.

splintor: > Add the prop to TagList directly

I think there is a misunderstanding here. `TagList` already has a `size` prop. The problem is that developers might set the `size` prop in the objects of its `tags` prop (like they do with other `Tag` props), and won't understand why setting the `size` doesn't affect the tags size. My suggestion is that if `TagList` `size` prop is not specified, and it is specified in the `tags` object, then it should be used.

PR #6324, opened 2020-11-26
Use places autocomplete

gendeld:
### 🔦 Summary

...

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6324

src/hooks/usePlacesAutocomplete/usePlacesAutocomplete.spec.js
```diff
@@ -0,0 +1,244 @@
...
+    const useInitFn = () => {
+      const [ready, setReady] = useState(false);
+      useEffect(() => {
+        setTimeout(() => setReady(true), 200);
+      }, []);
+      return { ready };
+    };
+    const client = { fetchPredictions, useInit: useInitFn };
+    const result = renderHelper({ client });
+    expect(result.current.ready).toBe(false);
```
shlomitc, line 89: what exactly is this `current` thing? It's repeated all across the tests. can't the `renderHelper` just return it?

src/hooks/usePlacesAutocomplete/usePlacesAutocomplete.js
```diff
@@ -0,0 +1,87 @@
...
+   * so we make latest options available through refs */
+  const requestOptionsRef = useLatest(requestOptions);
+  const clientOptionsRef = useLatest(clientOptions);
+
+  const clearPredictions = useCallback(
+    () => setFetchState(initialFetchState),
+    [],
+  );
+
+  const getPredictionsRef = useRef(
```
shlomitc, line 46: I don't like that a `getPredictionsRef` actually changes the prediction state.
we use the `get` prefix for getters.
Also, I think it's best that it would actually return a value and not the ref.

src/hooks/usePlacesAutocomplete/usePlacesAutocomplete.spec.js
```diff
@@ -0,0 +1,244 @@
...
+  let timeoutId;
+  return (...args) => {
+    clearTimeout(timeoutId);
+    timeoutId = setTimeout(() => {
+      timeoutId = null;
+      callback(...args);
+    }, delay);
+  };
+});
+jest.useFakeTimers();
```
shlomitc, line 18: Instead of mocking, I think it's better to make the original hook configurable. that way you don't care about the actual implementation

src/hooks/usePlacesAutocomplete/usePlacesAutocomplete.spec.js
```diff
@@ -0,0 +1,244 @@
...
+
+const wait = milliseconds =>
```
shlomitc, line 20: this probably repeats in multiple areas in the system, simply reuse an existing one

src/hooks/usePlacesAutocomplete/usePlacesAutocomplete.spec.js
```diff
@@ -0,0 +1,244 @@
...
+    const client = { fetchPredictions: fetchPredictionsFn, useInit };
+    const result = renderHelper({ client });
+    expect(result.current.loading).toEqual(false);
+    act(() => result.current.setValue('test'));
+    act(() => jest.advanceTimersByTime(defaultDebounceMs));
+    expect(result.current.loading).toEqual(true);
+    await act(async () => {
+      jest.advanceTimersByTime(fetchDelay);
+      await flushPromises();
+    });
```
shlomitc, line 109:
```suggestion
    await waitForDebounce(fetchDelay)
```

shlomitc (changes requested):

src/hooks/usePlacesAutocomplete/usePlacesAutocomplete.js
```diff
@@ -0,0 +1,87 @@
...
+
+      const newPredictions = await fetchPredictions(
+        val,
+        requestOptionsRef.current,
+        clientOptionsRef.current,
+      );
+
+      // Don't update state if new fetch request has been initiated or if component was unmounted
+      if (!checkUnmounted() && requestId === predictionsRequestId.current) {
+        setFetchState({ loading: false, predictions: newPredictions });
```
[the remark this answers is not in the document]
gendeld: As far as I know, React doesn't know to batch state updates inside asynchronous callbacks, so updating two states will cause an extra re-render.
Because these states are linked, it makes sense to combine them.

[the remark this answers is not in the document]
shlomitc: interesting. I obviously lack some hooks knowledge, but this sounds like something that can be solved, so let's discuss it later in private

src/hooks/usePlacesAutocomplete/usePlacesAutocomplete.spec.js
gendeld: `@testing-library/react-hooks` saves result as ref and updates it as it changes, so when we want to recheck return value of hook throughout a test we can always use `result.current` which will return the latest value.

src/hooks/usePlacesAutocomplete/usePlacesAutocomplete.js
```diff
@@ -0,0 +1,93 @@
...
+    loading,
+    /** (value: string, requestOptions?: RequestOptions) => void
+     * fetches predictions for given value from client (debounced)
+     * and sets results to prediction state.
+     *
+     * Can also receive requestOptions,
+     * which are client specific options to pass to the request */
+    updatePredictions,
+    /** function that clears predictions array and sets loading state to false */
+    clearPredictions,
```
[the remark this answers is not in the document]
gendeld: I think it's probably better to have clearing not debounced, I now also increase the request id on clear to stop async updates that were started before clear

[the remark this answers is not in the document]
shlomitc: yes, clearing should be immediate (and probably stop any denounced action)

PR #6326, opened 2020-11-27
<Radio /> - new component

Zivilekin:
Extract Radio button from RadioGroup in order to use it separately.


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6326

src/Radio/Radio.js
```diff
@@ -0,0 +1,76 @@
...
+  return (
+    <RadioButton
+      className={st(classes.root, { alignItems }, className)}
+      data-hook={dataHook}
+      checked={checked}
+      label={renderLabel}
+      onChange={onChange}
+      disabled={disabled}
+      name={name}
+    ></RadioButton>
```
mykas, line 37: RadioButton does not accept children so you can just write here as: `<RadioButton {...props etc} />`

src/Radio/Radio.st.css
```diff
@@ -0,0 +1,87 @@
...
+  -st-default: RadioButton;
+}
+
+.root {
+  -st-extends: RadioButton;
+  -st-states: alignItems(enum(top, center));
+  cursor: initial;
+  display: flex;
+  flex-direction: row;
+  cursor: inherit;
```
mykas, line 18: Duplication of cursor css

src/Radio/Radio.st.css
```diff
@@ -0,0 +1,87 @@
...
+  left: 0;
+  margin: 2px;
+}
+
+.root::label {
+  cursor: pointer;
+  padding-left: 12px;
+}
+
+:global([dir='rtl']) .root::label {
```
mykas, line 52: Please move this global to the bottom

src/Radio/Radio.st.css
```diff
@@ -0,0 +1,87 @@
...
+.root:focus-visible::icon {
+  border-color: value(THEME-COLOR-10);
+  box-shadow: 0 0 0 3px value(F00);
+}
+
+.root:not(:checked)::icon:hover:not(:active) {
+  background-color: value(THEME-COLOR-40);
+}
+
+.root:checked::icon::after {
```
mykas, line 66: Hm I don't think that after is a class in RadioButton in wix-ui-core. Pseudoclasses in stylable are targeted only with one `:`

src/Radio/Radio.st.css
```diff
@@ -0,0 +1,87 @@
...
+  cursor: pointer;
+  padding-left: 12px;
+}
+
+:global([dir='rtl']) .root::label {
+  padding-left: 0;
+  padding-right: 12px;
+}
+
+.root:focus-visible::icon {
```
mykas, line 57: Lets move different states of the `.root` element to top and the targeted classnames like `icon` to the bottom

src/Radio/Radio.uni.driver.js
```diff
@@ -0,0 +1,32 @@
...
+     * @returns {Promise<string>}
+     */
+    getCountText: () => findByHook(base, dataHooks.radioCount).text(),
+
+    /**
+     * Clicks the button
+     * @param {number} times Times to click
+     * @returns {Promise<void>}
+     */
+    clickButtonTimes: async times => {
```
mykas, line 19: Hm I believe that Radio component is not actually a button. Also why do you think we need to click many times the component?

src/Radio/Radio.js
```diff
@@ -0,0 +1,76 @@
...
+  label: PropTypes.node,
+
+  /** Unique name of radio */
+  name: PropTypes.node,
+
+  /** A callback when radio is selected */
+  onChange: PropTypes.func,
+
+  /** Align of radio button. Default is center */
+  alignItems: PropTypes.oneOf(['top', 'center']),
```
mykas, line 66: Thats not exactly align of radio, but rather radio alignment with label

src/Radio/Radio.js
```diff
@@ -0,0 +1,76 @@
...
+  );
+};
+
+Radio.displayName = 'Radio';
+
+Radio.propTypes = {
+  /** Applied as data-hook HTML attribute that can be used in the tests */
+  dataHook: PropTypes.string,
+
+  /** A css class to be applied to the component's root element */
```
mykas, line 47: We never document classNames in our docs and also do not include it in propTypes.

src/Radio/Radio.meta.js
```diff
@@ -0,0 +1,8 @@
+import Radio from './Radio';
```
mykas, line 1: This file is required for many purposes, but one of them is making sure that this component renders correctly in SSR. Lets add a simulation that would be tests by `npm run sanity` in our build step: 

```js
metadata.addSim({
   checked: true
});
```

src/Radio/docs/index.story.js
```diff
@@ -0,0 +1,102 @@
...
+          example({
+            title: 'Structure',
+            text:
+              'Radio button consists of a radio itself and a label on the side.',
+            source: examples.structure,
+          }),
+          example({
+            title: 'States',
+            text:
+              'Radio button has three states - default, `checked` and `disabled`.',
```
mykas, line 67: default is lonely here without the `

src/Radio/docs/index.story.js
```diff
@@ -0,0 +1,102 @@
...
+            source: examples.states,
+          }),
+          example({
+            title: 'Label',
+            text:
+              'Label of a radio is a `node`. Default value of it is a string, but it can be overridden with any other component.',
```
mykas, line 73: Label value accepts node or string based values. It is recommended to use string as it comply with design systems typography.

mykas (approved):

PR #6330, opened 2020-11-29
<BadgeSelectItem/> remove extra margin-right when subtitle is enabled and expose size and suffix props

bamai:
### 🔦 Summary

https://jira.wixpress.com/browse/DSM-1139


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6330

src/BadgeSelectItem/BadgeSelectItem.js
```diff
@@ -80,6 +95,9 @@ export const badgeSelectItemBuilder = ({
       selected={selected}
       highlighted={hovered}
       disabled={disabled}
+      size={size}
+      suffix={suffix}
+      prefixMarginRight={prefixMarginRight ? prefixMarginRight : '0px'}
```
TalZa1810, line 100: If it's a capability that we want to add to the `ListItemSelect`, we should add this check to the ListItemSelect itself and not here

src/BadgeSelectItem/docs/index.story.js
```diff
@@ -87,13 +87,33 @@ import { badgeSelectItemBuilder } from 'wix-style-react';
             source: examples.subtitleExample,
           }),
 
+          example({
+            title: 'Suffix',
+            text: '',
```
TalZa1810, line 92: Let's add some description here. You can consult @milkyfruit what's an appropriate description for this kind of example.

src/BadgeSelectItem/BadgeSelectItem.js
```diff
@@ -52,6 +58,12 @@ BadgeSelectItem.propTypes = {
   highlighted: PropTypes.bool,
   /** If true, the item is disabled */
   disabled: PropTypes.bool,
+  /** Any suffix */
+  suffix: PropTypes.node,
+  /** Changing text size */
+  size: PropTypes.oneOf(['small', 'medium']),
+  /** allows to set prefix's margin-right */
+  prefixMarginRight: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
```
TalZa1810, line 66: I'm not sure how flexible we want this to be. Maybe just `0px` or `6px` instead of being too flexible. 
@milkyfruit WDYT? 
If it's just two possible options, we would probably want to rename this as well.

src/ListItemSelect/ListItemSelect.js
```diff
@@ -139,14 +146,19 @@ class ListItemSelect extends React.PureComponent {
...
         {prefix && (
-          <Text
-            className={st(classes.prefix, { subtitle: Boolean(subtitle) })}
-            dataHook={dataHooks.PREFIX}
-            {...textProps}
-            ellipsis={false}
-          >
-            {prefix}
-          </Text>
+          <Box marginRight={prefixMarginRight}>
```
TalZa1810, line 149: Why using both Box and css?
We want to avoid using Box wrappers when possible (for the same reason we had to refactor `<ListItemSelect/>` for the `<AddressInputOption/>` issue).

src/ListItemSelect/docs/index.story.js
```diff
@@ -196,6 +196,42 @@ import { listItemSelectBuilder } from 'wix-style-react';
               `,
           }),
 
+          example({
```
TalZa1810, line 199: Are these the examples you've discussed with Domas? Can you please elaborate?

TalZa1810 (changes requested):
Hi Mai,
I've added some questions and notes for things I wasn't sure about.
Feel free to send me a message if something is not clear.

src/BadgeSelectItem/BadgeSelectItem.js
milkyfruit: if it would be `true` or `false` would it have the same default state for all the options or each option could have a different value?
Right now it's `0` for title option, and `6px` for title+subtitle option.

shlomitc: I'm not sure why this is needed. can't the BaseSelect component customize the ListItemSelect using css only?

TalZa1810: It's a capability that @milkyfruit wanted us to add to the `ListItemSelect` in general, not just the `BadgeSelectItem`. We wanted this capability to be customisable by the user. 

@milkyfruit do we have another example besides the `BadgeSelectItem` where the user would need/want to customise this?

milkyfruit: at the moment, no. But I see that the same issue might occur also using other prefixes.

shlomitc: @milkyfruit - by other prefixes you mean other predefined list items?
if so, then we can always encapsulate it in each list item.
we can always open such a prop, but I want to make sure it makes sense

milkyfruit: Bookings use subtitle only when additional info exists:
image

I see it can happen with `ListItemSelect` when used with avatar or other prefix:
image

Making it as a string instead of a boolean would save us a trouble, because it's already controlled by subtitle prop:
image
image

src/ListItemSelect/docs/index.story.js
milkyfruit: The example can be improved to this, to explain when this prop should be used:

- [ ] PrefixMarginRight
`prefixMarginRight` prop allows to control prefix's margin-right. It's used to align text when a dropdown contains items with and without a subtitle.

```
<Layout cols={1} gap={0}>
  <ListItemSelect
    title="Prefix margin right is 0 (default when no subtitle)"
    prefix={
      <Box>
        <Icons.Toolbox />
      </Box>
    }
  />
  <ListItemSelect
    title="Prefix margin right was overriden to 6px (now it aligns text aligns with bottom item)"
    prefixMarginRight="6px"
    prefix={
      <Box>
        <Icons.Toolbox />
      </Box>
    }
  />
  <ListItemSelect
    title="Prefix margin right is 6px (default when no subtitle)"
    subtitle="default when there ia a subtitle"
    prefix={
      <Box>
        <Icons.Toolbox />
      </Box>
    }
  />
</Layout>;
```

src/BadgeSelectItem/docs/index.story.js
milkyfruit: - [ ] Suffix
Component has a suffix area. If plain text or icon is inserted, component automatically inverts the color when selected.

src/BadgeSelectItem/BadgeSelectItem.js
bamai: Hi :)
So how should this be implemented? Do we want to expose a new prop or add changes in css?
@shlomitc @milkyfruit

shlomitc: What I see here is only when the subtitle appears, which is already covered in the `ListItemSelect` subtitle prop.
If that's the only case, then maybe the `BadgeSelectItem` should have a subtitle prop as well.

If no other objections, we can remove this prop

bamai:
Currently the additional prefix margin is removed only in BadgeSelectItem. Does it need to be changed in ListItemSelect as well?

src/BadgeSelectItem/BadgeSelectItem.js
```diff
@@ -14,13 +14,15 @@ const BadgeSelectItem = props => {
...
+    size,
+    suffix,
   } = props;
 
   const prefix = <div className={st(classes.marker, { skin })} />;
 
   return (
     <ListItemSelect
-      className={className}
+      className={st(classes.root, {}, className)}
```
TalZa1810, line 25:
```suggestion
      className={st(classes.root, className)}
```

With stylable 3 no need for an empty object.

TalZa1810:
We've decided that currently we'll fix this just with `BadgeSelectItem`. If we decide to fix this with `ListItemSelect`, we should create a dedicated task for it and update the zeplin spec.

milkyfruit (approved):

src/BadgeSelectItem/test/BadgeSelectItem.visual.js
```diff
@@ -69,6 +71,69 @@ const tests = [
       },
     ],
   },
+  {
+    describe: 'badge select sizes',
```
TalZa1810, line 75: let's add the size explicit here. You can do it like it's done with the `skins` describe

src/BadgeSelectItem/test/BadgeSelectItem.visual.js
```diff
@@ -69,6 +71,69 @@ const tests = [
...
+          suffix: 'suffix',
+        },
+      },
+      {
+        it: 'node suffix',
+        props: {
+          id: 3,
+          text: 'standard',
+          subtitle: `subtitle text`,
+          suffix: (
```
TalZa1810, line 126: Let's add a simple `<div/>` with inline styles here rather than a component. Than, if there's an issue with the `Badge` component, it will not make this test fail as well.

TalZa1810 (approved):
Two tiny comments. After you fix this, it can be merged :)

PR #6344, opened 2020-12-01
useCopyClipboard - new custom hook

daugerdas:
### 🔦 Summary

Add a custom hook `useCopyClipboard` primarily to be used in a new component (`<CopyClipboard/>`) to copy strings to clipboard

Description: [DSM-769](https://jira.wixpress.com/browse/DSM-769)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6344

src/providers/useCopyClipboard/index.d.ts
```diff
@@ -0,0 +1,17 @@
+export interface useCopyClipboardProps {
+  value: string;
+  onCopy?: Function;
+  stateResetInterval?: number;
```
mykas, line 4: Lets name this as resetInterval instead

src/providers/useCopyClipboard/useCopyClipboard.js
```diff
@@ -0,0 +1,89 @@
...
+    };
+
+    const reselectPrevious = toggleSelection();
+    const container = newContainer.current;
+    registerCopyEvent();
+    copy();
+    cleanup();
+  }, [value, onCopy, range, newContainer, selection]);
+
+  const reset = () => {
```
mykas, line 70: Is this function used somewhere?

src/providers/useCopyClipboard/useCopyClipboard.js
```diff
@@ -0,0 +1,89 @@
...
+        } else {
+          selection.removeAllRanges();
+        }
+        if (newContainer) document.body.removeChild(newContainer.current);
+      }
+      // Toggle selection is needed in case user has selected an item on the page on the time of executing copying to clipboard
+      reselectPrevious();
+    };
+
+    const reselectPrevious = toggleSelection();
```
mykas, line 63: What for is this variable? Why not just rename toggleSelection to reselectPrevious instead?

src/providers/useCopyClipboard/useCopyClipboard.js
```diff
@@ -0,0 +1,89 @@
...
+    const container = newContainer.current;
```
mykas, line 64: What for is this const here?

src/providers/useCopyClipboard/test/useCopyClipboard.e2eStory.js
```diff
@@ -0,0 +1,50 @@
...
+  return (
+    <Layout>
+      <Cell>
+        <Input
+          value={value}
+          onChange={event => {
+            setValue(event.target.value);
+          }}
+          dataHook={storySettings.dataHookInputCopy}
+          onInputClicked={() => copyToClipboard()}
```
mykas, line 26: Just pass copyToClipboard without the `()`

src/providers/useCopyClipboard/useCopyClipboard.js
```diff
@@ -0,0 +1,89 @@
...
+ * @prop {function} copyToClipboard - a return function that allows to initiate copying
+ * @prop {function} resetState - an optional return function
+ */
+
+/**
+ * A custom hook for copying to clipboard. Returns copying to clipboard status and function to initiate copying to clipboard
+ * @param {Object} props -  an object for value, onCopy, and resetInerval props
+ * @param {string} props.value -  a string that should be copied to clipboard
+ * @param {function} [props.onCopy] - an optional parameter to apply additional logic when copying to clipboard
+ * @param {number} [props.resetInterval] - an optional parameter to set interval which after copying to clipboard will reset status
```
mykas, line 16: I believe resetTimeout is the better name here. Interval is something between period of time and another. Timeout is sort of delay.

src/providers/useCopyClipboard/useCopyClipboard.js
```diff
@@ -0,0 +1,89 @@
...
+      container.style.msUserSelect = 'text';
+      container.style.userSelect = 'text';
+      container.addEventListener('copy', event => {
+        event.stopPropagation();
+        if (onCopy) {
+          onCopy();
+        }
+      });
+      document.body.appendChild(container);
+      return container;
```
mykas, line 41: Why do we need to return a container here?

src/providers/useCopyClipboard/test/useCopyClipboard.e2eStory.js
daugerdas: great, thank you!

src/providers/useCopyClipboard/useCopyClipboard.js
daugerdas: It is a function returned by `toggleSelection()` which can be used to restore previously highlighted/selected area. `reselectPrevious()` is then called after copying to clipboard. 

https://github.com/wix/wix-style-react/blob/ce82661e5da11f6291c8f972e7bf248a1e17d0e9/src/providers/useCopyClipboard/useCopyClipboard.js#L50-L60

`toggleSelection()` also unselects current selection - otherwise the selecion will copied instead of the defined value. I've now added a comment in the code next to it, but probably would be great to have a better naming for these

src/providers/useCopyClipboard/useCopyClipboard.js
daugerdas: `reset` is one of the return values of the hook, it could be used to manually reset `isCopied` value inside a component that uses this hook. 

I've now reused it in `useEffect()` for a timeout, perhaps there might be some use cases for this in a component, but it can be removed as there's already a timeout prop

https://github.com/wix/wix-style-react/blob/751093b32cacc6f2f06f8060860fbb53ddd8b18f/src/providers/useCopyClipboard/useCopyClipboard.js#L76-L78

src/providers/useCopyClipboard/useCopyClipboard.js
daugerdas: oh true, return value is not used currently, removed it. Thank you!

src/providers/useCopyClipboard/useCopyClipboard.js
daugerdas: It is only used in `registerCopyEvent()` to be reused for assigning properties, so moved it there now

https://github.com/wix/wix-style-react/blob/9635a6744e0f4df70dd7801725d010bfaf0156e0/src/providers/useCopyClipboard/useCopyClipboard.js#L27-L34

src/providers/useCopyClipboard/useCopyClipboard.js
mykas: Oh I did not notice that. Its worth reusing it inside useEffect that suppose to reset state too

src/providers/useCopyClipboard/useCopyClipboard.js
```diff
@@ -0,0 +1,88 @@
...
+    let timeout;
+    if (isCopied && resetTimeout) {
+      timeout = setTimeout(reset, resetTimeout);
+    }
+    return () => {
+      clearTimeout(timeout);
+    };
+  }, [isCopied, resetTimeout]);
+
+  useEffect(() => () => setCopied(null), [value]);
```
mykas, line 83: Lets use reset() here

src/providers/useCopyClipboard/useCopyClipboard.js
mykas: But when should it unselect the selection? Maybe during the copy command?

daugerdas: oh true, now I've added it in `copy()` before new content is selected, and added function to select previous content (after `copy` command is executed)

https://github.com/wix/wix-style-react/blob/aafcb9bc81a85192f3300d04f3c5b7333103df85/src/providers/useCopyClipboard/useCopyClipboard.js#L44-L51

src/providers/useCopyClipboard/useCopyClipboard.js
daugerdas: great, done

src/providers/useCopyClipboard/index.d.ts
```diff
@@ -0,0 +1,17 @@
...
+  resetTimeout?: number;
+}
+
+export interface useCopyClipboardReturn {
+  isCopied: boolean;
+  copyToClipboard: () => void;
+  resetState: () => void;
+}
+
+declare const useCopyClipboard: (
```
mykas, line 13: Export const

src/providers/useCopyClipboard/index.d.ts
```diff
@@ -0,0 +1,17 @@
...
+  props: useCopyClipboardProps,
+) => useCopyClipboardReturn;
+
+export default useCopyClipboard;
```
mykas, line 17: Lets delete this one

mykas (approved):
Bravo, great job 🚀

PR #6364, opened 2020-12-08
useCopyClipboard - fix and update hook

daugerdas:
### 🔦 Summary

Fix and update `useCopyClipboard` custom hook, related to https://github.com/wix/wix-style-react/pull/6344


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6364

mykas (approved):

src/providers/useCopyClipboard/useCopyClipboard.js
```diff
@@ -19,32 +19,26 @@ import { createToggleSelection } from './createToggleSelection';
...
     const registerCopyEvent = () => {
-      const container = newContainer.current;
-      container.style.all = 'unset';
-      container.textContent = value;
-      container.style.whiteSpace = 'pre';
-      container.style.MozUserSelect = 'text';
-      container.style.msUserSelect = 'text';
-      container.style.userSelect = 'text';
-      container.addEventListener('copy', event => {
+      copyElement.addEventListener('copy', event => {
```
gendeld, line 29: I see we are adding the event listener but not removing it.
event listeners can persist even after elements are removed from DOM - as long as there is a reference in modern browsers and possibly even causing memory leaks in less maintained ones -
so it is a good practice to remove them before removing element

daugerdas: great to know, thank you! Now added a removal for event listener before element is removed (456645f)

PR #6366, opened 2020-12-08
Suggestion - Add deprecation log to old drivers

EgoziE:
In order to push our consumers to move to uni-drivers, I suggest adding a deprecation warning to the old testkit.
This is just a suggestion, I'd like to hear your thoughts.

### DIFF
--- a/src/AddItem/AddItem.driver.js
+++ b/src/AddItem/AddItem.driver.js
@@ -2,7 +2,16 @@ import textDriverFactory from '../Text/Text.driver';
 import { tooltipTestkitFactory } from 'wix-ui-core/dist/src/testkit';
 import { dataHooks } from './constants';
 
-const addItemDriverFactory = ({ element, eventTrigger }) => {
+const addItemDriverFactory = ({ element, eventTrigger, muteWarning }) => {
+  if (console && !muteWarning) {
+    // eslint-disable-next-line no-console
+    console.warn(`
+    Warning! addItemDriverFactory is deprecated and will be removed, please use AddItemDriver instead.
+    Note that AddItemDriver is async!
+    (Hide this warning by setting muteWarning: true)
+    `);
+  }
+
   const byHook = hook => element.querySelector(`[data-hook*="${hook}"]`);
   const tooltipTestkit = tooltipTestkitFactory({
     wrapper: element,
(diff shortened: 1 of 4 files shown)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6366

src/AddItem/AddItem.driver.js
```diff
@@ -2,7 +2,16 @@ import textDriverFactory from '../Text/Text.driver';
 import { tooltipTestkitFactory } from 'wix-ui-core/dist/src/testkit';
 import { dataHooks } from './constants';
 
-const addItemDriverFactory = ({ element, eventTrigger }) => {
+const addItemDriverFactory = ({ element, eventTrigger, muteWarning }) => {
```
shlomitc, line 5: Consumers will not be able to mute it the change you did in `test/utills/react` effects only our library, not in the exported testkits.

I can suggest adding an environmental variable to make it more configurable. e.g.: `test: MUTE_WSR_LEGACY_DRIVERS_WARNING yoshi test`.

src/AddItem/AddItem.driver.js
```diff
@@ -2,7 +2,16 @@ import textDriverFactory from '../Text/Text.driver';
...
+  if (console && !muteWarning) {
```
shlomitc, line 6: This should probably be much more generic, even point to some guide and not per component

EgoziE:
@shlomitc Changed according to your comments.

Now when you run:
`jest --watch AddItem`
You'll get:
image

In order to mute warnings, run:
`MUTE_WSR_LEGACY_DRIVERS_WARNING=true jest --watch AddItem`

src/utils/WithDeprecationWarning.js
```diff
@@ -0,0 +1,11 @@
+export const WithDeprecationWarning = driver => {
+  if (console && !process.env.MUTE_WSR_LEGACY_DRIVERS_WARNING) {
+    // eslint-disable-next-line no-console
+    console.warn(`
+    Warning! ${driver.name} is deprecated and will be removed.
+    For more information please refer to https://github.com/wix/wix-style-react/blob/master/docs/usage/testing.md
```
shlomitc, line 6: I think this link is too generic. We need to have a doc explaining exactly how to migrate from the legacy drivers to the new ones, which also explains how to mute the warnings.

Also, a bullet to think of - once people will mute it, they will not see it, so it's all-or-nothing.
Maybe there's a way to do the migration gradually and remove the warnings one-by-one.

shlomitc (changes requested):

EgoziE:
I added an optional `wsr.config.js` file that the consuming project can use to mute warnings.
image

PR #6373, opened 2020-12-10
Typography - adjust font weight for a specific family

EgoziE:
This is a rough POC to set different weights for specific font families.

This is the result for weight 500 vs 800 
image


src/Foundation/stylable/typography.st.css
```diff
@@ -1,7 +1,9 @@
+@import url('https://fonts.googleapis.com/css2?family=Noto+Sans+JP:wght@100&display=swap');
```
shlomitc, line 1: this will cause the font to be added to all stylesheets.
This of a way that we can opt-int specific locales (hint, theming)

shlomitc (changes requested):

src/Foundation/stylable/typography.st.css
EgoziE: It was just a quick POC to see this is technically possible.
Now we can discuss the logistics :)

EgoziE:
Closing for inactivity, when time comes we'll create a new PR.

PR #6380, opened 2020-12-13
<CustomModal/> - Expose title typography

EgoziE:
CustomModal has a very specific typography that we don't want to expose as is. Therefore I exposed it specifically as CustomModal.Title


src/BaseModalLayout/LayoutBlocks/Header/Header.js
```diff
@@ -14,6 +14,8 @@ export const Header = ({
   showHeaderDivider,
 }) => {
   const { headerClassName, title, subtitle } = useBaseModalLayoutContext();
+  const isModalHeading =
+    typeof !!title && (title === 'string' || title.type === Header.Title);
```
shlomitc, line 18: is the `typeof` by mistake? it looks weird for not comparing it with anything

src/BaseModalLayout/LayoutBlocks/Header/Header.js
```diff
@@ -50,6 +52,8 @@ export const Header = ({
   );
 };
 
+Header.Title = ({ children }) => children;
```
shlomitc, line 55:
```jsx
Header.Title = props => <ModalHeading {...props}/>
```

I think it will be more explicit to expose exactly the `ModalHeading`.
And since this is always a non-textual value, you can remove the `title.type === Header.Title` condition from the code above

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6380

shlomitc (changes requested):

src/BaseModalLayout/LayoutBlocks/Header/Header.js
EgoziE: mistake! thanks

src/BaseModalLayout/LayoutBlocks/Header/Header.js
EgoziE: I thought about that initially but then I figured that in the process we might lose current props

EgoziE: Ok, I think I got it, changed.

lironch (approved):

PR #6383, opened 2020-12-14
utils - update util for story of props permutations

daugerdas:
### 🔦 Summary

Update visual test util (props permutation story) to include more stories in one line


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6383

EgoziE:
I don't think this is a good idea, when each test case is in a different line, only this specific line will break.
If the tests are wrapped, then when one breaks it moves all the others and there will be diffs all over the place.

If you find that you got too many permutations, just split into two tests

mykas:
> I don't think this is a good idea, when each test case is in a different line, only this specific line will break.
> 
> If the tests are wrapped, then when one breaks it moves all the others and there will be diffs all over the place.
> 
> 
> 
> If you find that you got too many permutations, just split into two tests

What do you mean that it will break others if one line breaks? The diff will be shown only on the one that broke.

mykas:
There is a posibility if component structure suddenly changes like for example width or some margin - then it might move other components to the next line, but this is very rare case. We usually test this with components that are in fixed width.

EgoziE:
This is exactly what I'm talking about:
image

When width change for some reason it will break all the other "good" visuals in the same line, not to mention a possibility it will push down an element and then break all other lines below.

mykas:
Ok, if you think this is popular case then we can go with the splitting tests thing.

EgoziE:
Take a look at Tag.visual.js
It have 2 different sets of permutations
One for themes and one for all other props

PR #6393, opened 2020-12-15
infra - remove scss tokens from unneeded file and update perfer

shlomitc:
1. Remove some scss tokens imports which causes prefer bundles to be created with 0kb.
2. Remove code generation from adding new entries for css bundles to new components
3. cleanup the perfer.config.js file.

Please note that some components are in the list even though they are not explicitly importing scss files, this is because they have some dependency that does, one of the more prominent is `Tag`.

To test it:
```bash
=> npm run build
=> npm run bundles:generate
```
you will see the output list and can compare it with `perfer.config.js` file entries


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6393

EgoziE (approved):

PR #6407, opened 2020-12-16
<Collapse/> - Implement with @kunukn/react-collapse

gendeld:
### 🔦 Summary

...

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test

### DIFF
--- a/package.json
+++ b/package.json
@@ -198,6 +198,7 @@
   "dependencies": {
     "@babel/runtime": "^7.12.5",
     "@date-fns/upgrade": "^1.0.3",
+    "@kunukn/react-collapse": "^2.2.9",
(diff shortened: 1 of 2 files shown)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6407

EgoziE:
I prefer not to use an external library if we can avoid it.
External libraries means we need to rely on them in case of bugs etc.

Let's think of a solution in wix-animations instead. Maybe "disable" animation prop or "disableRunOnMount"

PR #6409, opened 2020-12-17
<Input/> - refactor component

EgoziE:
A complete re-arrange for Input component.
I also removed irrelevant code selecting inner input text that was made in 2017 due to an old bug in Opera browser that was already fixed.


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6409

src/Input/Input.js
```diff
@@ -67,21 +67,192 @@ class Input extends Component {
...
+  _onChange = (event) => {
+    const { onChange } = this.props;
+
+    if (this._isValidInput(event.target.value)) {
+      onChange && onChange(event);
+    }
+  };
+
+  _onKeyPress = (e) => {
+    if (!this._isValidInput(e.key)) {
```
gendeld, line 172: For number this is ok, but maybe if we introduce more complex types in the future checking only last char isn't enough.
Maybe check
```
`${value}${e.key}`
```

src/Input/Input.js
```diff
@@ -67,21 +67,192 @@ class Input extends Component {
...
+      e.preventDefault();
+    }
+  };
+
+  _triggerOnChangeHandlerOnClearEvent = (event) => {
```
gendeld, line 177: This isn't being used anywhere. Can we remove it?

src/Input/Input.js
EgoziE: You're right!
I was trying to create a more "generic" validation function in order to enable future need (for example a credit card validation)

src/Input/Input.js
```diff
@@ -67,21 +67,192 @@ class Input extends Component {
...
+      inputProps,
+    );
+  };
+
+  /**
+   * Sets focus on the input element
+   * @param {FocusOptions} options
+   */
+  focus = (options = {}) => {
+    this._onFocus();
```
gendeld, line 222: Aren't we already passing _onFocus as `onFocus` prop for input?
Won't this cause _onFocus to be triggered twice?

src/Input/Input.js
```diff
@@ -67,21 +67,192 @@ class Input extends Component {
...
+
+    if (!this._isControlled) {
+      this.input.value = '';
+    }
+
+    onClear && onClear(event);
+  };
+
   render(props = {}) {
     const {
```
gendeld, line 257: `clearButton` and `onClear` are unused

src/Input/index.d.ts
```diff
@@ -68,10 +68,10 @@ export default class Input extends React.Component<InputProps> {
   static StatusLoading: InputStatusLoading;
   static StatusWarning: InputStatusWarning;
 
-  clear: (event: any) => void;
+  clear: (event?: React.ChangeEvent<HTMLInputElement>) => void;
```
gendeld, line 71: This is triggered by a click on the clear button in `InputSuffix` so the event will be a  `MouseEvent`, I think it's not optional either

src/Input/index.d.ts
```diff
@@ -68,10 +68,10 @@ export default class Input extends React.Component<InputProps> {
...
   blur: () => void;
-  focus: (options: any) => void;
-  handleSuffixOnClear: (event: any) => void;
+  focus: HTMLElement['focus'];
+  handleSuffixOnClear: (event?: React.ChangeEvent<HTMLInputElement>) => void;
```
gendeld, line 74: `MouseEvent` as well

gendeld: You can delete this

src/Input/test/Input.tsx
```diff
@@ -87,10 +87,10 @@ function InputWithAllProps() {
 function testInstanceMethods() {
   const instance = new Input({});
   instance.blur();
-  instance.clear(null);
-  instance.focus(null);
+  instance.clear();
+  instance.focus({ preventScroll: false });
   instance.select();
-  instance.handleSuffixOnClear(null);
+  instance.handleSuffixOnClear();
```
gendeld, line 93: This is an internal method

src/Input/index.d.ts
EgoziE: It's tricky!
All these functions `clear`, `blur`, `focus`, `select` are public and can be triggered manually, therefore you might not get any event.

EgoziE: Also, I deleted handleSuffixOnClear from here, just noticed it's here by mistake

gendeld (approved):

PR #6420, opened 2020-12-22
Radio add unidriver

Zivilekin:
1. Added unidriver (reused from wix-ui-core)
2. Added id as a prop
3. Fixed style for 'alignItems: top'


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6420

src/Radio/Radio.js
```diff
@@ -58,6 +60,9 @@ Radio.propTypes = {
   /** Unique name of radio */
   name: PropTypes.node,
 
+  /** Unique id of radio */
+  id: PropTypes.string,
```
EgoziE, line 64: Add "id" to type and type test files.
Also in type file, pay attention that a string is a subtype of React.ReactNode, so in this case:
```
label?: React.ReactNode | string;
```

this is enough:
```
label?: React.ReactNode;
```

EgoziE (changes requested):

EgoziE:
By the way, I see that also in the types file there's an error:
```
import {
  RadioButtonChangeEvent,
  RadioButtonClickEvent,
} from 'wix-ui-core/radio-button';
```
'wix-ui-core/radio-button' import is restricted from being used by a pattern.

Please change is to:
```
import {
  RadioButtonChangeEvent,
  RadioButtonClickEvent,
} from 'wix-ui-core/dist/src/components/radio-button';
```

EgoziE (approved):

PR #6001, opened 2020-09-18
remove `yoshi lint`, use `eslint` directly

argshook:
This PR removes `yoshi lint` and uses `eslint` directly.

It is done by directly depending on the last version of `eslint-config-yoshi` which is open source.

Temporarily this solution is good enough. Meanwhile we're in talks with
yoshi team to possibly open source `eslint-config-yoshi`.

PR #6421, opened 2020-12-22
Move `eslint-config-yoshi` from dependencies to dev dependencies

ronami:
### Summary

Since https://github.com/wix/wix-style-react/pull/6001, `wix-style-react` depends on `eslint-config-yoshi` directly.

This PR moves it away from `dependencies` and into `devDependencies`. This currently causes some projects to get an older version of `eslint-config-yoshi`, which then overrides the version Yoshi brings. This is a problem because the version currently at the root is the one that applies its rules on the project.

cc @ranyitz @amitdahan


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6421

shlomitc:
thanks @ronami

PR #6432, opened 2020-12-23
Tag - migrate to stylable

Zivilekin:


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6432

src/Tag/Tag.st.css
```diff
@@ -0,0 +1,338 @@
...
+  background-color: value(B50);
+  border-color: value(B30);
+}
+
+.root:disabled {
+  background-color: value(D10-05);
+  color: value(D10-30);
+}
+
+:global(.rtl) .root:size(tiny),
```
EgoziE, line 208: I see all our `:global(.rtl)` styles here are related to margin and padding.
Let's remove **all** of them in favor of:
- margin-block-start
- margin-block-end
- padding-block-start
- padding-block-end

Note, don't use `margin-block` or `padding-block` without the starts/end suffix

EgoziE (changes requested):

src/Tag/Tag.js
```diff
@@ -68,39 +72,45 @@ class Tag extends React.PureComponent {
...
-        [styles.clickable]: onClick !== noop,
-        [styles.hoverable]: !disabled && onClick !== noop,
-      },
-    );
-  }
-
-  render() {
-    const { id, onClick, maxWidth, dataHook } = this.props;
-
+    const clickable = onClick !== noop;
```
mykas, line 88: Lets remove this noop  usage and instead do: `!!onClick`

src/Tag/Tag.st.css
```diff
@@ -0,0 +1,205 @@
...
+.root {
+  -st-states: withRemoveButton, withThumb, disabled, clickable, hoverable,
+    size(enum(tiny, small, medium, large)),
+    theme(enum(standard, error, warning, dark, neutral, light, lightOutlined));
+  box-sizing: border-box;
+  position: relative;
+  display: inline-flex;
+  align-items: center;
+  transition: background-color ease 300ms;
+  max-width: -moz-available;
```
mykas, line 16: I think that browser based prefixes are handled by the webpack already. No need to add special handling

src/Tag/Tag.st.css
```diff
@@ -0,0 +1,205 @@
...
+  max-width: -webkit-fill-available;
+  max-width: stretch;
```
mykas, line 18: I think we have a problem here. `stretch` is not supported by most of the browsers like safari, firefox: https://developer.mozilla.org/en-US/docs/Web/CSS/max-width

Some kind of other solution must be found here

src/Tag/Tag.st.css
```diff
@@ -0,0 +1,205 @@
...
+}
+.root:size(large) .removeButton {
+  margin-inline-start: 12px;
+}
+.root:not(:disabled):clickable {
+  cursor: pointer;
+}
+
+/* Themes */
+.root:not(:disabled):theme(standard) {
```
mykas, line 151: I wonder if you remove this `:not(:disabled)`. Does it work?

src/Tag/Tag.st.css
mykas: to resolve the issue a separate task has been created.

mykas (approved):
Good job, Zivile! Really, I mean it!

PR #6475, opened 2020-12-31
<Accordion/> - Add accordionSectionItem

mjurfest:
### 🔦 Summary

- Change accordion section item style
- Change accordion item header padding to 30px

### ✅ Checklist

- [X] 👨‍🎨 UX change is approved by the Design System UX @ayeletr


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6475

src/Accordion/AccordionSectionItem/AccordionSectionItem.js
```diff
@@ -1,11 +1,13 @@
...
-import { classes } from './AccordionSectionItem.st.css';
+import { st, classes } from './AccordionSectionItem.st.css';
 import Text from '../../Text';
 
 function accordionSectionItem({ title, size }) {
   return (
-    <div className={classes.root}>
-      <Text>{title}</Text>
+    <div className={st(classes.root, { size })}>
+      <Text skin="standard" weight="bold" size="small">
```
EgoziE, line 8: Might need ellipsis by default

EgoziE (changes requested):

src/Foundation/stylable/colors.st.css
```diff
@@ -30,6 +30,7 @@
   B40: #DAEFFE; /* Tag(Default), Badge */
   B50: #EAF7FF; /* Floating Notification */
   B60: #F4FAFE;
+  B70: #F9FBFC; /* Accordion Section Item */
```
EgoziE, line 33: We decided not to add this color to the palette.

mjurfest: Do you mean to add the color directly to the `AccordionSectionItem.st.css` file or not to use it at all?
Kfir told me that the color was approved.

EgoziE: Not to use it at all.

EgoziE (approved):

milkyfruit (commented):
@ayeletr accordion title is the same size as the item:
image

Shouldn't it be the same size as card subheader?
image

Also, why do we need yet another component for a section title? We already have card.subheader.

ayeletr:
@EgoziE @mjurfest 
1. the height is incorrect. needs to be 48px.
2. there should be a line at the bottom like in the spec.
3. did you use text style definitions that are connected to the system? (It just look a little weird)
4. in the first example in SB - please show the header in a separate example with small size Accordion Item.

ayeletr:
@milkyfruit 
1. The size of the component is incorrect, but it's also not according to card section title.
it's what kfir defined in zeplin for his need. we can change it if we think it should be smaller.

2. according to dev we can't use here the card section title, we need a component for accordion. that's the answer I got.

mjurfest:
@ayeletr
3. Yes, I used `<Text skin="standard" weight="bold" size="small" ellipsis>`

PR #6482, opened 2021-01-03
<MultiSelectCheckbox/> - [testkit] fix getLabels function

EgoziE:
Internally we can see raw components props, but externally we don't expose them.
So I fixed this specifically in MultiSelectCheckbox.


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6482

src/MultiSelectCheckbox/MultiSelectCheckbox.driver.js
```diff
@@ -10,8 +10,8 @@ const multiSelectCheckboxDriverFactory = ({ element, component }) => {
     getNumOfLabels() {
       return this.getLabels().length;
     },
-    getLabels: () => {
-      return inputDriver.getValue().split(component.props.delimiter);
+    getLabels: delimiter => {
```
mykas, line 13:
```suggestion
    getLabels: (delimiter = `, `) => {
```

src/MultiSelectCheckbox/MultiSelectCheckbox.uni.driver.js
```diff
@@ -18,8 +13,8 @@ export const multiSelectCheckboxUniDriverFactory = (
       async getNumOfLabels() {
         return (await this.getLabels()).length;
       },
-      getLabels: async () =>
-        (await inputDriver.getValue()).split(component.props.delimiter),
+      getLabels: async delimiter =>
```
mykas, line 16:
```suggestion
      getLabels: async (delimiter = `, `) =>
```

src/MultiSelectCheckbox/MultiSelectCheckbox.driver.js
```diff
@@ -10,8 +10,8 @@ const multiSelectCheckboxDriverFactory = ({ element, component }) => {
...
+      return inputDriver.getValue().split(delimiter || ', ');
```
mykas, line 14: I believe this might be a breaking change 🤔  You see, what if user gave `delimiter` prop some custom delimiter and used this `getLabels` method to get it? So now it will not work, because if he will not provide same delimiter as in props - getLabels will not return options, no?

EgoziE: I know, I took that in mind...
But we're deprecating old drivers anyway and if someone breaks the fix is easy.

Do you have an idea how to reach the props without a mess of creating an attribute and target it?

EgoziE: Or, we can fix only the uni-driver, I don't think it's such a bad solution

mykas: For old driver, yes. I don't think its a problem, but we have the same situation for unidriver. In theory we could use data-attributes to get the given delimiter 🤷‍♂️  Do you think thats possible?

mykas: O you just said by not using attribute. Why do you think its messy?

mykas: I believe the method was quite cool that it could figure out the delimiter for itself

mykas: ohhh i see now...

mykas: I still believe its better to figure out a way to know from the structure. I see that it renders Input component. Maybe somehow passing the data-attribute there?

EgoziE: I call it messy because this is not a good practice to have a "test specific code" inside a component.
This code has no use in production of course but it still increases the bundle size and the DOM (input is used 9999 times)

mykas: Yeah but we do data attributes for testing purposes anyway :D

splintor:
Hi,
It seems this PR was merged too soon:
1. You didn't add an optional `delimiter` parameter to the `getLabelAt` testkit method
1. You didn't update https://github.com/wix/wix-style-react/blob/master/src/MultiSelectCheckbox/README.TESTKIT.md that `getLabels` (and `getLabelAt`?) now accepts an optional `delimiter` parameter

mykas:
Well, this is embarrassing. Thank you for your insights! I will fix those immediately in next PR.

PR #6486, opened 2021-01-04
<Checkbox /> - fix checkbox's disabled overrides tooltipProps

alexkubica:
### 🔦 Summary

There is a bug that when `Checkbox` is disabled it also disables the tooltip even if passed:
```
tooltipProps: {
	disabled: false
}
```
This unfinished PR is my attempt to fix it. I had 2 issues:
1. Changing `Checkbox`'s disabled CSS rule from `pointer-events: none` to `pointer-events: all` fixes the issue where the tooltip's `onMouseEnter` event isn't triggered but makes `Checkbox` focusable.
2. I couldn't implement the `isTooltipEnabled` method on the driver properly. This method is used on a new test I wrote to see it's failing before fixing the issue.

### ✅ Checklist

- [X] 👨‍💻 API change is approved by the UI Infra Developers 
- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [X] Other (explain) - added documentation for `isTooltipEnabled` on the testkit.
  - 
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6486

src/Checkbox/test/Checkbox.spec.js
```diff
@@ -136,6 +136,21 @@ describe('Checkbox', () => {
...
+        disabled: false,
+      };
+      const { driver } = render(
+        <Checkbox
+          tooltipContent={tooltipContent}
+          tooltipProps={tooltipProps}
+          disabled
+        />,
+      );
+      expect(await driver.isTooltipEnabled()).toBe(true);
```
alexkubica: the implementation for `isTooltipEnabled` (both sync and async) doesn't work correctly, if you pass `disabled={true}` [in the Tooltip inside Checkbox](https://github.com/wix/wix-style-react/blob/master/src/Checkbox/Checkbox.js#L109) the test still passes when it should fail
I tried using the approach like in [Tooltip's spec](https://github.com/wix/wix-style-react/blob/master/src/Tooltip/test/Tooltip.spec.js#L59) but it doesn't work either (maybe the test there isn't good too?)

alexkubica (commented):
the example in the playground works like magic :)

src/Checkbox/test/Checkbox.spec.js
EgoziE: But I fixed it, take a look at the spec file.
```
      it('should not show error message when tooltipProps disabled is true', async () => {
        const tooltipContent = 'tooltip content';
        const tooltipProps = {
          disabled: true,
        };
        const { driver } = render(
          <Checkbox
            tooltipContent={tooltipContent}
            tooltipProps={tooltipProps}
          />,
        );

        expect(await driver.isTooltipEnabled()).toBe(false);
      });

      it('should show error message when disabled but tooltipProps disabled is false', async () => {
        const tooltipContent = 'tooltip content';
        const tooltipProps = {
          disabled: false,
        };
        const { driver } = render(
          <Checkbox
            tooltipContent={tooltipContent}
            tooltipProps={tooltipProps}
            disabled
          />,
        );

        expect(await driver.isTooltipEnabled()).toBe(true);
      });
```

EgoziE: Am I missing something?

alexkubica: i checked again and i was wrong, the tests are ok

EgoziE (approved):

PR #6495, opened 2021-01-05
<Calendar /> -  Add day indication slot

sivanharel:
### 🔦 Summary

The Calendar component is used also in the wix-ui-tpa library. 
One of the vertical (the Bookings team) needs the ability to add indications for specific dates.
Since these indications are specific for the Bookings product we decided to add a prop 'renderDateIndication' which will allow them to render indication for each date.
This prop will probably be used by the Events team in the future.

**In this task, I added also some improvements:**
1. Remove the duplication from the renderDay method.
2. Remove the for loop with relevantModifier.
The dayCircle class was added for each day that has one of the following modifiers 'start' / 'end' / 'selected'.
And there is no need to check all of the modifiers since every time that 'start' and 'end' exist the 'selected' is added as well so we can just check if the selected modifier is exist.
In addition, we don't need to add a specific state or class we could just use the selector `.day:selected .dayWrapper`
3. Add the prop renderDateIndication and addapt the styling accordigly.

### ✅ Checklist

- 📚 Change is documented
  - [x] Story
  - [x] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [x] Visual test


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6495

EgoziE (approved):

PR #6528, opened 2021-01-10
<SelectorList/> -  add visual tests

bamai:


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6528

src/SelectorList/test/SelectorList.visual.js
```diff
@@ -3,6 +3,7 @@ import { storiesOf } from '@storybook/react';
 import { uniTestkitFactoryCreator } from 'wix-ui-test-utils/vanilla';
 import { selectorListUniDriverFactory } from '../SelectorList.uni.driver';
 import SelectorList from '../index';
+import { Text } from '../../Text';
```
TalZa1810, line 6: redundant import

src/SelectorList/test/SelectorList.visual.js
```diff
@@ -45,6 +46,22 @@ const Example = props => (
   />
 );
 
+const EmptyExample = props => (
```
TalZa1810, line 49: It's similar to the regular example. The only difference is the `dataSource` and `emptyState` props. so this can be added as props instead of creating a new example.

src/SelectorList/test/SelectorList.visual.js
```diff
@@ -72,22 +89,121 @@ const interactiveTests = [
...
+      {
+        it: 'should display subtitle',
+        props: { subtitle: 'selector list subtitle' },
+        componentDidMount: () => {},
       },
     ],
   },
 ];
 
+storiesOf('SelectorList', module)
```
TalZa1810, line 202: you can create another variable of `tests` instead of `interactiveTests` and run the tests in a similar why of the interactive ones, similar to what we have [here](https://github.com/wix/wix-style-react/blob/master/src/FeatureList/test/FeatureList.visual.js#L112). Also, in this case, you don't need to send `componentDidMount` but simply using the `prop`.

TalZa1810 (changes requested):
few tiny comments

src/SelectorList/test/SelectorList.visual.js
```diff
@@ -72,22 +72,126 @@ const interactiveTests = [
...
+    describe: 'search bar',
     its: [
       {
         it: 'should render selector list without search bar',
-        props: { multiple: true, withSearch: false },
-        componentDidMount: async () => {
-          await createDriver().toggleSelectorAt(1);
-        },
+        props: { withSearch: false },
+        componentDidMount: () => {},
```
TalZa1810, line 94: no need for `componentDidMount` here

src/SelectorList/test/SelectorList.visual.js
```diff
@@ -72,22 +72,141 @@ const interactiveTests = [
...
+        props: { withSearch: false },
+      },
+      {
+        it: 'should render selector list with search bar',
+        props: {},
+      },
+    ],
+  },
+  {
+    describe: 'image shape',
```
TalZa1810, line 136: let's create a constants file which has the following:
```
export const IMAGE_SHAPE = {
  rectangle: 'rectangle',
  circle: 'circle',
};
```

Afterwards, we can map them over. [Here's](https://github.com/wix/wix-style-react/blob/master/src/Notification/test/Notification.visual.js#L17) an example as a reference.

src/SelectorList/test/SelectorList.visual.js
```diff
@@ -72,22 +72,141 @@ const interactiveTests = [
...
+        props: { imageSize: 'portrait' },
+      },
+      {
+        it: 'should render cinema images',
+        props: { imageSize: 'cinema' },
+      },
+    ],
+  },
+  {
+    describe: 'circle image size',
```
TalZa1810, line 174: There's no need to check the combination of the props.

src/SelectorList/test/SelectorList.visual.js
```diff
@@ -72,22 +72,141 @@ const interactiveTests = [
...
+        props: { imageShape: 'rectangle' },
+      },
+      {
+        it: 'should render circle images',
+        props: { imageShape: 'circle' },
+      },
+    ],
+  },
+  {
+    describe: 'image size',
```
TalZa1810, line 149: sizes can be exported to a constant as well

TalZa1810 (changes requested):
Few more comments

TalZa1810 (approved):

PR #6556, opened 2021-01-14
[wix-style-react] - <Modal /> introduced screen prop for spacing between viewport and modal content

egidijusja:
### 🔦 Summary

...

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


EgoziE:
Modal is just a mechanism component, no styles should be applied to it.
By adding this padding prop (specially with that 48px default) you're going to break every modal our consumers have.

Are you sure this is correct place for this fix?

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6556

egidijusja:
> Modal is just a mechanism component, no styles should be applied to it.
> By adding this padding prop (specially with that 48px default) you're going to break every modal our consumers have.
> 
> Are you sure this is correct place for this fix?

Actually this should be draft PR, sorry for that. Actually we had a meeting with Domas and @mykas regarding this one. Modal design guidelines tell that modal should have margins of 48px. This behaviour right now is implemented in some modals (children) already:
https://github.com/wix/wix-style-react/blob/master/src/BaseModalLayout/BaseModalLayout.st.css#L60
https://github.com/wix/wix-style-react/blob/master/src/CustomModalLayout/CustomModalLayout.st.css#L24

Although, there is a problem with this approach, e.g. CustomModalLayout accepts props like width, height, maxHeight and if some of these props are provided margins calculation logic is overridden and modal doesn't follow design guidelines anymore. So the idea was to introduce padding prop for Modal component with default value `0px` and once we move library to new major version change default value to `48px` and fix other modal components styles which try to calculate margins themselves.

mykas (approved):
Looks very clean and good!

PR #6572, opened 2021-01-17
[wix-style-react] - <Calendar/> - add a11y attributes to month and year dropdown

bamai:
https://jira.wixpress.com/browse/DSM-1436


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6572

sivanharel (commented):
Awesome:)
I added some small comments.
In addition, I saw that in case that "linkTo" is passed the role="option" and aria-selected is added twice. Let's check it:)

packages/wix-style-react/src/DropdownLayout/DropdownLayout.js
```diff
@@ -500,7 +502,7 @@ class DropdownLayout extends React.PureComponent {
   };
 
   _renderOptionContent({ option, idx }) {
-    const { itemHeight, selectedHighlight } = this.props;
+    const { itemHeight, selectedHighlight, linkTo } = this.props;
```
sivanharel, line 505: `linkTo`  is always undefined

packages/wix-style-react/src/Calendar/DatePickerDropdown/DatePickerDropdown.js
```diff
@@ -68,11 +69,26 @@ export default class DropdownPicker extends React.Component {
...
+      options,
+      dataHook,
+      selectedId,
+      ariaLabel,
+      ariaLabelledBy,
+    } = this.props;
+
     const { open } = this.state;
 
+    const finalAriaLabel = ariaLabel ? `${ariaLabel} ${caption}` : caption;
```
sivanharel, line 84: We shouldn't add any aria-label in case that `ariaLabel` is not passed.
So the else should be ''

sivanharel (approved):
Great:)

PR #6624, opened 2021-01-21
[wix-style-react] extract `test-runtime` into dedicated package

argshook:
this PR moves `test-runtime` folder into a dedicated package:

```diff
-wix-style-react/packages/wix-style-react/test/e2e-runtime
+wix-style-react/packages/wix-style-react-test-runtime
```


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6624

shlomitc:
all the files in `.storybook`  folder should be handled with a non-relative path
like this one: https://github.com/wix/wix-style-react/blob/extract-test-runtime/packages/wix-style-react-test-runtime/.storybook/component-definitions.js

packages/wix-style-react-test-runtime/.storybook/config.js
```diff
@@ -1,4 +1,4 @@
-import { addParameters, configure } from '@storybook/react';
+import { addParameters, configure } from 'wix-style-react/node_modules/@storybook/react';
```
shlomitc, line 1: Looks like we need to install storybook here as well. Maybe at some point we can hoist the dependencies but I think it’s better to do it right

packages/wix-style-react-test-runtime/.storybook/component-definitions.js
```diff
@@ -1,16 +1,16 @@
...
-import MultiSelect from '../../../dist/es/src/MultiSelect';
-import AutoComplete from '../../../dist/es/src/AutoComplete';
-import Input from '../../../dist/es/src/Input';
-import Notification from '../../../dist/es/src/Notification';
-import FloatingHelper from '../../../dist/es/src/FloatingHelper';
-import GoogleAddressInput from '../../../dist/es/src/GoogleAddressInput';
-import VerticalTabs from '../../../dist/es/src/VerticalTabs';
-import SegmentedToggle from '../../../dist/es/src/SegmentedToggle';
+import { GmapsTestClient } from 'wix-style-react/dist/es/src/GoogleAddressInput/gmapsTestClient';
+import CropRotate from 'wix-style-react/node_modules/wix-ui-icons-common/CropRotate';
```
shlomitc, line 4: Install Wix-UI-icons-common or remove if not needed

packages/wix-style-react-test-runtime/package.json
```diff
@@ -0,0 +1,55 @@
...
+  "babel": {
+    "presets": [
+      "yoshi"
+    ]
+  },
+  "eslintConfig": {
+    "extends": "yoshi"
+  },
+  "dependencies": {
+    "react": "^16.13.1",
```
shlomitc, line 40: This can probably be a dev dependencies

packages/wix-style-react-test-runtime/.storybook/config.js
argshook: true, there are dependencies that could be hoisted.

however, would like to do that once wix-ui-tpa gets into the codebase.
This way we avoid possible double work, because if we hoist deps now, we
likely need to update them once wix-ui-tpa gets into the monorepo.

anyway, it should be done as a separate PR imo

packages/wix-style-react-test-runtime/.storybook/component-definitions.js
argshook: for now i'd like to reuse dependencies, because installing them again adds up to the bootstrap time.

but in this case, i'll just remove this icon, it's not mandatory

packages/wix-style-react-test-runtime/package.json
argshook: i'll move to devDeps, although this package is marked as private, so it doesn't have any impact

PR #6628, opened 2021-01-22
useCopyClipboard - introduce browser api polyfills to cover CopyClipboard in JSDOM

mykas:
Component CopyClipboard is rendered by our consumers in JSDOM environment. We should provide JSDOM based polyfills.


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6628

shlomitc:
Following @carmelc comments - even after adding the polyfills (which are good), he will still not be able to interact with the component in unit tests. So all of these are just not to fail the test, but need to keep that in mind

mykas:
Partially he will be able to do that, because CopyClipboard does include a div that can be run against with methods like: `exists`.

packages/wix-style-react/testkit/polyfills/selection-behavior.js
```diff
@@ -10,6 +10,7 @@ const install = () => {
     return {
       addRange: () => {},
       removeAllRanges: () => {},
+      selectNodeContents: () => {},
```
egidijusja, line 13: Do we really need this? I checked w3c specs and couldn't find that getSelection factory would have selectNodeContents function.

egidijusja (approved):
Looks good to me!

PR #6638, opened 2021-01-25
[wix-style-react] - Add swimm verify GitHub Action

OmerRosenbaum:
### 🔦 Summary

Adding `swimm verify` to the CI as a GitHub Action.
It fetches the latest version of `swimm verify` each time, and runs it.
In case of a failure, the log will provide a link to Swimm's `Status Page`, allowing for a quick fix.

### DIFF
--- a/.github/workflows/swimm_verify.yml
+++ b/.github/workflows/swimm_verify.yml
@@ -1,0 +1,29 @@
+name: Swimm
+
+on: pull_request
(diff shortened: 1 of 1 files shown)


shlomitc:
Thanks for the effort!
We’re not working with github actions so I prefer to not introduce additional CI flows to our current one.

PR #6645, opened 2021-01-26
[wix-style-react] - <Radio/> - add style/value props and legacy driver

Zivilekin:


packages/wix-style-react/src/Radio/Radio.js
```diff
@@ -27,14 +29,16 @@ const Radio = ({
 
   return (
     <RadioButton
+      style={style}
       className={st(classes.root, { alignItems }, className)}
       data-hook={dataHook}
       checked={checked}
-      label={renderLabel}
+      label={label ? renderLabel : null}
```
mykas, line 36: Lets move this `label` checking inside the `renderLabel` function. Also you forgot to call the renderLabel function

packages/wix-style-react/src/Radio/Radio.st.css
```diff
@@ -79,12 +84,7 @@
   margin: 2px;
 }
 
-.root::label {
+.root::label:not(:empty) {
```
mykas, line 87: Why no :empty?

Zivilekin: SelectableAccordion generates Radio without using label and in this case label's padding is present creating uneccessary space

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6645

packages/wix-style-react/src/Radio/Radio.st.css
mykas: But in case label is not present you are not rendering the label div anyway?

Zivilekin: yeah, but the span that label is in - is still generated and creates space as showed in the picture. I'm not rendering label so that span would remain empty and then with this css rule I can check that
image

mykas: IMO thats a bug in RadioButton of wix-ui-core. There shouldn't be any div related to label if label is not rendered. Lets fix this in wix-ui-core.

Zivilekin: that makes sense

mykas (approved):

mykas:
Nicely done! lets try to push this to RadioGroup now

PR #6656, opened 2021-01-27
<PulseAnimation/> - Add new component

mjurfest:
### 🔦 Summary

- New component: PulseAnimation

### ✅ Checklist

- [X] 👨‍💻 API change is approved by the UI Infra Developers @TalZa1810 
- [X] 👨‍🎨 UX change is approved by the Design System UX @ZilvinasMaciulis 
- 📚 Change is documented
  - [x] Story
  - [x] API description


TalZa1810:
Hi @ZilvinasMaciulis ,
Can you please review the examples? which examples should we add to this story page?

TalZa1810:
Hi @mjurfest ,
You should add the `children` prop to the component definition, like we have in `<BounceAnimation/>`:
https://github.com/wix/wix-style-react/blob/master/packages/wix-style-react/testkit/component-definitions.js#L39

Otherwise, unit tests are failing and it doesn't publish this to storybook.

Thanks :)

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6656

packages/wix-style-react/src/PulseAnimation/PulseAnimation.st.css
```diff
@@ -0,0 +1,177 @@
...
+    animation-timing-function: value(ease-3);
+  }
+  100% {
+    box-shadow: 0 0 0 value(mediumShadowSpread) rgba(0, 0, 0, 0);
+    animation-timing-function: value(ease-3);
+  }
+}
+
+.root {
+  -st-states: active, loop, size(enum(small, medium)), color(enum(B10, R10, P10, G10));
```
TalZa1810, line 118: Adding `width: fit-content` to the root class so that the animation is wrapping the children component. We need to make sure that there's also a border radius for the box shadow

packages/wix-style-react/stories/index.js
```diff
@@ -230,3 +230,6 @@ require('../src/Themes/editorX/docs');
 
 // BusinessDashboard
 require('../src/Themes/businessDashboard/docs');
+
+// TODO: move to correct position
+require('../src/PulseAnimation/docs/index.story');
```
TalZa1810, line 235: Please move this to the correct position (alphabetically)

packages/wix-style-react/src/PulseAnimation/docs/index.story.js
```diff
@@ -0,0 +1,105 @@
...
+              '`<PulseAnimation/>` component is a wrapper that contains predefined pulse animation. It is used to emphasize UI elements with pulse motion.',
+          }),
+
+          importExample("import { PulseAnimation } from 'wix-style-react';"),
+
+          divider(),
+
+          title('Examples'),
+
+          example({
```
TalZa1810, line 69: Hi @ZilvinasMaciulis 
Do we need all examples? Which examples should we add to the story page? Of which components?

Thanks

TalZa1810 (changes requested):

ZilvinasMaciulis:
The pulse effect (border-radius) should follow the visual shape of the component. Now the check toggle is round while the effect is square. See the image below. 

The effect should be applied to components without any external text. It should be used on buttons, badges but not on the inputs, dropdowns. Unless the effect is applied to the input itself.  See the example. For the "end" example please use the icon button with a tooltip

According to the inspect panel, animation takes around 700ms. It should be 1050ms. Please check if the animation setup meets specs. 

image

TalZa1810:
@ZilvinasMaciulis 
We're aware that the animation is ready yet. We just wanted a list of examples we would like to present in the storybook page and their descriptions.
Do we want to show this with the CheckToggle? Input? Which components should we use?

Thanks.

ZilvinasMaciulis:
@TalZa1810 

As discussed before, this component is meant to draw users' attention with a continuous loop. Loop lasts until the user provides the required action. In some rare cases, the loop might last a certain amount of time.  Also, it has a color. That being said I think there is a need to demonstrate that:

its default state (loop is a default prop)
It has a color
It can have a delay

The first application for this wrapper will be on the button. Therefore I think the button is the best example to use it on.

Use plain HTML button for play/stop animation.

Other options like play once, or show something on animation end should not be included in this example. 

image

TalZa1810:
> The pulse effect (border-radius) should follow the visual shape of the component. Now the check toggle is round while the effect is square. See the image below. 
> 
> The effect should be applied to components without any external text. It should be used on buttons, badges but not on the inputs, dropdowns. Unless the effect is applied to the input itself.  See the example. For the "end" example please use the icon button with a tooltip
> 
> According to the inspect panel, animation takes around 700ms. It should be 1050ms. Please check if the animation setup meets specs. 
> 
> image

Hey @mjurfest ,
Let's expose a `borderRadius` prop for the `<PulseAnimation/>` that can be customized by the user. 
Also, please make sure that the animation values are the same as mentioned in the specs. If you're not sure, please contact Zilvinas regarding the desired values. 

Thanks :)

packages/wix-style-react/src/Animate/Animate.js
```diff
@@ -46,6 +49,9 @@ Animate.propTypes = {
 
   /** set a delay before the animation execution. When provided a number- sets this as `ms`.*/
   delay: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+
+  /** Sets the border-radius css property of the animation */
+  borderRadius: PropTypes.string,
```
TalZa1810, line 54: This prop is relevant just for the `<PulseAnimation/>`. We're not sure if it's relevant for other types on animations as well.

Let's rename this to `animateInlineStyle` so that next time we will be able to add more inline styles when necessary.

packages/wix-style-react/src/Animate/index.d.ts
```diff
@@ -7,6 +7,7 @@ export interface AnimateProps {
   onStart?(): void;
   onEnd?(): void;
   delay?: string | number;
+  borderRadius?: React.CSSProperties['borderRadius'];
```
TalZa1810, line 10: same comment

packages/wix-style-react/src/PulseAnimation/PulseAnimation.js
```diff
@@ -0,0 +1,105 @@
...
+          active,
+          loop,
+          size: animationSize,
+          color,
+          borderRadius,
+        })}
+        onEnd={onEnd}
+        onStart={onStart}
+        ref={this.rootRef}
+        borderRadius={borderRadius}
```
TalZa1810, line 62: Let's rename this prop to `animateInlineStyle`. Moreover, I think that using css variable for the inline styling is a better solution. 

```suggestion
        animateInlineStyle={[ [vars.borderRadius]: borderRadius ] }
```


There's an example with usage of [css variable](https://github.com/wix/wix-style-react/blob/master/packages/wix-style-react/src/CardFolderTabs/CardFolderTabs.js#L23)

packages/wix-style-react/src/PulseAnimation/constants.js
```diff
@@ -0,0 +1,11 @@
+export const dataHooks = {};
```
TalZa1810, line 1: not used. can be removed

packages/wix-style-react/src/PulseAnimation/test/PulseAnimation.tsx
```diff
@@ -0,0 +1,45 @@
...
+  return (
+    <PulseAnimation
+      dataHook="dataHook"
+      children={<div/>}
+      active
+      loop
+      onStart={()=>{}}
+      onEnd={()=>{}}
+      delay={200}
+      color = "B10"
```
TalZa1810, line 23: add the borderRadius prop here as well

TalZa1810 (changes requested):
Hey,
I've added some comments. Let me know if something is not clear :)

packages/wix-style-react/src/Animate/index.d.ts
```diff
@@ -7,6 +7,7 @@ export interface AnimateProps {
   onStart?(): void;
   onEnd?(): void;
   delay?: string | number;
+  animateInlineStyle?: React.CSSProperties['animateInlineStyle'];
```
TalZa1810, line 10:
```suggestion
  animateInlineStyle?: React.CSSProperties;
```

packages/wix-style-react/src/PulseAnimation/docs/index.story.js
```diff
@@ -0,0 +1,92 @@
...
+
+import PulseAnimation from '..';
+import Button from '../../Button';
+import * as examples from './examples';
+
+const example = config => baseExample({ components: allComponents, ...config });
+
+const exampleChildren = [
+  {
+    label: 'Default example',
```
TalZa1810, line 27:
```suggestion
    label: 'Button',
```

packages/wix-style-react/src/PulseAnimation/docs/index.story.js
```diff
@@ -0,0 +1,92 @@
...
+];
+
+export default {
+  category: storySettings.category,
+  storyName: storySettings.storyName,
+
+  component: PulseAnimation,
+  componentPath: '..',
+
+  componentProps: { children: exampleChildren[0].value },
```
TalZa1810, line 39:
```suggestion
  componentProps: { 
    children: exampleChildren[0].value, 
    color:"B10",
    borderRadius: "18px",
    active: true 
},
```

TalZa1810 (changes requested):
Great job :) 
Few more tiny comments and I think that we can merge this.

packages/wix-style-react/src/PulseAnimation/docs/examples.js
```diff
@@ -0,0 +1,56 @@
...
+  )
+}
+`;
+
+export const colorExample = `() => {
+  const [ active, setActive ] = React.useState(false);
+  return (
+    <Layout>
+      <Cell span="8">
+        <PulseAnimation onEnd={()=>setActive(false)} active={active} color="P10" borderRadius="18px">
```
TalZa1810, line 25: The `onEnd` is irrelevant if the `loop` prop is true by default. It is not executed because it is defined as `infinite` within the css. So, we should add `loop={false}` so that the `onEnd` callback would be executed.

mjurfest: I agree. I will remove the onEnd and keep the default loop behavior

TalZa1810 (approved):

ZilvinasMaciulis (approved):

PR #6706, opened 2021-02-04
Add id prop to generator

EgoziE:
Motivation
In order to render components in the Editor, they require to have an `id` attribute on their root element.
We decided it's a good idea to make it a standard for all components.


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6706

PR #6731, opened 2021-02-09
sync package.json versions

argshook:
This PR touches all packages `package.json` so that common dependencies are specified with the same version.

`package.json` of `wix-style-react` had `"lodash": "^4.17.19"`
while `package.json` of `wix-ui-tpa` had `"lodash": "^4.17.11"`

both libraries will get the same lodash version in the end (because of
`^`), however, during `npm install` there would be two different
requests to artifactory, for each version.

syncing versions reduces the number of artifactory requests, which
improves overall build time.

---

different versions were found with `syncpack` tool
https://github.com/JamieMason/syncpack

only versions up to minor version were changed, different major versions
will remain.


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6731

argshook:
funnily, this made the build longer ~30sec, so closing for now

PR #6736, opened 2021-02-10
[wix-ui-tpa]: publish storybook

argshook:
This PR adds `storybook:release` to run after `test9` is successful

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "test6:sanity": "lerna run --scope wix-style-react test6:sanity",
     "test7:performance": "lerna run --scope wix-style-react test7:performance",
     "test8:chromatic": "lerna run --scope wix-style-react-test-runtime test",
-    "test9": "lerna run build --scope wix-ui-tpa && lerna run test --scope wix-ui-tpa",
+    "test9": "lerna run build --scope wix-ui-tpa && lerna run test --scope wix-ui-tpa && lerna run storybook:release --scope wix-ui-tpa",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -14,7 +14,7 @@
     "test6:sanity": "lerna run --scope wix-style-react test6:sanity",
     "test7:performance": "lerna run --scope wix-style-react test7:performance",
     "test8:chromatic": "lerna run --scope wix-style-react-test-runtime test",
-    "test9": "lerna run build --scope wix-ui-tpa && lerna run test --scope wix-ui-tpa",
+    "test9": "lerna run build --scope wix-ui-tpa && lerna run test --scope wix-ui-tpa && lerna run storybook:release --scope wix-ui-tpa",
```
jonathanadler, line 17: I would run them in parallel or at least put the storybook:release first, so that it will run even when tests will fail

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6736

package.json
argshook, line 17:
```suggestion
    "test9": "lerna run build --scope wix-ui-tpa && lerna run storybook:release --scope wix-ui-tpa && lerna run test --scope wix-ui-tpa",
```

i'd think maybe if tests fail then storybook is falsy and shouldn't be published?
but if that's ok, then you can commit this suggestion and merge

jonathanadler: but that's what happens with wsr as well. Janet is parallel

PR #6737, opened 2021-02-10
move wut scripts to root package.json (well aliases actually)

jonathanadler:


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6737

package.json
```diff
@@ -21,7 +22,10 @@
     "test13": "#",
     "test14": "#",
     "test15": "#",
-    "test16": "#"
+    "test16": "#",
+    "start:wut": "lerna run --scope wix-ui-tpa start --stream",
+    "test:wut": "lerna run test --scope wix-ui-tpa --stream && lerna run --scope wix-ui-tpa test:eyes-storybook --stream",
+    "wut": "lerna run --scope wix-ui-tpa --stream"
```
argshook, line 28: what is `wut` script for, what command will it run?

jonathanadler: this will let us run every command from the root folder:
`npm run wut -- build:perfer-bundles` for instance.

I think in general it's best to work from the root folder in these types of monorepos

argshook: ah i see, used as a shortcut.
sure why not. i think eventually we'll want a single `npm run build` and `npm run test` that understands which builds to build and which tests to run by checking what changed. but that's future future

argshook (approved):

PR #6747, opened 2021-02-11
Adjusting wut scripts, so that prebuild is called

jonathanadler:


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_6747/

shlomitc:
I'm a bit confused by the structure and naming...
I think we should aim that each package will separate its own commands in a "human" language.
Then, the root level will just do the mapping of a parallel task to the semantically named task.

Lastly, if you want to allow developers to run some commands from the root, you can aggregate them.

**wut:**
instead of `monorepo:test9` - simply call it the command that you need `test:e2e`, `test:eyes-storybook`, ...
the generic `test` command can be moved as is to the root

**root:**
`test9` -> call `test:e2e`
`test10` -> call `test:eyes-storybook`
`wut:test` -> call `test9` and `test10`

jonathanadler:
> I'm a bit confused by the structure and naming...
> I think we should aim that each package will separate its own commands in a "human" language.
> Then, the root level will just do the mapping of a parallel task to the semantically named task.
> 
> Lastly, if you want to allow developers to run some commands from the root, you can aggregate them.
> 
> **wut:**
> instead of `monorepo:test9` - simply call it the command that you need `test:e2e`, `test:eyes-storybook`, ...
> the generic `test` command can be moved to the root
> 
> **root:**
> `test9` -> call `test:e2e`
> `test10` -> call `test:eyes-storybook`
> `wut:test` -> call `test9` and `test10`

So this PR's purpose was to make it more readable actually 😄 

Our readable scripts still exist, as they were before.
The problem is that with lerna, the scripts in the root package.json become very very long and unreadable (each command needs to have `--scope wix-ui-tpa --stream` etc.).
So I made the scripts in the root package short, and added some scripts in `wix-ui-tpa/package.json` that are long but more readable if they were in the root.

I'm not sure I explained this well enough 🤔

package.json
```diff
@@ -15,8 +15,8 @@
     "test6:sanity": "lerna run --scope wix-style-react test6:sanity",
     "test7:performance": "lerna run --scope wix-style-react test7:performance",
     "test8:chromatic": "lerna run --scope wix-style-react-test-runtime test",
-    "test9": "lerna exec --scope wix-ui-tpa --stream \"npm run build && npm run test:basic\"",
-    "test10": "lerna exec --scope wix-ui-tpa --stream \"npm run build && npm run test:visual\"",
+    "test9": "lerna exec --scope wix-ui-tpa --stream -- \"npm run build && npm run test:basic\"",
```
argshook, line 18: these dashes are necessary only for flags, but in this case there are none

jonathanadler: 🤔  I see
I'm not sure I understand what flags though 😄

argshook: for example this would work
`lerna exec --scope wix-ui-tpa yoshi test`

but this wouldn't
`lerna exec --scope wix-ui-tpa yoshi test --watch`

the `--watch` would be passed to `lerna` and not to `yoshi`. for such cases the `--` is used:

`lerna exec --scope wix-ui-tpa -- yoshi test --watch`

jonathanadler: Oh, ok, I see now, thanks!

argshook:
for transparency:

i pushed a PR for this PR which makes the root scripts a bit cleaner imo
https://github.com/wix/wix-style-react/pull/6754

please let me know if it can be merged into this PR

argshook (approved):
LGTM, except that i didn't check all the flows in wut package.json, but i trust @jonathanadler to know those better :)

run-package-command.sh
```diff
@@ -0,0 +1,17 @@
...
+# npm run wut build
+# would run `npm run build` under `packages/wix-ui-tpa`
+#
+# this is nice, because without this helper it would be:
+# lerna --exec --scope wix-ui-tpa -- npm run build
+
+scope=$1
+script=$2
+rest_arguments="${@:3}"
+lerna exec --scope $scope -- npm run $script `[[ -n "$rest_arguments" ]] && echo "-- $rest_arguments"`
```
sivanharel, line 17: What does this file aim for? :)

sivanharel (approved):
Looks good!
But let's wait for approval from @argshook as well :)

argshook (approved):

nitayneeman (approved):

PR #6749, opened 2021-02-11
<InputWithOptions /> - add missing disabled prop to native select

iaroslavshvets:
### 🔦 Summary
Even though it looks disabled from UI point of view, select is pretty active and value can be changed.
passing disabled prop fixes this case.


EgoziE:
Thanks for your contribution!
Can you please add a unit test?

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_6749/

iaroslavshvets:
> Thanks for your contribution!
> Can you please add a unit test?

I've honestly tried, but did get nowhere. I've added a commit with non working test. I think the issue is the place I've marked https://github.com/wix/wix-style-react/blob/fdbbc80076687401587cdcc36d11b844d8fbd3c1/packages/wix-style-react/src/InputWithOptions/InputWithOptions.uni.driver.js#L36 . Despite passed `disabled` prop it still somehow manages to changes the select. Maybe ReactBase should be changed, like "if there is disabled - do nothing". Feels hacky. I don't know the proper fix. I've checked in real browser - it works as excepted - selection can't be changed. This issue is blocking several projects to make a move to WSR from deprecated custom ui-lib, so it would be nice if you could merge it with or without test. This bug is a total blocker for us :(

shlomitc:
@iaroslavshvets although looks hacky, I think it actually what needs to be done.
Let's take for example how the `jsdom` adapter of unidriver works for a click event:
https://github.com/wix-incubator/unidriver/blob/master/adapters/jsdom-react/src/index.ts#L176-L180
It simulates the focus event but checks first that the item is not disabled.
So jsdom kind of sucks in these cases...

It might something to consider adding to unidriver itself - @GabiGrin WDYT?

GabiGrin:
@shlomitc agree - UniDriver's place here is to abstract away these different quirks, like we do in your example.
also reminds me of https://github.com/wix-incubator/unidriver/issues/120

feel free to open an issue and I'll address it on the next blitz (meanwhile, PR-s welcome of course) :)

packages/wix-style-react/src/InputWithOptions/test/InputWithOptions.spec.js
```diff
@@ -963,6 +963,33 @@ describe('InputWithOptions', () => {
...
+        ];
+
+        const onSelect = jest.fn();
+
+        const { driver } = createDriver(
+          <InputWithOptions
+            options={sampleOptions}
+            onSelect={onSelect}
+            native={true}
+            disabled={true}
```
TalZa1810, line 985:
```suggestion
            native
            disabled
```

packages/wix-style-react/src/InputWithOptions/test/InputWithOptions.spec.js
```diff
@@ -954,7 +954,7 @@ describe('InputWithOptions', () => {
           <InputWithOptions
             options={sampleOptions}
             onSelect={onSelect}
-            native
+            native={true}
```
TalZa1810, line 957:
```suggestion
            native
```

TalZa1810 (approved):

TalZa1810:
fails on visuals which are not related to the change

PR #6754, opened 2021-02-15
Add package run command

argshook:
This PR is adjustment to https://github.com/wix/wix-style-react/pull/6747

i was typing comments until it became a PR :)

it would be better for `wix-ui-tpa` related scripts to be managed within
`packages/wix-ui-tpa/package.json` and not in the root `package.json`

IMO `npm run wut start` is easier to remember than `npm run start:wut`
and also scales nicer.

This way root `package.json` scripts are proxies and not a collection of custom scripts. Also, root `package.json` wouldn't need to be edited each time `wix-ui-tpa/package.json` scripts change.

the way to do it might seem with
```
"wut": "lerna exec --scope wix-ui-tpa npm run"
```

and then just, for example `npm run wut generate`

this works but passing `--flag`s say, `--component-name`, is awkward:

`npm run wut generate -- -- --component-name TheName`

need to use `--` twice, yuck!

to avoid that, a bash oneliner to help (inside `./run-package-command.sh`)

with it `npm run wut generate -- --compnent-name TheName` works as expected


package.json
```diff
@@ -23,10 +23,7 @@
     "test14": "#",
     "test15": "#",
     "test16": "#",
-    "wut": "lerna run --scope wix-ui-tpa --stream",
-    "start:wut": "npm run wut -- start",
-    "test:wut": "npm run wut -- test",
-    "generate:wut": "npm run wut -- generate"
+    "wix-ui-tpa": "./run-package-command.sh wix-ui-tpa"
```
argshook, line 26: this can be `wut` instead of `wix-ui-tpa` i just thought maybe using full name is not too much

packages/wix-ui-tpa/docs/internal/VERSION_RELEASE.md
```diff
@@ -26,7 +26,7 @@ It's pretty simple actually:
 
 1. First make sure that you're on the master branch
 2. Also, make sure the branch is clean and aligned to the `origin`
-3. run `npm run wut -- release-candidate -- <option>`, where `option` is the new version type:
+3. run `npm run wix-ui-tpa release-candidate <option>`, where `option` is the new version type:
```
argshook, line 29: because `<option>` is not a flag no need for any dashes here!

package.json
jonathanadler, line 26: maybe?
```suggestion
    "wut": "./run-package-command.sh wix-ui-tpa"
```
Less to type 😁

argshook: only for those who don't have autocomplete in the terminal :)

which is probably the vast majority :(
it can be `wut`, i'm just thinking for future when we have more packages if that wont become confusing. you know, `wsr`, `wut`, `wuf`, `wbu`, `wuc` etc. maybe too many abbreviations?

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_6754/

run-package-command.sh
```diff
@@ -0,0 +1,3 @@
+#/usr/bin/env bash
+
+lerna exec --scope $1 npm run $2 `[[ -n "${@:3}" ]] && echo "-- -- ${@:3}"`
```
jonathanadler, line 3: Isn't it supposed to be:
```suggestion
lerna exec --scope $1 -- npm run $2 `[[ -n "${@:3}" ]] && echo "-- ${@:3}"`
```
?

package.json
jonathanadler: I think it would be easier to use `wut`.
@sivanharel @nitayneeman what do you think?

run-package-command.sh
argshook: yeah that works too, not sure if there's difference. maybe easier to understand as opposed to double `-- --`?

package.json
argshook: let's have `wut` until further notice, let's not delay :)

jonathanadler (approved):

PR #6768, opened 2021-02-17
<VariableInput /> - Fix emoji utf splitting issue

tzachbon:
### 🔦 Summary

Fix utf splitting issue.

Related stack overflow issues:
* https://stackoverflow.com/questions/24531751/how-can-i-split-a-string-containing-emoji-into-an-array
* https://stackoverflow.com/questions/38345372/why-is-length-2


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_6768/

packages/wix-style-react/src/VariableInput/test/VariableInput.private.uni.driver.js
```diff
@@ -11,13 +11,17 @@ const getTagDriver = base =>
...
 export default (base, body) => {
-  return {
+  const driver = {
     ...publicDriverFactory(base, body),
-
     getPlaceholder: () => getPlaceholder(base).text(),
     isTagTiny: () => getTagDriver(base).isTiny(),
     isTagSmall: () => getTagDriver(base).isSmall(),
     isTagMedium: () => getTagDriver(base).isMedium(),
+    enterTextAsUser: async text => {
```
mykas, line 20: Why not reuse the same `enterText` method from the public uni driver?

tzachbon: Since I needed the `focus` and `blur` to trigger the parsing method of draft js.

I could explicitly use:
```js
      await driver.focus();
      await driver.enterText(text);
      await driver.blur();
```

But it doesn't have the context of why I'm calling focus and blur.
Let me know what do you think

mykas: I believe the idea is to provide public testkit method for our users where they could simulate it the same as in the web. Maybe then the original `enterText` is not good enough?

EgoziE:
@tzachbon Oh nice solution!
I didn't know Array.from can distinct icons.
I'm happy you finally found a solution for this problem /😃

packages/wix-style-react/src/VariableInput/test/VariableInput.private.uni.driver.js
tzachbon: I believe that 'enterText' is like raw enter text because you do not want to always blur after entering the text. Maybe the name could be better than enterTextAsUser.

mykas: But could user simulate the parsing somehow too? Thats the big question. Maybe user wants to test his input and output too.

mykas (approved):
Looks super good! We are very happy with this PR 🚀

PR #6819, opened 2021-02-22
hoist dev dependencies to the root package.json

argshook:
The purpose of this PR is to improve monorepo bootstrapping time (`npm install`)

currently it takes ~10mins locally and ~5mins on CI (because of cache) just
for `npm install`. 5mins on CI is not horrible, however 10mins locally
has a hit on dev velocity.

~This PR adds `--hoist` to `lerna bootstrap`. It makes lerna to move~
~common dependencies to the root `node_modules`. This way we reduce~
~multiple installations which saves bootstrapping time~

This PR does not add `--hoist` to `lerna bootstrap`. Unfortunately
adding it eventually increases bundle sizes of various wix-style-react
components. That's a nono, an internal change should not impact
distributed code quality.

Bundle size increase is solvable but requires more investigation and
effort. `--hoist` will likely be added in the future.

For now devDependencies are simply moved from nested `package.json`s to
the root `package.json`. This alone cuts local `npm install` time in
~half


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  Project Name: **wix-ui-tpa**

  * https://bo.wix.com/pages/wix-ui-tpa.pr_6819/

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6819

shlomitc:
Question - why are we hoisting all of the dependencies? I remember that at the times only shared dependencies were hoisted. If it is only used by one library - it shouldn’t be hoisted. Maybe there’s another reason for it?

argshook:
> Question - why are we hoisting all of the dependencies? I remember that at the times only shared dependencies were hoisted. If it is only used by one library - it shouldn’t be hoisted. Maybe there’s another reason for it?

i'm hoisting only `devDependencies` and `npm run build` needs all
`devDependencies` even if some of them is used in one package only.

This is the case also already, this PR does not add more `devDependencies`

---

we could argue that if you work locally on only one package that is not a
dependee of anything, and it has `devDependencies` that no other package
has, then yes, it's not totally necessary to define it in the root
package.json

but this approach is more harmful than benefitial, because you need to
track these special dependencies and we anyway encourage contributors to
`npm install` and `npm run ...` from the root, not from specific
package.

---

but that's not the main goal of this PR, the main is to reduce install
time (locally this hoisting cuts time for me in half) and as preparation
for `lerna --hoist` and eventual (cleaner) package-lock.json

scripts/post-install.sh
```diff
@@ -2,6 +2,11 @@
 
 lerna bootstrap --no-ci
 
+# symlink each package node_modules/.bin to root node_modules/.bin
+# this is needed to support CLI tools in each package,
+# without installing CLI tools separately, multiple times
+lerna exec "rm -r node_modules/.bin | true; ln -s \$LERNA_ROOT_PATH/node_modules/.bin node_modules/.bin"
```
jonathanadler, line 8: Interesting 🤔 
I would've thought that searching for the `.bin` is propagated
so that if the specific bin isn't found in the package's node_modules/.bin folder, it automatically uses the parent node_modules/.bin folder

jonathanadler (commented):
seems legit 👍 

Just a few things:
1) I always thought that lerna actually does this for us, meaning we keep using our "internal" package.json, but lerna installs them only once in the parent node_modules
2) The only real issue I can think of, is if a certain package needs a specific dependency version. But I guess in that case it could just be added to its package.json individually.
3) This means if someone needs to install a dev dependency, they'll need to add it to the root package.json. We should probably document this somewhere

scripts/post-install.sh
argshook: this depends on how the `$PATH` variable is set. running only, for
example `eyes-storybook` does not involve npm or anything, it's straight
up `eyes-storybook`. so for it to work it has to be available on the
`$PATH` either under some `/usr/local/bin` (for globally installed
packages) or `node_modules/.bin`

even with `npx`, it doesn't go up the directory tree to look for other
`.bin`s.

i believe this is not an issue with yarn, but we're not using it at the
moment

argshook:
@jonathanadler 

> 1. I always thought that lerna actually does this for us, meaning we keep using our "internal" package.json, but lerna installs them only once in the parent node_modules

yes, running `lerna bootstrap --hoist` does that. However, i prefer to
extract `devDependencies` to be at the top so that we can manage them
easier. For example, we have only one `yoshi` version to take care of
instead of multiple. It will also make it easier to move to falcon,
which has package-lock.json as a requirement

> 2. The only real issue I can think of, is if a certain package needs a specific dependency version. But I guess in that case it could just be added to its package.json individually.

yes, packages can still use `devDependencies` and will likely do so,
it's not some hard requirement to keep `devDependencies` only at the
root. It's more convenient to have it at the root, but not always
possible.

> 3. This means if someone needs to install a dev dependency, they'll need to add it to the root package.json. We should probably document this somewhere

it's only an issue if different packages have dependency `A` which is in
different versions per package. But probably some tooling could help us.

Not sure if it's an issue at the moment

jonathanadler (approved):

packages/wix-style-react/package.json
```diff
@@ -49,7 +49,6 @@
     "test:visual": "eyes-storybook -c ./.storybook/visual",
     "test:types": "tsc -p tsconfig.declarations.json",
     "test:swimm": "SWIMM_PATH=./node_modules/.bin/swimm_cli && ./scripts/get-swimm.sh $SWIMM_PATH $FORCE && ./scripts/run-swimm.sh $SWIMM_PATH",
-    "test:performance": "npm run transpile && npm run bundles:generate && npm install --no-save @wix/perfer && perfer --verbose",
```
argshook, line 52: this is not necessary anymore and it's enough to run `perfer --chunked --verbose --json` on any of the test scripts.

also, having `test:performance` would run wix-style-react package on
separate CI, without bootstrapping the monorepo, potentially missing
some dependencies (or in other words negating the whole bootstrapping
benefits)

PR #6824, opened 2021-02-23
<Calendar/> - Add the missing locales

sivanharel:
### 🔦 Summary

...

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  Project Name: **wix-ui-tpa**

  * https://bo.wix.com/pages/wix-ui-tpa.pr_6824/

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6824

sivanharel:
@jonathanadler, @shlomitc - 
According to the WUT Perfer tests:
Adding the new locales increases the bundle size by 🔺 +17.622kb so it changed from 121.879kb to 139.501kb.
For some reason, the Perfer of WSR wasn't failed. Maybe it doesn't run when it's failed on WUT. 🤔 

It sounds a lot but in the next task when we will find a solution to import only the relevant local the number of the languages shouldn't have an impact.
WDYT?

jonathanadler:
> @jonathanadler, @shlomitc -
> According to the WUT Perfer tests:
> Adding the new locales increases the bundle size by 🔺 +17.622kb so it changed from 121.879kb to 139.501kb.
> For some reason, the Perfer of WSR wasn't failed. Maybe it doesn't run when it's failed on WUT. 🤔
> 
> It sounds a lot but in the next task when we will find a solution to import only the relevant local the number of the languages shouldn't have an impact.
> WDYT?

WSR's threshold for the Calendar is 125Kb so maybe it's still high enough for it to not fail 🤔 
It does seem that WSR's perfer test didn't run, so it also might be that as you suggested it didn't get there because WUT failed.
Let's bump the bundle size and then see what happens.

shlomitc:
17kb is a small cost we will need to pay but since we plan on work the bundle size anyway, then I'm totally ok with adding this to WSR.

packages/wix-style-react/src/Calendar/index.d.ts
```diff
@@ -27,7 +60,7 @@ export interface CalendarProps {
   showYearDropdown?: boolean;
   showMonthDropdown?: boolean;
   shouldCloseOnSelect?: boolean;
-  locale?: string | Locale;
+  locale?: LanguageType | Locale;
```
jonathanadler, line 63: I think this will be a breaking change.
We might need to leave the `string` there as well

packages/wix-style-react/src/Calendar/Calendar.js
```diff
@@ -521,6 +521,18 @@ Calendar.propTypes = {
...
+      'el',
+      'fi',
+      'he',
+      'hi',
+      'hu',
+      'id',
+      'ms',
+      'ro',
+      'lt',
+      'vi',
```
jonathanadler, line 535: Can we maybe use the Locales from the LocaleUtils file so we won't have duplication?

packages/wix-style-react/src/CalendarPanel/index.d.ts
```diff
@@ -14,7 +15,7 @@ export interface CalendarPanelProps {
   showYearDropdown?: boolean;
   showMonthDropdown?: boolean;
   shouldCloseOnSelect?: boolean;
-  locale?: LanguageType | Locale;
+  locale?: CalendarLanguageType | Locale;
```
jonathanadler, line 18: This might also be a breaking change

packages/wix-style-react/src/DatePicker/index.d.ts
```diff
@@ -1,16 +1,18 @@
...
+export { LanguageType } from '../Calendar';
+
 export interface DatePickerProps extends CalendarProps {
   className?: string;
   customInput?: React.ReactNode;
   inputProps?: {};
   dateFormat?: string | Function;
   dateFormatV2?: string | Function;
-  locale?: string | Locale;
+  locale?: CalendarLanguageType | Locale;
```
jonathanadler, line 15: Also here

jonathanadler (commented):
Looks good!

Left some comments.

Also, do we want to add a test somehow?
Do we even have how?

packages/wix-style-react/src/Calendar/index.d.ts
sivanharel: It is not a breaking change. It is just a fix in the type file.
In the Calendar.js props type, it's defined as follow:
  /** DatePicker instance locale */
  locale: PropTypes.oneOfType([
    PropTypes.oneOf([
      'en',
      'es',
      'pt',
      'fr',
      'de',
      'pl',
      'it',
      'ru',
      'ja',
      'ko',
      'tr',
      'sv',
      'no',
      'nl',
      'da',
      'zh',
      'th',
      'cs',
      'uk',
      'ar',
      'bg',
      'el',
      'fi',
      'he',
      'hi',
      'hu',
      'id',
      'ms',
      'ro',
      'lt',
      'vi',
    ]),
    PropTypes.shape({
      code: PropTypes.string,
      formatDistance: PropTypes.func,
      formatRelative: PropTypes.func,
      localize: PropTypes.object,
      formatLong: PropTypes.object,
      match: PropTypes.object,
      options: PropTypes.object,
    }),
  ]),
  
  So also before my change a sreign which is not one of the the types was not allowd.

packages/wix-style-react/src/CalendarPanel/index.d.ts
sivanharel: In this file, it was exposed correctly I just added the missing locales and use the same type in order to prevent duplications.

packages/wix-ui-tpa/package.json
```diff
@@ -73,7 +73,6 @@
   },
   "dependencies": {
     "@stylable/dom-test-kit": "^3.5.0",
-    "@date-fns/upgrade": "^1.0.3",
```
sivanharel, line 76: I added this in my last PR by mistake.
It is not used in WUT.

jonathanadler (approved):

EgoziE (approved):

PR #6868, opened 2021-03-03
`<TableListItem/>` - enable ellipsis support for value property

mykas:
Currently if TabeListItem is used in this manner:
```
<TabelListItem options=[{ value: <Text ellipsis/>very very very very very very long message</Text> }]/>
```
The ellipsis will not work. 

The fix is to make the option value container act as parent which is css grid now. This overrides the parent css grid control over the content length.


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6868

PR #6888, opened 2021-03-07
add wuf dependency explicity and remove --no-install from scripts

jonathanadler:


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  Project Name: **wix-ui-tpa**

  * https://bo.wix.com/pages/wix-ui-tpa.pr_6888/

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6888

packages/wix-ui-tpa/package.json
```diff
@@ -143,6 +143,7 @@
     "wix-style-processor": "^3.1.14",
     "wix-ui-framework": "^3.5.2",
     "wix-ui-test-utils": "^1.0.0",
+    "wuf": "^0.0.1",
```
argshook, line 146:
```suggestion
```

this is not needed, `wuf` is coming from `wix-ui-framework`. 
`npm install wuf` is a differnet package

packages/wix-ui-tpa/.wuf/update-components.sh
```diff
@@ -16,28 +16,28 @@ npx --no-install wuf update \
 mkdir -p src/testkit
 
 # vanilla testkits
-npx --no-install wuf export-testkits \
+npx wuf export-testkits \
```
argshook, line 19: removing non-install is not good, because `npx` will try to install `wuf`, which is a different package. what we use is `wix-ui-framework`, which brings `wuf` CLI.

to really ensure that `wuf` is available in `postpublish`, there can be
`npm install wix-ui-framework` in `postpublish` script

argshook (changes requested):

packages/wix-ui-tpa/.wuf/update-components.sh
jonathanadler: oh ok cool 👍 
So maybe we can add `wix-ui-framework` to the devDependencies?
Shouldn't that be enough?
(Since we already run `npm install` in the `postpublish` script

jonathanadler: oh, but weird thing is that I see we do have `wix-ui-framework` in our dev dependencies.

jonathanadler: btw, maybe we don't need to use `npx` then?

argshook: yes, `wix-ui-framework` is already there. but because during
`postpublish` `npm install` does not happen, `wix-ui-framework` is not
installed and `wuf` is not available

in `wix-style-react` there's just `npm install` in `postpublish` script,
maybe this should happen in `wix-ui-tpa` too?
https://github.com/wix/wix-style-react/blob/master/packages/wix-style-react/package.json#L66

a bit longer publish time but no need to nitpick which packages need to
be installed

argshook: > btw, maybe we don't need to use `npx` then?

we do, and it should be with `--no-install`

`npx` helps resolve the path to executable. If a single package has
`node_modules/.bin/wuf`, just clean `wuf` (without `npx`) should be ok.

but i noticed there were some builds where `wuf` was not available as
executable (maybe some CI agent didn't have well configured `$PATH`).

in monorepo, when `wix-ui-framework` will get hoisted, a clean `wuf`
will not be enough even if `$PATH` is well configured. Something will
need to find that `wuf` is at some upper level `node_modules`. and
that's what `npx` can do

jonathanadler: I see 🤔 
But we're still getting the `not found: wuf` error, and we also run `npm install` in our postpublish script, so this was an effort to fix that

jonathanadler: https://github.com/wix/wix-style-react/blob/1d22f69602b662b756c5b21ce16f8fe8ffa7716d/packages/wix-ui-tpa/package.json#L62

argshook: hmm so `npx --no-install wuf` should really be enough.

and now that i merged the hoisting PR, `npx --no-install` is a must

jonathanadler: but it's not working 🤔  so I don't get it

PR #6895, opened 2021-03-09
<InputWithOptions/> - use props.value for highlighting

splintor:
### 🔦 Summary
Since `props.value` is what's used for rending the `value` in the `inputElement`, it only makes sense to use that value for highlighting.

See [this Slack thread](https://wix.slack.com/archives/C39FTGUBZ/p1615216858340000) for a use case where this makes a difference:
We use a controlled component and reset to non-empty `value` on `blur`. In that case, `inputValue` state is not reset, but we want the highlighting to be reset to the set `value`.


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6895

mykas (approved):
Good job! Thanks for fixing this. Lets wait for the build to get green.

mykas:
All good, just flakey.

PR #6896, opened 2021-03-09
Removed Grid usages from library

egidijusja:
### 🔦 Summary

...

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  Project Name: **wix-ui-tpa**

  * https://bo.wix.com/pages/wix-ui-tpa.pr_6896/

packages/wix-style-react/stories/Introduction/Cheatsheet/ComponentsCheatsheet.js
```diff
@@ -42,13 +42,15 @@ class ComponentsCheatsheet extends React.Component {
...
         />
         <Page.Content>
-          <Container>
+          <Layout>
             {componentsFamiliesArr.map((Family, i) => (
-              <Row key={`family-row-${i}`}>
-                <Family key={`family-${i}`} />
-              </Row>
+              <Cell key={`family-Cell-${i}`}>
+                <Box margin="0 -15px">
```
shlomitc, line 48: this looks wrong

egidijusja: I can agree, this looks weird. Earlier cheatsheet used Grid and one of its classes had negative padding styles:

Screenshot 2021-03-09 at 17 01 20

After we've changed Grid to Layout we don't have here negative paddings. So I've added here Box to keep old look. Anyway this could be removed, and then the width of cheatsheet page will be slightly narrower (for smaller screens), but it would still look good. So we have to decide which way should we go with this.

shlomitc: This could indicate something that will happen when people will change their implementation from `Grid` to `Layout`, so we should not just treat is a small issue, but rather investigate what and why are the differences

shlomitc: `Container`, `Row` and `Cell` all work together in order to achieve the required Grid behavior in our design system. It uses bootstrap's grid behind the scenes. `Container` also sets the min-max width and the others align them by 12 grid columns.

We should do some parity check to make sure that `Layout` answers the design guidelines when doing a layout of cards (in a page) and layout of form elements inside cards.

mykas: Hey @shlomitc yes, indeed we noticed these 2 things that `Container`, `Row` and `Cell` does:
1. Has restrictions like: min-width: 894px; max-width: 1254px;
2. Defined 12 grid columns just like Layout does too.

The layoting mechanism should not be the answer to width and height restrictions. Those max-width and max-height restrictions are actually defined in Page component. We double checked that with @milkyfruit and and according to their layout guidelines they do not include these restrictions for Layout components.

Other then that we have something called gutter defined per column which is also defined in Layout.

@egidijusj is gonna prepare a full blown migration doc which will say how to migrate from Grid to Layout. We gathered the cases in one doc.

packages/wix-style-react/stories/Spacing/index.story.js
```diff
@@ -21,25 +21,29 @@ export default {
...
-                <Col span={3}>
-                  <Text>=</Text>
-                </Col>
-                <Col span={4}>
-                  <Text secondary>
-                    {`${parseInt(name.substr(2)) * parseInt(stVars.Spacing)}px`}
-                  </Text>
-                </Col>
-              </Row>
+              <Cell key={name}>
```
mykas, line 28: Lets do a research on how css grid acts inside a restricted width div

packages/wix-style-react/stories/Introduction/Cheatsheet/ComponentsCheatsheet.js
mykas: Sorry @egidijusj for tagging you again xD

packages/wix-style-react/stories/Spacing/index.story.js
egidijusja: If we have parent div with restricted width our Layout component will take this into account only if div has `display: block`

mykas (approved):
Good job! You found lots of good insights on what we would like to explain to our users when writing the migration guide.

WixBuildServer:
### 🤖 Hi, I'm #Janet!

Project Name: **wix-ui-tpa**

* https://bo.wix.com/pages/wix-ui-tpa.pr_6896/

WixBuildServer:
### 🤖 Hi, I'm #Janet!

* https://preview.wix-style-react.com/pr_6896

PR #6897, opened 2021-03-09
Storybook stories - phase 2 - facelift

mykas:
This is a PR that will add new visuals to the storybook stories. Part of New Stories Guidelines effort.


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  Project Name: **wix-ui-tpa**

  * https://bo.wix.com/pages/wix-ui-tpa.pr_6897/

packages/wix-style-react/.storybook/preview-head.html
```diff
@@ -9,3 +9,10 @@
   rel="stylesheet"
   href="//static.parastorage.com/unpkg/@wix/wix-fonts@1.10.1/madefor.min.css"
 />
+
+<!-- Roboto font -->
+<link rel="preconnect" href="https://fonts.gstatic.com" />
+<link
```
shlomitc, line 15: why do we need Roboto? can't we just use Madefor?

mykas: Nope, this is agreed font with @milkyfruit design. The code part must be different to the description part.

shlomitc: Can't we use some native font? loading another font just for this sounds to me not worth it.
It's harder to maintain and keep track of what it relates to

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6897

packages/wix-style-react/.storybook/preview-head.html
mykas: In general I think all these configuration files should be held in one place then it will not be hard to maintain this.

shlomitc: let's see if we can inline the font in `wix-storybook-utils` instead

mykas: ok

jonathanadler:
I see that it's affecting our components as well.
Is it possible to use the fonts only for the storybook-part and isolate the components?

mykas:
@jonathanadler yes, this is a problem. One way is to move back the wix-storybook-utis package version to WUT and pin the version, but in general I believe your components should not be affected by global font family. How many components are being affected btw?

jonathanadler:
> @jonathanadler yes, this is a problem. One way is to move back the wix-storybook-utis package version to WUT and pin the version, but in general I believe your components should not be affected by global font family. How many components are being affected btw?

Yeah, we need to map the gaps and fix them sometime probably.
I need to check how many.

package.json
```diff
@@ -137,7 +137,6 @@
     "velocity": "^0.7.2",
     "webpack": "^4.44.2",
     "webpack-cli": "^3.3.12",
-    "wix-storybook-utils": "^3.0.139",
```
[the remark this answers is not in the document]
argshook: that's cool, just need not to forget to later upgrade wut also

PR #6902, opened 2021-03-09
infra - fix missing and incorrect types

shlomitc:
I've found these mismatches while playing around with bundling WSR to a `umd` module.
I used the `yoshi-flow-library` and converted the `index.js` file to `index.ts` file.
As a result, typescript errors were thrown and fixed through these changes. The experiment is on my machine so I can share it if needed

**Public API:**
1. add types of `FailWhale` - not critical, but needed.
2. refactor `RadioButton` as a standalone type and expose it

**Tests**:
1. change to use `esModuleInterop: true` (` TimeInput` to import `moment` as default)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6902

packages/wix-style-react/src/index.d.ts
```diff
@@ -307,6 +310,9 @@ export * from './PreviewWidget';
 export { default as Proportion } from './Proportion';
 export * from './Proportion';
 
+export { default as RadioButton } from './RadioGroup/RadioButton';
+export * from './RadioGroup/RadioButton';
```
EgoziE, line 314: I dunno if it's a good idea to expose it as a stand alone.
We already have Radio, and RadioGroup
Right now we recommend to use RadioGroup.RadioButton, not RadioButton directly
@mykas am I wrong?

shlomitc: It just that it was already exposed in the index file so I matched the correct types. If it shouldn’t be in the index we can remove it

mykas: We can't remove yet the RadioButton. Its being used widely in wix and we would break everybody. This is to delete from next major version.

EgoziE: I see

EgoziE (approved):

PR #6906, opened 2021-03-10
update reinstall script to not use lerna

argshook:
although bash scripts should be kept to a minimum, here we don't really
have a choice. `npm run reinstall` should work no matter if
`node_modules` are available or not, it should working in any kind of
state of the monorepo.

thus, we cannot use `lerna exec -- rm -rf ...`, which simplifies
removing of `dist` and similar. That's because user might try to `npm
run reinstall` when node_modules are corrupted, not installed or half
installed and `lerna` is not available.


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  Project Name: **wix-ui-tpa**

  * https://bo.wix.com/pages/wix-ui-tpa.pr_6906/

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6906

PR #6929, opened 2021-03-16
<Box /> - new prop gap

egidijusja:
### 🔦 Summary

...

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  Project Name: **wix-ui-tpa**

  * https://bo.wix.com/pages/wix-ui-tpa.pr_6929/

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6929

packages/wix-style-react/src/Box/Box.st.css
```diff
@@ -1,8 +1,15 @@
...
   -st-states: inline, direction(enum(horizontal, vertical)),
     alignItems(enum(left, center, right, space-between)),
     justifyContent(enum(top, middle, bottom, space-between));
   display: flex;
+  margin: 0 calc(-1 * var(--gap)) calc(-1 * var(--gap)) 0;
+}
+
+.root > * {
+  margin: 0 var(--gap) var(--gap) 0;
 }
```
shlomitc, line 13: take into consideration rtl support

packages/wix-style-react/src/Box/Box.st.css
```diff
@@ -1,8 +1,15 @@
+@st-global-custom-property --gap;
```
shlomitc, line 1: why not use stylable's hashing instead?

packages/wix-style-react/src/Box/Box.js
```diff
@@ -49,6 +49,7 @@ const formatSizeValue = value => {
 
 const Box = ({
   dataHook,
+  gap,
```
shlomitc, line 52: `gap` is a css api that is going to be adopted soon as a standard in all major browsers, so this means it will override the native one.
https://caniuse.com/?search=gap

maybe we should give it a different name

shlomitc (changes requested):

packages/wix-style-react/src/Box/Box.js
mykas: Hey Shlomi, the reason we decided to use this prop is basically because it is not supported in safari yet.

About changing the name - I don't think we want to change this to something else because this is universal name that is known not only in Box but also in Layout. 

@egidijusja soon will update this pr with new approach where we will use stylable mixins to check browser support and based on that either return native `gap` or apply custom solution with margins that would work in safari and other unsupported browsers.

packages/wix-style-react/src/Box/Box.st.css
mykas: Because we would not be able to pass css var from javascript otherwise. Check this out: https://github.com/wix/wix-style-react/pull/6929/files#diff-c01be18ed0796f31b8c348e248a78e2a91baf36c59ee1490c34190156838cd50R197

mykas: oh never mind. I just saw that in stylable you can reuse css vars in JS too.

packages/wix-style-react/src/Box/Box.js
shlomitc: I understand the intention, but in `Box` every API is handled as the original property as a style property.

Since it's not a 1-1 polyfill, we should be very careful with it.
See the differences between your implementation and the native behavior

image

```jsx
<>
  <Box gap="6px" flexWrap="wrap" width="150px">
    <Box border="1px solid black">Box</Box>
    <Box border="1px solid black">Box</Box>
    <Box border="1px solid black">Box</Box>
    <Box border="1px solid black">Box</Box>
    <Box border="1px solid black">Box</Box>
    <Box border="1px solid black">Box</Box>
  </Box>
  <hr />
  <div
    style={{ display: 'flex', gap: '6px', flexWrap: 'wrap', width: '150px' }}
  >
    <Box border="1px solid black">Box</Box>
    <Box border="1px solid black">Box</Box>
    <Box border="1px solid black">Box</Box>
    <Box border="1px solid black">Box</Box>
    <Box border="1px solid black">Box</Box>
    <Box border="1px solid black">Box</Box>
  </div>
</>;
```

packages/wix-style-react/src/Box/Box.st.css
egidijusja: Verified that with current solution rtl is supported.

mykas (approved):
Best solution in a year!

WixBuildServer:
### 🤖 Hi, I'm #Janet!

Project Name: **wix-ui-tpa**

* https://bo.wix.com/pages/wix-ui-tpa.pr_6929/

WixBuildServer:
### 🤖 Hi, I'm #Janet!

* https://preview.wix-style-react.com/pr_6929

PR #6965, opened 2021-03-24
fix `Uncaught [Error: Invalid hook call.`

argshook:
This PR fixes a build error:

```
Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:

1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app See for tips about how to debug and fix this problem.
```

more details about the error: https://reactjs.org/warnings/invalid-hook-call-warning.html

in our case it was `3.`. Due to recent updates in `wix-storybook-utils`
a new dependency chain was introduced:
```
- wix-storybook-utils
  - @storybook/addons
    - @storybook/api
      - react
```

our `node_modules` tree started to look something like this:

```
- node_modules
  - react
- packages
  - wix-style-react
    - node_modules
      - react
```

this is enough for various environments to get confused on which `react`
should be used. for example, runnin tests with jest would resolve
`react` to `packages/wix-style-react/node_modules/react`, but
`react-dom` would be used from `node_modules/react-dom` and some other
file possibly `require`s `node_modules/react`.

even if package versions are the same, the fact that there are two
instances of `react` causes it to throw an error.

Thus, to solve this we're in `wix-storybook-utils` we're moving
`@storybook/addons` to be a `peerDependency`. This then requires
consumer of `wix-storybook-utils` to install this dependency on their
own. But it also has a nice side effect of much easier to resolve
node_modules tree.

This way we no longer have multiple react versions


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  Project Name: **wix-ui-tpa**

  * https://bo.wix.com/pages/wix-ui-tpa.pr_6965/

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://preview.wix-style-react.com/pr_6965

env-github-token:
### 🤖 Hi, I'm #Janet!

Project Name: **wix-ui-tpa**

* https://bo.wix.com/pages/wix-ui-tpa.pr_6965/

env-github-token:
### 🤖 Hi, I'm #Janet!

* https://preview.wix-style-react.com/pr_6965

PR #6980, opened 2021-03-29
add v8 dependency

argshook:
after `@babel/core` update it became required. a suggested fix of setting `target: 'web'` in `webpack.config.js` does not seem to help.

https://github.com/babel/babel/issues/13067

~merging to get green builds~


WixBuildServer:
<!-- PRֹֹֹ_COMMENT_FOR:wix-ui-tpa !>
### 🤖 Hi, I'm #Janet!

Project Name: **wix-ui-tpa**

* https://bo.wix.com/pages/wix-ui-tpa.pr_6980/

WixBuildServer:
<!-- PRֹֹֹ_COMMENT_FOR:wix-style-react !>
### 🤖 Hi, I'm #Janet!

* https://preview.wix-style-react.com/pr_6980

package.json
```diff
@@ -139,6 +139,7 @@
     "stylelint-no-unsupported-browser-features": "^4.1.4",
     "ts-jest": "^22.0.3",
     "typescript": "^3.9.7",
+    "v8": "^0.1.0",
```
nicolo-ribaudo, line 142: This is not the `v8` thing Babel imports; Babel uses https://nodejs.org/api/v8.html (which is a native Node.js module).

The `v8` module from npm [is empty](https://unpkg.com/browse/v8@0.1.0/lib/v8.js).

argshook: hey, thanks for the headsup.

`v8` package in npm registry is indeed an empty package, not even available anymore on github.

however, it is enough for a build to pass.

`v8` is resolved, [this](https://github.com/babel/babel/blob/main/packages/babel-core/src/transformation/util/clone-deep.js#L5) gets falsy and [`clone-deep-browser`](https://github.com/babel/babel/blob/main/packages/babel-core/src/transformation/util/clone-deep-browser.js) is used resulting in a passable build.

none of this is a good solution, but it's too easy and can even get
suggested by some tools to just install a missing dep.

those who're too lazy to get deeper (like me) can fall for this
workaround.

perhaps you can suggest a better way for a fix? or maybe a pending fix
is here and will get released shortly? https://github.com/babel/babel/pull/13068

Thanks!

nicolo-ribaudo: Yeah, I plan to release a patch when babel/babel#13068 is merged! (probably this evening)

argshook: i suppose we should really be using `@babel/standalone` as we're using `@babel` in a browser to provide code sandbox

nicolo-ribaudo: Yes, `@babel/standalone` is already bundled with the browser-specific modules, and it should thus be safer.

argshook:
closing in favor of https://github.com/wix/wix-style-react/pull/6981

PR #6994, opened 2021-04-01
fix getColumnTextAt typo

kobiburnley:
getColum -> getColum**n**

types are correct: https://github.com/wix/wix-style-react/blob/bbb4feb53fc693edbadc00488eb41c7276dab6f6/packages/wix-style-react/src/TableListItem/TableListItem.uni.driver.d.ts#L4


shlomitc:
I see a couple of usages for the current method so renaming will be a breaking change.

Instead of breaking, let's add a new method with the correct spelling, alongside the existing one, and deprecate the old one.
You can also add [`deprecationLog`](https://github.com/wix/wix-style-react/blob/master/packages/wix-style-react/src/Checkbox/Checkbox.js#L58) to the old method letting users know of the new one.

kobiburnley:
> I see a couple of usages for the current method so renaming will be a breaking change.
> 
> Instead of breaking, let's add a new method with the correct spelling, alongside the existing one, and deprecate the old one.
> You can also add [`deprecationLog`](https://github.com/wix/wix-style-react/blob/master/packages/wix-style-react/src/Checkbox/Checkbox.js#L58) to the old method letting users know of the new one.

@shlomitc done 👍

packages/wix-style-react/src/TableListItem/TableListItem.uni.driver.js
```diff
@@ -27,7 +38,7 @@ export const tableListItemDriverFactory = (base, body) => {
     },
     toggleCheckbox: () => {
       const driver = getCheckboxDriver();
-      driver.click();
+      return driver.click();
```
mykas, line 41: I don't think there is something to return here. This is action based method

mykas (approved):

packages/wix-style-react/src/TableListItem/TableListItem.uni.driver.js
kobiburnley: @shlomitc I see you added this return statement here: https://github.com/wix/wix-style-react/pull/6994/commits/f9844e7cf53f960d7ad2adfeb84fd1a0fdf9cddc

I reverted it now, before I noticed it wasn't my change @mykas

kobiburnley: @mykas `.click()` returns a promise, I guess it makes sense to return it.

kobiburnley: It also align with the types: https://github.com/wix/wix-style-react/blob/3b94417c90adf7209d915a6c52bcf155835e377b/packages/wix-style-react/src/TableListItem/TableListItem.uni.driver.d.ts#L7

WixBuildServer:
### 🤖 Hi, I'm #Janet!

Project Name: **wix-ui-tpa**

* https://bo.wix.com/pages/wix-ui-tpa.pr_6994/

WixBuildServer:
### 🤖 Hi, I'm #Janet!

* https://preview.wix-style-react.com/pr_6994

packages/wix-style-react/src/TableListItem/TableListItem.uni.driver.js
shlomitc: yes, that was my change. wanted to retrigger the CI but also found this as a reasonable change.
Anyway, this method is not tested at all so probably if we add a test this return is needed.

PR #7005, opened 2021-04-05
docs - fix the root link to components folder path

shlomitc:


mykas:
The fix is great, but there is one problem with it - we have many stories hardcoded with links. I will take care of those in this same pr.

shlomitc:
> The fix is great, but there is one problem with it - we have many stories hardcoded with links. I will take care of those in this same pr.

This can be done as a separate task since it might take a while. this is a quick fix that will cover the majority of the stories

mykas:
ok

PR #7048, opened 2021-04-15
[wsr] Expose unidrivers for our users

mykas:
This is PR that introduces generator template for unidriver expose. We need to expose this so that our users could composite our components unidrivers and create other libraries just like Cairo project does.


packages/wix-style-react/testkit/unidriver.js
```diff
@@ -0,0 +1,199 @@
+/* eslint-disable */
+/* tslint:disable */
+/*
+ * DO NOT EDIT THIS FILE!
+ * YOUR CHANGES WILL BE OVERWRITTEN!
+ * FILE IS BASED ON .wuf/testkits/unidriver.template.ejs
+ * AND GENERATED BY `wuf export-teskits`
+ */
+module.exports = {
+  ...require("./../dist/src/Accordion/Accordion.uni.driver"),
```
argshook, line 10: add this file to `.gitignore`

packages/wix-style-react/.wuf/testkits/unidriver.template.ejs
```diff
@@ -0,0 +1,24 @@
...
+    const driverPath = component.uniDriverPath ||
+      [
+        './../dist/src/',
+        ...(component.deprecated ? ['deprecated/'] : []),
+        component.name,
+        '/',
+        component.pascalCase ? utils.toPascal(component.name) : component.name,
+        '.uni.driver'
+      ].join('');
+%>  ...require("<%= driverPath %>"),
```
argshook, line 22: i'd strongly suggest to explicitly name the exports instead of using the spread operator.

the `...require` is used in wix-ui-core and it's troublesome, because one require can potentially override another, there's a chance something random can get exported, plus the only way to know what's exported is to evaluate code. It should be enough to just read it.

before the exports file generation in wix-ui-core, there was a manually written file with all the `...require`s. thus, when changing it to be generated i left it with same spread, just to reduce the chance of broken builds. but it's not a good approach to take an example from.

i think this template should be similar to other testkit templates, with one major difference - no testkit factory function, just exported testkit.

packages/wix-style-react/testkit/unidriver.js
```diff
@@ -0,0 +1,199 @@
...
+  ...require("./../dist/src/AddItem/AddItem.uni.driver"),
+  ...require("./../dist/src/AddressInput/AddressInput.uni.driver"),
+  ...require("./../dist/src/AddressInputItem/AddressInputItem.uni.driver"),
```
argshook, line 13: checkout these newlines. not a big deal but by tweaking `.ejs` file you can avoid these

packages/wix-style-react/.wuf/testkits/unidriver.template.ejs
```diff
@@ -0,0 +1,24 @@
...
+    ((component.hasVanillaDriver || component.hasProtractorDriver) && component.hasUnidriver)
+    || (!component.hasVanillaDriver && !component.hasProtractorDriver)
+  )
+  .map(component => {
+    if(component.name === 'Themes') {
+      return ''
+    }
+    const driverPath = component.uniDriverPath ||
+      [
+        './../dist/src/',
```
argshook, line 15:
```suggestion
        '../dist/src/',
```

packages/wix-style-react/.wuf/testkits/unidriver.template.ejs
```diff
@@ -0,0 +1,24 @@
+module.exports = {
+<%
+  components
+  .filter(component => !component.noDriver)
+  .filter(component =>
+    ((component.hasVanillaDriver || component.hasProtractorDriver) && component.hasUnidriver)
+    || (!component.hasVanillaDriver && !component.hasProtractorDriver)
+  )
+  .map(component => {
+    if(component.name === 'Themes') {
```
argshook, line 10: since we already have two `.filter`s above, it would make sense to also use `.filter` for `Themes`

packages/wix-style-react/.wuf/testkits/unidriver.template.ejs
```diff
@@ -0,0 +1,24 @@
+module.exports = {
+<%
+  components
+  .filter(component => !component.noDriver)
+  .filter(component =>
+    ((component.hasVanillaDriver || component.hasProtractorDriver) && component.hasUnidriver)
```
argshook, line 6: `.hasUnidriver` does not exist in wix-style-react definitions.js

env-github-token:
### 🤖 Hi, I'm #Janet!

Project Name: **wix-ui-tpa**

* https://bo.wix.com/pages/wix-ui-tpa.pr_7048/

argshook (changes requested):
i understand the inspiration is taken from wix-ui-core https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/.wuf/testkits/unidriver.template.ejs

unfortunately it's not a good example there. the template in wix-ui-core directly mimics whatever was done manually before the generator to reduce the chance of broken builds. but what was done before generating from template is not a good approach. would be best not to repeat it in wix-style-react

env-github-token:
### 🤖 Hi, I'm #Janet!

* https://preview.wix-style-react.com/pr_7048

packages/wix-style-react/.wuf/testkits/unidriver.template.ejs
mykas: yes! totally make sense! thanks for pointing this out.

argshook (approved):

mykas:
Now the big question is - should we expose this through documentation? And where should be the right place. @shlomitc @EgoziE can you give us your opinion on this?

shlomitc:
Definitely expose it in our docs.
Regarding where it should sit - I'd say that the same way as other testkits are documented in the storybook, and also in this doc: https://github.com/wix/wix-style-react/blob/master/packages/wix-style-react/docs/usage/COMPONENTS_DRIVERS.md#supported-platforms

PR #7064, opened 2021-04-22
add all devDeps for individual packages

argshook:
This PR should be considered a quick fix as a workaround, to reduce the number of failures we have when publishing new versions.

---

This PR copy-pastes `devDependencies` from root `package.json` to `wix-style-react/package.json` & `wix-ui-tpa/package.json`.

Yes, it potentially increases install time and should not be necessary. However, due to the current CI setup, publish step does not work well.

Currently CI, when publishing an artifact, does not bootstrap the monorepo. But, because after publishing we also want to publish storybook, we need a bunch of `devDependencies` to allow doing that.

Ideally storybooks should be separated into dedicated packages, but that's a greater effort.


env-github-token:
### 🤖 Hi, I'm #Janet!

Project Name: **wix-ui-tpa**

* https://bo.wix.com/pages/wix-ui-tpa.pr_7064/

env-github-token:
### 🤖 Hi, I'm #Janet!

* https://preview.wix-style-react.com/pr_7064

sivanharel:
I wonder what will happen when the versions in `wix-style-react/package.json` & `wix-ui-tpa/package.json` won't be the same.
We probably should verify that whenever we are updating a version in one of the dependencies to do it in both files `wix-style-react/package.json` & `wix-ui-tpa/package.json` right?

argshook:
> I wonder what will happen when the versions in `wix-style-react/package.json` & `wix-ui-tpa/package.json` won't be the same.
> We probably should verify that whenever we are updating a version in one of the dependencies to do it in both files `wix-style-react/package.json` & `wix-ui-tpa/package.json` right?

although failures are possible, it's unlikely to happen, we lived like this for a while and the biggest cost is maintenance, not failures.

anyway, this is a temp workaround (one of many) to make legacy CI happy with monorepo.

unfortunately this temp workaround is failing on something unrelated, i'm checking (test6 timeout, `autotools sanity` is not happy for some reason)

PR #7093, opened 2021-04-29
<Tab/> - add prefix element prop

jjaskulis:
Adding support for prefix icon in Tab components, according to specs:
[(https://zeroheight.com/7sjjzhgo2/p/7404ef-03-navigation/t/662242](https://zeroheight.com/7sjjzhgo2/p/7404ef-03-navigation/t/662242

### 🔦 Summary

...

### ✅ Checklist

- 📚 Change is documented
  - [ ] Story
  - [ ] API description
  - [ ] Cheatsheet
  - [ ] Other (explain)
- 🔬 Change is tested
  - [ ] Component
  - [ ] Visual test


env-github-token:
### 🤖 Hi, I'm #Janet!

Project Name: **wix-ui-tpa**

* https://bo.wix.com/pages/wix-ui-tpa.pr_7093/

env-github-token:
### 🤖 Hi, I'm #Janet!

* https://preview.wix-style-react.com/pr_7093

packages/wix-ui-tpa/src/components/Tabs/Tab/Tab.st.css
```diff
@@ -61,3 +63,13 @@
...
+
+.tabContentWrapper{
+    display: flex;
+    align-items: center;
+    justify-content: center;
+}
+
+.tabContentWrapper > *:first-child{
+    margin-inline-end: 4px;
+}
```
jonathanadler, line 75: I understand this works well since the `title` is a `string`, so this only affects the prefix element.
However, it's a bit "undocumented" in this case.
Meaning, if for instance we, for some reason, add another element before the prefix, this will stop working.

So I wonder if maybe we should do something like this:
```jsx
const renderTabContent = () => (
    <div className={classes.tabContentWrapper}>
      {prefix ? <span className={classes.prefix}>{prefix}</span> : null}
      <span className={classes.title}>{title}</span>
    </div>
  );
```
and then
```css
.prefix {
    margin-inline-end: 4px;
}
```

wdyt?

jonathanadler (commented):
Very very nice 👌 

I just left one small comment, tell me what you think

packages/wix-ui-tpa/src/components/Tabs/Tab/Tab.st.css
jjaskulis: I agree, better to have a more straightforward way without any gotcha's. Updated

jonathanadler (approved):
🥳

